PR #47, opened 2026-06-17
feat(eve): support inline tool auth providers

allenzhou101:
## Summary
- add `ctx.getToken(provider, options?)` and `ctx.requireAuth(provider, options?)` for inline tool auth providers
- wrap all authored tools with an auth-capable context while preserving no-arg top-level `auth` behavior
- reuse scoped authorization for inline auth completion, cache eviction, challenge parking, and loop guarding
- document the API and add the original implementation plan under `research/`

## Verification
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/definitions/tool-auth.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/tool-auth.integration.test.ts`
- `pnpm docs:check`

Note: `pnpm --filter eve run test:integration -- src/execution/tool-auth.integration.test.ts` appears to run the full integration suite and hit an unrelated failure in `src/execution/workflow-entry.integration.test.ts`; the direct focused Vitest invocation for `tool-auth.integration.test.ts` passes.


docs/guides/auth-and-route-protection.md
```diff
@@ -262,15 +262,50 @@ export default defineTool({
...
+    if (repo.status === 401) ctx.requireAuth(githubAuth);
+
+    return updateLinearIssue(issueId, linearToken, await repo.json());
+  },
+});
+```
+
+The tool's `ctx` exposes two auth accessors:
 
 - `ctx.getToken()` resolves the bearer for the declared strategy, checking the per-step token cache first. With an interactive strategy, a cache miss suspends the turn on a framework-owned callback URL, shows a "Sign in" affordance, and re-runs the tool once the OAuth callback completes.
```
ctgowrie: Do we still need getToken?

docs/guides/auth-and-route-protection.md
```diff
@@ -262,15 +262,50 @@ export default defineTool({
...
+- `ctx.getToken(provider, options?)` resolves an inline provider such as `connect("github")`. It uses the same cache, callback, and sign-in machinery, but scopes the flow to that provider instead of the tool's top-level `auth`.
 - `ctx.requireAuth()` throws `ConnectionAuthorizationRequiredError` to gate the tool on authorization before any token resolves. The runtime turns that into the same consent prompt.
```
ctgowrie: Same question. Let's move towards requiring provider

docs/guides/auth-and-route-protection.md
allenzhou101: Hmm if we deprecate the auth key, not sure I would go there immediately tho. Think that shape is pretty intuitive

docs/guides/auth-and-route-protection.md
```diff
@@ -245,19 +245,20 @@ The first call that needs the connection kicks off an OAuth sign-in, surfaced as
...
 
-When one tool calls a service behind OAuth, it can declare its own `auth` and skip the separate connection. `auth` takes the same shapes: `connect("...")` for Vercel Connect-backed OAuth, a custom interactive definition, or a plain `{ getToken }` for static credentials.
+When one tool calls a service behind OAuth, keep the auth provider at the call site and skip the separate connection. Providers take the same shapes as connection `auth`: `connect("...")` for Vercel Connect-backed OAuth, a custom interactive definition, or a plain `{ getToken }` for static credentials.
 
 ```ts title="agent/tools/list_okta_groups.ts"
 import { connect } from "@vercel/connect/eve";
 import { z } from "zod";
 
+const oktaAuth = connect("okta");
```
dvoytenko: Nit: UIDs are always `okta/myagent`.

docs/guides/auth-and-route-protection.md
```diff
@@ -245,19 +245,20 @@ The first call that needs the connection kicks off an OAuth sign-in, surfaced as
...
+
 export default defineTool({
   description: "List the caller's Okta groups.",
   inputSchema: z.object({}),
-  auth: connect("okta"),
   async execute(_input, ctx) {
-    const { token } = await ctx.getToken();
+    const { token } = await ctx.getToken(oktaAuth, { displayName: "Okta" });
```
dvoytenko: We don't need `displayName` anymore?

docs/guides/auth-and-route-protection.md
```diff
@@ -266,14 +267,49 @@ export default defineTool({
...
 
-Declaring `auth` adds two accessors to the tool's `ctx`:
+This same inline shape naturally handles tools that need more than one credential:
+
+```ts title="agent/tools/sync_ticket.ts"
+import { connect } from "@vercel/connect/eve";
+import { defineTool } from "eve/tools";
+import { z } from "zod";
+
+const githubAuth = connect("github");
```
dvoytenko: ditto on UID

docs/guides/auth-and-route-protection.md
```diff
@@ -266,14 +267,49 @@ export default defineTool({
...
+
+    return updateLinearIssue(issueId, linearToken, await repo.json());
+  },
+});
+```
+
+The tool's `ctx` exposes provider-scoped auth accessors:
+
+- `ctx.getToken(provider, options?)` resolves an inline provider such as `connect("github")`. It uses the same cache, callback, and sign-in machinery, but scopes the flow to that provider instead of the tool's legacy top-level `auth`.
+- `ctx.requireAuth(provider, options?)` does the same for an inline provider. Use it after a downstream `401` rejects a token returned by `ctx.getToken(provider)`.
```
dvoytenko: the sentence is not clear. recommend to rewrite "does the sane for and inline...": what's "the same"?

dvoytenko: Also, `requiredAuth` seems a bit confusing too. Why not just `ctx.resetToken`? Though feel free to push back. Maybe I just need to get used to it.

docs/reference/typescript-api.md
```diff
@@ -56,13 +56,15 @@ A few non-`define*` helpers round out the set: `disableTool` and `ExperimentalWo
...
-| `ctx.getToken()`           | Resolve the bearer token for a tool's declared `auth` (throws without `auth`) |
-| `ctx.requireAuth()`        | Force the tool's authorization flow before proceeding                         |
+| Member                      | Use                                                                           |
+| --------------------------- | ----------------------------------------------------------------------------- |
+| `ctx.session`               | Current session, turn, auth, and optional parent lineage (read-only)          |
+| `ctx.getSandbox()`          | Live sandbox handle for the current agent                                     |
+| `ctx.getSkill(identifier)`  | Handle for a named skill visible to the current agent                         |
+| `ctx.getToken(provider)`    | Resolve a bearer token for an inline auth provider such as `connect("...")`   |
+| `ctx.requireAuth(provider)` | Re-challenge an inline provider, commonly after a downstream `401`            |
+| `ctx.getToken()`            | Deprecated shortcut for a tool's top-level `auth`                             |
```
dvoytenko: What's "deprecated shortcut"?

packages/eve/src/public/definitions/tool.ts
```diff
@@ -50,13 +51,29 @@ export type ToolAuthDefinition =
...
+ * Controls how an inline tool auth provider is scoped and presented.
+ *
+ * The scope keys token caches, callback URLs, pending authorization state,
+ * and authorization completion. When omitted, eve derives a stable
+ * tool-qualified scope from the provider metadata.
+ */
+export interface ToolAuthScopeOptions {
+  readonly connection?: ConnectionAuthorizationContext;
+  readonly displayName?: string;
+  readonly scope?: string;
```
dvoytenko: Only `scope`? In Connect, we have additional targeting with `resources: string[]` and more importantly `authorizationDetails: []` - this one is important because something like GitHub doesn't fall into `scope` easily, so it has to use RAR.

dvoytenko (changes requested):

docs/guides/auth-and-route-protection.md
allenzhou101: ya im not a huge fan of requireAuth. not sure if it should be like triggerReauth maybe. open to new naming, not sure about resetToken tho

AndrewBarba (approved):

docs/guides/auth-and-route-protection.md
dvoytenko: How about `requireReauth`?

dvoytenko (approved):

PR #61, opened 2026-06-17
feat(eve): multi-line input in the dev TUI prompt (bracketed paste + Shift+Enter)

ritiksahni:
## Problem

The dev TUI prompt is a single-line editor where `Enter` submits. A pasted newline was decoded as `Enter`, so pasting a multi-line message (a brief, a code block, logs) submitted only the first line and dropped the rest. There was also no way to compose a multi-line message by hand.

## What this does

- **Bracketed paste (DEC private mode 2004).** The prompt enables bracketed paste while it's active and decodes the `\x1b[200~ … \x1b[201~` wrapper in `nextKey` as a single `paste` token, reassembling it across reads (it returns `incomplete` until the end marker arrives). Content is sanitized: newlines and tabs preserved, CRLF/CR normalized to LF, embedded ESC and other C0/DEL dropped so a paste can't inject escape sequences. The paste inserts intact and `Enter` submits the whole buffer.
- **Shift+Enter / Alt+Enter insert a newline** instead of submitting, so you can compose multi-line input by hand. Decoded from the xterm modifyOtherKeys form (`\x1b[27;2;13~`), the kitty/CSI-u form (`\x1b[13;2u`), and `ESC`+CR/LF. A bare `Enter` still submits.
- Embedded newlines render inline as a dim `⏎` glyph so the single input row never writes a literal newline into the live region; the buffer and the submitted message keep their real newlines.

### DIFF
--- a/.changeset/tui-multiline-paste.md
+++ b/.changeset/tui-multiline-paste.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Multi-line input in the dev TUI prompt: pasting multi-line text now inserts it intact via bracketed paste (instead of truncating at the first line), and Shift+Enter / Alt+Enter insert a newline without sending. Embedded newlines show inline as `⏎` and are preserved on submit.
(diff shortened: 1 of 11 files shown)


ruiconti:
Thanks for the PR @ritiksahni. This was indeed lacking, but this patch was not working as expected - running this branch you wouldn't really get multiline besides a single-line with new line markers.

Feedback is good though, I ported it to #67 here with attribution to you.

PR #65, opened 2026-06-18
feat(eve): warn when Vercel build skips sandbox prewarm

ruiconti:
An agent ran `vercel deploy --prebuilt` on an eve app and shipped a build that broke at runtime — the deployment referenced sandbox templates that were never provisioned.

Why: sandbox template prewarming only runs when `VERCEL_DEPLOYMENT_ID` is set. A prebuilt build (`VERCEL` set, no deployment id) skips prewarming but still writes `.vercel/output` pointing at those templates, with no signal the output is unsafe to deploy.

Fix: `runVercelBuildPrewarm` now warns when it skips inside a Vercel build, pointing to `vercel deploy` (hosted source build, which sets the deployment id) instead of `--prebuilt`.

**What's here**
- Warning in the build-time prewarm hook.
- Note in the deployment guide.
- Scenario test for the skip-and-warn path.

**Warn, not fail** — the skip is correct for dev and one-off builds, so blocking the build would be wrong. Prewarm running and then failing still fails the build; this only covers prewarm being skipped. The remaining hazard is deploying that output with `--prebuilt`, which the build can't detect, so it flags the risk instead.

### DIFF
--- a/packages/eve/src/internal/nitro/host/vercel-build-prewarm.ts
+++ b/packages/eve/src/internal/nitro/host/vercel-build-prewarm.ts
@@ -2,6 +2,11 @@ import { prewarmAppSandboxes } from "#execution/sandbox/prewarm.js";
 
 type PrewarmAppSandboxesInput = Parameters<typeof prewarmAppSandboxes>[0];
 
+const VERCEL_BUILD_PREWARM_SKIPPED_WARNING =
+  "[eve] WARNING: Skipped Vercel sandbox template prewarm because VERCEL_DEPLOYMENT_ID is missing. " +
+  "The generated .vercel/output may reference sandbox templates that were not provisioned. " +
+  'Do not deploy it with "vercel deploy --prebuilt"; use "vercel deploy" so Vercel builds from source.';
+
 /**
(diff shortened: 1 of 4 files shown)


packages/eve/src/internal/nitro/host/vercel-build-prewarm.ts
```diff
@@ -30,6 +35,9 @@ export function shouldPrewarmVercelBuild(): boolean {
  */
 export async function runVercelBuildPrewarm(input: PrewarmAppSandboxesInput): Promise<boolean> {
   if (!shouldPrewarmVercelBuild()) {
+    if (process.env.VERCEL?.trim() && !process.env.VERCEL_DEPLOYMENT_ID?.trim()) {
+      console.warn(VERCEL_BUILD_PREWARM_SKIPPED_WARNING);
+    }
```
ijjk, line 40: I think the missing template from being warmed during build should be recoverable at runtime we should never fail fataly if a snapshot/template is missing for sandbox 🤔

ijjk (approved):

PR #67, opened 2026-06-18
feat(eve): multi-line, UTF-16-safe input in the dev TUI prompt

ruiconti:
The dev TUI prompt was a single-line editor where Enter submits. A pasted newline decoded as Enter, so pasting a multi-line message submitted the first line and dropped the rest, and there was no way to compose multi-line input by hand. This replaces it with a grapheme- and width-aware multi-line editor.

History note: this branch was force-pushed to a single clean commit. It supersedes an earlier line-based version (and the original bracketed-paste idea from #61, @ritiksahni). The prior approval is stale — it was against the old implementation, so this wants a fresh review. Base is `main`.

## What

- Bracketed paste, so a multi-line paste inserts intact instead of truncating.
- One visual row per line; the caret's row scrolls horizontally, tall input scrolls vertically.
- A block caret that doesn't shift text or flash a space.
- `↑`/`↓` move between lines, `Shift+Enter` inserts a newline, `Ctrl+C` clears the prompt.
- All editing is grapheme- and display-width correct, not UTF-16 code units.

## Bracketed paste and the key model

The prompt enables DEC mode 2004 while active, so the terminal frames a paste in `\x1b[200~ … \x1b[201~`; `nextKey` captures it as one token and waits for the close marker across reads. Keys are a unified `text` key carrying a `framing` tag (framed paste vs typed). Single-line inputs flatten a framed paste's newlines to spaces so they can't add rows; the prompt keeps them. Pasted content is sanitized (CRLF/CR to LF, drops C0/DEL/C1 introducers) so a paste can't smuggle an escape sequence.

## Layout, caret, and navigation

The buffer renders one visual row per logical line. The caret's row windows horizontally around the cursor (a long line scrolls instead of hiding the caret); a draft taller than the terminal windows vertically to the available height with ellipsis gutters. The block caret reverse-videos the grapheme under the cursor rather than inserting a bar, so text never shifts and the blink doesn't flash a space. `↑`/`↓` step between lines and fall back to prompt history at the top and bottom edge.

## UTF-16 / grapheme correctness

A shared `src/shared/text-boundaries.ts` (`previousGraphemeBoundary`) plus width-aware helpers (`inputTextWidth`, `renderInputText`) replace code-unit slicing. Moves, deletes, the caret cell, and the horizontal window all operate on grapheme clusters and display columns, so a wide CJK glyph, a combining mark, or a ZWJ emoji is never split or mismeasured. The same boundaries fix the setup-panel rename, the searchable-select filter, and the select-state backspace.

## Scope

Touches the prompt plus the setup-panel and select inputs that share the text-boundaries module. A fixture model swap in the working tree was intentionally left out. The whole change is one signed commit.


packages/eve/src/cli/dev/tui/terminal-renderer.ts
```diff
@@ -2698,16 +2732,75 @@ function clip(line: string, width: number): string {
...
-  const prompt = active ? c.cyan(theme.glyph.prompt) : c.dim(theme.glyph.prompt);
-  const body = active ? content : c.dim(content);
-  return [clip(`${prompt} ${body}`, width), ""];
+  const style = (segment: string): string =>
+    isCommand && segment.length > 0 ? c.blue(segment) : segment;
+
+  const layout = layoutPromptInput({ text, cursor });
+  const promptGlyph = c.cyan(theme.glyph.prompt);
+  // Columns left for row text after the two-column gutter (glyph + space).
+  const budget = Math.max(1, width - 2);
```
ijjk: The active prompt row no longer reserves a column for the block caret. With `budget = width - 2`, a line that exactly fills the prompt body produces `before` at the full body width, then appends the EOL caret space; the final `clip(..., width)` cuts that caret off. That means the caret can disappear right at the terminal edge, which is the case this windowing is meant to make visible. Could we reserve one cell for the caret on the active row, at least when the caret is at EOL?

packages/eve/src/cli/dev/tui/line-editor.ts
```diff
@@ -96,7 +97,10 @@ export function deleteWord(state: LineState): LineState {
 export function applyLineEditorKey(state: LineState, key: TerminalKey): LineState | undefined {
   switch (key.type) {
     case "character":
+    case "paste":
       return insert(state, key.value);
+    case "newline":
```
ijjk, line 160: Now that the prompt buffer can contain embedded newlines, the line-kill commands still operate on the whole buffer rather than the current logical line. For example, `Ctrl+K` on the first line of `one\ntwo` deletes the newline and `two`, and `Ctrl+U` from the second line deletes the prior line too. That conflicts with the documented/readline-style “kill line” behavior. Could we bound these to the nearest `\n` before/after the caret and add multi-line tests?

ijjk (dismissed):
Approved with two editor-behavior nits inline.

AndrewBarba (dismissed):

packages/eve/src/cli/dev/tui/terminal-renderer.ts
ruiconti: Fixed in 4949277. Reserved the column back: the caret row now windows to `width - 3` (gutter glyph + space + caret), matching what the prompt did before the rewrite. A line that fills the width keeps its EOL caret on-screen instead of having `clip` cut it. Reserving unconditionally rather than only at EOL since mid-line the caret overlays a real character, so the spare column just costs one column of horizontal headroom.

packages/eve/src/cli/dev/tui/line-editor.ts
ruiconti: Fixed in 4949277. `killToEnd` now stops at the next newline (or end of buffer) and `killToStart` at the previous one, so Ctrl+K only clears to the end of the current line (leaving the `\n` and the lines below) and Ctrl+U only back to the start of the current line. Added multi-line tests for both, including the no-op cases at a line's start/end. Left Ctrl+W (delete-word) crossing newlines, since that matches readline.

ruiconti:
Addressed the review bot's setup-panel suggestion in d37d28f: bracketed paste runs for the whole session, so a multi-line paste also hit the single-line setup inputs (agent-name rename, text questions) and inserted literal newlines. `applyLineEditorKey` now takes a `multiline` flag — single-line inputs flatten a paste's newlines to spaces and ignore Shift+Enter; the prompt opts in to keep real newlines.

packages/eve/src/cli/dev/tui/line-editor.ts
```diff
@@ -89,14 +103,35 @@ export function deleteWord(state: LineState): LineState {
...
+  state: LineState,
+  key: TerminalKey,
+  options?: LineEditorOptions,
+): LineState | undefined {
+  const multiline = options?.multiline ?? false;
   switch (key.type) {
     case "character":
       return insert(state, key.value);
+    case "paste":
+      return insert(state, multiline ? key.value : key.value.replaceAll("\n", " "));
```
ijjk, line 159: This text-question path still seems to drop pasted input. Bracketed paste is now enabled for the whole interactive TUI session, so when a user pastes into this field the terminal sends `\x1b[200~...\x1b[201~`, and `nextKey()` decodes that as `{ type: "paste", value }` instead of `{ type: "character", value }`.

The main prompt handles `paste` via `applyLineEditorKey(..., { multiline: true })`, and the setup-flow text/editable inputs handle it via `applyLineEditorKey()` with the default single-line behavior. This older `readInputQuestion` text mode only appends `character`, so a pasted freeform answer silently does nothing. Can this path use the line editor helper too, probably keeping the default single-line behavior so pasted newlines flatten to spaces?

nima54851:
Great to see Vercel enter the Agent framework space! The approach of building agents as first-class primitives in a framework is the right direction.

One question from an AI tooling perspective: how does eve handle **context window management** for long agent conversations? A common pain point in agent frameworks is context overflow in multi-session use cases.

Would love to see a comparison with similar open-source agent frameworks (LangChain, AutoGPT, etc.) in the README — it would help new users understand eve's unique value proposition more clearly.

Keep up the great work!

packages/eve/src/cli/dev/tui/line-editor.ts
ruiconti: I've removed the `"paste"` key type entirely so it is treated as regular text. We may want to revisit this in the future but it's fine for this PR's scope.

ijjk (approved):

PR #81, opened 2026-06-18
fix(harness): keep compacted prompt ending on a user turn

### LINKED ISSUE #80: Compaction can emit a prompt ending on an assistant turn → Anthropic rejects the next model call
### eve version

Observed on `eve@0.8.2`; the affected code path is unchanged on current `main` (`f44e16a`, `packages/eve/src/harness/compaction.ts` — untouched since the initial release).

### Node.js version

`v24.x` (Vercel Functions runtime)

### Where does the bug occur?

- Production (`eve start` / deployed) — direct Anthropic provider (`@ai-sdk/anthropic`)

### Describe the bug

After compaction fires, the very next model call can fail terminally on Anthropic models with:

```
This model does not support assistant message prefill. The conversation must end with a user message.
```

This ends the turn — the agent cannot recover, because every subsequent attempt re-compacts to the same assistant-terminal prompt.

It surfaces on agents that return large tool results (a single oversized tool-result turn, or a long run where `keep` decays to 0) and use a provider without assistant-prefill support (Anthropic).

### Root cause

`compactMessages` already has a trailing-assistant guard, but it only fires when the **kept recent window** trails with an assistant message — it never considers the case where `recent` is empty and the summary's own assistant block becomes the final message.

```ts
// packages/eve/src/harness/compaction.ts
const trailingAssistantGuard: ModelMessage[] =
  recent.length > 0 && recent.at(-1)?.role === "assistant"   // <-- gated on recent.length > 0
    ? [{ role: "user", content: "Continue." }]
    : [];

const compacted: ModelMessage[] = [
  { content: "Summary of our conversation so far:", role: "user" },
  { content: result.text, role: "assistant" },   // <-- becomes the LAST message when recent is empty
  ...recent,
  ...trailingAssistantGuard,
];
```

`recent` is empty whenever `selectRecentWindowSize` returns 0 — i.e. a single turn already exceeds the threshold, or the inner `keep -= 1` loop has decayed `keep` to 0 because even one recent message plus the summary still overflows. In both cases `compacted` ends on the summary's `assistant` message, the guard is skipped (`recent.length === 0`), and Anthropic rejects the prompt.

### Steps to reproduce

1. An agent on a direct Anthropic model (`anthropic("claude-opus-4-8")` / `claude-sonnet-4-6`).
2. A tool that returns a large result (e.g. a wide DB/log query) so one tool-result turn alone exceeds the compaction threshold.
3. Send a prompt that triggers that tool. Compaction folds the oversized turn into the summary, `recent` is empty, and the continuation model call fails with the assistant-prefill error above.

### Impact

Terminal, non-recoverable failure of the turn on Anthropic models — exactly the providers eve documents for direct use. The condition is reached by ordinary large tool outputs, not an edge configuration.

### Proposed fix

Move the guard off `recent` and onto the fully-assembled `compacted` array: append the synthetic `"Continue."` user message whenever the final message is an assistant turn, regardless of whether it came from a trailing recent message or the summary block itself.

```ts
const compacted: ModelMessage[] = [
  { content: "Summary of our conversation so far:", role: "user" },
  { content: result.text, role: "assistant" },
  ...recent,
];

if (compacted.at(-1)?.role === "assistant") {
  compacted.push({ role: "user", content: "Continue." });
}
```

PR with the fix + a regression test (and an updated existing test that had encoded the buggy assistant-terminal shape) attached below.

### Related

Distinct from #79 — that's an `AI_InvalidPromptError` (ModelMessage **schema** validation) on multi-tool-round transcripts thrown by `standardizePrompt` *before* the API call. This one is a provider-side rejection (prompt must end on a user turn) triggered specifically by compaction.

adolfo-tamayo:
Fixes #80.

### Problem

After compaction, the next model call can fail terminally on Anthropic models with *"This model does not support assistant message prefill. The conversation must end with a user message."*

The existing trailing-assistant guard only fires when the **kept recent window** trails with an assistant message (`recent.length > 0 && recent.at(-1)?.role === "assistant"`). It never covers the case where `recent` is empty — a single oversized tool-result turn, or `keep` decaying to 0 — so the summary's own assistant block becomes the final message, the guard is skipped, and Anthropic rejects the prompt. The turn cannot recover, since each retry re-compacts to the same assistant-terminal shape.

### Fix

Move the guard off `recent` and onto the fully-assembled `compacted` array: append the synthetic `"Continue."` user message whenever the final message is an assistant turn, however it got there.

```ts
const compacted: ModelMessage[] = [
  { content: "Summary of our conversation so far:", role: "user" },
  { content: result.text, role: "assistant" },
  ...recent,
];

if (compacted.at(-1)?.role === "assistant") {
  compacted.push({ role: "user", content: "Continue." });
}
```

### Tests

- Adds a regression test: `recentWindowSize: 0` forces an empty recent window, and the compacted result must still end on a user turn.
- Updates the existing `folds oversized recent tool results into the summary...` test, which had asserted the buggy assistant-terminal output; it now expects the trailing `"Continue."` user turn.

`pnpm run test:unit -- compaction` → all green (26 passing).

### DIFF
--- a/.changeset/compaction-trailing-assistant.md
+++ b/.changeset/compaction-trailing-assistant.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix compaction emitting a prompt that ends on an assistant turn, which providers without assistant-prefill support (e.g. Anthropic) reject with "the conversation must end with a user message". The trailing-assistant guard now runs against the fully-assembled compacted prompt instead of only the recent window, so it also covers the case where the recent window is empty (a single oversized turn, or `keep` decaying to 0) and the summary block itself is the final message.
--- a/packages/eve/src/harness/compaction.test.ts
+++ b/packages/eve/src/harness/compaction.test.ts
@@ -251,6 +251,37 @@ describe("compactMessages", () => {
     expect(result[4]).toEqual({ content: "Continue.", role: "user" });
   });
 
+  it("ends on a user turn when no recent messages are kept", async () => {
+    const { generateText } = await import("ai");
+
+    vi.mocked(generateText).mockResolvedValue({
+      text: "Summary of prior context",
+    } as Awaited<ReturnType<typeof generateText>>);
+
+    const messages: ModelMessage[] = [
+      { content: "old message 1", role: "user" },
+      { content: "old message 2", role: "assistant" },
+    ];
+
+    const model = {} as Parameters<typeof compactMessages>[1];
+    // recentWindowSize 0 forces an empty recent window, so the summary's
+    // assistant block would be the final message without the trailing guard.
+    const result = await compactMessages(messages, model, {
+      ...config,
+      recentWindowSize: 0,
+    });
+
+    expect(result.at(-1)).toEqual({ content: "Continue.", role: "user" });
+    expect(result[0]).toEqual({
+      content: "Summary of our conversation so far:",
+      role: "user",
+    });
+    expect(result[1]).toEqual({
+      content: "Summary of prior context",
+      role: "assistant",
+    });
+  });
+
   it("forwards provider options to the compaction model call", async () => {
     const { generateText } = await import("ai");
 
@@ -339,6 +370,10 @@ describe("compactMessages", () => {
       {
         content: "Summary of the large SQL result",
         role: "assistant",
+      },
+      {
+        content: "Continue.",
+        role: "user",
       },
     ]);
     expect(vi.mocked(generateText)).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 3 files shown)


AndrewBarba:
Thanks this looks great. We're finishing some setup on our side to be able to accept contributions. Will get this merged in the next day or two

adolfo-tamayo:
Closing this — #175 (make compaction the sole context manager) landed and already covers it. The new guard treats an empty kept tail (`lastKeptRole === undefined`) the same as a trailing assistant, which is exactly the empty-`recent` case this PR targeted, and the updated tests assert the `Continue.` tail. Good to see it handled more thoroughly as part of that refactor — thanks.

PR #85, opened 2026-06-18
Fix optional microsandbox auto-install reload

ijjk:
## Summary
- serialize optional sandbox engine auto-installs so concurrent first-use requests do not race
- after auto-install, load the installed engine through its package entrypoint file instead of retrying the cached bare specifier
- add EVE_INIT_PACKAGE_SPEC for local tarball/source validation of generated projects

## Verification
- pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/internal/application/optional-package-install.test.ts
- pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/cli/commands/init.integration.test.ts
- pnpm --filter eve test:integration
- pnpm --filter eve typecheck
- reduced fresh Bun project repro with repacked tarball

### DIFF
--- a/packages/eve/src/cli/commands/init.ts
+++ b/packages/eve/src/cli/commands/init.ts
@@ -22,6 +22,10 @@ import {
 } from "#setup/primitives/index.js";
 import { addAgentToProject } from "#setup/scaffold/create/add-to-project.js";
 import { ensureChannel, scaffoldBaseProject } from "#setup/scaffold/index.js";
+import {
+  DEFAULT_EVE_PACKAGE_CONTRACT,
+  type EvePackageContract,
+} from "#setup/scaffold/create/project.js";
 
 import { initAgentDevHandoff } from "./agent-instructions.js";
 import { tryInitializeGit } from "./init-git.js";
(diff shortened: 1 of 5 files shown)


packages/eve/src/cli/commands/init.ts
```diff
@@ -159,12 +170,16 @@ async function scaffoldProject(
...
+    const scaffoldOptions: Parameters<typeof scaffoldBaseProject>[0] = {
       projectName: stagedProjectName,
       model: DEFAULT_AGENT_MODEL_ID,
       packageManager,
       targetDirectory: stagingDirectory,
-    });
+    };
+    if (evePackage !== undefined) {
+      scaffoldOptions.evePackage = evePackage;
+    }
```
ruiconti: @AndrewBarba would like a word here 😁

```suggestion
    const scaffoldOptions = {
        projectName: stagedProjectName,
        model: DEFAULT_AGENT_MODEL_ID,
        scaffoldOptions.evePackage: evePackage,
        targetDirectory: stagingDirectory,
        packageManager,
    };
```

AndrewBarba (approved):

PR #88, opened 2026-06-18
Fix model version format in agent configuration

asnewman:
`model: anthropic("claude-opus-4.8")` results in the following error message:

```
Error
  MODEL_CALL_FAILED: model: claude-opus-4.8 was not found. Did you mean claude-opus-4-8?
```

Switching to `4-8` fixes it.


AndrewBarba:

PR #90, opened 2026-06-18
feat(eve): report microsandbox create progress and enrich failures

ruiconti:
When a microsandbox sandbox is slow to come up, all you see is the prewarm waiter printing `waiting for sandbox template prewarm to finish (Ns elapsed)`. The work behind it — `Sandbox.create()` pulling the image and booting the VM — runs with no progress and no timeout, so a slow pull or a wedged boot looks identical to a hang. On failure you get the raw SDK error with no hint.

**What**
- Replace the bare `Sandbox.create()` with `createWithPullProgress()` and report progress.
- Wrap create and prewarm failures with the phase that failed and a recovery hint.
- Add a repro script for the stall.

**Progress**
`createMicrosandboxWithProgress` maps pull events to four phases — resolving the image, downloading + materializing layers, assembling the filesystem, booting the VM — and logs each transition. A 10s heartbeat reprints the active phase with elapsed time, so the boot phase (which emits no pull events) is never silent. It reuses the existing prewarm log callback, so progress shows up inline with `creating template VM…`.

**Errors**
`enrichMicrosandboxError` adds the failing phase and a code-keyed hint:
- `database` → runtime binary and npm package are on different versions; use a clean `MSB_HOME` or migrate.
- `imageNotFound` → check the image and that registry credentials can pull it.
- `libkrunfwNotFound` → run `npx microsandbox install`.

It reads `code` off `MicrosandboxError` when the class is in hand, otherwise duck-types — but only trusts a `code` it has a hint for, so a stray `ENOENT` during prewarm isn't mislabeled as a microsandbox code. `prewarm()` failures get the same wrapper.

**Repro**
`pnpm --filter eve repro:microsandbox-create-stall` holds the prewarm lock with a create that stalls 32s, then asserts both sides: the waiter prints its heartbeats and the create prints `image ready; booting microsandbox VM (30s elapsed)`.

**Not here**
No create timeout, and no liveness check on the prewarm lock — a dead prewarmer still makes waiters wait out the 15-minute lock timeout. That's the next step; this makes the slow/failing path legible first.

### DIFF
--- a/.changeset/clear-microsandbox-prewarm.md
+++ b/.changeset/clear-microsandbox-prewarm.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Report image-pull and VM-boot progress during microsandbox creation, and include phase and provider-specific recovery guidance when prewarm fails.
--- a/packages/eve/src/execution/sandbox/bindings/microsandbox-create.ts
+++ b/packages/eve/src/execution/sandbox/bindings/microsandbox-create.ts
@@ -1,0 +1,166 @@
+import { toErrorMessage } from "#shared/errors.js";
+import type {
+  MicrosandboxError,
+  MicrosandboxErrorCode,
+  PullProgressEvent,
+  Sandbox as MicrosandboxSandbox,
+  SandboxBuilder as MicrosandboxSandboxBuilder,
+} from "microsandbox";
+
+type MicrosandboxCreatePhase =
+  | "resolving the image"
+  | "downloading and materializing image layers"
+  | "assembling the image filesystem"
+  | "image ready; booting microsandbox VM";
+
+export async function createMicrosandboxWithProgress(input: {
+  readonly builder: MicrosandboxSandboxBuilder;
+  readonly errorType: typeof MicrosandboxError;
+  readonly log?: (message: string) => void;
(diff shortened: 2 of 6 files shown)


ijjk (dismissed):

packages/eve/package.json
```diff
@@ -272,6 +272,7 @@
     "check:bin-runtime-dependencies": "node ./scripts/check-bin-runtime-dependencies.mjs",
     "dev": "pnpm run build:compiled && node ./scripts/copy-runtime-assets.mjs && tsgo -p tsconfig.dev.json --watch --preserveWatchOutput",
     "prepack": "pnpm run build && node ../../scripts/copy-package-license.mjs .",
+    "repro:microsandbox-create-stall": "pnpm run build:js && node ./scripts/reproduce-microsandbox-create-stall.mjs",
```
AndrewBarba: Do we need a new top level script like this?

AndrewBarba (changes requested):

ijjk (approved):

packages/eve/package.json
ruiconti: It's gone

AndrewBarba (approved):

PR #96, opened 2026-06-18
Add pip to sandbox Docker image

ijjk:
## Summary
- install python3-pip in the sandbox Docker image
- verify pip and pip3 during the image build alongside the existing Python checks

## Why pip instead of uv?
The image already includes Python, so python3-pip is the smallest standard package-installer addition.

Local arm64 image-size measurement:
- python3-pip: about +11.4 MiB
- optimized uv/uvx: about +48.0 MiB

## Verification
- docker build -t eve-add-pip-check -f Dockerfile .
- docker run --rm eve-add-pip-check bash -lc 'python --version; python3 --version; pip --version; pip3 --version'

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,6 +27,7 @@ RUN set -eux; \
     jq \
     python-is-python3 \
     python3 \
+    python3-pip \
     ripgrep \
     sudo \
     unzip \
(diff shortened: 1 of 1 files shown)


ruiconti (approved):
> The image already includes Python, so python3-pip is the smallest standard package-installer addition.

Sounds like a safe default. It is the most used[^1] package manager (though `uv` is very fast)

[^1]: https://wagtail.org/blog/uv-overtakes-pip-in-ci/

PR #102, opened 2026-06-19
fix(cli): skip Next.js, Nuxt, and SvelteKit build directories during dev snapshot copying

crossxr:
### Description

During local development running `eve dev` or `npm run dev`, the CLI recursively copies the source workspace to an immutable snapshot directory under **.eve/dev-runtime/snapshots/<session-id>/source/.**

When working in framework environments like Next.js, Nuxt, or SvelteKit, the CLI attempts to scan and recursively copy the transient build/cache directories (e.g. .next/, .nuxt/, .svelte-kit/). Because these frameworks dynamically create, modify, and delete files during startup and builds, a race condition often occurs: a file (like .next/lock) is deleted by the framework after the Eve CLI reads the directory contents but before it runs lstat or cp on that path. This causes the CLI copying logic to throw an ENOENT error, crashing the dev server startup.

**Solution:** Added .next, .nuxt, and .svelte-kit framework build/cache directories to the` SNAPSHOT_SKIP_NAMES` set in packages/eve/src/internal/nitro/dev-runtime-source-snapshot-copy.ts so they are bypassed during snapshot copying.

Also updated the **@vercel/eve-catalog** package clean script from rm -rf dist to a cross-platform command `(node -e "fs.rmSync(...)")` to allow project typechecking/building to succeed on Windows.

1. Added an integration test to `packages/eve/src/internal/nitro/dev-runtime-artifacts.integration.test.ts` verifying that .next, .nuxt, and .svelte-kit directories are skipped during staging.

2. Verified all integration tests pass:
```
bash
pnpm --filter eve exec vitest run --config vitest.integration.config.ts dev-runtime-artifacts
```
3. Verified the codebase typechecks, lints, formats, and invariant guards pass:
```
bash
pnpm typecheck
pnpm lint
pnpm fmt
pnpm guard:invariants
```

### PR Checklist


AndrewBarba:

crossxr:
Hey @AndrewBarba - just wanted to follow up on this. Looks like #181 was merged this week fixing the downstream GC issue. This PR addresses the upstream side, preventing framework` build dirs` from entering the snapshot in the first place, which also reduces the frequency of spurious rebuilds that trigger that GC path (noted in #172). Happy to rebase or make any changes needed.

samprice393:
+1 — hit this independently on a Next.js app, with a failure mode worth adding alongside the `ENOENT` race described here: **disk exhaustion**.

eve keeps multiple dev snapshots, and each one recursively copied the full `.next` build cache (~16 GB on this project). They stacked up until `.eve/dev-runtime` reached ~37 GB and the copy died mid-way with `ENOSPC: no space left on device`, which crashes the eve sidecar and takes the whole dev server down (exit 1):

```
Failed to copy development runtime source snapshot path ".../.next/dev/cache/turbopack/....sst" ...: ENOSPC: no space left on device
Error: eve server process exited before printing its server URL (code 1, signal null).
```

Adding `.next` to `SNAPSHOT_SKIP_NAMES` (exactly what this PR does) fixes that too. After the change on eve@0.11.10:

- per-snapshot size: **~7 GB → ~90 MB**
- `.eve/dev-runtime` total: **37 GB → ~178 MB**
- authored source (`app/`, `components/`, …) still snapshotted correctly

Same root cause as the race, just a second symptom. Would be great to see this merged. Thanks @crossxr!

byronogis:
In addition to some built-in default values, it is expected to support user input (override or increment), such as via environment variables

De-ASI-INTERFACE (approved):

cmpadden:
Closing as superseded by https://github.com/vercel/eve/pull/2262 but please re-open if you feel necessary.

Thank you for the contribution.

PR #105, opened 2026-06-19
fix(eve): resolve dev model module map from authored source

### LINKED ISSUE #104: Handler-form schedule → channel `receive()` fails in `eve dev` with `LoadCompiledModuleMapError`
### eve version

`eve@0.11.6`

### Node.js version

`v24.15.0`

### Where does the bug occur?

Local dev (`eve dev`)

### Describe the bug

A schedule with a `run` handler that imports an authored channel crashes the agent's first turn under `eve dev`:

```
Step "step//eve@0.11.6//turnStep" failed after 3 retries:
Cannot find module '…/.eve/dev-runtime/snapshots/<id>/source/agent/channels/probe.js'
imported from '…/.eve/dev-runtime/snapshots/<id>/source/agent/schedules/digest.ts'
LoadCompiledModuleMapError
  at loadCompiledModuleMap → loadSourceBackedRuntimeModelReference
  → resolveRuntimeModelReference → turnStep
```

Builds and deployed runs are fine. It isn't specific to channels: any module in the compiled module map (a `run`-handler schedule, a tool) that imports another authored module with a `.js` specifier hits this. No credentials needed; it fails during model resolution, before any delivery.

Root cause: `loadSourceBackedRuntimeModelReference` (`runtime/agent/resolve-model.ts`) calls `loadCompiledModuleMap`, whose disk branch does a raw `import()` of `module-map.mjs`. That pulls in the `hasRun` schedule and runs its `import "../channels/probe.js"`. The raw import skips eve's authored-module loader, so the `.js` specifier never maps to the snapshot's `.ts` file. Dev model resolution should use the authored-source loader here, like the rest of the dev runtime. I have a fix and a regression test; PR to follow.

### Steps to reproduce

`npx eve@latest init eve-repro && cd eve-repro`, then:

`agent/agent.ts` (a model instance, not a string id, so resolution loads the module map):

```ts
import { gateway } from "ai";

export default defineAgent({
  model: gateway("openai/gpt-5.5"),
  modelContextWindowTokens: 200_000, // avoids a gateway lookup at compile
});
```

`agent/channels/probe.ts` (minimal custom channel, no credentials):

```ts
import { defineChannel, POST } from "eve/channels";

export default defineChannel({
  routes: [POST("/probe", async () => Response.json({ ok: true }))],
  async receive(input, { send }) {
    return send(input.message, { auth: input.auth, continuationToken: "probe" });
  },
});
```

`agent/schedules/digest.ts` (`run` handler importing the channel):

```ts
import { defineSchedule } from "eve/schedules";
import probe from "../channels/probe.js";

export default defineSchedule({
  cron: "0 0 * * *",
  async run({ receive, waitUntil, appAuth }) {
    waitUntil(receive(probe, { message: "ping", target: { sessionRef: "probe" }, auth: appAuth }));
  },
});
```

Then run `eve dev` and trigger a turn:

```bash
curl -X POST http://localhost:2000/eve/v1/dev/schedules/digest
```

The session fails with the `LoadCompiledModuleMapError` above.

### Expected behavior

The turn resolves the model and runs, the same as a build.

oksr:
### Description

Fixes #104.

Under `eve dev`, model resolution loaded the compiled module map with the raw loader (`loadCompiledModuleMap`). That loader statically re-imports authored entry modules, so a `run`-handler schedule importing a channel via a `.js` specifier crashed the turn with `LoadCompiledModuleMapError` (native ESM can't map the `.js` specifier to the dev snapshot's `.ts` source). Builds and production were unaffected.

The fix routes model resolution through the runtime's existing `loadRuntimeCompiledModuleMap` (now exported from `compiled-agent-cache.ts`), which already hydrates source-backed/dev runs through the authored-source loader and falls back to the raw loader for bundled/production. One call swapped; production path unchanged.

- Added `packages/eve/test/scenarios/schedule-channel-model-resolution.scenario.test.ts`: spawns a real `eve dev` subprocess (with real, unmocked model resolution) for an agent whose `run`-handler schedule imports a channel, then asserts a turn starts and resolution does not raise `LoadCompiledModuleMapError`. It fails before the fix with the exact error and passes after. (In-process tests can't catch this: vitest's resolver maps `.js`→`.ts`, and the model mock seam skips the load.)
- Manually verified against a real agent running this branch's build: a `run`-handler schedule that delivers to a Telegram channel failed with `LoadCompiledModuleMapError` under `eve dev` before the fix, and after it the session resolved the model, ran the turn, and delivered the message.
- `pnpm typecheck`, `pnpm lint`, `pnpm guard:invariants`, `pnpm test:unit`, `pnpm test:integration`, `pnpm test:scenario` all pass.

### PR Checklist

### DIFF
--- a/.changeset/dev-schedule-channel-model-resolution.md
+++ b/.changeset/dev-schedule-channel-model-resolution.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix `eve dev` failing with `LoadCompiledModuleMapError` when resolving an agent's model for an app that has a `run`-handler schedule (or any module-map entry) which statically imports an authored channel. Runtime model resolution now loads the dev module map through the authored-source hydrator, matching the rest of the dev runtime, so authored `.js` import specifiers resolve to their `.ts` sources. Production builds are unaffected.
--- a/packages/eve/src/runtime/agent/resolve-model.ts
+++ b/packages/eve/src/runtime/agent/resolve-model.ts
@@ -8,7 +8,7 @@ import {
 } from "#internal/authored-module.js";
 import type { ModuleDefinitionExport } from "#public/definitions/source.js";
 import type { RuntimeCompiledArtifactsSource } from "#runtime/compiled-artifacts-source.js";
-import { loadCompiledModuleMap } from "#runtime/loaders/module-map.js";
+import { loadRuntimeCompiledModuleMap } from "#runtime/sessions/compiled-agent-cache.js";
 import type { RuntimeModelReference } from "#runtime/agent/bootstrap.js";
 import { resolveBootstrapRuntimeModel } from "#runtime/agent/bootstrap-model.js";
 import {
@@ -60,9 +60,7 @@ async function loadSourceBackedRuntimeModelReference(
     );
   }
 
-  const moduleMap = await loadCompiledModuleMap({
-    compiledArtifactsSource: input.compiledArtifactsSource,
-  });
+  const moduleMap = await loadRuntimeCompiledModuleMap(input.compiledArtifactsSource);
   const moduleNamespace =
     moduleMap.nodes[ROOT_COMPILED_AGENT_NODE_ID]?.modules[reference.source.sourceId];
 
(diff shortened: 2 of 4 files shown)


AndrewBarba:

oksr:
Superseded by #156 (shipped in 0.13.1), which fixes the same `LoadCompiledModuleMapError` in source-backed model resolution by reading the model from the already-hydrated bundle module map instead of re-loading it via the raw loader. Closing.

PR #115, opened 2026-06-19
[codex] [EXPERIMENT] add bounded sandbox file downloads

remiconnesson:
## TL;DR

Adds an Eve-native, backend-independent path for handing bounded sandbox files to browser users.

## What changed

- adds the default `download_file` tool for files under `/workspace`, capped at 1 MiB
- validates the payload and projects it as a typed `EveFilePart` while keeping base64 out of model and TUI output
- renders downloadable file cards in the default Web Chat template
- documents the limit and large-artifact fallback, with unit, scenario, browser, TUI, and live eval coverage

## Why

Eve could create and read sandbox files, but it had no framework-native way to hand one to an end user. Carrying a small file through the authenticated session stream keeps the behavior portable across sandbox backends and durable after sandbox suspension without introducing artifact storage.

## Checks

- `pnpm test:unit`
- `pnpm test:integration`
- `pnpm test:scenario`
- `pnpm test:tui`
- `pnpm typecheck`
- `pnpm build`
- `pnpm docs:check`
- `pnpm lint`
- `pnpm exec oxfmt --check`
- `pnpm guard:invariants`
- `pnpm check:deps`
- `MSB_HOME=/tmp/eve-msb-download-test pnpm --filter agent-tools-sandbox exec eve eval --strict --skip-report sandbox/download-file`
- browser verification of rendered filename, `download` attribute, and downloaded bytes


remiconnesson:
closing: slop

PR #117, opened 2026-06-19
fix(eve): fix Vercel Sandbox file reads

### LINKED ISSUE #120: Vercel Sandbox readFile violates Eve’s Web stream contract
## Summary

The Vercel Sandbox backend can return a Node.js readable stream from `SandboxSession.readFile()`, even though Eve's public sandbox contract exposes a Web `ReadableStream<Uint8Array>`.

Code using the documented Web stream API therefore fails for valid, readable files.

## Reproduction

1. Create an Eve session using the Vercel Sandbox backend.
2. Write a regular file inside `/workspace`.
3. Call `session.readFile({ path })`.
4. Consume the result with `stream.getReader()`.

The file can be listed and read through sandbox commands, but the returned value does not implement the Web stream API.

## Expected behavior

`SandboxSession.readFile()` returns `ReadableStream<Uint8Array> | null` consistently across every backend.

## Actual behavior

The Vercel backend passes through the provider's Node.js readable stream. Consumers fail with an error such as:

```text
TypeError: stream.getReader is not a function
```

Applications may then incorrectly report the existing file as unreadable.

## Root cause

Eve's vendored Vercel Sandbox declaration describes `readFile()` as returning a Web `ReadableStream`, while the provider SDK returns `NodeJS.ReadableStream`. The backend trusts the vendored declaration and does not adapt the provider value to Eve's public contract.

## Proposed fix

Normalize the provider response at the Vercel backend boundary:

- pass through an existing Web `ReadableStream`
- convert a Node.js `Readable` with `Readable.toWeb()`
- preserve `null` for missing files
- propagate provider errors unchanged
- reject unsupported response shapes explicitly

Add a regression test that consumes the result through Eve's public `SandboxSession.readFile()` API.

Related fix: #117.

remiconnesson:
## Summary

- adapt Vercel Sandbox's Node file stream to Eve's `ReadableStream<Uint8Array>` contract
- correct Eve's vendored Vercel Sandbox declaration to match the provider SDK
- add a regression test that consumes the returned stream through Eve's public `SandboxSession.readFile()` API

## Root cause

Vercel Sandbox returns a Node.js readable stream from `readFile()`. Eve's vendored declaration described that value as a Web `ReadableStream`, and the Vercel backend passed it through unchanged.

Consumers therefore received a value without `getReader()`. Download helpers mapped that resulting `TypeError` to a generic unreadable-file error even though the file existed and was readable inside the sandbox.

## User impact

Vercel-backed Eve agents can stream existing sandbox files through `SandboxSession.readFile()`, including bounded download flows.

Fixes #120.

## Verification

- regression test failed before the fix because the returned value was not a `ReadableStream`
- 3,631 unit tests pass, including the 700-line structural guard
- Eve typecheck, formatting, and lint pass

Also tried the same kind of change locally [here](https://github.com/remiconnesson/eve-single-user-agent-starter/pull/17), and it worked 
image


remiconnesson:
## Vercel `readFile()` verification (`1f3f8cc`)

The regression was reproduced before the fix with a Node `Readable` returned by the Vercel SDK. Eve passed that value through, and the test failed because it was not an instance of `ReadableStream`—the same condition that makes the starter's bounded reader fail at `stream.getReader()`.

After the fix:

- the provider Node stream is converted at the Vercel adapter boundary
- Eve's public `SandboxSession.readFile()` returns a Web `ReadableStream`
- the regression consumes the stream and matches the original bytes
- provider errors are not caught or replaced by the adapter

The real Vercel Sandbox pass also traversed this adapter successfully: 45 bytes matched with SHA-256 `cee7126090710957909be4c25dfff28428432d7416752a2baa4d4f4066400557`. Sandbox `pr117-vercel-mqlb9eu3-038458` was then deleted and lookup confirmed it was absent.

Current branch verification:

- 3,631 unit tests passed
- focused Vercel suite: 38 tests passed
- typecheck, lint, and formatting passed
- `vercel.ts` remains at the repository's 700-line structural limit

Reproduction: remiconnesson/eve-single-user-agent-starter#18.

ctgowrie (commented):
Let's see if readFile solves the requirements now to keep surface area small

remiconnesson:
Follow-up documentation is in #121. It documents the existing `readFile()` Web-stream contract, reader selection, and frontend rendering for generated images and file downloads. The docs PR targets `main` separately and should merge after this runtime fix.

PR #118, opened 2026-06-19
feat(eve): cancel active turns server-side

remiconnesson:
## Summary

- add authenticated per-turn cancellation tokens and a server cancellation route
- propagate the active turn `AbortSignal` through model generation and authored tool execution
- retry server-side workflow-hook registration without client polling
- keep the durable parent session waiting and ready for the next turn after cancellation

## Stack

1. #118 — core server cancellation
2. #127 — client, framework, and TUI controls
3. #128 — delegated local and remote runtime-action cancellation

This PR was reduced to the core server primitive so each layer can be reviewed and landed independently.

## Validation

- `pnpm test:unit` (3,639 tests)
- `pnpm test:integration` (318 tests)
- `pnpm build`

### DIFF
--- a/.changeset/native-turn-cancellation.md
+++ b/.changeset/native-turn-cancellation.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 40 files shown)


stellanhaglund:
Any progress on this?

PR #130, opened 2026-06-20
restore sandbox network policy after github actions

caleb-vercel:
### Description

`checkoutGitHubRepository` brokers the GitHub App installation token at the sandbox firewall — it swaps the session's network policy for one that injects an `Authorization` header on egress to `github.com` / `codeload.github.com`, so the token never enters the sandbox process. It then never puts the old policy back.

Two consequences, for the rest of the session:

- the token transform stays attached to every later request the agent makes to GitHub, well past the fetch it was minted for;
- the broker policy's `"*": []` rule leaves general egress open, so a session an author had locked down (`deny-all`, or a narrow allowlist) silently comes back wide open after the first checkout — and the agent can then fetch more code.

This PR scopes the broker policy to the fetch window: read the session's policy up front, restore it in a `finally` so a failed fetch can't strand the sandbox in the widened state either.

Reading that prior policy needs a session that tracks it, so `buildSandboxSession` now takes a `SandboxNetworkPolicyRef` — seeded with the policy the sandbox was created under, updated on every *accepted* `setNetworkPolicy` — and `SandboxSession` exposes `getNetworkPolicy()`. The Vercel and microsandbox handles share one session and one ref with their `useSessionFn`, which applies `use({ networkPolicy })` straight to the provider SDK and so has to write the ref itself.

Adding a method to the sandbox surface reachable from `ctx.getSandbox()` bumps the `tool`, `dynamicTool`, `connection`, `hook`, and `state` capability epochs. Each previous epoch is retained (the addition is structurally backward compatible for authors) with a compatibility fixture, mirroring #1801, which added `stop()` to the same surface.

- New unit tests: policy restored after a successful checkout, restored after a failed fetch, and restored to a *restrictive* prior policy rather than assumed `allow-all`.
- New `buildSandboxSession` tests for the seeded ref, the post-`setNetworkPolicy` value, the shared-ref path, and that a rejected apply (just-bash) leaves the tracked policy untouched.
- `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm check:deps`, `pnpm docs:check`, `pnpm guard:invariants`, `pnpm build`, `pnpm test:unit` (6136), `pnpm test:integration` (610), `pnpm test:scenario`.

### PR Checklist


caleb-vercel:
Uhh gonna start over on this

PR #140, opened 2026-06-20
fix(eve): resolve code-defined model imported from a sibling module in dev

### LINKED ISSUE #92: `eve dev` fails every turn with LoadCompiledModuleMapError when the agent uses a code-defined (direct-provider) model
**Environment:** eve `0.11.6` (also repro'd on `0.11.3`) · Node 24 · macOS · `@ai-sdk/anthropic` `4.0.0-beta.67` / `ai` `7.0.0-beta.178`

## Summary

When the agent's `model` is a **code-defined provider object** (the documented direct-provider path, e.g. `model: anthropic("claude-sonnet-4-6")` from `@ai-sdk/anthropic`), **every turn under `eve dev` fails** during `turnStep` with a `LoadCompiledModuleMapError`. The same app **builds and runs fine** via `eve build` + `eve start`. Switching `model` to a **gateway string** (`"anthropic/claude-sonnet-4.6"`) makes `eve dev` work again.

## Error

```
Step "step//eve@0.11.6//turnStep" failed after 3 retries: Cannot find module
'.../.eve/dev-runtime/snapshots/<id>/source/.../agent/lib/helper.js'
imported from '.../agent/tools/ping.ts'
```

Stack (abridged): `loadCompiledModuleMap` ← `loadSourceBackedRuntimeModelReference` ← `resolveRuntimeModelReference` ← `turnStep`.

The failing import is an ordinary NodeNext `.js` specifier between authored modules — a tool importing a local `lib/` helper, or a schedule importing a channel. The `.ts` files are present in the dev snapshot; the loader looks for the `.js` and doesn't map it to the `.ts`.

## Root cause (from eve's runtime)

`resolveRuntimeModelReference` routes to `loadSourceBackedRuntimeModelReference` when `isSourceBackedRuntimeModelReference(e)` is true, and that predicate is:

```js
function isSourceBackedRuntimeModelReference(e){ return e.source !== void 0 }
```

A **code-defined** model reference carries `.source` (the authored module that defines it); a **gateway string** does not (the resolver `return`s the bare id). So only code-defined models reach the source-backed loader → `loadCompiledModuleMap`, which does `await import(moduleMapPath)` against the **`.ts`-only `eve dev` snapshot** without applying the NodeNext `.js`→`.ts` mapping. `eve build` (rolldown) produces bundled artifacts and resolves correctly, which is why production / `eve start` is unaffected.

Relevant source: `packages/eve/src/runtime/agent/resolve-model.ts` (`resolveRuntimeModelReference` / `isSourceBackedRuntimeModelReference` / `loadSourceBackedRuntimeModelReference`) and `packages/eve/src/runtime/loaders/module-map.ts` (`loadCompiledModuleMap`).

## Minimal repro

1. `eve init` a fresh app (or any app with one tool that imports a sibling `lib/` helper via a `.js` specifier).
2. Set `agent/agent.ts` → `model: anthropic("claude-sonnet-4-6")` (direct provider).
3. `eve dev`, send any message → `LoadCompiledModuleMapError: Cannot find module '.../lib/helper.js' imported from '.../tools/<tool>.ts'`.
4. Change `model` to the gateway string `"anthropic/claude-sonnet-4.6"` → the same message now runs (no module error).
5. With the direct provider, `eve build` + `eve start` also run fine.

Happy to provide a standalone minimal repo if useful.

## Related (not duplicates)

Other direct-provider reports, all describing different symptoms: #27 (closed — provider package not pre-installed), #70 (open — Vercel-login warning under `eve dev`), Discussion #37 (direct providers working; `modelContextWindowTokens`). None describe this runtime module-map / `.js`-specifier failure.

## Expected

`eve dev` should resolve authored-module `.js` imports for code-defined models the same way `eve build` does — or the docs should state that code-defined / direct-provider models aren't supported under `eve dev`.

## Workarounds

- Serve the build: `eve build` + `eve start --host 127.0.0.1` (`localDev()` authorizes loopback, so localhost turns work with no auth change).
- Or use a gateway model string for `eve dev`.

dang-w:
Fixes #92.

## Problem

`eve dev` (and therefore `eve eval`, which runs through the dev runtime) fails at turn
time with `LoadCompiledModuleMapError` when an agent's `model` is a **code-defined provider
object** (e.g. `anthropic("claude-opus-4-8")`) imported into the agent config from a
**sibling module**:

```
LoadCompiledModuleMapError: Cannot find module '.../source/agent/lib/models.js'
  imported from '.../source/agent/agent.ts'
```

`eve build` + `eve start` resolve the same agent fine; only the dev runtime is affected.

## Root cause

A non-string `model` can't be serialised into durable state, so it becomes a _source-backed
model reference_ that is re-imported at turn time to read `.model` back out
(`resolve-model.ts` → `loadSourceBackedRuntimeModelReference`). That path called the basic
`loadCompiledModuleMap` **directly**, which imports `module-map.mjs` and re-imports the
authored config from raw `.ts` source — without the NodeNext `.js`→`.ts` specifier mapping.

Meanwhile the normal agent-loading path (`compiled-agent-cache.ts`) already branches on
`moduleMapLoaderPath` and uses the **authored-source loader** (`loadAuthoredModuleNamespace`)
in dev, which resolves those specifiers correctly. The two paths diverged; the model
resolver used the wrong one.

## Fix

Extract that branching logic into a single shared `loadRuntimeCompiledModuleMap` in
`runtime/loaders/module-map.ts`, and route **both** the compiled-agent cache and source-backed
model resolution through it. So a code-defined model imported from a sibling module now
resolves identically in dev and production. Net change is a small consolidation
(+49/−39) that removes the duplicated loader.

## Verification

- **Repro fixed end-to-end.** A fresh `eve init` app with `export const model =
anthropic("claude-opus-4-8")` in `agent/lib/models.ts`, imported by `agent/agent.ts`, fails
  on `eve eval` before this change. After it, model resolution succeeds and the run proceeds to
  the actual model call (no more `LoadCompiledModuleMapError`).
- **Regression test added** (`runtime/agent/resolve-model.test.ts`): asserts source-backed
  model resolution loads the module map through `loadRuntimeCompiledModuleMap` (the
  authored-source-aware loader), not `loadCompiledModuleMap`. It fails on the previous code and
  passes on the fix. (A note on why it's a unit test: the runtime failure only reproduces under
  real Node — vitest's loader transparently remaps `.js`→`.ts`, so an end-to-end scenario test
  passes either way and can't guard this regression. The unit test pins the loader the resolver
  must use.)
- **No regressions.** `typecheck`, `build`, and the existing scenario + integration tests for
  the authored-source loading path (`authored-module-loader.scenario`, `cache-key.scenario`,
  `dev-runtime-artifacts.integration`) all pass.

### DIFF
--- a/.changeset/fix-dev-source-backed-model-resolution.md
+++ b/.changeset/fix-dev-source-backed-model-resolution.md
@@ -1,0 +1,12 @@
+---
+"eve": patch
+---
+
+Fix `eve dev` and `eve eval` failing with `LoadCompiledModuleMapError` when an
+agent's `model` is a code-defined provider object (e.g. `anthropic("claude-opus-4-8")`)
+imported into the agent config from a sibling module. Source-backed model
+resolution now loads the compiled module map through the same authored-source
+loader the rest of the dev runtime uses, so NodeNext `.js` import specifiers
+between authored source files resolve to their `.ts` sources. Previously this
+path imported the compiled module map directly and skipped that mapping, which
+only surfaced in the dev runtime (`eve build` + `eve start` were unaffected).
--- a/packages/eve/src/runtime/agent/resolve-model.test.ts
+++ b/packages/eve/src/runtime/agent/resolve-model.test.ts
@@ -1,0 +1,85 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// The module-map loader is mocked so we can assert *which* loader source-backed
+// model resolution uses without standing up a compiled app on disk.
+const { loadRuntimeCompiledModuleMap, loadCompiledModuleMap } = vi.hoisted(() => ({
+  loadCompiledModuleMap: vi.fn(),
+  loadRuntimeCompiledModuleMap: vi.fn(),
+}));
+
+vi.mock("#runtime/loaders/module-map.js", () => ({
+  loadCompiledModuleMap,
+  loadRuntimeCompiledModuleMap,
+}));
+
+import { ROOT_COMPILED_AGENT_NODE_ID } from "#compiler/manifest.js";
+import type { RuntimeModelReference } from "#runtime/agent/bootstrap.js";
+import type { RuntimeCompiledArtifactsSource } from "#runtime/compiled-artifacts-source.js";
+import { resolveRuntimeModelReference } from "#runtime/agent/resolve-model.js";
+
+describe("resolveRuntimeModelReference", () => {
+  beforeEach(() => {
+    // The deterministic authored-model mock activates under NODE_ENV=test and
+    // would short-circuit before the source-backed path; bypass it here.
+    vi.stubEnv("NODE_ENV", "production");
+    loadRuntimeCompiledModuleMap.mockReset();
+    loadCompiledModuleMap.mockReset();
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
+
+  it("resolves a source-backed model through the authored-source-aware module-map loader", async () => {
+    // Regression (vercel/eve#92): a code-defined model imported into the agent
+    // config from a sibling module is a source-backed reference, re-imported at
+    // turn time. That re-import must go through loadRuntimeCompiledModuleMap —
+    // which uses the authored-source loader in dev and so resolves NodeNext
+    // `.js` specifiers between authored source files — not loadCompiledModuleMap,
+    // which imports the compiled module-map directly and fails to resolve them
+    // in the dev runtime.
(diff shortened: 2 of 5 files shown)


ruiconti:
@dang-w thanks for the PR! This patch does fix the problem but still incurs an unnecessary dynamic import at every step, which is unnecessary given that the graph has already been loaded. Fixed in #156.

dang-w:
Ahhh, nice - thanks @ruiconti , no worries! 

Took a look at #156 - cleaner solution 🎉

PR #163, opened 2026-06-21
fix(eve): set framework preset only for standalone projects

allenzhou101:
## Summary

Follow-up to #151: set the Eve framework preset only when creating standalone Eve projects.

If the root app already declares a host framework that Eve integrates with today (Next.js, Nuxt, or SvelteKit), the new Vercel project leaves `framework` unset so Vercel can auto-detect the top-level app framework on deploy. Eve still runs as the sibling service through the generated `vercel.json` / Build Output config.

Detector alignment was checked against `vercel/vercel`: the CLI project setup path uses `detectProjects()` from `@vercel/fs-detectors` over `@vercel/frameworks`; `matchPackage` detection checks the package root manifest. This PR mirrors the relevant package detectors for Next.js (`next`), Nuxt (`nuxt`, `nuxt3`, `nuxt-edge`, `nuxt-nightly`), and SvelteKit (`@sveltejs/kit`) without inferring a framework from hoisted `node_modules` or unrelated sibling package manifests.

## Verification

- `fnm exec --using=24 pnpm --filter eve run typecheck`
- `fnm exec --using=24 pnpm --filter eve run test:integration -- src/setup/scaffold/index.integration.test.ts`
- `fnm exec --using=24 pnpm --filter eve run test:unit -- src/setup/vercel-project.test.ts`
- `git diff --check`
- Real web-chat smoke from the local Eve package:
  - Project creation left `framework` absent before deploy.
  - Deployment detected Next.js, completed successfully, and project metadata reported `framework: "nextjs"` after deploy.
  - A control smoke with forced `framework=eve` failed after `next build` with missing `.output`, confirming why host-framework apps must remain unset.

### DIFF
--- a/packages/eve/src/setup/scaffold/update/channels.ts
+++ b/packages/eve/src/setup/scaffold/update/channels.ts
@@ -31,6 +31,14 @@ const DEFAULT_TYPES_REACT_PACKAGE_VERSION = "__TYPES_REACT_VERSION__";
 const DEFAULT_TYPES_REACT_DOM_PACKAGE_VERSION = "__TYPES_REACT_DOM_VERSION__";
 const CONNECT_PACKAGE_NAME = "@vercel/connect";
 const NEXT_PACKAGE_NAME = "next";
+const VERCEL_HOST_FRAMEWORK_PACKAGE_NAMES = [
+  "@sveltejs/kit",
+  NEXT_PACKAGE_NAME,
+  "nuxt",
+  "nuxt3",
+  "nuxt-edge",
+  "nuxt-nightly",
+] as const;
 const PACKAGE_DEPENDENCY_FIELDS = [
   "dependencies",
   "devDependencies",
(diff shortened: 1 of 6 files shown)


packages/eve/src/setup/scaffold/update/channels.ts
```diff
@@ -167,6 +175,18 @@ async function hasPackageDependency(
...
+async function hasAnyPackageDependency(
+  packageJsonPath: string,
+  dependencyNames: readonly string[],
+): Promise<boolean> {
+  if (!(await pathExists(packageJsonPath))) return false;
+
+  const parsed: unknown = JSON.parse(await readFile(packageJsonPath, "utf8"));
+  if (!isJsonObject(parsed)) return false;
+
+  return dependencyNames.some((dependencyName) => packageJsonHasDependency(parsed, dependencyName));
```
ijjk, line 182: We might want to attempt actual resolving to handle monorepo cases with multiple package files

ijjk (dismissed):

packages/eve/src/setup/scaffold/update/channels.ts
allenzhou101: @ijjk I looked into this and I think resolving could lead to false positives in monorepos, since a package can resolve deps from parent `node_modules` that belong to another workspace/app. 

Vercel CLI also seems to do manifest based rather than require.resolve so kept to match

AndrewBarba (approved):

PR #167, opened 2026-06-21
fix(eve): slack - restrict HITL responses to turn caller

AndrewBarba:
## Summary

- bind Slack HITL controls to the current turn caller using the Slack `user_id`
- send mismatched responders an ephemeral explanation, then ignore the interaction
- render non-actionable prompts when no Slack caller is available
- document the behavior and add a patch changeset

## Why

Slack interactions were authenticated, but any authenticated workspace user could respond to a pending HITL request. This makes caller-only responses the Slack channel default while keeping the prompt visible in the thread.

## Validation

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm fmt`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm test:unit`
- `pnpm test:integration`
- `pnpm test:scenario`


packages/eve/src/public/channels/slack/interactions.ts
```diff
@@ -200,12 +202,22 @@ export async function handleInteractionPost(
...
     return ack;
   }
 
   const continuationToken = slackContinuationToken(interaction.channelId, interaction.threadTs);
-  const inputResponses = interaction.actions
+  const hitlActions = interaction.actions.filter((action) => isHitlAction(action.actionId));
+  const unauthorizedAction = hitlActions.find((action) => !canRespondToHitlAction(action));
+  if (unauthorizedAction) {
+    notifyUnauthorizedHitlResponder(interaction, unauthorizedAction.user.id, ctx.waitUntil, deps);
+  }
```
ruiconti, line 218: Shouldn't we return early here as well?

packages/eve/src/public/channels/slack/hitl.ts
```diff
@@ -72,6 +76,20 @@ interface SlackHitlAction {
   readonly selectedOptionValue?: string;
 }
 
+/** Returns a request-unique Block Kit block id that authorizes one Slack user. */
+export function hitlResponderBlockId(userId: string, requestId: string): string {
+  return `${HITL_RESPONDER_BLOCK_PREFIX}${userId}:${requestId}`;
```
ruiconti, line 81: Can this blow up the 255 limit? https://docs.slack.dev/reference/block-kit/blocks/actions-block/

PR #181, opened 2026-06-22
fix: preserve parked dev runtime snapshots

### LINKED ISSUE #172: Parked HITL turn crashes on resume with `LoadCompiledManifestError` after `eve dev` rebuild/restart GCs its snapshot
### eve version

`eve@0.11.10`

### Node.js version

`v24.17.0` (macOS; also relevant on Linux)

### Where does the bug occur?

Local dev (`eve dev --no-ui`)

### Describe the bug

A **parked HITL turn pins the id of the dev-runtime snapshot it was compiled against** into its serialized (durable) turn context. When `eve dev` later rebuilds (a watched source file changes) or is restarted, it builds a **new** snapshot under `.eve/dev-runtime/snapshots/<new-id>/` and the **old snapshot directory is removed**. Resuming the parked turn then deserializes a context that points at the now-deleted snapshot and crashes while loading its compiled manifest:

```
Step "step//eve@0.11.10//turnStep" failed after 3 retries:
ENOENT: no such file or directory, open
'.../.eve/dev-runtime/snapshots/<old-id>/source/.eve/compile/compiled-agent-manifest.json'

LoadCompiledManifestError: ENOENT: ... compiled-agent-manifest.json
    at loadCompiledManifest (.eve/nitro/dev/index.mjs)
    at async loadFullBundle
    at async getCompiledRuntimeAgentBundle
    at async deserializeContext
    at async turnStep
```

After 3 retries this bubbles out of the parent workflow as `FatalError` / `code USER_ERROR`. The session that owns the parked turn (`workflow//eve//workflowEntry`) stays `running`, so any client that re-attaches to it (e.g. a panel that auto-resumes the account's latest session) re-triggers the resume and hits the same crash repeatedly. Builds / deployed runs are unaffected — this is dev-runtime only.

### Root cause (analysis)

- The snapshot id is serialized into the **turn** context (`workflow//eve//turnWorkflow` run `input`), not the session context. I confirmed this by decoding the durable run payloads in `.workflow-data/<store>/runs/`: the failing turn's `input` (and `error`) contain exactly one snapshot id, `<old-id>`, and that directory is gone from `.eve/dev-runtime/snapshots/` while several newer snapshot dirs exist.
- On resume, `deserializeContext` → `getCompiledRuntimeAgentBundle` → `loadFullBundle` → `loadCompiledManifest` does a direct read of `<old-id>/source/.eve/compile/compiled-agent-manifest.json`, which no longer exists → `ENOENT`.
- #169 ("load dev snapshots with authored-source resolver") preserves the *loader path* across durable steps, but does not address the snapshot **directory itself being garbage-collected** out from under a still-resumable durable turn. This is the same dev-runtime-snapshot durability area as #104 / #105 (`LoadCompiledModuleMapError`), but the trigger here is a missing snapshot rather than import resolution inside a present one.

### Steps to reproduce

1. `eve dev --no-ui`.
2. Start a session whose first turn parks on human input (`ask_question` / any HITL approval). Leave it unanswered. It parks at `input.requested` / `session.waiting`.
3. Force a new snapshot: either restart `eve dev`, or edit any watched source file so the dev watcher rebuilds. Observe that `.eve/dev-runtime/snapshots/` now contains a new snapshot id and the previously-active one has been removed.
4. Answer the HITL request (deliver the input response) so the parked turn resumes.
5. The resumed turn crashes with `LoadCompiledManifestError` (ENOENT on the old snapshot's `compiled-agent-manifest.json`), retried 3× then surfaced as `FatalError` / `USER_ERROR`.

(In a SvelteKit app this is easy to hit unintentionally: Vite continuously rewrites `.svelte-kit/tsconfig.json`, which the dev watcher treats as a tsconfig dependency and rebuilds on — see #102 — so parked HITL turns get orphaned without any manual restart.)

### Expected behavior

Resuming a durable HITL turn under `eve dev` after a rebuild/restart should not hard-crash on a missing snapshot. Reasonable options:

- Don't garbage-collect a snapshot directory while a non-terminal (parked / `session.waiting`) durable turn still references it; or
- On resume, if the pinned snapshot is gone, fall back to the current snapshot (and/or restart the turn) instead of throwing `LoadCompiledManifestError`; or
- At minimum, fail the *turn* with a clear, recoverable "snapshot no longer available, start a new turn" error rather than an opaque `FatalError` / `USER_ERROR`, and surface it to the client (related: #134, #136).

ijjk:
## Summary
- Add regression coverage for stale dev-runtime snapshots referenced by durable workflow data.
- Preserve snapshots that are still present in local workflow payloads so parked HITL turns can resume after `eve dev` rebuilds.
- Handle JSON-escaped Windows paths when scanning durable workflow data.

## Tests
- `pnpm exec vitest run --config vitest.integration.config.ts src/internal/nitro/dev-runtime-artifacts.integration.test.ts -t "preserves stale snapshots referenced by durable workflow data"`

Closes: https://github.com/vercel/eve/issues/172

### DIFF
--- a/packages/eve/src/internal/nitro/dev-runtime-artifacts.ts
+++ b/packages/eve/src/internal/nitro/dev-runtime-artifacts.ts
@@ -11,6 +11,7 @@ const DEV_RUNTIME_ARTIFACTS_DIRECTORY = "dev-runtime";
 const DEV_RUNTIME_ARTIFACTS_POINTER_VERSION = 2;
 const DEV_RUNTIME_SNAPSHOT_RECENT_WINDOW_MS = 15 * 60 * 1000;
 const DEV_RUNTIME_SNAPSHOT_RETAIN_COUNT = 5;
+const DEV_RUNTIME_WORKFLOW_DATA_MAX_SCAN_BYTES = 1024 * 1024;
 
 interface DevelopmentRuntimeArtifactsPointerV1 {
   readonly appRoot: string;
@@ -180,7 +181,10 @@ export async function pruneDevelopmentRuntimeArtifactsSnapshots(input: {
   const pointer = readDevelopmentRuntimeArtifactsPointer(
     resolveDevelopmentRuntimeArtifactsPointerPath(input.appRoot),
   );
-  const protectedPaths = collectProtectedSnapshotPaths(pointer);
+  const protectedPaths = [
+    ...collectProtectedSnapshotPaths(pointer),
+    ...(await collectWorkflowDataSnapshotPaths({ appRoot: input.appRoot, snapshotsDirectory })),
+  ];
   const now = input.now ?? Date.now();
   const recentWindowMs = input.recentWindowMs ?? DEV_RUNTIME_SNAPSHOT_RECENT_WINDOW_MS;
   const retainCount = input.retainCount ?? DEV_RUNTIME_SNAPSHOT_RETAIN_COUNT;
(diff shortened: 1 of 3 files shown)


AndrewBarba (approved):

packages/eve/src/internal/nitro/dev-runtime-artifacts.ts
```diff
@@ -292,6 +296,94 @@ function collectProtectedSnapshotPaths(
...
+
+  for (const match of normalizedSource.matchAll(pattern)) {
+    const snapshotName = match[1];
+
+    if (snapshotName !== undefined && snapshotName.length > 0) {
+      snapshotPaths.add(join(snapshotsDirectory, snapshotName));
+    }
+  }
+
+  return [...snapshotPaths];
```
ruiconti, line 380: This would retain snapshots from sessions that have already completed, as this is just looking for whether a snapshot exists in a workflow run. For instance

```ts
{
    "status": "completed",
    "input": {
      "serializedContext": {
        "eve.bundle": {
          "source": {
            "appRoot": ".../.eve/dev-runtime/snapshots/A/source"
          }
        }
      }
    }
  }
```

would still be incorrectly retained.

ijjk: Good catch cleaning up completed/failed in https://github.com/vercel/eve/pull/187

PR #204, opened 2026-06-23
feat(tools): add defineClientTool for client-resolved (HITL) tools

### LINKED ISSUE #203: No way to author a HITL / client-resolved tool: authored tools require `execute`, which yields a duplicate `tool_result` when the same call parks for input
# No way to author a HITL / client-resolved tool: authored tools require `execute`, which yields a **duplicate `tool_result`** when the same call also parks for input

## Summary

There is no first-class way to author a tool that participates in eve's
human-in-the-loop (HITL) **input-request** flow (`input.requested` →
`session.waiting` → resume via `inputResponses`) the way the built-in
`ask_question` does.

- The built-in `ask_question` works because it has **no `execute`** — the model
  emits the call, nothing runs it, the harness parks it, and the user's answer
  becomes its single `tool_result`.
- Any **authored** tool is *required* by the compiler to have an `execute`
  function. That `execute` produces a `tool_result` for the call. If the same
  call also parks for input (e.g. an authored override of `ask_question`, or any
  attempt at an HITL tool), the resumed turn then carries **two `tool_result`
  blocks for one `tool_use` id**, and the provider rejects it:

  ```
  GatewayInternalServerError: messages.18.content.1: each tool_use must have a
  single result. Found multiple `tool_result` blocks with id:
  toolu_01QQFSTbsnsdR6qXFt5EFaCS
  ```

Net effect: you cannot widen/customize `ask_question` (its input schema is fixed
and strict), and you cannot build your own parking/HITL tool, because authoring
forces an `execute` whose result collides with the input-response result.

## Environment

- `eve` **0.11.6**
- Next.js 16 (App Router, Turbopack), `useEveAgent` over a same-origin
  `/eve/v1/*` proxy (`withEve`)
- Model: `anthropic/claude-sonnet-4.6` via AI Gateway (`@ai-sdk/gateway`)
- Default harness, conversation mode (`ask_question` enabled)

## Goal / use case

We want rich, typed HITL pickers in a chat UI — e.g. "choose media", "pick a
template", "pick an aspect ratio", a multi-select, or a multi-question form. The
ideal is the exact semantics of `ask_question`, but with a **richer, typed input
schema** so the client can render a dedicated widget from the parked request,
and the user's structured choice resumes the *same* turn.

`ask_question`'s built-in input schema is fixed and `.strict()`
(`{ prompt, options?, allowFreeform? }`,
`runtime/framework-tools/ask-question.js` → `inputRequestSchema.omit(...)`), so
extra fields can't be smuggled through it. The natural approach is to **author
`agent/tools/ask_question.ts` to override it with a superset schema** — which the
graph supports (an authored tool with a framework name replaces the framework
tool, see `runtime/resolve-agent-graph.js`).

The override's **schema half works perfectly**: the model calls our widened
`ask_question` with a typed `ui` payload, the harness emits `input.requested`
with our input intact, the client renders the widget, the user answers, and the
turn **resumes** (`turn.started`). The only failure is the duplicate
`tool_result` described below.

## Reproduction

1. Author `agent/tools/ask_question.ts` overriding the built-in with a wider
   `inputSchema` (any superset is fine). A no-op `execute` is required to satisfy
   the compiler:

   ```ts
   import { defineTool } from 'eve/tools'
   import { z } from 'zod'

   export default defineTool({
     description: 'Ask the user a question (widened).',
     inputSchema: z.object({
       prompt: z.string(),
       options: z.array(z.object({ id: z.string(), label: z.string() })).optional(),
       allowFreeform: z.boolean().optional(),
       ui: z.object({ kind: z.string() }).passthrough().optional(), // the widening
     }),
     // Required by the compiler. This is the whole problem — see below.
     execute: () => ({ status: 'ignored' as const }),
   })
   ```

2. Have the agent call `ask_question` (with or without the `ui` field) and answer
   it from the client via `agent.send({ inputResponses: [{ requestId, text }] })`.

3. The turn parks and resumes, then **fails** on the next model call with
   `each tool_use must have a single result … multiple tool_result blocks with
   id <ask_question callId>`.

The built-in `ask_question` (no override) does **not** fail — because it has no
`execute`.

## Observed event trace (client `onEvent`)

```
… reasoning/message/step events …
input.requested            ← harness parks the ask_question call
turn.completed
session.waiting            ← stream closes, status → ready  ✅ (parking works)
onFinish                   ← client resumes with inputResponses
turn.started               ← resume accepted  ✅
step.started
step.failed                ← model call rejected: duplicate tool_result  ❌
turn.failed
session.failed
```

So pause/resume is functioning; the failure is purely the **two results for one
`tool_use`** in the reconstructed history sent to the model.

## Root-cause analysis (source references, eve 0.11.6)

1. **Authored tools must have an `execute` (function).**
   `compiler/normalize-tool.js` → `compileToolEntry` →
   `normalizeToolDefinition` (`internal/authored-definition/schema-backed.js`)
   calls `expectFunction(n.execute, t)`. Omitting `execute` (or setting it to
   `undefined`) fails compilation with *"Expected the tool export 'default' …
   to match the public eve shape."* So an executeless authored tool is
   impossible.

2. **The built-in `ask_question` has no `execute`.**
   `runtime/framework-tools/ask-question.js` defines
   `ASK_QUESTION_TOOL_DEFINITION` with only `description` / `inputSchema` /
   `outputSchema` — no `execute`. In `harness/tools.js` `buildToolSet`,
   `execute: wrapToolExecute(s)` and `wrapToolExecute` returns `undefined` when
   `s.execute === undefined`. So the model-facing built-in tool has no executor
   and is never auto-run; it only ever resolves via the input-request path.

3. **The override replaces the framework tool *with execute*.**
   `runtime/resolve-agent-graph.js` builds the registry as
   `createRuntimeToolRegistry({ tools: [...framework.filter(notOverridden), ...authored] })`,
   so the authored `ask_question` (which, per (1), *has* an `execute`) replaces
   the built-in. Now `buildToolSet` wraps a real executor onto the model-facing
   `ask_question`, and a `tool_result` is produced for the call **in addition
   to** the input-request resolution.

4. **`ask_question` is excluded from eve's own action runner, but that does not
   prevent the result.** `harness/tool-loop.js` passes
   `excludedActionToolNames: new Set([ASK_QUESTION_TOOL_NAME, CODE_MODE_TOOL_NAME, FINAL_OUTPUT_TOOL_NAME])`
   to `emitStepActions`, and `harness/input-extraction.js`
   `extractQuestionInputRequests` keys the park on `toolName === ASK_QUESTION_TOOL_NAME`.
   So the call is *both* parked for input *and* (because the tool now carries an
   `execute`) resolved with an execute-derived `tool_result`. On resume the
   input answer adds a second `tool_result` for the same id → the provider 400.

In short: **HITL parking and "authored tool ⇒ has `execute` ⇒ produces a
result" are mutually exclusive, and there is no supported way to opt an authored
tool out of producing a result.**

## Why the obvious workarounds don't work

- **Omit `execute`** → compiler rejects the tool (requirement (1)).
- **`execute: undefined as never`** → same compiler rejection (`expectFunction`).
- **`execute` returns `{}` / throws / no-ops** → it still yields a `tool_result`
  block; the duplicate remains.
- **Smuggle rich fields through the built-in `ask_question`** → its input schema
  is fixed and `.strict()`; unknown keys (`ui`, etc.) aren't part of the
  model-facing schema and don't survive.
- **`needsApproval` instead of `ask_question`** → approval defers `execute`
  until after the park (so no duplicate), **but** `execute` only receives the
  original tool input via `ToolContext`; the user's structured answer
  (`inputResponse` / approval `text`) is **not** exposed to `execute`
  (`public/definitions/tool.d.ts` `ToolContext` = `SessionContext` + token
  accessors only). So an approval-gated tool cannot return the user's selection
  as its result.
- **`awaiting_selection` + relay** (execute returns a marker, turn *completes*,
  client relays the choice as a follow-up `message`) does avoid the duplicate,
  but it abandons native HITL: it relies on a synthetic out-of-band user message
  instead of `inputResponses`, loses the deterministic single-turn pause/resume,
  and pollutes session history.

## Proposed solutions (any one would unblock this)

1. **First-class authored HITL / client-resolved tools (preferred).**
   Allow an authored tool to declare it has *no executor* and is resolved by the
   client/HITL channel — e.g. `defineClientTool({...})`, a `clientResolved: true`
   flag, or letting `defineTool` accept a definition without `execute` (compiler
   allows it; `buildToolSet`/`wrapToolExecute` already handle
   `execute === undefined`). Such a tool would behave exactly like the built-in
   `ask_question`: park on call, resolve via `inputResponses`, single result.

2. **Allow customizing the built-in `ask_question` input schema.**
   A config hook to widen/replace `ASK_QUESTION_INPUT_SCHEMA` (e.g. on
   `defineAgent`, or a `defineAskQuestion({ inputSchema })` helper) so apps can
   pass typed payloads through the existing, working (executeless) HITL tool
   without authoring a replacement.

3. **Suppress the auto-result when a call is parked for input.**
   When `extractQuestionInputRequests` (or an approval) parks a tool call, ensure
   the harness does **not** also emit/persist an `execute`-derived `tool_result`
   for that same call id — i.e. parking wins and the input response is the sole
   result. This would make an authored `ask_question` override "just work."

4. **Expose the input/approval response to `execute`.**
   Pass the resolved `inputResponse` (`{ optionId?, text? }`) into `ToolContext`
   so a `needsApproval`/HITL tool's `execute` can read the user's structured
   answer and return it. Enables clean authored HITL tools whose result is the
   user's choice.

## Acceptance criteria

- An app can author a tool that:
  - parks the turn on call (`input.requested` / `session.waiting`),
  - carries a **custom, typed input schema** the client can render from,
  - resolves via `agent.send({ inputResponses })`, and
  - produces **exactly one** `tool_result` for its call id (no provider 400 on
    resume).
- Equivalently: a supported way to widen `ask_question`'s input schema, **or** to
  opt an authored tool out of producing an `execute` result when it is parked.

## Appendix — exact provider error

```
GatewayInternalServerError: messages.18.content.1: each tool_use must have a
single result. Found multiple `tool_result` blocks with id:
toolu_01QQFSTbsnsdR6qXFt5EFaCS
  at createGatewayErrorFromResponse (@ai-sdk/gateway/src/errors/create-gateway-error.ts:121)
  …
upstreamType: 'AI_APICallError'
gatewayType: 'internal_server_error'
resolvedProvider: 'anthropic', canonicalSlug: 'anthropic/claude-sonnet-4.6'
```

keesvandorp:
Fixes #203.

## Problem

Authored tools are required to provide an `execute`: both the compiler
(`normalizeToolDefinition` → `expectFunction(record.execute)`) and the runtime
(`resolveToolDefinition` → `expectFunction(resolvedRecord.execute)`) reject a
tool without one. That makes it impossible to author a tool that participates in
the human-in-the-loop input flow the way the built-in `ask_question` does — no
executor, the model emits the call, the harness parks it, and the user's answer
becomes its single `tool_result`.

The practical consequence (from #203): overriding `ask_question` to widen its
(fixed, `.strict()`) input schema for typed HITL pickers forces an `execute`,
whose auto-result collides with the input response. The resumed turn then
carries **two `tool_result` blocks for one `tool_use` id** and the provider
rejects it:

```
each tool_use must have a single result. Found multiple `tool_result` blocks with id: toolu_…
```

## Change

Add **`defineClientTool({ description, inputSchema, outputSchema? })`** — an
authored tool with **no `execute`**, stamped `clientResolved: true`. eve never
runs it; the call parks for input and resolves out-of-band, producing exactly
one result.

- `internal/authored-definition/schema-backed.ts` — allow omitting `execute`
  when `clientResolved`; every other tool still requires it.
- `runtime/resolve-tool.ts` — skip reattaching a live `execute` for
  client-resolved tools.
- `public/definitions/tool.ts` — `defineClientTool` + `ClientToolDefinition`;
  passing `execute` throws.
- `public/tools/index.ts` — export from `eve/tools`.

No harness change is needed: the runtime already surfaces executeless tools as
client-side (`buildToolSet` / `wrapToolExecute` return `undefined`) and
`ResolvedToolDefinition.execute` is already `Optional`. This PR just lets
authored tools reach that existing path. `defineTool` is unchanged and still
requires `execute`.

Authoring `agent/tools/ask_question.ts` with `defineClientTool` overrides the
built-in question tool with a wider, typed schema while keeping native
pause/resume — the parked `input.requested` carries the full typed input, so a
client can render a dedicated widget from it.

```ts
import { defineClientTool } from "eve/tools";
import { z } from "zod";

export default defineClientTool({
  description: "Ask the user to pick a template.",
  inputSchema: z.object({
    prompt: z.string(),
    ui: z.object({ kind: z.literal("template_picker") }).passthrough(),
  }),
});
```

## Tests

- `defineClientTool` brands the definition, marks it `clientResolved`, carries
  no `execute`, and throws when `execute` is supplied.
- `normalizeToolDefinition` accepts a client-resolved tool without `execute` and
  still rejects a non-client tool that omits it.
- `pnpm --filter eve typecheck` / `oxlint` / unit tests green.

Verified end-to-end in a downstream app (Next.js + `useEveAgent`): an
executeless `ask_question` override parks (`input.requested` → `session.waiting`),
resumes from the user's structured answer, produces a single `tool_result`, and
the turn continues — no duplicate-result `400`.

## Notes

- Scope is intentionally minimal (issue #203 solution 1). Generalising the
  input-extraction park to executeless tools with *arbitrary* names (so HITL
  tools needn't be named `ask_question`) is a natural follow-up, kept out of
  this PR to keep the primitive focused.

### DIFF
--- a/.changeset/client-resolved-tools.md
+++ b/.changeset/client-resolved-tools.md
@@ -18,4 +18,6 @@ provider rejected the resumed turn with "each tool_use must have a single
 result". See #203.
 
 `defineTool` is unchanged and still requires `execute`; only `defineClientTool`
-may omit it.
+may omit it. The two shapes are mutually exclusive: the compiler and runtime
+reject a client-resolved tool that also defines `execute` (and a non-client tool
+that omits it).
--- a/.changeset/client-resolved-tools.md
+++ b/.changeset/client-resolved-tools.md
@@ -1,0 +1,21 @@
+---
+"eve": minor
+---
+
+Add `defineClientTool` for authoring client-resolved (human-in-the-loop) tools.
+
+A client-resolved tool has **no `execute`**: eve surfaces it to the model, parks
+the turn when the model calls it, and resolves the call from the client/HITL
+channel (e.g. an `inputResponses` answer) rather than running server code — the
+same mechanism the built-in `ask_question` uses, now available to authored
+tools.
+
+This unblocks widening/overriding `ask_question` with a richer, typed input
+schema (typed HITL pickers) without the duplicate `tool_result` that a
+`defineTool` override produced — authoring previously forced an `execute`, so a
+parked call yielded two `tool_result` blocks for one `tool_use` id and the
+provider rejected the resumed turn with "each tool_use must have a single
+result". See #203.
+
+`defineTool` is unchanged and still requires `execute`; only `defineClientTool`
+may omit it.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,51 @@ jobs:
       - name: Run unit tests
         run: pnpm test:unit
 
+  client-resolved-hitl-construction:
+    # Gateway-free, merge-blocking proof of the client-resolved / HITL
+    # CONSTRUCTION contract (vercel/eve#203): a client-resolved tool may omit
+    # `execute`, the mixed `clientResolved + execute` shape is rejected at both
+    # compile and runtime, and the authored HITL fixture (the `ask_question`
+    # override + the approval-vs-client-resolved fixture) typechecks and builds
+    # against the authored surface. The gateway-backed `eve eval`
+    # (e2e-local / e2e-vercel) stays as the RUNTIME evidence — that the resumed
+    # history yields exactly one result and keeps approval-gated execution
+    # separate from client-resolved input. This job guarantees no executor-less
+    # path ships without a fixture compiled against it.
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@0e279bb959325dab635dd2c09392533439d90093 # v6
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
+        with:
+          node-version-file: .nvmrc
+          cache: "pnpm"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Build eve
+        run: pnpm --filter eve build
+
+      - name: Construction guards (omit-execute + mixed-shape rejection)
+        # Also covered by the full suite in test-unit; pinned here so this job is
+        # a self-contained statement of the construction contract.
+        run: >-
+          pnpm --filter eve exec vitest run --config vitest.unit.config.ts
+          src/public/definitions/define-client-tool.test.ts
+          src/internal/authored-definition/schema-backed.test.ts
+
+      - name: Build + typecheck the authored HITL fixture
+        run: pnpm --filter agent-tools-hitl run typecheck
+
   test-integration:
     name: test-integration (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
--- a/docs/tools/human-in-the-loop.md
+++ b/docs/tools/human-in-the-loop.md
@@ -56,6 +56,27 @@ The built-in `ask_question` tool lets the model pause and ask the user, rather t
 
 `ask_question` is part of the [default harness](/docs/concepts/default-harness), so it is available without you defining anything. It produces the same `input.requested` pause as an approval, and resumes the same way.
 
+### Custom client-resolved tools
+
+`ask_question` is a _client-resolved_ tool: it has no `execute`, so eve never runs it — the model emits the call, the turn parks, and the user's answer becomes its single result. Author your own with `defineClientTool` when you want that exact pause-and-resume but with a **richer, typed input schema** — a "pick a template", "choose media", or multi-field form that a channel or your frontend renders as a dedicated widget:
+
+```ts title="agent/tools/ask_question.ts"
+import { defineClientTool } from "eve/tools";
+import { z } from "zod";
+
+export default defineClientTool({
+  description: "Ask the user to pick a template.",
+  inputSchema: z.object({
+    prompt: z.string(),
+    ui: z.object({ kind: z.literal("template_picker") }).passthrough(),
+  }),
+});
+```
+
+Naming the file `ask_question.ts` overrides the built-in question tool with your wider schema while keeping the same parking behavior (parking is keyed on the `ask_question` name). The parked `input.requested` carries your full typed input, so the client renders the picker from it and resumes the turn with the user's structured choice.
+
+A `defineClientTool` tool must **not** declare an `execute` — that is the whole point, and passing one is rejected at compile and runtime. [`defineTool`](/docs/tools) is the opposite: it always requires an `execute` and runs on the server. The two shapes are mutually exclusive, so a parked call always resolves to exactly one result.
+
 ## How pause and resume works
 
 Approvals and questions share one protocol:
--- a/docs/tools/overview.mdx
+++ b/docs/tools/overview.mdx
@@ -63,7 +63,7 @@ export default defineTool({
 });
 ```
 
-Approval is one half of eve's [human-in-the-loop](./human-in-the-loop) model — the page covers the `always/once/never` helpers, input-dependent predicates, and how a gated call pauses and resumes durably.
+Approval is one half of eve's [human-in-the-loop](./human-in-the-loop) model — the page covers the `always/once/never` helpers, input-dependent predicates, and how a gated call pauses and resumes durably. For the other half — a tool with **no** `execute` that parks for a typed user answer (`defineClientTool`, e.g. a custom picker) — see [client-resolved tools](./human-in-the-loop#custom-client-resolved-tools).
 
 ## Shape what the model sees with `toModelOutput`
 
--- a/e2e/fixtures/agent-tools-hitl/agent/tools/ask_question.ts
+++ b/e2e/fixtures/agent-tools-hitl/agent/tools/ask_question.ts
@@ -1,0 +1,25 @@
+import { defineClientTool } from "eve/tools";
+import { z } from "zod";
+
+/**
+ * Authored override of the built-in `ask_question`, widened with a typed `ui`
+ * payload. This is the exact shape that regressed in #203: as a plain
+ * `defineTool` it was forced to carry an `execute`, so a parked call produced a
+ * second `tool_result` for its `tool_use` id and the resumed turn was rejected
+ * ("each tool_use must have a single result"). As a `defineClientTool` it has no
+ * executor — the call parks for input and resolves to exactly one result.
+ *
+ * The schema is a strict superset of the built-in (prompt / options /
+ * allowFreeform), so the existing `ask-question-select` eval keeps passing while
+ * exercising the authored path.
+ */
+export default defineClientTool({
+  description:
+    "Ask the user a question and wait for their answer. Plain choice: prompt (+ options, allowFreeform). Rich input: set a typed `ui` payload to render a picker.",
+  inputSchema: z.object({
+    prompt: z.string(),
+    options: z.array(z.object({ id: z.string(), label: z.string() })).optional(),
+    allowFreeform: z.boolean().optional(),
+    ui: z.object({ kind: z.string() }).passthrough().optional(),
+  }),
+});
--- a/e2e/fixtures/agent-tools-hitl/agent/tools/note.ts
+++ b/e2e/fixtures/agent-tools-hitl/agent/tools/note.ts
@@ -1,0 +1,19 @@
+import { defineTool } from "eve/tools";
+import { z } from "zod";
+
+/**
+ * Trivial downstream executable tool. Used by the client-resolved regression to
+ * continue the turn AFTER a parked `ask_question` resumes — exercising the
+ * reconstructed provider history (a duplicate result for the parked call would
+ * reject that follow-up model call).
+ */
+export default defineTool({
+  description:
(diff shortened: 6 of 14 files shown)


keesvandorp:
Thanks — addressed in d839f7b.

**Reject mixed shapes (compiler + runtime).** A client-resolved tool that also defines `execute` now throws at both `normalizeToolDefinition` (compile) and `resolveToolDefinition` (runtime), rather than silently dropping the executor. The non-client-without-marker direction was already rejected. (`defineClientTool` still throws at authoring time too.) Unit-tested.

**Regression 1 — the one that matters.** New `e2e/fixtures/agent-tools-hitl` eval `client-resolved-question`: an authored, widened `ask_question` (`defineClientTool` + typed `ui`) parks, resumes from a structured answer, and continues into a downstream `note` tool. The single-result invariant is asserted operationally — before the fix the authored override carried an `execute`, so the resume reconstructed two `tool_result` blocks for one id and the provider 400'd; a green `expectOk()` + downstream `calledTool('note')` + `completed()` can only happen with exactly one result for the call.

**Regression 2 — separation.** New eval `approval-vs-client-resolved`: `guarded-echo` (approval-gated executable) parks for *approval* and resolves via its *executor* (token in the result); `ask_question` (client-resolved) parks for *input* and the user's answer *is* the result, no executor. Same parking machinery, opposite result sources.

Verified: eve typecheck + unit tests (incl. the mixed-shape rejection) + oxlint; the HITL fixture typechecks and `eve build`s with the override. I haven't run the evals themselves (they need a gateway) — they mirror the existing `ask-question-select` / `approve-then-no-regate` evals.

rpelevin:
Nice update. The important thing I would preserve before merge is that the mixed-shape rejection and the resume regression stay paired.

The compiler and runtime checks close the construction path, but the fixture is what proves the runtime history is actually reconstructing one result for the original parked call.

The remaining review lens I would use is:

1. The client-resolved marker is the only path that allows no executor.
2. Any exported execute on that shape fails before it can be silently dropped.
3. The resumed structured input is bound to the same call id.
4. The reconstructed provider history contains exactly one result for that call.
5. The approval-gated executable path still proves the opposite source of truth: approval permits the executor to run; client input supplies the result.

If the gateway-backed evals cannot run in ordinary CI, I would at least keep the fixture build/typecheck plus the mixed-shape unit tests as merge blockers, and treat live eval execution as release evidence before closing the original issue.

Boundary: architecture and test feedback only; no claim about using this project or running its code.

keesvandorp:
Agreed on all five, and on keeping the rejection and the resume regression paired — the compiler/runtime checks guard *construction*, the fixture proves the *reconstructed history* settles to one result for the parked call. Neither substitutes for the other.

The gating maps cleanly onto the existing CI:

- **Deterministic merge blockers (no gateway):** `ci.yml` runs `lint` → `typecheck` → `test-unit` (`pnpm test:unit`, where the mixed-shape rejection tests live) → `test-integration` on every PR. That covers invariants 1–2 (only the marker permits no executor; an exported `execute` on that shape *fails*, not silently dropped) plus the construction surface.
- **Release evidence (gateway-backed):** `e2e-local.yml` triggers on the `e2e/**` change — it builds eve + the HITL fixture (so construction errors still gate there), then runs `eve eval --strict`. The eval *execution* (invariants 3–5: same-call-id binding, exactly-one-result on resume, and the approval path proving the opposite source of truth) needs a model gateway, so I'd treat a green eval run as the evidence before closing #203 rather than a PR gate.

Locally verified on the branch: `pnpm --filter eve typecheck` + `test:unit` (incl. the mixed-shape rejection) + `oxlint`, and the `agent-tools-hitl` fixture both `tsc`-typechecks and `eve build`s with the `defineClientTool` override.

If it'd help, I'm happy to split the fixture's typecheck/build into an explicit gateway-free job so the construction guarantee gates independently of eval execution — just say the word.

rpelevin:
Yes, I would split that gateway-free job.

The useful boundary is:

1. A merge-blocking deterministic job proves construction: client-resolved tools can omit execute, mixed client-resolved plus execute fails, and the authored HITL fixture typechecks and builds.
2. Gateway-backed eval remains release evidence: it proves resumed history creates exactly one result for the parked call and keeps approval-gated executable output separate from client-resolved input.
3. Closing the original bug should wait for both pieces: deterministic construction gate green, and strict eval evidence green.

That split keeps CI fast while preserving the invariant that no executor-less path ships without a concrete fixture compiled against the authored surface.

I would make the job name explicit enough that future maintainers know what it protects, for example client-resolved-hitl-construction, and keep it pinned to the authored ask_question override plus the approval-vs-client-resolved fixture build.

Boundary: architecture and test feedback only; no claim about using this project or running its code.

keesvandorp:
Done — split out as a dedicated merge-blocking job in `32782ea`: **`client-resolved-hitl-construction`** (in `ci.yml`, gateway-free).

It proves construction, deterministically:
- the omit-`execute` + mixed-shape rejection unit guards (`define-client-tool` + `schema-backed`), and
- `pnpm --filter agent-tools-hitl run typecheck` (`eve build && tsc`) — so the authored `ask_question` override and the approval-vs-client-resolved fixture must compile and build against the authored surface.

The gateway-backed `eve eval` (`e2e-local` / `e2e-vercel`) stays as runtime release evidence: exactly one result on resume, and approval-gated execution kept separate from client-resolved input. So closing #203 waits for both — construction gate green here, strict eval green as release evidence.

Job name is intentionally explicit and pinned to that fixture pair, with a header comment stating what it protects, so it's legible to future maintainers. Verified locally: unit guards 18/18, and the fixture `eve build && tsc` clean.

nima54851:
Solid contribution! The rate limiting implementation looks clean.

Quick feedback: the error handling in agent_core.py could benefit from more granular retry logic for transient failures. Also, has the team considered adding structured logging from the start? It'll save a lot of debugging pain later.

Happy to help if you want a hand with any of this!

PR #220, opened 2026-06-23
feat(eve): support structured output for GitHub turns

### LINKED ISSUE #214: Allow GitHub webhook and receive(github, ...) turns to request structured output

A GitHub maintenance agent needs to turn each issue or pull-request delivery into one schema-validated assessment, persist it, and optionally notify downstream systems. The GitHub session must remain durable and conversational so maintainers can continue talking to the same agent later.

The application also claims webhook deliveries and retries them through `receive(github, ...)`. Initial webhook turns and scheduled retry turns therefore need identical structured-output behavior.

eve already supports turn-scoped `outputSchema` in conversation mode and emits the finalized value as `result.completed`. However:

- [`GitHubInboundResult`](https://github.com/vercel/eve/blob/963b601308329945bc67521f1a799adda95ada4a/packages/eve/src/public/channels/github/githubChannel.ts#L93-L101) only carries `auth` and `context`.
- [`CrossChannelReceiveOptions`](https://github.com/vercel/eve/blob/963b601308329945bc67521f1a799adda95ada4a/packages/eve/src/channel/cross-channel-receive.ts#L11-L21) only carries `message`, `target`, and `auth`.
- The underlying [`SendPayload`](https://github.com/vercel/eve/blob/963b601308329945bc67521f1a799adda95ada4a/packages/eve/src/channel/routes.ts#L35-L50) already supports `outputSchema`.

The current workaround is a custom submission tool. That is not equivalent to framework-managed output: the model can omit or repeat a normal tool call, and trusted persistence becomes part of the model-visible tool surface. Switching the session to task mode is also unsuitable because it would discard the durable conversation.

### Proposed solution

Add optional `outputSchema` to the two public turn-entry surfaces:

```ts
export type GitHubInboundResult = {
  readonly auth: SessionAuthContext | null;
  readonly context?: readonly string[];
  readonly outputSchema?: JsonObject;
} | null;

export interface CrossChannelReceiveOptions<TTarget = Record<string, unknown>> {
  readonly message: string | UserContent;
  readonly target: TTarget;
  readonly auth: SessionAuthContext | null;
  readonly outputSchema?: JsonObject;
}
```

Expected behavior:

- GitHub webhook dispatch forwards the handler’s schema through `SendPayload`.
- Cross-channel receive applies the caller’s schema to the target channel’s framework-provided `send`. Individual `receive` implementations and routing targets should not need schema-specific plumbing.
- The session remains in conversation mode.
- Successful structured turns emit one finalized `result.completed`, persist the structured assistant turn, clear the schema, and return to `session.waiting`.
- Unfulfilled schemas emit the existing recoverable `step.failed` → `turn.failed` → `session.waiting` sequence without `result.completed`.
- A fresh delivery that omits `outputSchema` must not inherit a schema left by an earlier failed turn. Internal continuation steps should still preserve the active schema. The current [`resolveEffectiveOutputSchema`](https://github.com/vercel/eve/blob/963b601308329945bc67521f1a799adda95ada4a/packages/eve/src/execution/workflow-steps.ts#L434-L461) does not make that distinction.
- Applications can subscribe to `result.completed` with an agent hook, revalidate against trusted current session auth, and perform idempotent persistence or delivery. No new channel event callback is needed.

The sketch uses `JsonObject` to match `SendPayload` and keep the change forwarding-focused. Accepting Standard Schema implementations and normalizing them at the boundary would also be reasonable if consistency with the client API is preferred.

This overlaps [#207](https://github.com/vercel/eve/pull/207), which modifies the same cross-channel send wrapper and `final_output` lifecycle for `allowEmptyDelivery`. The current #207 implementation rejects combining `allowEmptyDelivery` with `outputSchema`; that conflict should remain explicit unless the #207 design changes.

<details>
<summary>Expected coverage</summary>

- GitHub issue and pull-request handlers forward the exact schema.
- Scheduled and route-level `receive(github, ...)` calls forward it for both new and existing sessions.
- Successful structured output emits one `result.completed` and does not affect the next plain turn.
- An unfulfilled schema produces the recoverable failure sequence.
- A plain delivery after a failed schema turn does not inherit the old schema.
- Internal continuations preserve an active schema.
- Interaction with #207’s `allowEmptyDelivery` is tested.
- GitHub and schedule documentation plus a patch changeset are included.

</details>

Would maintainers support this as a narrow follow-up to #207? I have not started implementation pending agreement on the public API and landing order.

### Alternatives considered

- **Keep the custom submission tool:** does not guarantee a framework-final result and exposes trusted side effects as a model action.
- **Use agent-level `outputSchema` or task mode:** changes session semantics and prevents later conversation.
- **Configure one schema on the GitHub channel:** too broad; schemas must be selected per delivery.
- **Put the schema in `GitHubReceiveTarget`:** mixes turn execution policy with platform routing.
- **Add a GitHub `result.completed` callback:** hooks already provide the result, current auth, session ID, and turn ID.
- **Generalize every built-in inbound-result type:** broader than the concrete need; a private shared helper is sufficient if useful internally.

marcoshernanz:
### Description

Adds turn-scoped structured output to GitHub channel deliveries.

- Lets `GitHubInboundResult` return `outputSchema` for webhook issue, pull-request, and comment turns.
- Lets cross-channel `receive(channel, ...)` pass the same schema for scheduled or proactive retries.
- Accepts Standard Schema or raw JSON Schema and normalizes it into durable JSON data.
- Scopes schemas to the current conversation turn while preserving runtime continuations and task-mode HITL contracts.
- Documents consuming validated results from the existing `result.completed` agent hook with trusted session auth.

Sessions remain durable and in conversation mode. This does not add a channel-specific result callback.

Closes #214.

### Relationship to #207

#207 modifies cross-channel receive for empty deliveries. This PR touches the same receive path; whichever lands second should rebase and resolve that small overlap without changing either behavior.

- `pnpm lint`
- `pnpm fmt`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test` — 3,895 unit tests and 369 integration tests passed
- `pnpm test:scenario` — 260 passed, 15 platform-specific skips
- `pnpm docs:check`
- `pnpm guard:invariants`
- `pnpm guard:fixtures`
- `pnpm check:deps`
- `pnpm changeset status`
- `NODE_ENV=test pnpm --filter agent-channels test:e2e` — 2 evals and 5 gates passed

Credentialed workspace-wide E2E, model-driven TUI, and Vercel deployment checks could not complete because the required AI Gateway and Vercel credentials are unavailable locally. The relevant channel E2E passed in deterministic mock mode.

### PR Checklist

- [x] I added a changeset because this touches the published `eve` package

### DIFF
--- a/.changeset/github-turn-output-schema.md
+++ b/.changeset/github-turn-output-schema.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Allow GitHub webhook handlers and cross-channel `receive(...)` calls to request turn-scoped structured output with Standard Schema or raw JSON Schema. GitHub webhook turns and scheduled retries can now share one schema-validated `result.completed` contract while keeping sessions in conversation mode.
--- a/docs/channels/github.mdx
+++ b/docs/channels/github.mdx
@@ -42,16 +42,32 @@ Inbound hooks return `{ auth }` to dispatch, or `null` to ignore. Use `defaultGi
 
 ```ts
 import { defaultGitHubAuth, githubChannel } from "eve/channels/github";
+import { z } from "zod";
+
+const assessmentSchema = z.object({
+  summary: z.string(),
+  priority: z.enum(["low", "medium", "high"]),
+});
 
 export default githubChannel({
   botName: "my-agent",
   // Replaces the @mention gate. ctx.conversation.kind is "issue", "pull_request", or "review_thread".
   onComment: (ctx, comment) => ({ auth: defaultGitHubAuth(ctx) }),
   // Opt in; no default dispatch on these events.
-  onIssue: (ctx, issue) => (issue.action === "opened" ? { auth: defaultGitHubAuth(ctx) } : null),
-  onPullRequest: (ctx, pr) => (pr.action === "opened" ? { auth: defaultGitHubAuth(ctx) } : null),
+  onIssue: (ctx, issue) =>
+    issue.action === "opened"
+      ? { auth: defaultGitHubAuth(ctx), outputSchema: assessmentSchema }
+      : null,
+  onPullRequest: (ctx, pr) =>
+    pr.action === "opened"
+      ? { auth: defaultGitHubAuth(ctx), outputSchema: assessmentSchema }
+      : null,
 });
 ```
+
+Add `outputSchema` when a delivery needs a schema-validated result. It accepts a Standard Schema implementation such as Zod or a raw JSON Schema object and applies only to that turn. A successful structured turn emits the finalized value as `result.completed`; the durable session stays in conversation mode, and later turns do not inherit the schema.
+
+Consume `result.completed` in an [agent hook](../guides/hooks), where `ctx.session.auth.current` provides the trusted caller for validation and side effects. Keep those side effects idempotent and handle their failures inside the hook. `message.completed` is not the structured-result signal, and the GitHub channel does not need a separate result callback. See [Output Schema](../guides/client/output-schema) for the event lifecycle and failure behavior.
 
 ### Delivery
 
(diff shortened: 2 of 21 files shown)


marcoshernanz:
Superseded by #221, which uses an upstream branch so the credential-gated CI jobs can run in trusted repository context.

PR #230, opened 2026-06-24
feat(eve): cancellation - add turn and session cancellation

AndrewBarba:
## Summary

- add authenticated turn- and session-scoped cancellation to the eve HTTP API, TypeScript client, custom channel routes, authored runtime context, and eval driver
- propagate cooperative cancellation through Workflow v5's durable `AbortController`, including model calls, tools, local descendants, and remote descendants
- add explicit `turn.cancelled` and `session.cancelled` stream events, public documentation, and a minor changeset

## Stack

- This PR contains the implementation, documentation, changeset, and only the existing-test compatibility migrations required for standalone CI.
- #249 is stacked on this branch and contains new cancellation tests, test support, the cancellation fixture, and executable eval cases.

## Why

This is phase 1 of session reset support: the public cancellation API and end-to-end runtime plumbing. Both turn and session cancellation address the session in the request path. No cancellation capability or token is introduced.

Cancellation is intentionally best effort. A turn cancellation received while the session is parked is an accepted no-op. Start and delivery do not wait for hook registration, cancellation does not poll or wait for cleanup, and unavailable or completed sessions return `409`. eve adds no cancellation readiness streams; real-time abort propagation is owned by Workflow's `AbortController` implementation.

## Public behavior

- `POST /eve/v1/session/:sessionId/cancel` accepts exactly `{ "scope": "turn" }` or `{ "scope": "session" }`
- message responses retain the existing session and continuation identifiers; no cancellation token is returned
- `MessageResponse.cancel()` cancels the currently running turn and `ClientSession.cancel()` cancels the session
- custom channel handlers receive `cancelTurn({ sessionId })` and `cancelSession({ sessionId })`
- authored code receives `ctx.abortSignal` and `ctx.cancel({ scope })`
- evals can retain an active turn with `t.start()`, cancel the turn with `active.cancel()`, or cancel its session with `t.cancel()`

## Validation

Parent branch:

- `pnpm typecheck`
- `pnpm test:unit` — 3,892 tests
- `pnpm test:integration` — 368 tests
- relevant scenario suite, including a full serial scenario run through the stacked tree

Complete stacked tree:

- `pnpm fmt`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test:unit` — 3,931 tests
- `pnpm test:integration` — 371 tests
- scenario tests — 261 passed, 15 skipped


packages/eve/src/execution/cancellation-step.ts
```diff
@@ -0,0 +1,93 @@
...
+  session = clearPendingRuntimeActionBatch(session);
+  session = clearPendingCodeModeInterrupt(session);
+  session = clearPendingInputBatch(session);
+  session = clearPendingAuthorization(session);
+  session = clearProxyInputRequests(session);
+
+  const prior = input.sessionState.emissionState;
+  const turnId = prior.turnId || `turn_${prior.sequence}`;
+  const events: HandleMessageStreamEvent[] = [
+    createTurnCancelledEvent({ sequence: prior.sequence, turnId }),
```
ruiconti: If the session is parked (ie no active turn), would this still emit a cancellation turn event?

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -226,53 +250,43 @@ async function runDriverLoop(input: {
...
-    return await finalizeDone({
-      action,
-      driverWritable: input.driverWritable,
-    });
-  }
+  const hooksToClaim: Hook<unknown>[] = [
+    authHook,
+    sessionCancellationHook,
+    activeTurnCancellation.hook,
+  ];
```
ruiconti: Grouping them in a collection like this is great for readability 👍🏻

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -304,20 +320,30 @@ async function runDriverLoop(input: {
...
+          if (runtimeResults.kind === "cancelled") {
+            const transition = await applyTurnOutcome(runtimeResults);
+            if (transition.kind === "done") return transition.result;
+            action = transition.action;
+            break;
+          }
 
-          await rekeyHook(action.sessionState.continuationToken);
+          const transition = await applyTurnOutcome(
+            await dispatchAndAwaitTurn({
```
ruiconti: Nit: Could we move `outcome` to its own binding e.g., `const outcome = await dispatchAndAwait(...)`

packages/eve/src/execution/workflow-entry-helpers.ts
```diff
@@ -0,0 +1,309 @@
...
+    return { action: payload.action, kind: "action" };
+  } finally {
+    completion.dispose();
+  }
+}
+
+async function awaitHookPayload<T>(hook: Hook<T>): Promise<T> {
+  for await (const value of hook) return value;
+  throw new Error("Turn completion hook closed before delivering a result.");
+}
```
ruiconti, line 125: Nit: Could we inline this?

packages/eve/src/execution/workflow-entry-helpers.ts
```diff
@@ -0,0 +1,309 @@
...
+  try {
+    results = (await accumulateRuntimeActionResults({
+      bufferedDeliveries: input.bufferedDeliveries,
+      async getNext() {
+        while (true) {
+          const nextOrCancellation = await Promise.race([
+            input.cancellation(),
+            input.getNextPromise().then((next) => ({ kind: "next" as const, next })),
+          ]);
+          if (nextOrCancellation.kind === "cancelled") throw nextOrCancellation;
```
ruiconti: Kind of tricky to follow through code execution when exceptions have semantic meaning. For example, here we do throw `nextOrCancellation`, whereas `applyTurnOutcome` will use it as a value.

Could we align on not using exceptions semantically? e.g


```suggestion
          if (nextOrCancellation.kind === "cancelled") return nextOrCancellation;
```

packages/eve/src/execution/workflow-entry-helpers.ts
```diff
@@ -0,0 +1,309 @@
...
+    })) as readonly RuntimeSubagentResultActionResult[] | null;
+  } catch (error) {
+    if (isCancellationSignal(error)) {
+      return {
+        ...error,
+        serializedContext: currentSerializedContext,
+        sessionState: currentSessionState,
+      };
+    }
+    throw error;
```
ruiconti: [cont] so then we wouldn't be spreading out control flow here

packages/eve/src/execution/workflow-entry-helpers.ts
```diff
@@ -0,0 +1,309 @@
...
+  };
+}
+
+export async function awaitCancellationSignal(iterator: AsyncIterator<void>): Promise<void> {
+  const result = await iterator.next();
+  if (!result.done) return;
+  throw new Error("Cancellation hook closed before receiving a signal.");
+}
+
+export function raceTurnCancellation(
```
ruiconti:
```suggestion
export function raceSessionCancellation(
```
?

packages/eve/src/execution/workflow-runtime.ts
```diff
@@ -142,6 +147,29 @@ export function createWorkflowRuntime(config: {
       }
     },
 
+    async cancelTurn(input): Promise<void> {
+      const resumed = await resumeOwnedCancellationHook({
+        cancellationHookId: createActiveTurnCancellationHookId(input.sessionId),
+        expectedRunId: input.sessionId,
+        ownerHookId: createActiveTurnCancellationHookId(input.sessionId),
+      });
```
ruiconti: What happens if a client calls `cancelTurn` but the session is parked (i.e it should've called `cancelSession`)?

packages/eve/src/execution/cancellation.ts
```diff
@@ -0,0 +1,39 @@
...
+  return `${completionToken}:cancel`;
+}
+
+export function createSessionCancellationHookId(sessionId: string): string {
+  return `${sessionId}:cancel-session`;
+}
+
+export function createActiveTurnCancellationHookId(sessionId: string): string {
+  return `${sessionId}:cancel-turn`;
+}
```
ruiconti, line 42: These names are very similar to one another, could we also maybe clarify their lifetimes?


```suggestion
/* Used internally by the driver to cancel one child turn workflow run */
export function createTurnWorkflowCancellationHookId(completionToken: string): string {
  return `${completionToken}:cancel`;
}

/* Used to cancel the whole session, including a parked session i.e., no active turns */
export function createSessionCancellationHookId(sessionId: string): string {
  return `${sessionId}:cancel-session`;
}

/* Exposed externally (ie to delivery) to cancel the currently active workflow turn  */
export function createActiveTurnCancellationHookId(sessionId: string): string {
  return `${sessionId}:cancel-turn`;
}
```

packages/eve/src/execution/subagent-adapter-state.ts
```diff
@@ -0,0 +1,23 @@
...
+/** Validates adapter state after its workflow serialization round trip. */
+export function isSubagentAdapterState(value: unknown): value is SubagentAdapterState {
+  return (
+    isObject(value) &&
+    isNonEmptyString(value.callId) &&
+    isNonEmptyString(value.parentContinuationToken) &&
+    typeof value.parentSessionId === "string" &&
+    isNonEmptyString(value.subagentName)
+  );
+}
```
ruiconti, line 23: This whole file could be just a zod schema

packages/eve/src/execution/cancellation-step.ts
AndrewBarba: Fixed. Cancellation finalization now emits turn.cancelled only when the persisted emission state identifies an active turn. A parked session emits session.cancelled directly; active cancellation keeps the turn boundary. Unit and integration coverage is in stacked PR #249.

packages/eve/src/execution/cancellation-step.ts
```diff
@@ -0,0 +1,118 @@
...
```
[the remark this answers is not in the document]
AndrewBarba: Fixed by the same lifecycle change. Parked session cancellation now emits only session.cancelled, with regression coverage in stacked PR #249.

packages/eve/src/execution/cancellation.ts
AndrewBarba: Done. The three hook-ID helpers now document their child-turn, session-lifetime, and active-turn lifecycles. Their wire token strings are unchanged.

packages/eve/src/execution/subagent-adapter-state.ts
AndrewBarba: I am leaving this as the small eve-owned type guard. This module is imported by delegated-parent-result.ts, which runs in the workflow driver body; replacing it with a Zod schema would pull the runtime parser into that bundle for four string checks. The current guard keeps that architecture boundary lightweight and preserves the open adapter-state record.

PR #232, opened 2026-06-24
feat(eve): workflow hooks - adopt ownership checks

AndrewBarba:
## Summary

- claim continuation hooks with `getConflict()` before the first agent turn
- claim replacement hooks before releasing the active hook during rekeying
- atomically resume deliveries and report the owner from `resumeHook()`
- add unit, integration, production-bundle, documentation, and changeset coverage

## Why

Continuation hooks use shared tokens, but eve previously relied on delayed hook registration and a separate owner lookup during delivery. That left ownership conflicts to surface after work could begin and introduced a lookup-to-resume race.

This adopts the Workflow SDK's ownership check as an explicit barrier. A competing run now fails before performing an agent turn, while the existing owner remains active and resumable. Tokenless sessions retain their existing behavior and claim ownership after the first turn establishes a token.

## Validation

- `pnpm build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm fmt`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm test:unit` — 3,884 passed
- `pnpm test:integration` — 369 passed
- `pnpm test:scenario` — 261 passed, 15 skipped
- `pnpm exec eve eval --strict` from `e2e/fixtures/agent-tools` was attempted, but the environment has neither `AI_GATEWAY_API_KEY` nor `VERCEL_OIDC_TOKEN`, so its 15 model-backed evals could not authenticate

### DIFF
--- a/packages/eve/src/execution/hook-ownership.ts
+++ b/packages/eve/src/execution/hook-ownership.ts
@@ -1,0 +1,77 @@
+import type { Hook } from "#compiled/@workflow/core/index.js";
+
+export async function claimHookOwnership<T>(hook: Hook<T>): Promise<void> {
+  let conflict: Awaited<ReturnType<Hook<T>["getConflict"]>>;
+  try {
+    conflict = await hook.getConflict();
+  } catch (error) {
+    return await disposeAndThrow(hook, normalizeHookClaimError(error, hook.token));
+  }
+
+  if (conflict !== null) {
+    return await disposeAndThrow(hook, createHookConflictError(hook.token, conflict.runId));
+  }
+}
+
+export async function closeHookIterator<T>(iterator: AsyncIterator<T>): Promise<void> {
+  if (typeof iterator.return === "function") {
+    await iterator.return(undefined);
+  }
+}
+
+export async function disposeHook(hook: {
+  dispose?: () => unknown;
+  [Symbol.dispose]?: () => unknown;
+}): Promise<void> {
+  const explicitDispose = hook.dispose;
+  if (typeof explicitDispose === "function") {
+    await explicitDispose.call(hook);
+    return;
+  }
+
+  const symbolDispose = hook[Symbol.dispose];
+  if (typeof symbolDispose === "function") {
+    await symbolDispose.call(hook);
+  }
+}
+
+async function disposeAndThrow(hook: Hook<unknown>, error: unknown): Promise<never> {
+  try {
+    await disposeHook(hook);
+  } catch {
+    // The claim failure is authoritative; cleanup must not replace it.
+  }
+  throw error;
+}
+
+function normalizeHookClaimError(error: unknown, token: string): unknown {
+  if (!isHookConflictError(error)) {
+    return error;
+  }
+
+  return createHookConflictError(
+    typeof error.token === "string" ? error.token : token,
+    typeof error.conflictingRunId === "string" ? error.conflictingRunId : undefined,
+  );
+}
+
+function isHookConflictError(
+  error: unknown,
+): error is Error & { readonly conflictingRunId?: unknown; readonly token?: unknown } {
+  return error instanceof Error && error.name === "HookConflictError";
+}
+
+function createHookConflictError(
+  token: string,
+  conflictingRunId?: string,
+): Error & {
+  readonly conflictingRunId?: string;
+  readonly token: string;
+} {
+  const owner = conflictingRunId === undefined ? "" : ` (run "${conflictingRunId}")`;
+  return Object.assign(new Error(`Hook token "${token}" is already in use${owner}`), {
+    conflictingRunId,
+    name: "HookConflictError",
+    token,
+  });
+}
--- a/packages/eve/src/execution/hook-ownership.ts
+++ b/packages/eve/src/execution/hook-ownership.ts
@@ -49,6 +49,7 @@ function normalizeHookClaimError(error: unknown, token: string): unknown {
     return error;
   }
 
+  // Legacy worlds reject here when they cannot identify the owning run.
   return createHookConflictError(
     typeof error.token === "string" ? error.token : token,
     typeof error.conflictingRunId === "string" ? error.conflictingRunId : undefined,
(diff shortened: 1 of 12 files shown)


packages/eve/src/execution/workflow-runtime.ts
```diff
@@ -126,8 +126,7 @@ export function createWorkflowRuntime(config: {
         payloads: [input.payload],
       };
       try {
-        const hook = normalizeWorkflowHook(await getHookByToken(input.continuationToken));
-        await resumeHook(input.continuationToken, hookPayload);
+        const hook = normalizeWorkflowHook(await resumeHook(input.continuationToken, hookPayload));
```
pranaygp, line 129: Nice — folding `getHookByToken` + `resumeHook` into one call removes the lookup-to-resume race. ✅ Confirmed against `@workflow/core@5.0.0-beta.24`: `resumeHook` returns `Promise<Hook>`, and the `@workflow/world` `Hook` carries `runId: string`, so `hook.runId` is sound.

Non-blocking: this return contract is now load-bearing for `deliver()` but is only asserted via a mock (`resumeHookMock.mockResolvedValue({ runId })`) in `workflow-runtime.test.ts`. The integration tier drives `resumeHook` directly rather than through `runtime.deliver`, so the real round-trip of `deliver()` → owning `runId` isn't covered against the local world. An integration assertion would lock this cross-package contract down so a future SDK bump can't silently break it.

pranaygp (commented):
Reviewed the full diff and traced the design against the actual Workflow SDK (`@workflow/core@5.0.0-beta.24`, `@workflow/world@5.0.0-beta.13`) by reading the published `.d.ts`. **No blocking issues** — this is a clean, well-tested change and the semantics line up precisely with the SDK contract.

What I verified:
- `getConflict()` resolves with the conflicting `Run` (`.runId: string`) or `null` once registration is committed, and only *rejects* with `HookConflictError` in the rare old-world case where the owner id was never recorded. `claimHookOwnership` handles both, and `Run.runId` is a plain string, so `conflict.runId` is sound.
- `resumeHook()` returns `Promise<Hook>` and the world `Hook` carries `runId: string`, so collapsing `getHookByToken` + `resumeHook` into a single call is correct and removes the lookup-to-resume race. `HookNotFoundError` is still thrown by `resumeHook`, so the `RuntimeNoActiveSessionError` normalization holds.
- `claim-before-release` ordering in `rekeyHook`, the consolidated `try/finally` cleanup, and the moved `closeHookIterator`/`disposeHook` helpers all check out. The integration test exercises the real local world end-to-end (contender fails before any turn work; owner stays resumable).

A few non-blocking notes left inline (mostly test-coverage / readability). One out-of-diff cleanup: `getHookByToken` is no longer used by any production code, but `packages/eve/src/execution/workflow-steps.test.ts:108` still mocks it — vestigial, safe to drop in a follow-up.

Note: I reviewed statically and could not execute the suites in this worktree (no installed deps); relying on the author's reported green runs + CI.

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -226,46 +214,59 @@ async function runDriverLoop(input: {
 
   const rekeyHook = async (nextToken: string): Promise<void> => {
     if (!nextToken || (hook !== undefined && nextToken === parkToken)) return;
-    await closeParkHook();
-    createParkHook(nextToken);
+
+    // Claim the replacement before releasing the current token. A failed
+    // claim leaves the active hook intact until normal failure cleanup.
+    const nextHook = createHook<HookPayload>({ token: nextToken });
+    await claimHookOwnership(nextHook);
```
pranaygp, line 235: `getConflict()` is a workflow suspension point — per the SDK docs, awaiting it "suspends the workflow to commit the hook registration." Since `rekeyHook` now runs before the first turn (for tokened sessions) and on every token change, channels that re-key per turn (e.g. Slack thread tokens) pay one extra durable commit per re-key versus the old synchronous `createParkHook`. That's the intended claim-before-release tradeoff — just flagging to confirm it was weighed for high-re-key channels.

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -226,46 +214,59 @@ async function runDriverLoop(input: {
...
+
+    try {
+      await closeParkHook();
+    } catch (error) {
+      await disposeHook(nextHook);
```
pranaygp: This branch — claim succeeded but releasing the old hook threw — doesn't appear to be covered. The sibling failures are (conflicting claim before first turn; conflicting rekey candidate disposed before touching the active hook), but a throwing `closeParkHook()` after a successful claim isn't.

Also worth a look: if `closeParkHook()` throws partway (e.g. `closeHookIterator(iterator)` throws), it leaves `hook`/`iterator` still pointing at the old hook, so the outer `finally` will call `closeParkHook()` again → a second `disposeHook(hook)` on the old hook. Harmless if the SDK's `dispose` is idempotent, but worth a test or a `best-effort` note here.

packages/eve/src/execution/hook-ownership.ts
```diff
@@ -0,0 +1,77 @@
...
+async function disposeAndThrow(hook: Hook<unknown>, error: unknown): Promise<never> {
+  try {
+    await disposeHook(hook);
+  } catch {
+    // The claim failure is authoritative; cleanup must not replace it.
+  }
+  throw error;
+}
+
+function normalizeHookClaimError(error: unknown, token: string): unknown {
```
pranaygp, line 47: Per the SDK, `getConflict()` only *rejects* with `HookConflictError` in the rare old-world case where the owning run id was never recorded — i.e. whenever this catch path runs, `conflictingRunId` is by definition absent, so the normalized error will always have `conflictingRunId: undefined` (which the test confirms). The field names are correct (the SDK's `HookConflictError` does carry `token`/`conflictingRunId`), so this is fine as defensive code — a one-line comment noting "this path implies an unidentifiable owner" would save the next reader a trip through the SDK source to understand why the run id is dropped.

PR #257, opened 2026-06-24
fix(vercel): honor author-supplied runtime instead of forcing the eve image

### LINKED ISSUE #239: vercel() sandbox backend ignores the runtime option and always runs vercel/eve:latest
### Summary

`vercel()` sandbox backend ignores the `runtime` option and always runs `vercel/eve:latest`.

When a sandbox is defined with `defineSandbox({ backend: vercel({ runtime: "python3.13" }) })`, the requested `runtime` is silently dropped. The Vercel binding hardcodes the base image to `vercel/eve:latest` and does not forward `runtime` to the `@vercel/sandbox` SDK `create()` call, so the sandbox never runs the requested runtime — with no error or warning.

### Version

eve `0.13.1`

### Where (from the published dist)

- `dist/src/execution/sandbox/bindings/vercel-create-sdk.js` sets `__image: VERCEL_EVE_SANDBOX_IMAGE` (= `"vercel/eve:latest"`) on the SDK create.
- `runtime` does not appear in that create path (it is not forwarded from `vercel({ runtime })` to the SDK).

### Reproduce

1. `defineSandbox({ backend: vercel({ runtime: "python3.13" }) })`.
2. In a tool: `const s = await ctx.getSandbox(); await s.run({ command: "cat /etc/os-release; python3 --version" })`.
3. Observed: the base is `vercel/eve:latest` (currently Ubuntu) with its own bundled Python — NOT the `python3.13` runtime that was requested.

### Impact

1. **Silent option drop** — `runtime` is accepted by `vercel()` but ignored at create time; no error/warn. A footgun: callers believe they're on the runtime they asked for.
2. **Wrong base for native-toolchain workloads** — anything that depends on the requested runtime's base (packages, package manager, Python version) gets `vercel/eve:latest` instead. (Concretely: a bootstrap that assumed the python3.13 base's tooling broke because `vercel/eve:latest` lacks `bzip2`.)
3. **Reproducibility** — `vercel/eve:latest` is a mutable tag with no way to pin a digest, so the sandbox base can drift between deploys (we observed it as a different distro than expected). For workloads that must reproduce results exactly, an unpinnable base is a real problem.

### Ask

Either honor the `runtime` option in the Vercel binding, **or** document explicitly that `vercel()` always uses `vercel/eve:latest` **and** expose a way to (a) select the runtime and (b) pin the base image by digest for reproducible builds.

serhiizghama:
Closes #239.

`vercel({ runtime: "python3.13" })` was silently ignored. `createVercelEveImageSandbox` always spread `__image: "vercel/eve:latest"` onto the create call, and since `__image` takes precedence over `runtime` in the SDK, the requested runtime never reached the sandbox — no error, no warning, just the default image.

The eve image is only meant to be a default base, so I made it conditional: when the author supplies a `runtime`, the image is left off and the runtime flows through to `Sandbox.create` as-is. This mirrors how `source` already works as a template base — `runtime` becomes the base for the template create and is stripped from session creates by the existing snapshot path. The base setup script only requires bash, which the supported runtimes provide, so nothing else changes.

There was already a test asserting runtime reaches the SDK, but it injected a mock `createSandbox` that bypasses the `__image` binding, so the real drop went unnoticed. Added a test that goes through the default binding and checks the runtime is forwarded and `__image` is absent.

### DIFF
--- a/.changeset/vercel-honor-runtime.md
+++ b/.changeset/vercel-honor-runtime.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+The Vercel sandbox backend now honors an author-supplied `runtime` (e.g. `vercel({ runtime: "python3.13" })`) instead of always forcing the `vercel/eve:latest` image. The default eve image is still used when no runtime is requested.
--- a/packages/eve/src/execution/sandbox/bindings/vercel-create-sdk.ts
+++ b/packages/eve/src/execution/sandbox/bindings/vercel-create-sdk.ts
@@ -24,11 +24,17 @@ export async function createVercelEveImageSandbox(input: {
   readonly createOptions: VercelSandboxCreateParams;
   readonly sandboxModule: VercelModule;
 }): Promise<VercelSandbox> {
-  const createOptions: VercelSandboxCreateParams = {
-    ...input.createOptions,
-    __image: VERCEL_EVE_SANDBOX_IMAGE,
-  };
+  // The eve base image is only a default. `__image` takes precedence over
+  // `runtime` in the SDK, so forcing it unconditionally silently drops an
+  // author-supplied runtime — apply it only when no runtime was requested.
+  const createOptions: VercelSandboxCreateParams = hasAuthorRuntime(input.createOptions)
+    ? input.createOptions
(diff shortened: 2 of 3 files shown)


serhiizghama:
Issue #239 has been closed since, so this isn't needed anymore.

PR #259, opened 2026-06-24
fix(eve): workflow runtime - centralize queue namespace

AndrewBarba:
## Summary

Centralize eve's Workflow queue namespace initialization behind one runtime boundary so every queue-producing Workflow API observes `WORKFLOW_QUEUE_NAMESPACE=eve` automatically.

- set the environment variable in exactly one module: `queue-namespace.ts`
- expose the Workflow runtime through one eve-owned `runtime.ts` facade
- route static consumers, generated Workflow aliases, callback handlers, and tests through that facade
- reject future direct Workflow runtime imports or namespace writes with `guard:invariants`
- remove the redundant Vercel function environment configuration
- include a patch changeset for the published `eve` package

## Root cause

The namespace was previously assigned at individual call sites. Correctness therefore depended on every new Workflow consumer remembering to set the environment variable before calling `start`, `resumeHook`, `getRun`, or another queue-producing API.

That was structurally unsafe: a valid runtime import could bypass namespace initialization, and the same setup was duplicated across source code, tests, and generated deployment configuration.

## Implementation

`src/internal/workflow/queue-namespace.ts` now owns the only assignment to `WORKFLOW_QUEUE_NAMESPACE`. `src/internal/workflow/runtime.ts` imports that module before re-exporting the vendored Workflow runtime.

All runtime consumers use the single facade. Call sites that previously used inline dynamic imports were tested independently with top-level imports; all four passed the build-output scenario, so no second lazy facade is needed.

Historical generated specifiers such as `workflow/api` and `workflow/runtime` resolve to the same eve facade. The intermediate Workflow wrapper imports that resolved facade, ensuring generated deployments cannot bypass initialization.

Invariant rule 33 makes this boundary mechanical:

- raw value imports of the vendored Workflow runtime are rejected outside the facade
- direct `workflow/api` and `workflow/runtime` imports are rejected in source code
- writes to `WORKFLOW_QUEUE_NAMESPACE` are rejected outside the canonical namespace module
- type-only imports remain allowed

The generated `workflowEntrypoint` still receives `{ namespace: "eve" }`; that controls Workflow handler metadata and is derived from the same central constant rather than being another environment assignment.

## Build pipeline

The Workflow builder intentionally writes an intermediate `workflows.mjs` wrapper with `bundleFinalOutput: false`. eve then injects step registrations, rewrites runtime imports and the workflow-code literal, mirrors the wrapper into the target-specific Nitro directory, and lets Nitro or the Vercel function emitter perform the actual final bundle.

This PR does not change Rolldown's code-splitting configuration or the deployed bundle shape. The integration test covers the production path by asserting that the intermediate wrapper imports the canonical eve runtime facade instead of bare `workflow/runtime`.

## User and developer impact

- Existing Workflow queues continue using the `eve` prefix.
- Runtime consumers no longer need to configure or remember the namespace.
- New code that bypasses the canonical facade fails the invariant guard in CI.
- Vercel deployment output no longer carries a duplicate environment setting.
- Deployment bundling behavior and bundle size are unchanged.

## Validation

- `pnpm fmt`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm typecheck`
- `pnpm build`
- `pnpm test:unit` — 3,992 tests passed
- `pnpm test:integration` — 370 tests passed
- full scenario tier with the serial runner — 262 passed, 15 skipped
- `bin-build-output.scenario.test.ts` after each dynamic-to-static import conversion and after the combined change — 5 tests passed
- focused intermediate-wrapper integration test — 4 tests passed

The relevant fixture eval was attempted, but this environment does not provide `VERCEL_OIDC_TOKEN` or `AI_GATEWAY_API_KEY`, so model-backed e2e evaluation could not complete.


packages/eve/src/internal/workflow-bundle/builder-support.ts
```diff
@@ -408,6 +409,7 @@ export const POST = workflowEntrypoint(workflowCode, { namespace: ${JSON.stringi
     plugins: [createWorkflowVirtualEntryPlugin(workflowFunctionCode)],
     write: false,
     output: {
+      codeSplitting: false,
```
ruiconti: This is a bit sus as this could blow up bundle sizes? Why is this needed?

AndrewBarba: Updated PR description with details

ruiconti (approved):

PR #269, opened 2026-06-25
fix(slack): ack view_submission with an empty 200 body so the modal closes

ya2s:
### Description

Free-form HITL input in Slack (an `ask_question`/input request with no options) renders a modal. When the user submits it, `handleViewSubmission` accepts the value and delivers it to the session via `ctx.send({ inputResponses: [...] })`, **but acknowledges the `view_submission` with a 200 whose body is `"ok"`**.

A `view_submission` must be acknowledged with an **empty 200 body** (or a `response_action` JSON). A non-empty body like `"ok"` is invalid, so Slack shows a generic error and **leaves the modal open** — even though the submission was accepted and the answer was already delivered to the agent. From the user's side it looks like the value went through but the form errored / the modal won't close.

**Fix:** return an empty body (`new Response(null, { status: 200 })`) from `handleViewSubmission`. Block Actions ignore the response body, so the `"ok"` ack in `handleInteractionPost` is intentionally left unchanged.

Repro: trigger a free-form HITL question over Slack, type a value, submit → value reaches the agent but the modal shows an error and stays open. Still reproduces on the latest release (0.13.6).

- Added a unit test asserting `handleInteractionPost` acks a `view_submission` with `status 200` and an **empty** body.
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack/interactions.test.ts` → 2 passed.
- `pnpm --filter eve run typecheck` → passes. `pnpm fmt` / `pnpm lint` clean.

### PR Checklist

> Note: no prior issue — happy to open one if you'd prefer. This is a small, self-contained correctness fix with the repro above.

### DIFF
--- a/.changeset/slack-view-submission-empty-ack.md
+++ b/.changeset/slack-view-submission-empty-ack.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Acknowledge Slack `view_submission` (the free-form HITL input modal) with an empty 200 body instead of `"ok"`. A non-empty body is invalid for a view submission, so Slack showed a generic error and left the modal open even though the submitted value was accepted and delivered. Returning an empty body dismisses the modal. Block Actions acks are unchanged (they ignore the response body).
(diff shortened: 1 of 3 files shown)


ya2s:
Superseded by #271, which carries the same fix as a verified-signed commit (this branch's commit was unsigned). Continuing review there.

PR #270, opened 2026-06-25
feat(eve): add /connect setup flow

ruiconti:
## What

- Add a searchable `/connect` command to the local dev TUI for Linear and Notion MCP servers through Vercel Connect.
- Try the provider's canonical connector first. If it cannot attach, let the user search compatible connectors or create one with a chosen name.
- When the directory is unlinked, selecting a server runs the same create-or-link project flow used by `/model`. Logged-out rows still point to `/vc:login`.
- Use the standard setup pulse. Browser authorization waits use a yellow pulse and highlight the word `browser` as pending user action.
- Update Linear's catalog entry to its current `/mcp` endpoint.

### Flow ownership

`/connect` owns the MCP catalog, project linking when required, connector provisioning, and the agent connection file. It calls the existing `runLinkFlow` with `create-or-link`; there is no second project picker or separate link command. Cancelling the nested picker returns to the MCP catalog.

### Connector setup

`/connect` writes the attached connector UID into `agent/connections/<name>.ts` and installs `@vercel/connect` before the dev server sees the file. A connector created by the current attempt is removed if attachment or connection-file setup later fails.

The catalog currently shows Linear and Notion because this setup path only covers user-authorized Connect providers. Runtime principal resolution, workflow protocol, and public-channel behavior are unchanged.


ruiconti:
Closed in favor of #327.

PR #271, opened 2026-06-25
fix(slack): ack view_submission with an empty 200 body so the modal closes

ya2s:
### Description

Free-form HITL input in Slack (an input request with no options) renders a modal. On submit, `handleViewSubmission` accepts the value and delivers it to the session via `ctx.send({ inputResponses: [...] })`, **but acknowledges the `view_submission` with a 200 whose body is `"ok"`**.

A `view_submission` must be acknowledged with an **empty 200 body** (or a `response_action` JSON). A non-empty body like `"ok"` is invalid, so Slack shows a generic error and **leaves the modal open** — even though the submission was accepted and the answer was already delivered to the agent. From the user's side the value goes through but the modal errors / won't close.

**Fix:** return an empty body (`new Response(null, { status: 200 })`) from `handleViewSubmission`. Block Actions ignore the response body, so the `"ok"` ack in `handleInteractionPost` is intentionally left unchanged. Still reproduces on the latest release (0.13.6).

- Added a unit test asserting `handleInteractionPost` acks a `view_submission` with status 200 and an **empty** body.
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack/interactions.test.ts` → 2 passed.
- `pnpm --filter eve run typecheck` passes; `pnpm fmt` / `pnpm lint` clean.

### PR Checklist

> Supersedes #269 (re-opened with a verified-signed commit). No prior issue — happy to open one if preferred.

### DIFF
--- a/.changeset/slack-view-submission-empty-ack.md
+++ b/.changeset/slack-view-submission-empty-ack.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Acknowledge Slack `view_submission` (the free-form HITL input modal) with an empty 200 body instead of `"ok"`. A non-empty body is invalid for a view submission, so Slack showed a generic error and left the modal open even though the submitted value was accepted and delivered. Returning an empty body dismisses the modal. Block Actions acks are unchanged (they ignore the response body).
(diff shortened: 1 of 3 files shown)


AndrewBarba:
Thanks for the fix. We reopened this under a branch in this repository as #274 so it can go through the normal merge flow for outside-contributor patches.

The replacement PR keeps the same behavior change and includes you as a co-author on the commit. We will merge #274 instead of this PR.

ya2s:
@AndrewBarba 
Thank you! I really appreciate it.

I’ve found a few other bugs as well and would be happy to submit fixes. Would you prefer that I open PRs directly, or would you rather I open issues first and have the changes go through your usual workflow?

Issue #77, opened 2026-06-18
Emit actions.requested before inline tool execution so long-running tools show Running state in the stream


sahkyo:
### What problem are you trying to solve?

Eve’s stream protocol supports a clear tool lifecycle (`actions.requested` → `action.result`), and the web UI maps `input-available` to a **Running** badge. That works well for runtime actions (subagents, HITL, etc.) where the turn parks and events arrive while work is in flight.

**Inline authored tools** (e.g. `generate_image`, `edit_image`) behave differently. In `emitStepActions`, `actions.requested` and `action.result` are emitted **back-to-back after the AI SDK step finishes** — i.e. after the tool has already executed. During a 12–70s FAL/image call, the client typically sees:

- Generic “Working…” (or nothing tool-specific)
- No Running tool card for the actual operation
- A Completed card only at the very end (if at all)

This gets worse in **multi-step flows**. The provider may finish (FAL dashboard shows success), but the frontend can remain stuck in a busy state because:

1. There are no stream events during the long silent execution window
2. The UI cannot tell “model thinking” from “tool running”
3. Reconnection / turn-boundary logic has nothing to anchor on until the step completes

App authors are forced to fake Running UI by inferring state from `message.completed` (`finishReason: "tool-calls"`) and guessing which tool will run — fragile and incomplete.

### Proposed solution

Split inline tool streaming into start and finish, similar to runtime actions:

1. **On tool-call resolution (before `execute` runs):** emit `actions.requested` with the pending batch so clients can render Running immediately.
2. **On tool completion:** emit `action.result` as today.
3. **(Optional)** For tools over a threshold (or opt-in via tool metadata), park the turn as a durable runtime action so long jobs survive disconnects and match subagent UX.

### Acceptance criteria

- After `message.completed` with `finishReason: "tool-calls",` clients receive `actions.requested` before the tool blocks on I/O.
- `useEveAgent` / message reducer can show Running → Completed without app-side hacks.
- Multi-step turns (skill load → image gen) don’t leave the UI stuck while the backend has already finished.
- Behavior for subagents / existing runtime actions stays unchanged.

### Alternatives considered

- **Wrap slow tools in subagents** — works today but wrong abstraction (extra session, no attachments, delegation overhead).
- **App-side optimistic UI** — what we’re doing now; doesn’t fix stuck state or missing `action.result` visibility.
- **New progress events** (`action.progress`) — nice-to-have on top of early `actions.requested`.

cmpadden:
[...]

PR #291, opened 2026-06-25
fix(eve): stream tool calls and results

ruiconti:
## What

- Emit actions.requested for each eligible fullStream tool call.
- Emit the matching terminal action.result at that stream position and skip the duplicate post-step event.
- Render tool calls by callId in the TUI so repeated payloads do not collapse into one row.

## Why this shape

[Issue #77](https://github.com/vercel/eve/issues/77) exposed the gap: a slow inline executor was invisible until it returned because the request event came from the completed step. fullStream already yields the call and terminal result in source order. Projecting that order gives clients the running state and keeps the renderer timeline accurate.

Approval pauses and runtime dispatch keep their current owners. Consumers correlate incremental requests and results by callId, not a step-wide batch.

The local regression holds a real execute call and asserts that its actions.requested event exists when execution begins.


packages/eve/src/harness/emission.ts
```diff
@@ -408,6 +452,33 @@ export async function emitStreamContent(
     );
   };
 
+  const observeStreamToolCall = async (toolCall: TypedToolCall<ToolSet>): Promise<void> => {
```
AndrewBarba: is observe the right name? seems like this is actually emitting something

ruiconti: 🤮 can't believe I missed this

AndrewBarba (approved):

PR #306, opened 2026-06-25
docs(eve): document self-hosted workflow world version pin and reverse-proxy routes

pranaygp:
## Summary

Documents two undocumented gotchas that tripped up a self-hoster wiring up the Postgres-backed workflow world for self-hosting. Both stem from a self-hosting writeup.

### 1. Reverse-proxy gap (`/.well-known/workflow/`)

eve serves an internal workflow dispatch route at `/.well-known/workflow/v1/flow` that the workflow world uses for run callbacks. A reverse proxy that forwards only `/eve/*` (the natural instinct) lets sessions **start** but makes runs **silently stall forever** — the callbacks never reach eve. The code already treats `eve/` and `.well-known/workflow/` as co-equal required prefixes (`EVE_VERCEL_FUNCTION_PREFIXES`), but the docs never told self-hosters this. Now they do: a proxy/ingress in front of eve must forward **both** `/eve/` and `/.well-known/workflow/`.

### 2. World version trap (version pin)

The docs referenced `experimental.workflow.world: "@workflow/world-postgres"` with no install command and no version guidance. `pnpm add @workflow/world-postgres` resolves the npm `latest` tag, which can be **incompatible** with the `@workflow/*` line this eve release bundles (currently the `5.0.0-beta` line). A mismatched world (e.g. a `4.x` world against a `5.x` core) throws a cryptic `ZodError: invalid_union` during run replay. The docs now give a concrete install command with an explicit pin (`pnpm add @workflow/world-postgres@5.0.0-beta.x`), warn that `latest` can lag, and make the Postgres credentials-in-env guidance concrete.

## Changes

- `docs/guides/deployment.md` (§8 "Deploy without Vercel"): expanded the custom-world bullet with the version-pin guidance, added a reverse-proxy bullet, and updated the closing HTTP-contract sentence to name the `/.well-known/workflow/` dispatch route.
- `docs/agent-config.md` ("Workflow world"): added the install block with the explicit pin, the `latest`-tag warning, and concrete Postgres credentials guidance.
- `docs/concepts/execution-model-and-durability.md` ("Selecting custom workflow world packages"): added a short compatibility note cross-linking to the deployment guide for the install command.

## Notes

- Docs-only change; no changeset.
- `pnpm docs:check` passes (69 files validated, snippets resolve, MDX compiles).
- Pairs with a sibling runtime PR (branch `pgp/eve-self-host-workflow-runtime`) that makes `eve start` work with a custom world.

[...]

PR #308, opened 2026-06-25
fix(eve): self-hosted workflow queue handler + world version-compat guard

pranaygp:
Two related fixes for the self-hosted Postgres workflow-world story. Pairs with the sibling docs PR on branch `pgp/eve-self-host-workflow-docs`.

## FIX 1 — `eve start` registers the workflow queue handler for self-hosted (non-Vercel) production

**Bug:** With a custom/local workflow world configured (e.g. `@workflow/world-postgres`) and the documented production path `eve build && eve start`, the direct queue→bundle handler binding was never registered. Jobs the world dispatched to `/.well-known/workflow/v1/flow` came back `400 {"error":"Unhandled queue"}` and runs sat in `pending` forever. The only workaround was `eve dev --no-ui`, which is semantically wrong for production.

**Root cause:** In `packages/eve/src/internal/nitro/host/configure-nitro-routes.ts` the `directHandlerEntries` were gated on `nitro.options.dev`. The HTTP flow route is registered in production, but its queue→bundle binding (generated only when `directHandlerEntries` is non-empty) was not. The real axis is not dev-vs-prod; it is *who drives the queue*: the local/configured world drives it in-process in both `eve dev` and self-hosted `eve start`, whereas Vercel dispatches through its managed queue trigger over HTTP.

**Fix:** Register the binding whenever the local world drives the queue — `dev`, OR a non-Vercel build (`!isVercelBuildEnvironment()`) with a configured custom world (`manifest.config.experimental.workflow.world` set) — and never for Vercel-managed deploys. `isVercelBuildEnvironment()` is now exported from `packages/eve/src/internal/application/paths.ts`. The Vercel path is preserved exactly (binding still absent when `VERCEL` is set at build).

**Final gate condition:**
```ts
const hasConfiguredWorkflowWorld =
  preparedHost.compileResult.manifest.config.experimental?.workflow?.world !== undefined;
const localWorldDrivesQueue =
  nitro.options.dev || (!isVercelBuildEnvironment() && hasConfiguredWorkflowWorld);
```
The added custom-world clause keeps non-custom-world production builds byte-for-byte identical to today (no binding); only custom-world self-hosted prod — the bug — changes.

### End-to-end wiring question (resolved)
I traced `@workflow/world-local`'s queue dispatch (`queue.js`): if a direct handler is registered for the queue prefix it calls it **in-process** and short-circuits the HTTP path entirely; only when no direct handler exists does it fall back to `fetch(${baseUrl}/.well-known/workflow/v1/...)` where `baseUrl` comes from `WORKFLOW_LOCAL_BASE_URL`/`PORT`. So registering the binding (FIX 1) is the **complete** fix for self-hosted runs and does **not** require the `installWorkflowLocalQueueEnvironment` env wiring used by `eve dev` — the world never reaches the HTTP fallback. As a belt-and-suspenders, `startProductionServer` already sets `PORT` on the spawned child, so even the fallback would resolve `http://localhost:${PORT}`. Full end-to-end validation against a real Postgres world remains out of scope for unit tests (no Postgres in CI); the in-process dispatch path is exercised by the existing/added unit tests.

### Tests (FIX 1)
- Retargeted `does not register direct workflow queue handlers in production builds` → `...for Vercel production builds` (stubs `VERCEL=1` via `vi.stubEnv`).
- Added `registers direct workflow queue handlers for self-hosted production builds with a configured world`.
- Added `does not register direct workflow queue handlers for self-hosted production builds without a configured world`.
- `beforeEach` now `vi.unstubAllEnvs()` to keep env from leaking; `paths.js` is mocked to a real `isVercelBuildEnvironment` that reads `process.env.VERCEL`, so `vi.stubEnv` still drives behavior while avoiding the heavy import graph.

## FIX 2 — boot-time workflow world version-compat guard

**Trap:** `pnpm add @workflow/world-postgres` resolves to npm `latest` (4.x), incompatible with the `@workflow/*` 5.0.0-beta line eve bundles (`@workflow/core@5.0.0-beta.24`). The mismatch surfaced as a cryptic `ZodError: invalid_union` deep in workflow replay.

**Fix:** A simple, low-risk eve-side guard that fails loud and early. New pure, exported, doc-commented helper `assertWorkflowWorldCompatibility` (`packages/eve/src/internal/workflow/world-compatibility.ts`) takes already-parsed JSON and throws only on a **definite** major / prerelease-line mismatch; it no-ops when the world declares no `@workflow/*` dep or the range is unparseable (avoids false-positive boot failures). The expected line comes from a single source of truth — eve's own `package.json` via new `resolveExpectedWorkflowVersion()` in `packages/eve/src/internal/application/package.ts` — so no version is hardcoded. The world package name is threaded from the generated bootstrap (`compiled-artifacts.ts`) into `installConfiguredWorkflowWorld({ module, packageName })`, which reads the world's `package.json` via `createRequire(import.meta.url).resolve(...)` and calls the pure helper. No new runtime dependency (no semver lib; a tiny major/prerelease-line parse). The deeper, fully version-aware fix is being proposed upstream in `@workflow/core` via a separate issue.

Example thrown message:
`Configured Workflow world "@workflow/world-postgres" targets @workflow/core 4.x, but this eve release requires the @workflow/core 5.0.0-beta line. Install a matching world, e.g. \`pnpm add @workflow/world-postgres@5.0.0-beta.24\`.`

### Tests (FIX 2)
- `world-compatibility.test.ts` (pure helper): older major → throws actionable message (asserts name + line + install command); same prerelease line / caret range → passes; falls back to `@workflow/world` and reads `peerDependencies`; no `@workflow/*` dep / unparseable range / unparseable expected → no-op.
- `package.test.ts`: `resolveExpectedWorkflowVersion` reads eve's `@workflow/core` line.
- `compiled-artifacts.test.ts`: bootstrap body assertion updated to include `packageName`.

## Files changed
- `packages/eve/src/internal/nitro/host/configure-nitro-routes.ts` — new gate; import `isVercelBuildEnvironment`.
- `packages/eve/src/internal/application/paths.ts` — export `isVercelBuildEnvironment` (+ doc).
- `packages/eve/src/internal/nitro/host/configure-nitro-routes.test.ts` — retargeted/added tests; env hygiene; mock `paths.js`.
- `packages/eve/src/internal/workflow/world-compatibility.ts` — new pure helper.
- `packages/eve/src/internal/workflow/world-compatibility.test.ts` — new helper tests.
- `packages/eve/src/internal/workflow/configure-world.ts` — boot-time compat check; thread `packageName`.
- `packages/eve/src/internal/application/package.ts` — new `resolveExpectedWorkflowVersion`.
- `packages/eve/src/internal/application/package.test.ts` — test for the above.
- `packages/eve/src/internal/application/compiled-artifacts.ts` — pass `packageName` to bootstrap call.
- `packages/eve/src/internal/application/compiled-artifacts.test.ts` — updated bootstrap assertion.
- `.changeset/self-host-workflow-runtime.md` — patch changeset.

## Validation
- `pnpm install --frozen-lockfile` ✓
- `pnpm --filter eve typecheck` ✓
- `oxlint packages/eve/src` ✓ (clean) · `pnpm guard:invariants` ✓
- Unit: touched files — 89 passed (9 files). Integration/scenario/e2e left to CI.

### DIFF
--- a/.changeset/self-host-workflow-runtime.md
+++ b/.changeset/self-host-workflow-runtime.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Self-hosted `eve start` now registers the workflow queue handler for custom (non-Vercel) worlds, so jobs dispatched by a configured world no longer return `Unhandled queue` or leave runs stuck `pending` — and you no longer need `eve dev --no-ui` to run a local world in production. eve also fails fast at boot with an actionable error when a configured workflow world's `@workflow/*` version is incompatible with the line eve bundles, instead of surfacing a cryptic `ZodError` deep in workflow replay.
--- a/packages/eve/src/internal/application/compiled-artifacts.test.ts
+++ b/packages/eve/src/internal/application/compiled-artifacts.test.ts
@@ -59,7 +59,7 @@ describe("createCompiledArtifactsBootstrapSource", () => {
 
     expect(source).toContain('import * as workflowWorldModule from "@acme/eve-world";');
     expect(source).toContain(
-      "await installConfiguredWorkflowWorld({ module: workflowWorldModule });",
+      'await installConfiguredWorkflowWorld({ module: workflowWorldModule, packageName: "@acme/eve-world" });',
     );
   });
 });
--- a/packages/eve/src/internal/application/compiled-artifacts.ts
+++ b/packages/eve/src/internal/application/compiled-artifacts.ts
@@ -174,7 +174,10 @@ function createWorkflowWorldBootstrapBody(
     return [];
   }
 
-  return ["", "await installConfiguredWorkflowWorld({ module: workflowWorldModule });"];
+  return [
+    "",
+    `await installConfiguredWorkflowWorld({ module: workflowWorldModule, packageName: ${JSON.stringify(world)} });`,
+  ];
 }
 
 function createInstrumentationPluginSource(input: {
--- a/packages/eve/src/internal/application/package.test.ts
+++ b/packages/eve/src/internal/application/package.test.ts
@@ -1,6 +1,9 @@
 import { describe, expect, it } from "vitest";
 
-import { resolveWorkflowModulePath } from "#internal/application/package.js";
+import {
+  resolveExpectedWorkflowVersion,
+  resolveWorkflowModulePath,
+} from "#internal/application/package.js";
 
 describe("resolveWorkflowModulePath", () => {
   it("resolves historical workflow specifiers to narrowed runtime modules", () => {
@@ -17,5 +20,13 @@ describe("resolveWorkflowModulePath", () => {
     expect(resolveWorkflowModulePath("workflow/runtime")).toMatch(
       /\/src\/internal\/workflow\/runtime\.ts$/,
     );
+  });
+});
+
+describe("resolveExpectedWorkflowVersion", () => {
+  it("reads the @workflow/core line from eve's own package.json", () => {
+    // Single source of truth: eve declares the workflow line it bundles in its
+    // own package.json, so this resolves to a concrete prerelease version.
+    expect(resolveExpectedWorkflowVersion()).toMatch(/^\d+\.\d+\.\d+/);
   });
 });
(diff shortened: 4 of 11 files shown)


pranaygp:
Part of the self-hosted workflow-world improvements (from a self-hosting writeup). Pairs with the docs in #306 (reverse-proxy `/.well-known/workflow/` forwarding + world version-pin guidance). The shallow boot-time guard added here is a stopgap; the durable core↔world compatibility check is proposed upstream in vercel/workflow#2638.

AndrewBarba (approved):

PR #320, opened 2026-06-26
refactor(eve): turns - own subagent waits inside the active turn

AndrewBarba:
## Summary

Restores a core lifecycle invariant: **one `turnWorkflow` owns one complete logical turn — until it completes or parks for the next user turn.**

Previously, dispatching a subagent ended the turn: the wait moved into `workflowEntry` and a *replacement* turn consumed the result. That split one logical turn across multiple durable workflows and made the parking boundary depend on subagent capability (local vs. nested vs. dynamic Workflow vs. HITL). Now the turn stays alive across every local, dynamic-workflow, and remote subagent wait, and child completion resumes the same turn.

## Responsibility matrix

| Responsibility | Previous | New |
| --- | --- | --- |
| Session lifetime | `workflowEntry` | `workflowEntry` |
| Public delivery hook | `workflowEntry` | `workflowEntry` |
| Logical turn lifecycle | Split across driver and turn | Entirely inside `turnWorkflow` |
| Child dispatch and waiting | Driver-owned | Turn-owned |
| Child HITL state | Driver-owned | Turn-owned |
| Child callbacks | Public session continuation token | Private turn inbox |
| Terminal session handling | Driver | Driver |

Session-level responsibilities stay with the driver; everything that *continues the current turn* moves under the active turn.

## Why now

Durable ownership only gets more expensive to change as behavior accumulates on it. Cancellation, timeouts, retries, nested delegation, HITL, and tracing each have to answer "which workflow owns this, and who may mutate the session?" — and under the split model every one of them needs cross-workflow state transfer and race-specific compatibility logic. Fixing the boundary now keeps the driver small, stops `NextDriverAction` from growing into a general-purpose turn-continuation protocol, and gives future lifecycle work one place to enforce ordering and cleanup. After 1.0, correcting it would mean supporting both ownership models for in-flight workflows.

## Owners

The boundary is explicit instead of spread across callsites:

- **`TurnControlReceiver`** — the driver's control hook, pending reads, continuation-token rekeys, delivery races, buffering, and cleanup for one turn
- **`TurnExecutionCursor`** — the turn's durable session state, serialized context, continuation-token sync, step inputs, and terminal transition
- **`TurnControlPayload`** — the private protocol between them

The durable turn input keeps its `completionToken` field (the wire shape older pinned drivers emit), so no migration or version bump is needed; `controlToken` is just the internal name past that boundary.

## Delivery and race safety

Child HITL is the only path crossing the public-session ↔ active-turn boundary. It uses a request / forward / accept-or-cancel handshake: the turn requests one public delivery, `workflowEntry` forwards it provisionally, and the turn either accepts it or cancels (e.g. the child finished first) — a cancelled delivery stays buffered, in arrival order, for the next turn. Request IDs keep repeated step execution idempotent, and the turn claims its private inbox before dispatching any child so early completions and callbacks queue safely before it waits.

Public-hook rekeying is lossless and ordered: the driver claims the replacement before releasing the old hook, keeps reads armed on both, and drains committed payloads in Workflow event order across disposal. Each `resumeHook` event is observed exactly once, or loses the race to `hook_disposed` and gets `HookNotFoundError`. A real-Workflow integration test pauses disposal and verifies both commit orders.

## Compatibility

Turn input stays at version 1; the new behavior is negotiated through an optional capability:

- new drivers send `driverCapabilities.turnInbox: true` and take the turn-owned path
- older pinned drivers omit it, so latest turn code falls back to the legacy entry-owned path

Forward deployment needs no lockstep rollout or migration.

## Scope & validation

Covers local subagents, nested subagents with proxied HITL, dynamic Workflow calls, remote-agent callbacks, immediate remote-dispatch failures, and concurrent child-result / public-delivery races.

Validated by the unit suite (incl. delivery-hook ordering/cleanup), workflow-entry and real-Workflow rekey-race integration tests, the subagent-HITL proxy suite, local + Vercel e2e (subagents, dynamic Workflow, HITL), and typecheck / lint / invariant guard / bundle analysis.

---

> [!NOTE]
> **High Risk**
> This rewrites core durable session/turn orchestration—hook ownership, delivery routing, and child callback targets—with a large blast radius; regressions would show up as lost messages, stuck turns, or broken subagent/HITL flows.
> 
> **Overview**
> **Moves subagent dispatch, child callbacks, and proxied HITL off the session driver and into the active `turnWorkflow`**, so one logical turn stays in one durable child workflow until it `done`s or `park`s instead of ending the turn and starting a replacement after each child wait.
> 
> New drivers negotiate **`driverCapabilities.turnInbox`** on turn input. When set, `turnWorkflow` claims a private **turn inbox**, resolves pending runtime actions in-line (local, dynamic-workflow, and remote), and talks to `workflowEntry` through **`TurnControlPayload`** (`turn-delivery-request` / forward / accept-or-cancel, continuation-token rekeys, terminal `turn-result`). **`TurnControlReceiver`** and **`TurnExecutionCursor`** own the driver-side relay and turn-side state cursor; **`createSessionDeliveryHook`** replaces ad-hoc park-hook rekeying so deliveries committed during token rotation are not lost. Child and remote-agent callbacks target the **turn inbox** via optional `parentContinuationToken` / `callbackToken` rather than the public session continuation token. **Duplicate concurrent turn runs** exit quietly on inbox `HookConflictError` instead of failing the session. Older pinned drivers without `turnInbox` keep the **legacy entry-owned** `dispatch-runtime-actions` path.
> 
> Adds **`agent-workflow-stress`** e2e (mock model, 100 sequential / 50×2 concurrent turns), shared **`coalesceDeliverPayloads`** / **`routeDeliverToChildren`**, and broad unit/integration coverage for the new protocol and delivery-hook races.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8127ba5bda8e8688480b38eeae483732fef903b1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/eve/src/execution/turn-workflow.ts
```diff
@@ -1,39 +1,328 @@
...
+          completionToken: input.completionToken,
+          inbox,
+          iterator,
+          parentWritable: currentStepInput.parentWritable,
+          pendingActionKeys,
+          serializedContext: result.serializedContext,
+          sessionState: result.sessionState,
+          syncToken,
+          workflowInterrupt: result.action === "dispatch-workflow-runtime-actions",
+        });
```
ruiconti: I don't love the scattered files with shallow functions. Makes it harder parse; by doing this we'd remove several functions and potentially a file by inlining it here.


```suggestion
        const input.workflowInterrupt = 
          result.action === "dispatch-workflow-runtime-actions"
        const dispatch = input.workflowInterrupt
          ? dispatchWorkflowRuntimeActionsStep
          : dispatchRuntimeActionsStep;
          
        const dispatchResult = await dispatch({
          callbackBaseUrl: resolveVercelProductionCallbackBaseUrl() ?? getWorkflowMetadata().url,
          parentContinuationToken: input.parentContinuationToken,
          parentWritable: input.parentWritable,
          serializedContext: input.serializedContext,
          sessionState: input.sessionState,
        });
        
        await input.syncToken(dispatchResult.sessionState.continuationToken);
        return waitForRuntimeActionResults({
            ...input,
            initialResults: dispatchResult.results,
            sessionState: dispatchResult.sessionState,
          }); 
```

AndrewBarba: Inlined the single-use dispatch/wait helper; kept the dedicated step boundary so environment-specific callback resolution stays outside replayed workflow code.

ruiconti (approved):

PR #325, opened 2026-06-26
fix(eve): degrade missing attachment bytes on resume instead of failing the turn

### LINKED ISSUE #276: Resuming a session with a prior file attachment fatally fails (Sandbox-ref FilePart references missing file)
### Summary
Resuming a durable session whose history contains an inbound file attachment (e.g. a Slack image) fails fatally because the staged attachment file no longer exists on disk. On resume, eve re-validates/re-hydrates the historical `FilePart` against the sandbox and throws when the file is gone, instead of degrading gracefully. The turn ends in `session.failed`, so the bot never replies.

### Environment
- Slack channel, **Vercel Sandbox** backend.
- Observed on eve **0.11.5**. No related fix found in the 0.11.6–0.13.6 changelog (`handleViewSubmission`/staging unchanged), so it appears to still apply on 0.13.6.

### Repro
1. In Slack, @mention the bot with an **image attached**. eve stages it to `/workspace/attachments/<id>/<name>` and the first turn runs fine (the image is used).
2. Later, @mention again **in the same thread** (a follow-up). eve resumes the session.
3. The turn fails fatally; the bot produces **no response**.

### Error
```
Step "step//eve@0.11.5//turnStep" failed after 3 retries: Sandbox-ref FilePart
references missing file: "/workspace/attachments/<id>/crayon.png".
The staging pipeline invariant (every eve-sandbox: ref has bytes on disk) was violated.
```
(workflow loop → terminal `session.failed`, code `FatalError` / `USER_ERROR`.)

### Expected
Resuming a session whose historical attachment bytes are no longer staged should **not** fatally fail. The missing sandbox-ref file part should be dropped or treated as a recoverable miss (the model already consumed it on the original turn), so follow-up turns in the thread keep working.

### Impact
Any Slack thread that ever included a file/image attachment becomes **unresumable** — every later @mention in that thread silently yields no reply.

serhiizghama:
Fixes #276. Resuming a durable session whose history contains an inbound file attachment fatally fails when the staging sandbox has been torn down between turns — an ephemeral backend means the bytes behind a historical `eve-sandbox:` ref are gone, and `hydrateMessageContent` threw on the missing read, so the turn ended in `session.failed` and the bot never replied.

Hydration now degrades a missing ref to a text part (`"Attached file <path> (<type>) is no longer available in this session."`) and logs a warning, instead of throwing. The run continues, and the model is told the attachment is unreachable rather than chasing a path that no longer resolves. I gave it a distinct message from the normal non-inlinable text reference on purpose — on resume the path really is gone, so pointing the model at it would just invite a failed read.

This only affects the resume case where the bytes are already missing; a freshly staged attachment in the current turn still hydrates as before.

Tests: rewrote the existing "missing file" case to assert the degrade (the leading text survives, the file ref becomes the unavailability notice, no file part reaches the model) and added a regression that stages an image into one sandbox and hydrates the resulting ref against a fresh one — the same shape as a resume after teardown — and asserts the turn survives. Reverting the fix fails both. Full attachment-staging integration + unit suites green, typecheck/oxlint/oxfmt/guard:invariants clean, changeset (patch).

### DIFF
--- a/.changeset/honor-missing-attachment-on-resume.md
+++ b/.changeset/honor-missing-attachment-on-resume.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Resuming a durable session that referenced an inbound file attachment no longer fails the turn when the staging sandbox has been torn down. The missing attachment now degrades to a text reference noting the file is unavailable, so the run continues instead of ending in `session.failed`.
--- a/packages/eve/src/harness/attachment-staging.integration.test.ts
+++ b/packages/eve/src/harness/attachment-staging.integration.test.ts
@@ -611,12 +611,16 @@ describe("hydrateSandboxAttachments (integration)", () => {
     expect(hydrated).toBe(messages);
   });
 
-  it("throws a descriptive error when an inlinable sandbox ref points at a missing file", async () => {
+  it("degrades to a text reference when an inlinable sandbox ref points at a missing file", async () => {
+    // Resuming a durable session whose staging sandbox was torn down
+    // leaves historical attachment refs pointing at bytes that are gone.
+    // Hydration must not fail the whole turn over it — it degrades to a
+    // text part so the run survives (#276).
     const sandbox = mockSandbox({ id: "sbx_missing" });
     const runtime = createTestRuntime();
 
     // Ref pointing at a path that was never written. Use an
-    // inlinable media type so the error path (byte read) fires —
+    // inlinable media type so the byte-read path fires —
     // non-inlinable refs never touch the sandbox.
     const danglingRef = new URL(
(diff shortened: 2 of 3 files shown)


ruiconti:
Thanks for this — the approach is right and we carried it forward in #507 with both of your commits cherry-picked, authorship and sign-off intact (they needed re-signing to satisfy the repo's signed-commit requirement, which a fork PR can't do).

#507 adds two things on top: a corrected root cause (the sandbox is never torn down — session sandbox names embed the deployment id, so any redeploy points the resumed session at a fresh sandbox; we traced the #276 incident to specific deploys) and a recovery hint in the model-facing text. There's also a follow-up plan in #509 to keep the bytes alive across deploys using sandbox drives.

Closing in favor of #507.

PR #327, opened 2026-06-26
feat(eve): add /connect setup flow

ruiconti:
## What

- Add `/connect` to the local dev TUI. It searches the Vercel Connect MCP catalog, links or creates the required Vercel project, attaches the canonical connector when possible, and falls back to choosing or creating one.
- Scaffold the resolved connector UID into `agent/connections`, install `@vercel/connect`, clean up a connector created by a failed attempt, and return to the main prompt after success.
- Keep OAuth in the same durable TUI session. The authorization panel replaces stale challenge details when the callback completes.
- Use the linked project's Vercel OIDC user before the local-dev fallback, so local Connect authorization has the user identity it needs.

## Runtime activation

`/connect` previously printed `Connections added` while the next turn could still run against the old dev snapshot. The rebuild POST went through a virtual Nitro route, whose module context could not reach the watcher held by the development host.

The rebuild endpoint now runs as a Nitro dev handler that closes over that watcher. `/connect` waits for the forced rebuild and rotates an active session before reading the next prompt. The command remains local-only; `eve dev --url` does not start or own a callback host.

The PR also updates the Connect catalog service identifiers, the weather fixture, docs, and the patch changeset.


apps/templates/web-chat-next/agent/channels/eve.ts
```diff
@@ -3,10 +3,10 @@ import { localDev, placeholderAuth, vercelOidc } from "eve/channels/auth";
 
 export default eveChannel({
   auth: [
-    // Open on localhost for `eve dev` and the REPL; ignored in production.
-    localDev(),
     // Lets the eve TUI and your Vercel deployments reach the deployed agent.
     vercelOidc(),
+    // Open on localhost for `eve dev` and the REPL; ignored in production.
+    localDev(),
```
ruiconti, line 9: This is being done because now the dev server attempts to mint an OIDC token based on `.vercel/project.json` or `vc` auth state.

AndrewBarba (approved):

PR #329, opened 2026-06-26
docs(connections): document gating MCP tools by name or input

bensabic:
Documents how to gate specific MCP connection tools behind human approval — by tool name or by tool input — instead of putting every call behind `always()`. The new "Gate specific tools by name or input" section in `docs/connections/mcp.mdx` shows a custom `approval` policy and calls out the two connection-specific details: the qualified `<connection>__<tool>` name format (match the bare name with `.includes()`/`.endsWith()`), and that `toolInput` is untyped and may be `undefined`.

```ts title="agent/connections/social.ts"
import { defineMcpClientConnection } from "eve/connections";

// Bare tool names whose effects are irreversible — always gate these.
const DELETE_TOOLS = ["delete_draft", "delete_thread"];
// Tools that can publish — gate only when the call schedules a post.
const PUBLISH_TOOLS = ["create_draft", "edit_draft"];

// Read `requestBody.publish_at` without trusting the input's shape.
const publishesNow = (input: unknown): boolean => {
  const body = (input as { requestBody?: { publish_at?: unknown } })?.requestBody;
  return typeof body?.publish_at === "string" && body.publish_at.length > 0;
};

export default defineMcpClientConnection({
  url: "https://mcp.example.com/mcp",
  description: "Social publishing: draft, schedule, and manage posts.",
  auth: { getToken: async () => ({ token: process.env.SOCIAL_API_KEY! }) },
  approval: ({ toolName, toolInput }) => {
    if (DELETE_TOOLS.some((t) => toolName.includes(t))) return "user-approval";
    if (PUBLISH_TOOLS.some((t) => toolName.includes(t))) {
      return publishesNow(toolInput) ? "user-approval" : "not-applicable";
    }
    return "not-applicable";
  },
});
```


docs/connections/mcp.mdx
```diff
@@ -143,6 +143,50 @@ export default defineMcpClientConnection({
 
 ## Troubleshooting
 
+The `once()`, `always()`, and `never()` helpers from `eve/tools/approval` gate every tool the connection serves the same way. See [Connections](/docs/connections#per-connection-approval) for those, and [Human-in-the-loop](/docs/human-in-the-loop) for the full pause-and-resume contract.
```
ijjk, line 144: This new custom approval-policy section is currently nested under `## Troubleshooting`, which makes the troubleshooting table no longer follow its own heading and frames the approval guidance as troubleshooting content. Can we move lines 146-189 up into the `## Approval gates` section, before `## Troubleshooting`, so the table stays directly under the troubleshooting heading?

ijjk (approved):

PR #347, opened 2026-06-26
feat(eve): cancellation - add active turn cancellation

AndrewBarba:
## Summary

- add authenticated turn cancellation to the eve channel, TypeScript client, custom-channel runtime surface, and eval driver
- keep cancellation ownership in the terminating root turn workflow while propagating its abort signal through models, tools, local children, and remote descendants
- settle cancelled descendant work through existing runtime-action results and emit `turn.cancelled` followed by `session.waiting`
- add deterministic unit, integration, nested-agent, race, and E2E coverage

## Architecture

The root `turnWorkflow` owns the only cancellation hook and `AbortController`. Inherited child turns use the supplied signal and deterministic turn inboxes without creating additional raced hooks. Remote cancellation waits for dispatch adoption so it targets the server-issued session identity.

## Validation

- `pnpm --filter eve build`
- `pnpm --filter agent-cancellation typecheck`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/evals/runner/execute-task.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/workflow-entry.integration.test.ts`
- `pnpm guard:invariants`

The new fixture E2E eval is intentionally left for CI, per repository policy.


ziggy-orbital:
this is exactly what I need !!! thank you for picking it up! 🙏

we're building a Discord bot on top of an eve agent. Users chat with it and it
runs long, multi-step tool chains (uploads, DB/project-system lookups... some turns
go 15+ steps over several minutes). The one thing we can't do today is let a user
**_stop a chain that's already running_** the client `stop()` only hangs up our
stream while the durable turn keeps executing server-side, and a follow-up "stop"
message just queues as its own turn instead of interrupting the active one.

Turn-scoped cancellation that aborts the in-flight model/tool work and then emits
`turn.cancelled` → `session.waiting` is perrrfect TYSM!

PR #348, opened 2026-06-26
fix(eve): workflows - fence replayed turn children

AndrewBarba:
## Summary

- derive each turn inbox token from the durable `dispatchTurnStep` step ID
- resolve the inbox's actual owning workflow run after starting a latest-deployment child
- require a separate, post-commit activation addressed to that owner before `turnStep` can execute
- keep the wire additive so children started by older pinned drivers retain the existing behavior

This closes the window where a turn child could finish and release its conflict hook before the dispatch step committed, allowing a replay or late concurrent dispatch attempt to start and execute another child. A late claimant may receive an old activation, but the expected run ID fences it from doing turn work.

This is intentionally narrower than, and intended to supersede, #335. It does not filter client events or claim to make individual step effects exactly-once. Concurrent attempts of the same `turnStep` still require effect-level idempotency.

## Testing

- `pnpm fmt`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm typecheck`
- `pnpm test:unit` (4,145 passed, 1 skipped)
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/workflow-entry.integration.test.ts` (8 passed)

### DIFF
--- a/.changeset/quiet-turn-replays.md
+++ b/.changeset/quiet-turn-replays.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fence replayed turn child workflows so only the inbox owner durably selected by the session driver can execute the turn.
--- a/packages/eve/src/execution/activate-turn-step.test.ts
+++ b/packages/eve/src/execution/activate-turn-step.test.ts
@@ -1,0 +1,33 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { activateTurnStep } from "#execution/activate-turn-step.js";
+import { resumeHook } from "#internal/workflow/runtime.js";
+
+vi.mock("#compiled/@workflow/core/runtime.js", () => ({
+  resumeHook: vi.fn(),
+}));
+
+afterEach(() => {
+  vi.clearAllMocks();
+});
+
+describe("activateTurnStep", () => {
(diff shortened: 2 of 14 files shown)


AndrewBarba:
CI monitoring results (corrected after inspecting the per-session artifact):

- All deterministic checks passed: lint, typecheck, unit, scenario, TUI, and Linux/Windows integration.
- `e2e-local (agent-workflow-stress)` failed twice. Its concurrent case passed all 156 gates both times. The sequential case showed two concurrent attempts of the **same child run's** `turnStep` emitting the same turn twice (turn 13 on the first run, turn 10 on the rerun).
- `e2e-vercel (agent-workflow-stress)` completed all 100 sequential turns (205/205 gates). In the concurrent case, one workflow run/session ID emitted `session.started` and its first turn twice; the eval client then consumed the duplicate boundary instead of the requested second turn. This is also same-run replay, not two separate top-level sessions.
- `e2e-vercel (agent-tools)` had an unrelated assertion mismatch: its log says an `action.result` was observed while the matcher reported zero.

The stress failures are real externally observable protocol failures even though neither workflow reached `session.failed`: a duplicate boundary was mistaken for completion of the next request. The child-owner fence in this PR handles distinct replayed child runs, but same-run attempts need idempotent event/effect handling and request-correlated client boundaries.

PR #385, opened 2026-06-28
fix(eve): session with first turn in streaming cannot be stored in fr…

lchrennew:
### Description

Cannot get session id during first turn in streaming


lchrennew:
@ruiconti @ijjk 

I got a detailed description generated by LLM to express my idea in native English:
```
Title: Eagerly surface sessionId after POST completes

Description:

Previously, session.state.sessionId was only updated after the full event stream was consumed (in the finally block of #createEventStream ). This meant consumers who needed the session ID early — for example to persist it or display it in the UI — had to reach into MessageResponse.sessionId directly and could not rely on session.state .

This change updates ClientSession.send() to write the session ID into #state immediately after the POST response returns, before the event stream is created. It also adds an early onSessionChange notification in EveAgentStore.send() right after the session POST resolves, so framework integrations (Vue, React) can react to the new session ID before any stream events arrive.

Changes:

- packages/eve/src/client/session.ts : Eagerly assign sessionId to #state after #postTurn resolves.
- packages/eve/src/client/eve-agent-store.ts : Fire onSessionChange and #publish() immediately after session.send() returns, before iterating the event stream.
- packages/eve/src/client/session.test.ts : Added test verifying state.sessionId is available right after send() resolves.
- packages/eve/src/vue/use-eve-agent.test.ts : Updated onSessionChange expectations — now fires twice per turn (once after POST with initial sessionId, once after stream completes with final cursor).
Behavior:

onSessionChange now fires at two points during a turn:

1. After POST completes — sessionId is set, streamIndex is still 0
2. After stream consumption — continuationToken and streamIndex are finalized
```

PR #407, opened 2026-06-29
feat(eve): add subagent runtime limits

allenzhou101:
## Summary

Adds default runtime guardrails for subagent delegation:

- caps nested subagent depth at 4 by default
- caps subagent calls from a single model step at 4 by default
- returns failed `subagent-result` action results when calls are rejected
- adds `defineAgent({ limits: { subagents: { maxDepth, maxCallsPerStep } } })` overrides
- persists effective subagent limits in session state and propagates them through parent lineage so children inherit the root policy and can only tighten it

The limits apply to the built-in `agent` tool, authored subagents, remote agents, and Workflow-launched subagent calls because enforcement happens at runtime-action dispatch.

## Testing

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/subagent-limits.test.ts src/internal/authored-definition/core.test.ts src/compiler/manifest.test.ts src/execution/session.test.ts src/execution/workflow-steps.test.ts`
- `pnpm --filter eve run build`
- `test-agent`: `npm install --no-save --package-lock=false ../eve-runtime-limits/packages/eve`
- `test-agent`: confirmed `node_modules/eve` resolves to `/Users/allenzhou/Desktop/code/eve/eve-runtime-limits/packages/eve`
- `test-agent`: `npm run typecheck`
- `test-agent`: `npm run build`
- `test-agent`: `npm run start -- --host 127.0.0.1 --port 2117`, then `curl -fsS http://127.0.0.1:2117/eve/v1/info`

### DIFF
--- a/.changeset/runtime-delegation-limits.md
+++ b/.changeset/runtime-delegation-limits.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add default runtime limits and agent config overrides for subagent depth and per-step child-agent fan-out.
--- a/e2e/fixtures/agent-subagent-limits/.gitignore
+++ b/e2e/fixtures/agent-subagent-limits/.gitignore
@@ -1,0 +1,11 @@
+node_modules
+.env*
+.eve
+.vercel
+.workflow-data
+.next
+.output
+.nitro
+dist
+.DS_Store
+*.tsbuildinfo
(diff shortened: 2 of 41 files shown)


allenzhou101:
Superseded by the split draft stack: #415 for subagent depth/tool hiding and #416 for per-step fan-out limits.

PR #409, opened 2026-06-29
fix(eve): cap recursive subagent delegation

ruiconti:
## What

- Track local delegated subagent depth through run context, workflow session creation, and durable session hydration.
- Hide delegated subagent tools at the configured max depth and block stale delegated runtime actions before another child session can start.
- Expose the authoring limit as `defineAgent({ limits: { maxSubAgentDepth } })`, with `@default 3` in the emitted type declarations.

## Why

Recursive self-delegation could keep creating child sessions because the parent-facing delegation tools stayed visible at every nested child. The model-visible guard handles the normal path: once a child reaches the configured max depth, those tools are no longer advertised.

The dispatch guard covers the forced-call path. If a stale `subagent-call` or `remote-agent-call` still reaches execution at the limit, eve logs a warning and returns a `SUBAGENT_DEPTH_LIMIT_REACHED` `subagent-result` instead of starting another child workflow.

## Boundary

The depth cap is counted as local child-session hops. Root sessions are depth 0. The default max depth is 3, so child sessions may run at depths 1 through 3. With `limits.maxSubAgentDepth: 4`, child sessions may run at depths 1 through 4.

At the configured max depth, the harness removes delegated runtime-action tools from the advertised tool map before provider tool construction. That covers declared local subagents, remote-agent tools, the built-in `agent` tool, and `Workflow` when its remaining host surface would only expose those hidden tools.


packages/eve/src/channel/types.ts
```diff
@@ -272,6 +272,17 @@ export interface RunInput {
...
+   * Framework-owned depth of delegated local subagent sessions. Root sessions
+   * omit this and are treated as depth 0; each local child receives
+   * parent depth + 1.
+   */
+  readonly subagentDepth?: number;
+  /**
+   * Optional maximum delegated subagent depth inherited by this run. When
+   * omitted, the session uses its resolved agent config or eve's default.
+   */
+  readonly subagentMaxDepth?: number;
```
allenzhou101, line 285: i did this as the below to account for future, but seems reasonable

```
/**
 * Limits for agent-to-agent delegation. These apply to the built-in `agent`
 * tool, authored subagents, remote agents, and subagent calls launched by
 * Workflow.
 */
export interface AgentSubagentLimitsDefinition {
  /**
   * Maximum depth for nested subagent calls. Root sessions are depth 0.
   */
  readonly maxDepth?: number;
}

/**
 * Runtime safety limits authored in `agent.ts`.
 */
export interface AgentLimitsDefinition {
  readonly subagents?: AgentSubagentLimitsDefinition;
}
```

allenzhou101 (approved):

packages/eve/src/channel/types.ts
ruiconti: I was debating between

```ts
defineAgent({ subagents: { maxDepthLimit }}) // wrong shape - this is more about limits rather than subagents
defineAgent({ limits: { subagents: { maxDepth }}}) // works but too much nesting
defineAgent({ limits: { subagentsMaxDepth }}) // works

// tools/agent.ts
import { defineSubagentTool } from "@eve/tools"
export defineSubagentTool({ maxDepth })
// like it but exposes how subagents are implemented which I'm not a fan of
```

.changeset/cap-recursive-subagent-delegation.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": patch
+---
+
+Cap recursive subagent delegation at three child-session levels by default, configurable with `defineAgent({ limits: { maxSubAgentDepth } })`. At the limit, eve no longer advertises subagent tools and blocks stale delegated calls before starting another child session.
```
AndrewBarba: Lets rename to `maxSubagentDepth` which is more grammatically correct

AndrewBarba (approved):

PR #413, opened 2026-06-29
Route Next eve service through Build Output

ijjk:
## Summary

- Update `withEve()` so Vercel builds keep the Next.js app as the default app and route public `/eve/v1/**` requests to a generated eve service.
- Write generated Build Output `services` plus service `routes` instead of adding Next.js rewrites for Vercel deploys.
- Teach the helper to write generated output next to the linked Vercel project settings, including monorepo templates where `.vercel/` lives at the repo root.
- Update Next.js docs/templates and focused tests for the new Vercel topology.

Relies on: https://github.com/vercel/vercel/pull/16889

## Tests

```sh
pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/next/index.test.ts
pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/public/next/index.integration.test.ts
```


packages/eve/src/public/next/index.integration.test.ts
```diff
@@ -173,26 +170,26 @@ describe("withEve Vercel config", () => {
     const outputConfig = await readJsonFile(join(appRoot, ".vercel", "output", "config.json"));
 
     expect(outputConfig).toEqual({
-      version: 3,
-      experimentalServices: {
+      experimentalServicesV2: {
```
ctgowrie: can just use services now

ijjk: moved back to stable "services" as upstream fix has landed

ctgowrie (approved):

PR #434, opened 2026-06-30
fix(sandbox): strip `runtime` param from Sandbox

AndyW22:
Sandbox package has been updated to a new version that requires a few changes:

- `__image` is now `image
- `runtime` and `image` are mutually exclusive
- `image` and `snapshot` are mutually exclusive

`runtime` is passed through in eve but is unused hence the param being removed here.

### DIFF
--- a/packages/eve/src/execution/sandbox/bindings/vercel.test.ts
+++ b/packages/eve/src/execution/sandbox/bindings/vercel.test.ts
@@ -156,7 +156,7 @@ describe("createVercelSandbox", () => {
     expect(sandboxModule.Sandbox.create).toHaveBeenCalledTimes(1);
     expect(sandboxModule.Sandbox.create).toHaveBeenCalledWith(
       expect.objectContaining({
-        __image: "vercel/eve:latest",
+        image: "vercel/eve:latest",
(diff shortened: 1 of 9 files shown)


packages/eve/src/execution/sandbox/bindings/vercel.test.ts
```diff
@@ -581,7 +581,7 @@ describe("createVercelSandbox", () => {
 
     const backend = createTestVercelSandbox({
       createOptions: {
-        runtime: "node22",
+        ports: [3000],
```
marc-vercel, line 584: Why do we need to specify the ports now, and before we didn't?

marc-vercel (approved):
LGTM, left a comment.

AndrewBarba (approved):

packages/eve/src/execution/sandbox/bindings/vercel.test.ts
AndyW22: We don't need to, but i noticed we were lacking test coverage of the `ports` param so replaced it here.

AndrewBarba (approved):

PR #446, opened 2026-06-30
feat(sandbox): forward supported just-bash SandboxOptions through justbash()

### LINKED ISSUE #431: justbash() backend can't enable just-bash's bundled Python (capability passthrough)
## Summary

The `justbash()` sandbox backend cannot enable just-bash's bundled interpreters (Python, and by the same mechanism sqlite3 / js-exec). `python3` returns **exit 127** under the just-bash backend even though the CPython-WASM runtime ships inside the `just-bash` package.

## Details

`just-bash` (3.0.2) bundles CPython-WASM at `vendor/cpython-emscripten/python.wasm` and exposes it through a typed constructor flag — `BashOptions.python: boolean` ("Enable python3/python commands. Python is disabled by default as it introduces additional security surface…"). Likewise it has flags/config for sqlite3, js-exec (QuickJS), and network.

But eve's `justbash()` backend (eve 0.17.1) constructs the interpreter with filesystem / network / env / cwd options only and **never sets `python`**, and `JustBashSandboxCreateOptions` exposes only `autoInstall`. There is no passthrough for the capability flags, so `python` stays at its default (off).

Result: the bundled runtime is present but unreachable — `python3` is "command not found".

## Reproduction

1. Author a sandbox that pins the backend:

   ```ts
   // agent/sandbox/sandbox.ts
   import { defineSandbox } from "eve/sandbox";
   import { justbash } from "eve/sandbox/just-bash";
   export default defineSandbox({ backend: justbash() });
   ```

2. Have the agent run any Python in the sandbox, e.g. `python3 -c "print(1+1)"`.
3. Observed: `sandbox command finished (exit 127)` (command not found).
   Expected: `2`.

## Impact

just-bash is uniquely valuable as the zero-dependency, no-daemon, no-VM backend — the natural choice for KVM-less CI and lightweight local runs. But any agent that computes in-sandbox (a Python/sqlite/js helper) silently fails on just-bash and must fall back to Docker or microsandbox, defeating that purpose. In our case we fell back to microsandbox solely because just-bash couldn't run a stdlib-only Python helper.

## Request

Let `justbash({ ... })` pass through just-bash's capability options — at minimum `python`, and ideally the relevant `BashOptions` subset (sqlite, js-exec, network) — so callers can opt in, with the security caveat surfaced in the docs (it's off-by-default in just-bash for a reason).

## Related

#239 ("`vercel()` sandbox backend ignores the runtime option and always runs `vercel/eve:latest`") looks like the same class of issue: a backend factory not forwarding an author-supplied option to the underlying runtime. May be worth addressing as a consistent "backends forward their options" pass.

## Versions

eve 0.17.1, just-bash 3.0.2.

arimxyer:
## What

Forward the just-bash `SandboxOptions` fields that `justbash()` currently hardcodes or
omits, through `JustBashSandboxCreateOptions` → `Sandbox.create(...)`:

- `timeoutMs`
- `maxCallDepth`
- `maxCommandCount`
- `maxLoopIterations`
- `defenseInDepth`

Today `JustBashSandboxCreateOptions` exposes only `autoInstall`, and `createBashSandbox`
calls `Sandbox.create({ cwd, env, fs, network })` with everything else hardcoded — so an
author has no way to set the interpreter's execution limits, an overall timeout, or
toggle defense-in-depth.

## Why

This is the "backends should forward their options" generalization (cf. #239): the
just-bash backend should let authors set the `SandboxOptions` just-bash actually accepts,
rather than silently fixing them. Each new field is optional and forwarded as `undefined`
when omitted, falling back to just-bash's own default — so existing apps are unaffected.

## Relationship to #431 (capability flags / Python)

This does **not** yet close #431. #431 asks for `python3` (and js-exec) through
`justbash()`, but those capability flags live on just-bash's `BashOptions` (the
`new Bash(...)` constructor) and just-bash's `Sandbox.create` — the call eve uses — drops
them. That's an upstream gap, filed and fixed in **vercel-labs/just-bash#284**. Once that
change ships in a just-bash release eve depends on, this same passthrough extends to
`python` / `javascript` / `commands` / `customCommands` / `fetch` to genuinely close #431.
The `JustBashSandboxCreateOptions` doc comment notes this explicitly.

The limits/timeout/defense-in-depth forwarding here is complete and self-contained —
mergeable on its own as the #239-class "backends forward their options" passthrough. The
bundled-interpreter capability flags (`python`, `javascript`, …) are a follow-up that
extends this same `JustBashSandboxCreateOptions` surface once just-bash#284 ships in a
release eve depends on; that follow-up is what closes #431.

## Backward compatibility

No behavior change when the new fields are omitted — each forwards as `undefined`, i.e.
just-bash's existing default (network still defaults to full internet access exactly as
before).

## Tests

`packages/eve/src/execution/sandbox/bindings/just-bash-capability-options.integration.test.ts`
(new): mocks the `just-bash` module's `Sandbox.create`, asserts (a) author-supplied
options are forwarded into the call, and (b) omitting them leaves the call unchanged from
today (new fields `undefined`, network default intact).

## Validation

- `pnpm --filter eve run typecheck` — clean
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts just-bash-capability-options` — pass

Refs #431, #239. Upstream prerequisite for the capability flags: vercel-labs/just-bash#284.

### DIFF
--- a/.changeset/justbash-forward-sandbox-options.md
+++ b/.changeset/justbash-forward-sandbox-options.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
+Forward the just-bash `SandboxOptions` that the `justbash()` backend previously hardcoded or omitted — `timeoutMs`, `maxCallDepth`, `maxCommandCount`, `maxLoopIterations`, and `defenseInDepth` — through `JustBashSandboxCreateOptions`. Each is optional and falls back to just-bash's own default when omitted, so existing apps are unaffected. (The bundled-interpreter capability flags like `python` are not reachable through `Sandbox.create` and remain gated upstream — see vercel-labs/just-bash#284.)
--- a/packages/eve/src/execution/sandbox/bindings/just-bash-capability-options.integration.test.ts
+++ b/packages/eve/src/execution/sandbox/bindings/just-bash-capability-options.integration.test.ts
@@ -1,0 +1,101 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { createJustBashSandboxBackend } from "#execution/sandbox/bindings/just-bash.js";
+import { useTemporaryDirectories } from "#internal/testing/use-temporary-app-roots.js";
+
+// Capture every `Sandbox.create(options)` call so we can assert which
+// capability options the just-bash backend forwards. `vi.hoisted` runs
+// before the `vi.mock` factory, so the array is in scope inside it.
+const created = vi.hoisted(() => ({ options: [] as unknown[] }));
+
+// A minimal functional stand-in for the optional `just-bash` package:
+// just enough surface for `createBashSandbox` to build a sandbox handle
+// without a real interpreter. `Sandbox.create` records its options.
+vi.mock("just-bash", () => {
+  class ReadWriteFs {
+    async mkdir(): Promise<void> {}
+    async writeFile(): Promise<void> {}
+    async readFileBuffer(): Promise<Uint8Array> {
+      return new Uint8Array();
+    }
+    async rm(): Promise<void> {}
+  }
+  const Sandbox = {
+    create: vi.fn(async (options: unknown) => {
+      created.options.push(options);
+      return {
+        bashEnvInstance: { getEnv: () => ({}) },
+        async runCommand() {
+          return {};
+        },
+        async stop() {},
+      };
+    }),
+  };
+  return { ReadWriteFs, Sandbox };
+});
+
+const createScratchDirectory = useTemporaryDirectories();
+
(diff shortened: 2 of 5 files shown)


arimxyer:
Looks like [#284](https://github.com/vercel-labs/just-bash/pull/284) just landed in the just-bash repo

arimxyer:
Superseded by #607. Now that the upstream fix (vercel-labs/just-bash#284) shipped in just-bash 3.1.0, #607 does this in one pass off fresh `main` — the limits/timeout forwarding from here **plus** the `python` / `javascript` capability flags — and closes #431. Closing this in favor of it.

PR #447, opened 2026-06-30
Use Vercel CLI for project creation

### LINKED ISSUE #438: Vercel preview deployments ignore deploymentEnabled for Eve apps
### eve version

`eve@0.12.0`

### Node.js version

`v24.12.0`

### Where does the bug occur?

Other: Vercel Git integration / PR preview deployments for a deployed Eve app

### Describe the bug

In a monorepo with multiple Vercel apps, our Eve app still gets Vercel PR status checks / automatic preview deployments even though its Vercel config disables deployments for all branches except `master`.

The Eve app has this `vercel.ts`:

```ts
import type { VercelConfig } from "@vercel/config/v1";

export const config: VercelConfig = {
	git: {
		deploymentEnabled: {
			"*": false,
			master: true,
		},
	},
};
```

The same `git.deploymentEnabled` pattern works for our other apps in the same monorepo. For those apps, Vercel does not create automated PR preview deployments/status checks for disabled branches. With the Eve app, Vercel appears to behave differently and still reaches the preview deployment/status-check path.

A build ignore script can prevent the build from running, but this config should prevent the automatic preview deployment from being created in the first place. We use this to avoid spending build minutes unless a deployment is manually invoked.

### Steps to reproduce

1. Create or deploy an Eve app in a Vercel monorepo project.
2. Add a build script such as:

```json
{
	"scripts": {
		"build": "eve build"
	}
}
```

3. Add `vercel.ts` with `git.deploymentEnabled` disabling all branches except `master`:

```ts
import type { VercelConfig } from "@vercel/config/v1";

export const config: VercelConfig = {
	git: {
		deploymentEnabled: {
			"*": false,
			master: true,
		},
	},
};
```

4. Open a pull request from a non-`master` branch.
5. Observe that Vercel still creates PR status checks / attempts an automatic preview deployment for the Eve app.

### Expected behavior

Vercel should honor `git.deploymentEnabled` for Eve apps the same way it does for the sibling Next.js apps in the monorepo: non-`master` PR branches should not create automatic preview deployments or PR status checks unless manually invoked.

ijjk:
## Summary
- create new linked Vercel projects through Vercel CLI link instead of direct project API calls
- read the resulting .vercel/project.json metadata after successful linking
- keep detected host frameworks when project sources already reference the matching eve integration (`eve/next`, `eve/nuxt`, or `eve/sveltekit`)
- ensure new projects use the eve framework preset when Vercel returns no framework, an unsupported framework, or a rejected ambiguous host framework
- prompt only in interactive runs when a supported host framework is detected without a matching eve integration import; headless ambiguous detections patch back to eve
- use the shared Vercel project request timeout for both framework inspection and framework PATCH calls
- keep the Vercel project orchestration file under the source line-count cap by moving framework handling into its own setup helper
- refresh stale deploy comments and add a patch changeset

## Tests
- pnpm --filter eve run test:unit
- pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/internal/structural-tests/file-length.test.ts src/setup/vercel-project.test.ts src/setup/boxes/link-project.test.ts src/setup/boxes/deploy-project.test.ts
- pnpm --filter eve run typecheck
- pnpm lint

## Issues
Fixes #438

### DIFF
--- a/packages/eve/src/setup/vercel-project-framework.ts
+++ b/packages/eve/src/setup/vercel-project-framework.ts
@@ -1,0 +1,248 @@
+import { createPromptCommandOutput } from "#setup/cli/index.js";
+import { captureVercel } from "#setup/primitives/index.js";
+import type { Dirent } from "node:fs";
+import * as fs from "node:fs/promises";
+import { extname, join } from "node:path";
+import { z } from "zod";
+
+import type { Prompter } from "./prompter.js";
+import { isForbiddenApiFailure, normalizeVercelApiResult } from "./vercel-api-failure.js";
+import {
+  parseVercelJson,
+  requireVercelTeamAccess,
+  type VercelProjectOperationOptions,
+  VERCEL_PROJECT_REQUEST_TIMEOUT_MS,
+} from "./vercel-project-api.js";
+
+const VercelProjectFrameworkSchema = z.object({
+  framework: z.string().min(1).nullish(),
+});
+
+const EVE_FRAMEWORK_PRESET = "eve";
+
+interface EveFrameworkIntegration {
+  readonly label: string;
+  readonly importSpecifier: string;
+}
+
+const EVE_FRAMEWORK_INTEGRATIONS: Readonly<Record<string, EveFrameworkIntegration>> = {
+  nextjs: { label: "Next.js", importSpecifier: "eve/next" },
+  nuxt: { label: "Nuxt", importSpecifier: "eve/nuxt" },
+  nuxtjs: { label: "Nuxt", importSpecifier: "eve/nuxt" },
+  sveltekit: { label: "SvelteKit", importSpecifier: "eve/sveltekit" },
+};
+
+const FRAMEWORK_INTEGRATION_SOURCE_EXTENSIONS = new Set([
+  ".cjs",
+  ".cts",
+  ".js",
+  ".json",
+  ".jsx",
+  ".mjs",
+  ".mts",
+  ".svelte",
+  ".ts",
+  ".tsx",
+  ".vue",
+]);
+
+const FRAMEWORK_INTEGRATION_IGNORED_DIRECTORIES = new Set([
+  ".eve",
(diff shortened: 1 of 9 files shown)


packages/eve/src/setup/vercel-project.ts
```diff
@@ -581,31 +542,48 @@ export async function pickNewProjectName(
...
+    await assertNewProjectNameAvailable(projectRoot, spec.team, spec.project, options);
+    const linked = await withSpinner(
       prompter,
       `Creating Vercel project "${spec.project}" in ${spec.team}...`,
-      async () => {
-        await assertNewProjectNameAvailable(projectRoot, spec.team, spec.project, options);
-        return createProject(projectRoot, spec.team, spec.project, onOutput, options);
-      },
+      () =>
+        runVercel(["link", "--project", spec.project, ...scope, "--yes"], {
```
ruiconti: Think we want to have `assertNewProjectNameAvailable` as part of the spinner callback as it is bounded by a network call.

packages/eve/src/setup/vercel-project-framework.ts
```diff
@@ -0,0 +1,248 @@
...
+  return keepDetected ? "keep" : "switch-to-eve";
+}
+
+function sourceMentionsPackageSpecifier(source: string, importSpecifier: string): boolean {
+  return (
+    source.includes(`"${importSpecifier}"`) ||
+    source.includes(`'${importSpecifier}'`) ||
+    source.includes(`\`${importSpecifier}\``)
+  );
+}
```
ruiconti, line 167: This is lossy but should be fine? I also don't think we should be parsing all modules, specially given that not all of the frameworks are in TSX.

ijjk: Ya don't think we need to parse since the import has to be statically detectable

ruiconti (approved):

PR #449, opened 2026-06-30
fix(eve): keep text and reasoning blocks when a step resumes after auth

### LINKED ISSUE #436: Default message reducer overwrites text when OAuth resumes in same step
### eve version

eve@0.17.0

### Node.js version

v24.17.0

### Where does the bug occur?

Local dev (`eve dev`), Production (`eve start` / deployed)

### Describe the bug

I have a chat app using eve, it has a connection to Outlook to read my email. When the model uses a tool which uses outlook it triggers an OAuth pop up. This works great, but there's a slight UI bug. I'm using `useEveAgent` and I render an entry in the chat UI to indicate that the auth pop up is pending. But then after I log in in the pop up, *the next agent message overwrites the message before the auth-indicator-entry instead of going after the auth-indicator-entry.

In eve events I see:
```
session.started

turn.started              turn_0 seq=0
message.received          turn_0 seq=0  "pls summarize my email from today"

step.started              turn_0 stepIndex=0 seq=0
reasoning.completed       turn_0 stepIndex=0 seq=0  "The user wants to summarize their emails from today. I need to use the Outlook connector to fetch today's emails. Let me search for the right tool to get emails."
message.completed         turn_0 stepIndex=0 seq=0  finishReason=tool-calls  "Sure! Let me search for the right Outlook tool to fetch today's emails."
actions.requested         turn_0 stepIndex=0 seq=0  connection_search {"keywords":"list messages inbox received today","connection":"outlook"}
action.result             turn_0 stepIndex=0 seq=0  connection_search completed
step.completed            turn_0 stepIndex=0 seq=0  finishReason=tool-calls

step.started              turn_0 stepIndex=1 seq=0
reasoning.completed       turn_0 stepIndex=1 seq=0  "I'll use `outlook__me_ListMessages` to list messages from today. I'll filter by today's date using an OData filter on `receivedDateTime`. Today is 2025-07-15. I'll fetch up to 50 messages, filtered to today, and select the key fields needed for a summary."
message.completed         turn_0 stepIndex=1 seq=0  finishReason=tool-calls  "I'll use `outlook__me_ListMessages` and filter for messages received today, selecting the key fields needed for a clean summary."
actions.requested         turn_0 stepIndex=1 seq=0  outlook__me_ListMessages {"filter":"receivedDateTime ge 2025-07-15T00:00:00Z and receivedDateTime lt 2025-07-16T00:00:00Z","orderby":["receivedDateTime desc"],"select":["subject","from","receivedDateTime","bodyPreview","isRead","importance"],"top":50}
step.completed            turn_0 stepIndex=1 seq=0  finishReason=tool-calls

authorization.required    turn_0 stepIndex=1 seq=0  name=outlook description="Authorization required for outlook"
authorization.completed   turn_0 stepIndex=1 seq=0  name=outlook outcome=authorized

step.started              turn_0 stepIndex=1 seq=0
reasoning.completed       turn_0 stepIndex=1 seq=0  "I found the `outlook__me_ListMessages` tool which lists messages in the user's mailbox. I'll use it to fetch today's emails with a filter for today's date. Let me use the receivedDateTime filter to get only today's messages. I need to know today's date first.\n\nToday's date - I'll use a filter like `receivedDateTime ge 2025-01-XX`. Let me get today's date."
message.completed         turn_0 stepIndex=1 seq=0  finishReason=tool-calls  "Got it. Now I'll fetch today's messages using `outlook__me_ListMessages` with a `receivedDateTime` filter for today. Let me first get today's date."
actions.requested         turn_0 stepIndex=1 seq=0  bash {"command":"date -u +%Y-%m-%d"}
action.result             turn_0 stepIndex=1 seq=0  bash completed stdout="2026-06-30\n"
step.completed            turn_0 stepIndex=1 seq=0  finishReason=tool-calls
```

But in the UI I see this

```
pls summarize my email from today

thinking

Sure! Let me search for the right Outlook tool to fetch today's emails.

connection_search keywords="list messages inbox received today" connection="outlook"

thinking

Got it. Now I'll fetch today's messages using outlook__me_ListMessages with a receivedDateTime filter for today. Let me first get today's date.

outlook__me_ListMessages filter="receivedDateTime ge 2025-07-15T00:00:0"… orderby=[1] select=[6] top=50

Sign-in · Outlook
Signed in.

bash command="date -u +%Y-%m-%d"
```

Note that the `I'll use outlook__me_ListMessages to list messages from today. I'll filter...` message isn't visible in the UI, and as the messages stream in I can see it being overwritten by the next `Got it. Now I'll fetch today's messages using outlook__me_ListMessages...` message.

**Below is Codex's analysis of the issue**, it looks reasonable to me. The message reducer should be adjusted to handle the events eve emits correctly or eve should give the event a new step index for the message after the oauth thing is done. Not sure what makes most sense for the project.

---

`useEveAgent`'s default message reducer can drop an assistant text block when an OAuth authorization resumes in the same step.

The raw stream contains both assistant text blocks. The reduced `data.messages` projection only keeps the later one.

The likely cause is that text and reasoning parts are keyed only by `stepIndex`:

```ts
text:${part.stepIndex ?? 0}
reasoning:${part.stepIndex ?? 0}
```

If auth resumes with the same `turnId` and `stepIndex`, later text replaces earlier text.

### Steps to reproduce

(Codex-written)

Observed with this stream shape:

```txt
message.completed        turn_0 stepIndex=1  "I'll use outlook..."
actions.requested        turn_0 stepIndex=1  outlook__me_ListMessages
authorization.required   turn_0 stepIndex=1
authorization.completed  turn_0 stepIndex=1
step.started             turn_0 stepIndex=1
message.completed        turn_0 stepIndex=1  "Got it. Now I'll fetch..."
```

The first `message.completed` is present in the raw event stream but missing from `useEveAgent().data.messages`.

A minimal test could feed these events through `defaultMessageReducer()` and assert that both text parts remain.

### Expected behavior

(Codex-written)

Both assistant text blocks should remain in `data.messages[].parts`, in stream order.

Reasoning parts likely need the same treatment.

It may also be worth checking whether the runtime should advance `stepIndex` before resuming after auth. A second `step.started` with the same `turnId` and `stepIndex` is surprising. But the reducer should still tolerate existing streams with this shape.

serhiizghama:
Closes #436.

When an OAuth authorization resolves mid-step, the runtime resumes the same step — a second `step.started`/`message.completed` arrives with the same `turnId` and `stepIndex`. `defaultMessageReducer` keyed text and reasoning parts by `stepIndex` alone, so the resumed block overwrote the earlier one and `useEveAgent().data.messages` lost the first assistant message even though it was in the raw stream.

The reducer now updates the current text/reasoning block only while it's still streaming; once a block is `done`, the next event of the same kind in that step starts a new block instead of replacing it. `removeTextPart` (null completion) is scoped to the trailing in-progress block too, so an empty delivery can't wipe a block kept from before the resume.

Added a test that feeds the issue's event shape (reasoning + message, auth required/completed, step resumes at the same index) and asserts both text and reasoning blocks survive in order.

### DIFF
--- a/.changeset/reducer-keep-text-on-step-resume.md
+++ b/.changeset/reducer-keep-text-on-step-resume.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix `defaultMessageReducer` dropping an assistant text or reasoning block when a step resumes after an authorization (OAuth) flow. A resumed step reuses the same `stepIndex`, so the second block no longer overwrites the first — both now stay in `data.messages[].parts` in stream order.
--- a/packages/eve/src/client/message-reducer.test.ts
+++ b/packages/eve/src/client/message-reducer.test.ts
@@ -691,4 +691,115 @@ describe("defaultMessageReducer", () => {
       { type: "step-start" },
     ]);
   });
+
+  it("keeps both text and reasoning blocks when a step resumes after authorization", () => {
+    const reducer = defaultMessageReducer();
+    let data = reducer.reduce(
+      reducer.initial(),
+      createStepStartedEvent({ sequence: 0, stepIndex: 0, turnId: "turn_0" }),
+    );
+    data = reducer.reduce(
+      data,
+      createReasoningCompletedEvent({
+        reasoning: "Let me list today's messages.",
+        sequence: 1,
+        stepIndex: 0,
+        turnId: "turn_0",
+      }),
+    );
+    data = reducer.reduce(
+      data,
+      createMessageCompletedEvent({
+        message: "I'll use outlook to fetch today's emails.",
+        sequence: 2,
+        stepIndex: 0,
+        turnId: "turn_0",
+      }),
+    );
+    data = reducer.reduce(
+      data,
+      createActionsRequestedEvent({
+        actions: [
+          { callId: "call_1", input: { top: 50 }, kind: "tool-call", toolName: "outlook__list" },
+        ],
+        sequence: 3,
+        stepIndex: 0,
+        turnId: "turn_0",
+      }),
+    );
+    data = reducer.reduce(
+      data,
+      createAuthorizationRequiredEvent({
+        description: "Authorization required for outlook",
+        name: "outlook",
+        sequence: 4,
+        stepIndex: 0,
+        turnId: "turn_0",
+      }),
+    );
+    data = reducer.reduce(
+      data,
+      createAuthorizationCompletedEvent({
+        name: "outlook",
+        outcome: "authorized",
+        sequence: 5,
+        stepIndex: 0,
+        turnId: "turn_0",
+      }),
+    );
+
+    // The runtime resumes the same step after the OAuth flow completes.
+    data = reducer.reduce(
+      data,
+      createStepStartedEvent({ sequence: 6, stepIndex: 0, turnId: "turn_0" }),
+    );
+    data = reducer.reduce(
+      data,
+      createReasoningCompletedEvent({
+        reasoning: "Now I'll fetch today's messages.",
+        sequence: 7,
+        stepIndex: 0,
+        turnId: "turn_0",
+      }),
+    );
+    data = reducer.reduce(
+      data,
+      createMessageCompletedEvent({
+        message: "Got it. Fetching today's messages now.",
+        sequence: 8,
+        stepIndex: 0,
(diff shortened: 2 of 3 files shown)


serhiizghama:
#973 covered this — closing.

PR #463, opened 2026-07-01
docs(research): mounted extensions API

ctgowrie:
## Summary

Adds a research doc proposing an **extensions API** for eve, using the "mount" model.

An extension is authored as an *agent-shaped directory with no agent config*, so a folder can be lifted out of an agent into an extension without editing a file. A consumer **mounts** it with a one-line re-export in `agent/extensions/`:

```ts
// agent/extensions/crm.ts
export { default } from "@acme/crm";
```

Nothing is copied — the package is the single source of truth and upgrades through the package manager.

## What the doc covers

- **Authoring** — same directories and `define*` calls as an agent; author owns stable, qualified names (`tools/crm_lookup.ts` → `crm_lookup`), so authored self-references survive mounting with **no consumer-side prefixing**.
- **Publishing** — `eve build` walks the source with the same discovery an agent uses and generates the default export (the mount target) plus a code-free static manifest for import-free discovery.
- **Importing** — discovery statically resolves the re-export, merges contributions under their authored names, and the runtime imports the default export for executable code.
- **Auto-scoped state** — `defineState("budget", …)` is automatically namespaced by the extension's package name at module-eval time, so durable state never collides across extensions and needs no author config.

Design decisions are recorded inline, with the residual forks left in an open-questions section.

## Notes

- `issue:` frontmatter is `TBD` — to be linked before this moves past draft.
- Docs-only / research change: no changeset, no published-package impact.

### DIFF
--- a/research/mounted-extensions.md
+++ b/research/mounted-extensions.md
@@ -1,0 +1,232 @@
+---
+issue: TBD
+last_updated: "2026-07-01"
+status: proposed
+---
+
+# Mounted extensions
+
+## Summary
+
+An extension packages eve concepts — tools, skills, connections, instructions, hooks, schedules,
+subagents — as an npm or local package. It is authored as an agent tree without agent config, so a
+folder can move from an agent into an extension unchanged. A consumer mounts it with a file in
+`agent/extensions/` that calls the extension's factory:
+
+```ts
+// agent/extensions/crm.ts   → namespace "crm"
+import { crm } from "@acme/crm";
+export default crm({ apiKey: process.env.CRM_API_KEY });
+```
+
+The extension is compiled once into eve's normal build artifact. The consumer's build composes
+that compiled node into its graph, the same way it composes subagents. Nothing is copied; upgrades
+go through the package manager.
+
+This works within four constraints: discovery is path-based and never imports authored modules; a
+compiled agent graph is already multi-node (root plus one node per subagent); durable state is a
+process-wide name-keyed registry; per-session limits are enforced on the session, not per tool.
+
+## Authoring
+
+An extension is an agent-shaped directory without `agent.ts` or `sandbox`. Every other slot works
+and uses the same `define*` functions, with names derived from paths.
+
+```
+@acme/crm/
+  package.json
+  ext/
+    config.ts
+    instructions/policy.md
+    tools/search.ts
+    tools/create_deal.ts
+    connections/api.ts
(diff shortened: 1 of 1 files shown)


research/mounted-extensions.md
```diff
@@ -0,0 +1,232 @@
...
+    config.ts
+    instructions/policy.md
+    tools/search.ts
+    tools/create_deal.ts
+    connections/api.ts
+    skills/triage/SKILL.md
+    hooks/audit.ts
+    schedules/nightly_sync.ts
+    lib/http.ts
+```
```
ruiconti, line 48: I'd expect the next section here to be about the mounting process as that's an important piece of the integration IMO e.g., the scope function of `crm()` above, ideally also highlighting the inherent config surface. Could refer to https://github.com/vercel/eve/pull/491 as an example.

research/mounted-extensions.md
```diff
@@ -0,0 +1,232 @@
...
+  import { defineConfig } from "eve/extension";
+  export default defineConfig({
+    apiKey: { type: "string", secret: true, required: true },
+    baseUrl: { type: "string", default: "https://api.acme.example" },
+  });
+  ```
+
+  `eve build` generates the factory from this and records it so `eve add` can prompt. Tools read
+  config with `config.get()`. Config is bound once at mount and constant for the session;
+  per-caller values belong in connection auth.
```
ruiconti, line 86: 👍🏻

research/mounted-extensions.md
```diff
@@ -0,0 +1,232 @@
...
+- A consumer file shadows a contribution of the same name. To override fields, import the base and
+  re-define it; `execute` still reads the extension's scoped config:
+
+  ```ts
+  // agent/tools/crm__search.ts
+  import { search } from "@acme/crm/tools";
+  import { defineTool } from "eve/tools";
+  import { always } from "eve/tools/approval";
+  export default defineTool({ ...search, approval: always() });
+  ```
```
ruiconti, line 172: I like that this is consistent with conventions, but it feels that the DX is not that great e.g., an agent names the file `agent/tools/crm_search.ts` or `crm___search.ts` and now you've doubled the `search` tool at runtime.

We could tag extension exported tools at build time (eg with a Symbol) so they're detectable and we can emit a warning or even bail if an invariant is violated (eg incorrect filename convention).

PR #470, opened 2026-07-01
feat(eve): mounted extensions

ctgowrie:
## Summary

**Mounted extensions** ([`docs/extensions.md`](https://github.com/vercel/eve/blob/feat/mounted-extensions/docs/extensions.md), [research/mounted-extensions.md](https://github.com/vercel/eve/blob/main/research/mounted-extensions.md)): author an extension as an agent-shaped package (tools, connections, skills, instructions, hooks), mount it with a one-line file under `agent/extensions/`, and the consumer's build composes its contributions into the agent under a namespace. Nothing is copied — the extension ships as source and the consumer's eve recompiles it, with `eve` as a **peer dependency** so one eve lives in the app.

## What's in this PR

**Authoring**
- `eve/extension` subpath with `defineExtension(options?)` — the declaration in `ext/extension.ts`. Its default export is both the mount factory a consumer calls and the handle the extension's own modules read config off (`import extension from "../extension.js"; extension.config`).
- Config is an optional **Standard Schema** (e.g. Zod, same family as a tool's `inputSchema`), validated at mount; `extension.config` returns the parsed value with defaults applied. No ambient global, no codegen — types flow schema → handle.
- `defineState` inside an extension is auto-scoped to the extension.

**Mounting & overrides**
- `agent/extensions/` discovery slot: a flat file (`crm.ts`) or a directory (`crm/extension.ts`) with co-located **override slots** (`crm/tools/search.ts`). Namespace = file basename / directory name; contributions compose as `<namespace>__<name>`.
- An override slot **shadows** the extension's same-named contribution, or **`disableTool()`** opts out of one (errors if it targets a tool the extension doesn't contribute). The `<namespace>__` prefix is reserved — an agent-root contribution using it is a discovery error, so an extension can't be shadowed from outside its mount.

**Resolution & composition**
- Mount resolution statically reads each mount's import specifier (never executes it), resolves the package (walking `node_modules`, honoring `package.json#eve.extension`), and discovers its agent-shaped source tree.
- Compiler composition merges tools, hooks, skills, connections, dynamic tools/skills/instructions, and instruction fragments under `<namespace>__` names, rebasing `logicalPath` so the module-map codegen loads them from the package.
- Namespace scoping for `defineState` keys + config binding is **package-derived** and baked at build time, so identically-named state never collides across extensions and the mount binding agrees with the extension's tools on the config key.

**Build & publish**
- `eve build` compiles the package's Node-facing entrypoints (`.`, `./tools`) to **self-contained runnable JavaScript** with the namespace baked in, emits **type declarations**, and manages the `exports` map — so a published extension installs and mounts with no second compiler. (An installed package resolves under `node_modules`, where those entrypoints are externalized and loaded by Node; re-exporting `.ts` there fails with `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`.)

**Guardrails**
- eve enforces each extension's `peerDependencies.eve` range when it is mounted.
- An extension cannot declare a `sandbox`, agent config, schedules, or limits, or mount other extensions — those are the consuming agent's to own.

## Testing

Unit + integration (discovery, compiler composition, authoring API, eve-version compatibility) and scenario tier:
- **build output** — entrypoints are self-contained runnable JS, namespace-scoped, with declarations and a `{ types, default }` exports map;
- **installed under `node_modules`** — build the package, place it under a consumer's real `node_modules`, compile, and execute a `extension.config`-reading tool (the path workspace-linked fixtures bypass);
- **directory-mount override** — a co-located override shadows one tool and `disableTool()`s another.

The `extensions` e2e fixture runs in CI.


packages/eve/src/internal/nitro/host/build-extension.ts
```diff
@@ -0,0 +1,177 @@
...
+
+function toolExportName(logicalPath: string): string {
+  let name = logicalPath;
+  for (const extension of SUPPORTED_AUTHORED_MODULE_FILE_EXTENSIONS) {
+    if (name.endsWith(extension)) {
+      name = name.slice(0, name.length - extension.length);
+      break;
+    }
+  }
+  return name.replace(/^tools\//, "").replaceAll("/", "_");
```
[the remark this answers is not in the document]
ctgowrie: Fixed in 21e2397a. `toolExportName` now coerces the binding to a valid JS identifier via a shared `safeJsIdentifier` helper (non-identifier chars → `_`, leading-digit guard), so `get-weather.ts` emits `export { default as get_weather } from "../../ext/tools/get-weather.mjs"` — the binding is valid while the source path keeps the kebab name. Added an integration test covering the kebab-case case. Good catch, thanks!

e2e/fixtures/toolkit-extension/ext/extension.ts
```diff
@@ -0,0 +1,12 @@
+import { defineExtension } from "eve/extension";
+import { z } from "zod";
+
+// The extension declaration: its default export is the mount factory the
+// consuming agent calls (`toolkit({ apiKey, tier })`), and the extension's own
+// tools read the bound, validated config off this handle via `extension.config`.
+export default defineExtension({
```
ruiconti, line 4: I wonder if exposing parts of the manifest to version control would make sense for change management? Analogous to committing a db schema, when changes are being made to the surface of an extension, I'd want an easy way to spot them and a public immutable manifest could be a good approach.

e2e/fixtures/extensions/.gitignore
```diff
@@ -0,0 +1,11 @@
+node_modules
```
ruiconti, line 1: Def a nit but I'd rename `agent-extensions` to just `extensions`. All tests are agent-bound so we could just gut it

packages/eve/src/compiler/normalize-extension.ts
```diff
@@ -0,0 +1,247 @@
...
+ * Exported for unit testing: passing the consumer overrides as `primary` and
+ * the extension's own contributions as `secondary` yields consumer-wins
+ * shadowing on name collision.
+ */
+export function mergeContributions(
+  primary: CompiledExtensionContributions,
+  secondary: CompiledExtensionContributions,
+): CompiledExtensionContributions {
+  return {
+    tools: dedupeBy([...primary.tools, ...secondary.tools], (tool) => tool.name),
```
ruiconti, line 311: Isn't this shadowing user-authored overrides? e.g.,

```ts
// agent/extensions/crm/tools/get_prospect.ts
import { disableTool } from "eve/tools"
export default disableTool() 
```

e2e/fixtures/toolkit-extension/ext/extension.ts
ctgowrie: Decided to use semver validation on eve app build to assert extension version is compatible. This will prevent various silent failures. In the future, we can scope packages to tools, skills, etc. to minimize scope of breaking API changes. 

I do not think it's worth a manifest for extension APIs (a la protobuf) quite yet. Once we have popular extensions we may want better version management, but can do later.

e2e/fixtures/extensions/.gitignore
ctgowrie: Done

AndrewBarba:
What's the publishing contract? Is it: an extension author runs `eve build`, publishes the package, and consumers can install it without another compiler or custom export setup.

Currently `eve build` emits `.mjs` entrypoints that re-export authored `.ts` files. This works for the workspace-linked fixtures because eve bundles their source, but a packed/installed package is externalized and Node 24 follows the entrypoint into TypeScript under `node_modules`, failing with `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`. The generated exports also lack declarations.

Should we make `eve build` emit runnable JavaScript plus TypeScript declaration? Or document a different strategy for how to build and publish extensions?

packages/eve/src/compiler/normalize-extension.ts
ctgowrie: Opposite, actually — before this commit the `disabled` kind fell through the compose loop, so the disable was silently dropped and the extension's tool survived (the bug). Now a `disableTool()` in the override slot removes the extension's same-named tool, so the consumer wins.

It also can't clobber a `defineTool` override: one file per slug, and a file is either a define or a disable — so the disable only ever removes the extension's *base* tool of that name, never a redefinition. If the extension has no such tool, it throws (listing what it does contribute) instead of no-op'ing.

ctgowrie:
> What's the publishing contract? Is it: an extension author runs `eve build`, publishes the package, and consumers can install it without another compiler or custom export setup.
> 
> Currently `eve build` emits `.mjs` entrypoints that re-export authored `.ts` files. This works for the workspace-linked fixtures because eve bundles their source, but a packed/installed package is externalized and Node 24 follows the entrypoint into TypeScript under `node_modules`, failing with `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`. The generated exports also lack declarations.
> 
> Should we make `eve build` emit runnable JavaScript plus TypeScript declaration? Or document a different strategy for how to build and publish extensions?

Yes, that's the contract: run eve build, publish, and a consumer just installs and mounts it, no extra compiler or export wiring.

The mentioned bug was accurate as the entrypoints were re-exporting the raw .ts, which is fine in the workspace but breaks the moment it's a real install. I fixed in 19172ab: eve build now compiles the two Node-facing entry points, . and ./tools, down to plain JS with the namespace baked in, emits .d.ts, and fills the exports map. Installed extensions load as-is now. There's also a new scenario test that builds a package, installs it under real node_modules, and runs a config-reading tool.

@AndrewBarba

AndrewBarba (approved):

PR #477, opened 2026-07-02
Emit an action.result when an ask_question is answered

mikkel-bergmann:
## Problem

Answering an `ask_question` leaves its tool call spinning forever in the dev TUI. When the user picks an option, the answer is written into model history, but no `action.result` stream event is emitted for it — only *denied* tool approvals emit one (`buildRejectedActionBatch` → the `rejectedActions` loop in `tool-loop.ts`). The TUI settles a tool block to `done` only when it receives a `tool-result` for that call (`terminal-renderer.ts`), so the `ask_question` block never settles and keeps its running spinner for the rest of the turn.

## Fix

Emit a successful `action.result` for each resolved **question** request (answered or ignored), mirroring the existing denied-approval path:

- `buildAnsweredActionBatch()` builds one tool-result per non-approval request via `createRuntimeToolResultFromValue` (the shared `action.result` coercion point).
- `resolvePendingInput()` returns it as `answeredActions`.
- The tool loop emits each as a non-rejected `action.result`.

Approval requests are intentionally excluded: an approved call runs and emits its own result, and a denied call is already surfaced by `buildRejectedActionBatch`.

## Testing

- New unit test in `input-requests.test.ts`: a resolved question produces an `answeredActions` batch (and no `rejectedActions`).
- `pnpm --filter eve typecheck`, `pnpm lint`, and the harness unit suite (`vitest … src/harness/`, 409 tests) pass locally.

### DIFF
--- a/.changeset/ask-question-settles-after-answer.md
+++ b/.changeset/ask-question-settles-after-answer.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix `ask_question` tool calls appearing to run forever in the dev TUI after the user answers. Resolved question requests now emit a successful `action.result` event (previously only denied tool approvals did), so consumers settle the call and its spinner clears once answered.
--- a/packages/eve/src/harness/input-requests.test.ts
+++ b/packages/eve/src/harness/input-requests.test.ts
@@ -558,6 +558,56 @@ describe("resolvePendingInput", () => {
     });
   });
 
+  it("returns an answered action for a resolved question", () => {
+    const session = setPendingInputBatch({
+      event: { sequence: 5, stepIndex: 1, turnId: "turn_0" },
+      requests: [
+        {
+          action: {
+            callId: "question-call",
+            input: { prompt: "Proceed with this plan?" },
+            kind: "tool-call",
+            toolName: "ask_question",
+          },
+          allowFreeform: false,
+          display: "select",
+          options: [
+            { id: "go", label: "Start researching" },
+            { id: "revise", label: "Adjust the plan" },
+          ],
+          prompt: "Proceed with this plan?",
+          requestId: "question-1",
+        } satisfies InputRequest,
+      ],
+      responseMessages: [],
+      session: createHarnessSession(),
+    });
+
+    const result = resolvePendingInput({
+      stepInput: {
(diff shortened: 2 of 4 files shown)


ruiconti:
@mikkel-bergmann could you please file an issue with repro steps? I cannot reproduce it on latest, but this could be an event ordering issue hence the need for repro steps.

mikkel-bergmann:
@ruiconti Added a harness regression test (`tool-loop.test.ts`) — fails on `main`, passes with this PR. Minimal repro:

```ts
const session = setPendingInputBatch({
  event: { sequence: 0, stepIndex: 0, turnId: "turn_0" },
  requests: [{
    action: { callId: "question-call", input: {}, kind: "tool-call", toolName: "ask_question" },
    display: "select",
    options: [{ id: "go", label: "Go" }, { id: "revise", label: "Revise" }],
    prompt: "Proceed?",
    requestId: "question-1",
  }],
  responseMessages: [{
    content: [{ input: {}, toolCallId: "question-call", toolName: "ask_question", type: "tool-call" }],
    role: "assistant",
  }],
  session: createTestSession(),
});
setupMockAgent({ finishReason: "stop", response: { messages: [{ content: "ok", role: "assistant" }] }, text: "ok", toolCalls: [], toolResults: [] });

const { emit, events } = createEventCollector();
await createToolLoopHarness(createTestConfig("conversation", emit))(session, {
  inputResponses: [{ requestId: "question-1", optionId: "go" }],
});

const settled = events.filter((e) => e.type === "action.result").map((e) => e.data.result.callId);
expect(settled).toContain("question-call"); // main: [] — never emitted
```

Run:
```
pnpm --filter eve build:compiled
  src/harness/tool-loop.test.ts -t "parked ask_question is answered"
```

On `main`: `expected [] to include 'question-call'`. `resolvePendingInput` writes the answer to history, but `tool-loop.ts` emits `action.result` only for `rejectedActions` (denied approvals) — answered questions emit nothing, so a consumer that settles a call on its result (the TUI marks a tool block done only on `tool-result`) never settles it.

mikkel-bergmann:
@ruiconti Sorry, I don't know how to re-open - should I create a new one instead?

PR #488, opened 2026-07-02
feat(eve): surface subagent usage to callers

ruiconti:
### Description

Subagent results carry no token usage — the caller can't attribute a child agent's spend. This adds usage reporting for **all three subagent types**: local, runtime, and remote. It supersedes #363, which covered remote only (deviations from it below).

- One usage shape everywhere: `shared/token-usage.ts` exports `TokenUsage` (`inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheWriteTokens`) with its zod schema. It replaces the harness's `TokenUsageTotals` and is the type on subagent results, callback payloads, and the driver's `done` action.
- Usage is computed once, where the session is live: `turnStep`'s `done` result carries the session-lifetime totals, which flow through the turn-result control payload onto the `done` driver action (optional fields on existing arms are forward-compatible by the `NextDriverAction` contract). `finalizeDone` hands the value to both terminal steps — no `DurableSessionState` threading, no re-read of a session the turn just finished.
- Local and runtime children attach it in the delegated-parent hook (`notifyDelegatedParentStep`); remote children transport the same totals in the `session.completed` callback body, and the caller's callback route validates them and projects them onto the result.
- The parent's turn step emits one `invoke_agent <name>` span per usage-bearing result, with `gen_ai.operation.name=invoke_agent`, `gen_ai.agent.name`, and `gen_ai.usage.*` attributes — all from the OpenTelemetry GenAI semantic conventions (https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/, Development stability; `invoke_agent` is a well-known `gen_ai.operation.name` value, and both cache attributes — `cache_read.input_tokens`, `cache_creation.input_tokens` — are registry attributes).
- Failed sessions never carry usage. Malformed callback usage is dropped while the result still resumes; unknown usage keys from a newer callee are stripped, not rejected. Remote usage needs both sides on this version.

**Two deviations from #363.** First: session totals, not final-turn totals. The `done` action carries `turnUsageState.session` — the session-lifetime accumulation — where #363 read the flat per-turn fields. The two only coincide for single-turn children; a child that parked (HITL, auth) would under-report otherwise. Second: one span site instead of two. #363 emitted the remote span at the callback route, which detaches it from any session trace. Emitting where the parent step ingests `runtime-action-result` nests attribution under the caller's session for every subagent type, and the child's own model spans stay in its own trace.

**Why usage rides the result:** the parent-side receipt is where child spend can later debit the parent's per-session token limits (#424), so per-fan-out child budgets (`remaining / N`) conserve the family total instead of resetting at every dispatch.

- `pnpm --filter eve typecheck`, `pnpm guard:invariants`, `pnpm test:unit`
- New unit tests: `delegated-parent-notification.test.ts` and `session-callback-step.test.ts` (usage attach/transport, error paths), `session-callback-route.test.ts` (projection, unknown-key strip, malformed drop), `subagent-usage-span.test.ts` (emission, skip rules, tracer failure)

### PR Checklist


packages/eve/src/execution/delegated-parent-notification.ts
```diff
@@ -9,15 +9,21 @@ import { ChannelKey } from "#runtime/sessions/runtime-context-keys.js";
...
  * No-op for root sessions.
+ *
+ * `usage` — the completed child's session-total token spend — is
+ * attached to success results so the caller can attribute the
+ * subagent's tokens. Error results never carry usage.
  */
 export async function notifyDelegatedParentStep(input: {
   readonly result: RuntimeSubagentResultActionResult | undefined;
   readonly serializedContext: Record<string, unknown>;
+  readonly usage?: TokenUsage;
```
ctgowrie, line 26: Why optional?

ruiconti: Usage state is stored as a key in `SessionStateMap` and read from it when calling `notifyDelegatedParentStep` which is not strongly typed and could assume undefined values:

https://github.com/vercel/eve/blob/2623682c4efc9f4d89c0a268f6f4dce0751a45af/packages/eve/src/harness/turn-tag-state.ts#L63-L65

AndrewBarba (approved):

PR #491, opened 2026-07-02
research(eve): local self-modification design

ruiconti:
Design doc for letting a dev-mode agent edit its own source tree, per repo convention in `research/`.

**What**

- A self-modifying subagent is declared as a module under `agent/subagents/`, mirroring `defineRemoteAgent`:

  ```ts
  // agent/subagents/selfmod.ts
  import { defineSelfModifyingAgent } from "eve/extensions/selfmod";
  export default defineSelfModifyingAgent({ development: true });
  ```

  The tool name derives from the file slug; `development: true` is a required literal, so the dev-only scoping is an explicit authored statement. Optional `model`, `reasoning`, `instructions` configure the subagent, defaulting to the root model.
- The capability ships as a framework-owned **agent extension**: authored as files inside the eve package (`extensions/self_modification/` — `agent.ts`, `instructions.md`, `sandbox.ts`), discovered with the normal subagent grammar and grafted into the compiled manifest + module map **at compile time** when the target is development.
- Hosted compiles lower the mount to nothing: production artifacts contain no extension node and no extension modules, and the bundler has nothing to trace. Dev and prod runtimes load and resolve artifacts identically — no runtime-only assembly path.
- Its sandbox is the just-bash backend mapped over the real project directory (`ReadWriteFs` root = project root, no template/cache copy), so it carries the unchanged default harness tools, `/workspace` is the live source tree, and `bash` is the pure-JS interpreter — no host binaries, no host process execution. Self-edits take effect through the existing dev source watcher.

**Internal contract**

The mount lowers through an internal `AgentExtension`/`AgentArtifact` API: the producer receives the authored definition and compile target and returns `null` for hosted, or an `AgentArtifact` — an ordinary `CompiledSubagentNode` plus module-map entries and parent edge — that graph resolution, the subagent registry, depth capping, and collision guards consume with zero special-casing. The one new compiler input is `target: "development" | "hosted"`, threaded from the existing `prepareApplicationHost` dev/build fork and folded into compile metadata and runtime cache keys.

**Open points for review**

- Dev and hosted compiles of the same tree now produce different manifests; target must participate in cache keys (specced, worth scrutiny).
- No real binaries inside the subagent (`git`, `pnpm`, `tsc`); allowlisted host commands are a possible follow-up, out of scope here.
- `.eve/` is visible to the subagent (prompt-discouraged, not blocked) in v1.

Tracking issue: #742.

Doc-only; no changeset.

### DIFF
--- a/research/local-self-modification.md
+++ b/research/local-self-modification.md
@@ -1,5 +1,5 @@
 ---
-issue: TBD
+issue: https://github.com/vercel/eve/issues/742
 last_updated: "2026-07-02"
 status: proposed
 ---
--- a/research/local-self-modification.md
+++ b/research/local-self-modification.md
@@ -8,225 +8,137 @@ status: proposed
 
 ## Summary
 
-An agent running under `eve dev` should be able to read and modify its own source tree — its
-instructions, tools, skills, and app code — so a developer can iterate on the agent by talking to
-it. This is host filesystem access, so it is opt-in per agent and exists only in local development.
-Production builds never contain this capability, regardless of configuration.
-
-The author declares it as a subagent module, the way remote agents are declared. The capability
-itself ships as a framework-owned **agent extension**: a subagent authored as files inside the eve
-package, grafted into the compiled manifest and module map at compile time when the compile target
-is development. Hosted compiles never graft it, so its modules are absent from production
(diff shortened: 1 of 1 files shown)


ruiconti:
Closing in favor of #1632

Issue #483, opened 2026-07-02
No way to cancel an in-flight turn: stop() only detaches the client; the model call keeps running (and billing)


manxz:
### Environment

- eve `0.18.1` (dev sidecar via `withEve` in a Next.js 16 host; same harness code path exists in `eve build` output)
- `ai` 7.0.2, `@ai-sdk/openai` 4.0.0, Node 24
- Model: `openai("gpt-5-mini")`, streamed turns over the HTTP channel

### Summary

There is currently no way — client or server — to cancel a running turn. `stop()` detaches the client, and because in-flight turns are deliberately resumable across disconnects (`session.stream({ startIndex })`), detaching can never double as cancelling. The result: the in-flight provider call keeps running after a user hits Stop, so **users are billed for generations they explicitly cancelled**, and in one observed mode the abandoned pipeline stalls without ever settling.

### Details

Client side, `stop()` only aborts the client↔server HTTP stream:

```js
stop(){this.#i?.abort()}   // dist/src/client/eve-agent-store.js
```

Server side, nothing forwards any cancellation to the model call:

- The emission loop swallows the dead-socket write error and returns:
  ```js
  catch(e){if(e instanceof TypeError)return;throw e}   // dist/src/harness/emission.js
  ```
- No `abortSignal` is threaded into model calls anywhere in the harness. The call site is `packages/eve/src/harness/tool-loop.ts:750` — `await agent.stream({ messages: callMessages })` with no signal — and `protocol/` has no cancel action to connect one to. (The workflow interrupt machinery in `harness/workflow-interrupt-state.ts` is for sandbox interrupts, not turn cancellation.)

The abandoned call then goes one of two ways (both observed live, timing-dependent):

1. **Background completion** — the call runs to completion server-side. Verified via usage metering: a turn stopped mid-stream still delivered its full post-call usage (`inputTokens.total=24674 / outputTokens.total=2469`) ~10s after the stop. Consistent with the documented resumable design, but it means every Stop bills the full generation.
2. **Stall** — the pipeline never settles: no `finish` part, no error, no cancellation ever reaches a `wrapLanguageModel` middleware wrapped around the model (observed >6 minutes with no stream end of any kind). This mode also defeats the resumable-turn promise for that turn.

### Impact

- Users pay for every generation they cancel.
- Stream-observing `wrapLanguageModel` middleware (we meter usage this way) can see no end signal at all in the stall case — we had to ship a watchdog timer to account for abandoned calls.
[...]

PR #494, opened 2026-07-02
feat(eve): cancellation - propagate turn abort signals through the harness

AndrewBarba:
## Summary

Layer 0 of the turn-cancellation stack ([plan](https://github.com/vercel/eve/blob/barba/turn-cancellation-harness-abort/research/turn-cancellation-harness-abort.md), #483): the tool-loop harness accepts an optional `AbortSignal` and honors it end to end. **No trigger exists yet — nothing creates or aborts the signal in production, so runtime behavior is unchanged.** Later layers add the durable controller in `turnWorkflow`, the HTTP cancel route, descendant cascade, and client APIs, per the ship ladder in the plan.

Five prior attempts (#118/#127/#128/#135, #230, #347) chained this propagation layer to the trigger/cross-process layers that kept breaking (raced hooks in the non-terminating driver, hook reclaim conflicts, duplicate finalizer steps). This PR banks the converged propagation layer alone: no hooks, no routes, no events, no upstream Workflow dependencies.

## What's inside

**Canonical semantics** — new `harness/turn-cancellation.ts`:
- `TurnCancelledError`: plain, cause-free, matched by `name` so it survives step serialization
- `isTurnCancellation()`: cause-chain walk; deliberately does **not** match generic `AbortError`/`TimeoutError` (a tool's internal fetch timeout is a failure, not a cancellation)
- `throwIfTurnAborted()`: cancellation is detected from the signal, never from thrown error shapes

**Harness propagation**:
- `ToolLoopHarnessConfig.abortSignal` → `agent.stream`/`agent.generate` and compaction's `generateText`
- retry loop checks the signal before each attempt and in each catch — an abort surfacing as a retryable transport error is never retried
- recovery pipeline guarded on both sides; `classifyModelCallError` returns `terminal` for cancellations (pins the abort-vs-failure invariant that leaked in #347's Windows run)
- `abort` stream part throws immediately so the harness never awaits an `onStepFinish` step result the SDK will never resolve

**Tool propagation**:
- `wrapToolExecute` forwards the AI SDK execute options — the single splice point for every tool kind
- public `ToolContext.abortSignal`, always present (inert signal until a trigger exists) so authored code is written once; dynamic tools get it via new `buildBaseToolContext`

**Framework executors**:
- `bash`/`grep`/`glob` turn the previously-dead sandbox `abortSignal` plumbing live (`sandbox.run`); `read_file`/`write_file` forward it into file I/O
- `web_fetch` composes the signal with its timeout via `AbortSignal.any`
- MCP `executeTool` threads it into the JSON-RPC request; OpenAPI into its `fetch`

**Execution ingress** — `TurnStepInput.abortSignal?` → `turnStep` → `createExecutionNodeStep`, documented as the seam layer 1's durable `AbortController` plugs into.

Tool cancellation is best effort by design: a tool that ignores the signal runs to completion, but the surrounding model-call promise settles on abort and the straggler result is discarded.

## Testing

- new `turn-cancellation.test.ts` (error identity, cause-chain, signal normalization) and `classifyModelCallError` cancellation cases
- `tool-loop.test.ts`: streamed cancellation settles without failure events; aborted calls are never retried or recovered; already-aborted signal never starts a model call
- `tools.test.ts`: options forwarding, `ctx.abortSignal` identity, inert-signal fallback
- new `tool-loop-cancellation.integration.test.ts` drives a **real `ToolLoopAgent`** over a cancellable `MockLanguageModelV3`: mid-stream abort settles with the canonical error and zero failure events; an executing tool observes a live signal; an inert signal produces output identical to no signal
- full suites green: 4338 unit + 375 integration, `typecheck`, `lint`, `fmt`, `guard:invariants`, `docs:check`

#347's e2e fixture (`agent-cancellation` / `wait-for-cancellation`) stays shelved until layer 2 provides a trigger — there is nothing to cancel with yet.

## Docs

- `docs/tools/overview.mdx`: `ctx.abortSignal` in the tool context list
- `research/turn-cancellation-harness-abort.md`: the layered plan this PR executes
- patch changeset


packages/eve/src/execution/sandbox/bash-tool.ts
```diff
@@ -81,10 +83,11 @@ export async function executeBashOnSandbox(
 async function runWithDevelopmentSandboxProgress(
   sandbox: SandboxSession,
   command: string,
+  abortSignal?: AbortSignal,
 ): Promise<Awaited<ReturnType<SandboxSession["run"]>>> {
   logDevelopmentSandboxCommand(`eve: starting sandbox command: ${formatCommand(command)}`);
   if (!isEveDevEnvironment()) {
-    return await sandbox.run({ command });
+    return await sandbox.run({ abortSignal, command });
```
ctgowrie: Should we wrap the sandbox and ensure abortSignal is always sent so authors don't have to do this manually? (While still allowing override)

manxz: That's what I'd expect as the default

AndrewBarba: Done in c5c5dffb — and it turned out to be more than a convenience: `ctx.getSandbox()` hands authors a raw session, so author code calling `sandbox.run({ command })` directly previously got no cancellation unless it remembered to thread `ctx.abortSignal`.

New `bindSandboxAbortSignal(session, signal)` wraps all nine I/O methods and injects the signal by default. It's applied at the two delivery points — `requireSandboxSession(abortSignal)` for the framework tools and `ctx.getSandbox()` via `buildBaseToolContext` — which let me delete the manual threading from all five executors (`SandboxToolExecuteOptions` is gone) and pick up previously-unsignalled calls like the ripgrep availability probe for free.

One deliberate deviation from "allow override": a per-call signal *composes* with the turn signal via `AbortSignal.any` instead of replacing it. A caller-supplied timeout still works, but nothing can opt sandbox work out of turn cancellation — a leaked long-running command after cancel is exactly the bug class this PR exists to prevent. Shout if you actually wanted full replacement semantics.

[automated reply from opencode]

ctgowrie (approved):

PR #495, opened 2026-07-02
[bisect] usage plumbing probe — do not merge

ruiconti:
Diagnostic probe for the dynamic-workflow e2e regression on #488. Reverts the done-action usage plumbing while keeping the Usage type consolidation. Only the `e2e-vercel (agent-subagents)` result matters. Will be closed without merging.


ruiconti:
Superseded: this probe run merged in newer main commits (model-catalog + #493), contaminating the signal. PR #496 is the clean attribution probe.

PR #500, opened 2026-07-02
fix(eve): drain Microsandbox sessions on dispose

scottschindler:
## Related issue

- Related issue: #504

## Motivation

While investigating a local eve development run, the machine rapidly drained battery and the fan spun up. The immediate cause was one or more `msb sandbox` processes continuing to consume CPU after the process that started the eve session had exited.

Those `msb` processes were detached Microsandbox VMs. eve creates Microsandbox sessions in detached mode so they can survive SDK client reconnects, but sandbox handle disposal only released the SDK client with `detach()`. If the normal development cleanup path does not run, the VM can remain independently running even though eve has released the session handle.

This draft change makes disposal drain the Microsandbox VM before detaching the SDK client. The direction needs maintainer confirmation because `dispose()` may be part of the reusable detached-session contract, while terminal development cleanup may need a separate stop/snapshot path.

## Summary

- drain Microsandbox VMs before detaching the SDK client when eve disposes a sandbox handle
- add unit coverage for handle disposal and drain-before-detach ordering
- add a patch changeset for eve

## Open questions for maintainers

I am not certain these lifecycle decisions are mine to make as a contributor, so I would appreciate maintainer direction before this is considered final:

- Should `dispose()` for a Microsandbox handle preserve the reusable detached-session contract, or is it the right place to signal that eve is done with the VM and it should leave the running state?
- If disposal should terminate or wind down the VM, is `drain()` the right Microsandbox primitive, or should eve use a bounded shutdown path such as `stopWithTimeout()` or the existing development cleanup/snapshot flow?
- How should reconnect handle sandboxes whose provider status is `draining`? Should eve continue treating them as connectable live sessions, or should it restore/start from durable state instead?
- Should this behavior apply to all Microsandbox handle disposal paths, or only to development cleanup cases where the user is exiting `eve dev` and wants leftover VMs stopped?
- What level of scenario coverage would maintainers want here: mocked call-order unit tests only, or a real Microsandbox scenario that proves dispose/reconnect or dispose/shutdown behavior end to end?

## Verification

Completed locally:

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/sandbox/bindings/microsandbox-lifecycle.test.ts src/execution/sandbox/bindings/microsandbox-runtime.test.ts`
- `pnpm lint`

Pending before this is ready for final review/merge:

- broader `pnpm test` / maintainer-requested scenario coverage after the lifecycle direction is confirmed

### DIFF
--- a/.changeset/drain-microsandbox-sessions.md
+++ b/.changeset/drain-microsandbox-sessions.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Drain Microsandbox VMs when disposing eve sandbox handles so detached sessions do not remain running after the client is released.
--- a/packages/eve/src/execution/sandbox/bindings/microsandbox-lifecycle.test.ts
+++ b/packages/eve/src/execution/sandbox/bindings/microsandbox-lifecycle.test.ts
@@ -148,6 +148,28 @@ describe("createMicrosandboxHandle", () => {
     });
   });
 
+  it("requests a VM drain before detaching a disposed session handle", async () => {
+    const vm = createFakeMicrosandboxVm("session-key");
+    runtimeMocks.createPreparedMicrosandbox.mockResolvedValue(vm);
+    const options = resolveMicrosandboxOptions({ image: MICROSANDBOX_DEFAULT_IMAGE });
+
+    const handle = await createMicrosandboxHandle({
+      backendName: "microsandbox",
+      createInput: {
+        runtimeContext: { appRoot: "/tmp/eve-app" },
+        sessionKey: "session-key",
+        templateKey: "template-key",
+      },
+      options,
+      optionsHash: "options-hash",
+    });
+
+    await handle.dispose();
+
+    expect(vm.drainAndDetach).toHaveBeenCalledTimes(1);
+    expect(vm.detach).not.toHaveBeenCalled();
+  });
+
   it("reports a missing template snapshot race as not provisioned", async () => {
     runtimeMocks.createPreparedMicrosandbox.mockRejectedValueOnce(
       new Error("snapshot template-snapshot not found"),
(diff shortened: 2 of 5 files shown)


scottschindler:
@AndrewBarba @ijjk @ruiconti would you be willing to give this an early review and tell me whether the direction looks right before I keep iterating?

I am especially looking for maintainer direction on the lifecycle questions in the PR description: whether Microsandbox handle `dispose()` should preserve reusable detached sessions, actively wind down the VM, or delegate that to a separate cleanup path.

Why I tagged each of you:

- @AndrewBarba: blame points to you for the original Microsandbox backend lifecycle implementation, including the handle `dispose()`/`detach()` path, dev `captureState()`, detached VM creation, and development cleanup flow.
- @ijjk: blame points to you for the missing sandbox backend state/reconnect handling, which is directly related to whether `draining` sandboxes should still be treated as reconnectable live sessions.
- @ruiconti: you have recent Microsandbox runtime work around create progress and failure handling, so I would value your read on whether the cleanup/error-handling direction fits the current runtime conventions.

No rush; I mainly want to avoid making the wrong framework-level lifecycle call as an outside contributor.

ijjk (commented):
Thanks for digging into this, the leaked `msb` processes were a real gap.

After tracing the lifecycle, I took a different approach in #510:

1. **`dispose()` has no production callers.** The only path that could invoke it (`SandboxAccess.dispose()` in `ensure.ts`) is never called by the runtime, so draining there couldn't actually fix the leak, the code path only runs in tests.
2. **Your question about the detached-session contract was the right one.** `dispose()` meant "release the client, keep the session warm" on other backends, and reconnect treats `draining` sandboxes as connectable, draining on dispose would have conflicted with that, exactly as you suspected.
3. **The real gap was that nothing triggered cleanup in production.** The built server had no signal handling at all, so nothing stopped VMs or containers regardless of what `dispose()` did.

#510 adds an explicit `shutdown()` contract on sandbox handles (stop the compute, keep the session reattachable), a registry of live handles, and a runtime plugin that drains it on `SIGTERM`/`SIGINT` in production. Since `dispose()` was dead surface with inconsistent semantics, I removed it entirely.

PR #511, opened 2026-07-02
Expose Gateway cost metadata

JohnPhamous:
- add Gateway generation id and costUsd to step.completed when AI Gateway returns provider metadata
- persist cumulative turn cost as $eve.cost_usd only after Gateway reports cost
- bump stream version to 17

### DIFF
--- a/packages/eve/src/harness/step-hooks.ts
+++ b/packages/eve/src/harness/step-hooks.ts
@@ -3,6 +3,7 @@ import type {
   LanguageModelUsage,
   ModelMessage,
   PrepareStepFunction,
+  ProviderMetadata,
   StepResult,
   ToolSet,
   ToolResultPart,
@@ -13,6 +14,7 @@ import {
   createActionResultEvent,
   createActionsRequestedEvent,
   createStepCompletedEvent,
+  type StepCompletedProviderMetadata,
 } from "#protocol/message.js";
 import {
   createRuntimeToolResultFromToolError,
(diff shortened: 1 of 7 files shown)


packages/eve/src/harness/step-hooks.ts
```diff
@@ -417,3 +436,46 @@ function extractStepUsage(usage: LanguageModelUsage | undefined):
...
+  }
+  return undefined;
+}
+
+function readGatewayGenerationId(providerMetadata: unknown): string | undefined {
+  const generationId = readGatewayMetadata(providerMetadata)?.generationId;
+  return typeof generationId === "string" && generationId.length > 0 ? generationId : undefined;
+}
+
+function readGatewayMetadata(providerMetadata: unknown): Record<string, unknown> | undefined {
```
ruiconti: Could we type `providerMetadata`? AFAIK `ai` exposes it

packages/eve/src/protocol/message.ts
```diff
@@ -987,10 +993,16 @@ export function createStepStartedEvent(input: {
  */
 export function createStepCompletedEvent(input: {
   readonly finishReason: AssistantStepFinishReason;
+  readonly providerMetadata?: {
+    readonly gateway?: {
+      readonly generationId?: string;
+    };
+  };
```
ruiconti: Also could this be inferred from `ai` types?

ruiconti (approved):

PR #515, opened 2026-07-02
feat(eve): add `experimental_chatgpt` model support

### LINKED ISSUE #44: Support for agent runtimes like Codex/Claude?

Would love to use Codex to drive the agent inside the sandbox instead of defining a fully custom one.

Being able to define skills and MCPs for the agent runtime to use based on agent definition would be very useful.

### Proposed solution

Add support for alternative agent runtimes (Codex App Server or Agents SDK, Claude Agents SDK)

### Alternatives considered

_No response_

ruiconti:
Closes #44.

Lets an agent use a ChatGPT subscription as its model, billed through the local `codex login` instead of an API key:

```ts
import { experimental_chatgpt } from "eve/models/openai";

export default defineAgent({
  model:
    process.env.NODE_ENV === "production"
      ? "anthropic/claude-sonnet-4.6"
      : experimental_chatgpt("gpt-5.5"),
  modelContextWindowTokens: 200_000,
});
```

**What**

- new `eve/models/openai` subpath exporting `experimental_chatgpt(slug)`
- codex transport: auth file read/refresh, fetch wrapper, model wrapper, all under `public/models/openai/codex/`
- direct provider API errors now report the upstream message instead of a generic retries-exhausted error

The export is named for the ChatGPT subscription that authorizes and bills it — the axis that differs from an API-key model, and the brand the codex CLI itself uses (`auth_mode: "chatgpt"`). It lives under the `eve/models/openai` provider namespace; the `codex/` internals keep their name because Codex is the backend that serves the subscription.

`experimental_chatgpt` returns a plain AI SDK language model, so it rides the existing source-backed instance path: the compiler classifies it as external and the runtime reloads it from `agent.ts`. No compiler, runtime, or TUI branches. An earlier draft integrated codex into all of those (a `model:` descriptor, compile-mode fallback selection, `/model` rewriting, status-line states) and came out three times this size. Shipping it as a self-contained import keeps the framework neutral on a backend that is not a public API contract. No new runtime dependencies; `@ai-sdk/openai` was already vendored.

## Transport internals

`public/models/openai/codex/` holds the three pieces the model needs. The directory is self-contained: it imports only `#compiled` vendored deps and two `#shared` utils, so extracting it to its own package later is a move plus an import swap.

- `auth.ts` reads `~/.codex/auth.json`, refreshes expired ChatGPT access tokens, and writes the refreshed file back the way the codex CLI does.
- `transport.ts` wraps fetch: routes to the Codex Responses backend, attaches the access token and account id, retries once after a refresh on 401.
- `model.ts` wraps the vendored `@ai-sdk/openai` responses model. The backend rejects stored responses and server-side item ids, so every call forces `store: false` and strips `openai.itemId` from the prompt (copy-on-write, long conversations are shared rather than cloned per call).

The backend enforces model availability per account at call time. An unknown slug fails the turn with the upstream message, e.g. `The 'foo' model is not supported when using Codex with a ChatGPT account.`

## Provider error messages

`summarizeKnownModelCallRequestError` only summarized gateway 400s; a direct provider rejection produced "retries exhausted" with no detail. It now returns the upstream API error message for terminal failures from any provider, and keeps the generic framing for transient statuses so throttles don't read as configuration bugs. This is what makes the ChatGPT rejection above debuggable, and it applies to every direct-provider model, not just this one.

## What this skips

The status line shows the model as external with no connectedness claim, and `/model` cannot switch these models because eve does not rewrite authored source. Both would need a public model-endpoint extension point; worth doing when a second subscription backend shows up.

Verified against the live backend: an agent on this branch compiles to `codex/gpt-5.5` (external routing) and completes a turn through the ChatGPT subscription; asking for `foo` fails the turn with the 400 above.


AndrewBarba:
`codex` is a harness not a model so I think we need to rename this and move the import:

```
import { chatGPT } from "eve/models/openai"
```

AndrewBarba (approved):

PR #516, opened 2026-07-02
feat(eve): add experimental_codex authoring

ruiconti:
Stack:
- Previous: #515
- Base: `rui/codex-transport-core`
- Next: #517

**What**
- Exports `experimental_codex(model, fallback?)` as a `defineAgent({ model })` value.
- Compiles `experimental_codex("model")` to `openai/<model>` plus `transport: "codex"` in development.
- Keeps production builds on deployable model credentials.
- Carries the transport marker through the compiled manifest and runtime model resolution.
- Adds the model-subscription guide and TypeScript API entry.

**Authoring contract.**
`experimental_codex(...)` is a model value, not an agent-wide flag. That ties local Codex transport to the exact model that needs it and lets production builds choose a deployable path for that same model slot.

**Compiler behavior.**
Development builds skip AI Gateway catalog validation for the Codex-backed model. Codex compatibility belongs to the Codex backend, and authors can still provide `modelContextWindowTokens` when they want compaction to use a known window.

Production builds never use a local Codex login. They keep `openai/<model>` on AI Gateway when the catalog confirms it, compile the fallback when it does not, and fail when no deployable path exists.

**Manifest and runtime.**
Compiled model references can now carry `transport: "codex"`. Runtime resolution uses that marker to ask the transport-core endpoint for the model. Source-backed production fallbacks are reloaded from the authored module and validated as a normal model value.

**Not here.**
The `/model` flow, status line, setup issues, and source rewrite UX are in #517.


ruiconti:
Superseded by #515. The authoring value experimental_codex now ships there as a plain AI SDK language model under eve/extensions/model, riding the existing source-backed external-instance path. That removes everything this PR added: the descriptor kind, the compiler normalization with compile-mode fallback selection, the manifest transport marker, and the runtime unwrap. The dev/prod split is a ternary in agent.ts instead of compiler behavior.

PR #517, opened 2026-07-02
feat(eve): wire Codex models into dev UX

ruiconti:
Stack:
- Previous: #516
- Base: `rui/codex-author-compile`

**What**
- Teaches `/model` and setup model selection how to read Codex-backed model state.
- Lets development model selection accept Codex model ids without using the AI Gateway catalog as the authority.
- Rewrites the first argument inside `experimental_codex(...)` while preserving the fallback argument.
- Reports Codex, Gateway, and external model endpoint state from `/eve/v1/info`.
- Updates TUI setup/status copy and client parsing for Codex endpoint states.

**Model selection.**
Codex-backed development models are validated against the local Codex auth path, not the Gateway catalog. Gateway models still use the normal catalog picker and credential checks.

**Agent info.**
Endpoint status now comes from the compiled model facts when the agent-info response is built. The status line and setup issue detection read that response instead of guessing from environment variables or source text.

**Source rewrite.**
`/model <provider/model-id>` updates a normal string model as before. When the authored source uses `experimental_codex(...)`, it rewrites the Codex model argument and leaves the production fallback in place.

**Not here.**
The Codex transport is in #515. The public authoring and compiler contract is in #516.


ruiconti:
Superseded by the extension reshape of #515. This layer wired the codex descriptor into /model, the status line, and setup issues; the descriptor and the endpoint factories it dispatched on were removed when experimental_codex became a plain AI SDK model exported from eve/extensions/model. The status line now shows codex models as external with no connectedness claim, and /model does not switch them, because eve does not rewrite authored source. Bringing this UX back means designing a public model-endpoint extension point first; worth doing when a second subscription backend shows up.

PR #520, opened 2026-07-03
fix(runtime): flatten top-level union in MCP tool input schemas

sandwichlegend:
## Problem

A connection (MCP) tool whose `input_schema` uses a top-level `oneOf`/`allOf`/`anyOf` union takes down the **entire turn**. Anthropic's Messages API rejects such a schema and fails the whole model request with an HTTP 400 **before the agent produces a response**:

```
tools.N.custom.input_schema: input_schema does not support oneOf, allOf, or anyOf at the top level
```

Because eve passes third-party MCP tool schemas straight through (`mcp-client.ts` → `toolsFromDefinitions` + metadata), one non-conforming external tool poisons the request for every other tool in it. The failure surfaces to the user only as the generic `summarizeKnownModelCallRequestError` message ("AI Gateway rejected the model request before the agent produced a response"), with no indication of which tool caused it.

This was hit in production against a live third-party MCP server whose tool schema had a top-level union. eve can't author or pin external schemas — and a server can change them at any time — so the runtime has to defend against it.

## Fix

Sanitize at MCP ingestion, before the schema reaches the model:

- New `flattenTopLevelUnionToolInputSchema` (`shared/json-schema.ts`) rewrites a top-level union into a single object schema — merging the branches' `properties` (first branch wins on collision, all optional) with `additionalProperties: true`, so no valid call is rejected. Top-level `description`/`title`/`$schema` are preserved; **nested** unions are left untouched (Anthropic only forbids them at the top level). The server's own validation still enforces the real union at call time.
- `McpConnectionClient` applies it to every listed tool, feeding the sanitized definitions to **both** `toolsFromDefinitions` and the connection metadata, and **logs the offending connection + tool name** so a recurrence is diagnosable at a glance instead of only as an opaque `tools.N.custom.input_schema` rejection.

## Tests

- `shared/json-schema.test.ts` — unit coverage for anyOf/oneOf/allOf, no-op passthrough (by reference), property-collision precedence, non-object branches, annotation preservation, and nested-union passthrough.
- `mcp-client.test.ts` — wiring test asserting a listed tool's top-level union is flattened in both the exposed metadata and the definitions handed to `toolsFromDefinitions`, and a conforming tool is untouched.

`pnpm --filter eve typecheck`, `pnpm lint`, `pnpm fmt`, and the two test files (50 tests) pass locally. Changeset included (`patch`).

### DIFF
--- a/.changeset/mcp-tool-schema-top-level-union.md
+++ b/.changeset/mcp-tool-schema-top-level-union.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+MCP connection tools whose input schema uses a top-level `oneOf`/`allOf`/`anyOf` union are now flattened to a single object schema before the model call. Anthropic rejects a top-level union in a tool's `input_schema` and fails the entire request with an HTTP 400 before the agent can respond, so one non-conforming third-party tool previously took down every tool in the turn. The flattened schema merges the union branches' properties (left optional, `additionalProperties` open) so no valid call is rejected, and the offending connection and tool name are logged. Nested unions are left untouched.
--- a/packages/eve/src/runtime/connections/mcp-client.test.ts
+++ b/packages/eve/src/runtime/connections/mcp-client.test.ts
@@ -343,6 +343,65 @@ describe("McpConnectionClient authorization recovery", () => {
   });
 });
 
+describe("McpConnectionClient tool input schema sanitization", () => {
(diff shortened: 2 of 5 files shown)


sandwichlegend:
Withdrawing — opened this prematurely. The consuming project is several minor versions behind and there's no active breakage to justify it upstream right now. Apologies for the noise.

PR #530, opened 2026-07-03
fix(eve): preserve batched Slack HITL cards

ruiconti:
What changed:

- Slack HITL button prompts now render as raw Slack `card` blocks.
- Approval prompts render as `Deny` then primary `Allow`, matching the card shape we want in Slack.
- The answered-message update path now finds HITL actions in both legacy `actions.elements` blocks and card `actions` arrays.
- When a card is answered, Eve removes only that card's actions, keeps its body, and leaves sibling cards in the same Slack message clickable.
- Card body text is truncated to Slack's 200-character `body.text` limit: https://docs.slack.dev/reference/block-kit/blocks/card-block

Why this shape:

The original bug was caused by updating the whole Slack message after one HITL answer. In a batched message, that removed the other approvals. Card blocks make each prompt a single block, so the update can replace only the clicked card and preserve the rest of the batch.

Select and freeform HITL prompts stay on their existing Slack controls. This PR changes the button prompt path because that is the path used by approvals.

| Patch | Before click | After click |
|-------|--------------|-------------|
| Current | image | image |
| After fixing | 01-before-click | 02-after-first-click |


AndrewBarba (approved):

rpelevin:
The important invariant here is to make the approval, not the Slack message, the unit of mutation.

I would keep four cases explicit:

1. one click consumes exactly one pending requestId
2. sibling approval cards stay actionable until they reach their own terminal state
3. Slack action_id, requestId, and inputResponses entry stay bound across chat.update
4. duplicate or stale clicks on an already answered card are terminal no-effect, not a second resume

For regression coverage, I would include approve-first/deny-second and deny-first/approve-second in the same Slack message, plus a replay of the already answered action_id. The assertions I would want are both the outgoing inputResponses payload and the remaining Slack block action ids, so runtime state and UI state cannot diverge.

That keeps batched HITL as independent authority records rather than one message-level mutation.

rpelevin:
The new e0 batched Slack approval test is the right regression shape.

I would keep one remaining boundary explicit before treating the PR as resolved: runtime success and UI preservation should be asserted as separate support.

The new test appears to cover the observed e0 batched escalation approval path. The invariant I would preserve is:

- one click consumes exactly one pending requestId
- sibling cards remain actionable until their own terminal state
- each Slack action_id stays bound to the same requestId and inputResponses entry
- stale or duplicate clicks are terminal no-effect
- the terminal Slack block state and the emitted inputResponses payload are checked together

The remaining red e2e-vercel agent-basic-runtime check may be unrelated to this Slack path, but I would not let a green local HITL test substitute for the cross-runtime proof if the failing Vercel path can touch the same request lifecycle.

So the merge-safe contract I would want is:

1. batched sibling approvals are preserved in local tests;
2. deployed/Vercel runtime does not collapse or replay the same pending approval set;
3. DCO/signoff is clean;
4. a stale click after one sibling completes cannot mutate another sibling request.

That keeps this as approval-record preservation rather than only Slack card rendering.

Boundary: architecture and test-contract feedback only; no claim about running this code or using this project.

PR #559, opened 2026-07-07
fix(eve): pin session sandboxes per durable session

ruiconti:
On Vercel, a durable session loses its sandbox `/workspace` on every redeploy. Session keys embed a scope hashed from `VERCEL_DEPLOYMENT_ID`, so a new deployment derives a new key, the persisted reconnect metadata stops matching, and `ensure()` provisions a fresh sandbox. `docs/sandbox.mdx` promises the opposite: "eve preserves the filesystem and resumes the sandbox on the next message as if nothing happened, even days later."

**What**

- Session keys now use a deploy-stable scope (Vercel project id) instead of the deployment id
- Session keys fold in the sandbox definition's version hash, so editing the sandbox rotates the session onto the rebuilt template
- `onSession` re-runs when the key rotates, so per-session setup lands on the replacement sandbox
- Project id resolves from `VERCEL_PROJECT_ID` or the `project_id` claim of `VERCEL_OIDC_TOKEN`, via a new `#shared/vercel-project` module that `vercel-credentials` also uses now
- The unreachable `source-graph` template plan and the deployment scope mode it forced are deleted; scope resolution has one rule

## Key derivation

Old: `eve-sbx-ses-{backend}-{deployment scope}-{sessionId}-{node}`. New: `eve-sbx-ses-{backend}-{project scope}-{version}-{sessionId}-{node}`.

The version segment reuses the template version hash (bootstrap source hash, `revalidationKey`, workspace seed content hash). A redeploy keeps the session sandbox. Changing the sandbox definition rotates it, since the old filesystem was built from a template that no longer exists. The eve package version does not participate, so upgrading eve keeps session state.

Both keys build from one `deriveRuntimeSandboxKeyParts` value, so the coupling holds by construction: the session key rotates exactly when the template content rotates, and a future edit cannot silently decouple them.

Nothing keys on the deployment id anymore. `createRuntimeSandboxTemplatePlan` could never return the `source-graph` plan that used it, so this PR deletes the plan kind, the `scopeKind` parameter, and the deployment branch of scope resolution outright. The remaining rule: Vercel project id (env var or OIDC token claim), else `realpath(appRoot)`, else the artifacts cache key. The OIDC claim is how sandbox creation authenticates in the first place, so any deployment where the backend works can resolve it, and the claim is identical at build-time prewarm and deployed runtime.

## onSession re-run

`ensure()` used to trust the persisted `initialized` flag, so after any key rotation the fresh sandbox skipped `onSession` and ran without network policy or per-user credentials. The persisted record now narrows to `reattachSession` once; a mismatch resets `initialized` at the point the mismatch is known.

## Scope

- The key format changed, so existing session sandboxes rotate once when this ships. Pre-1.0, no migration.
- Derived key strings are otherwise unchanged for every reachable input; only test-constructed `source-graph` keys differ.
- `dev-client` has its own OIDC claim decoder with a structured failure taxonomy for setup UX. Left alone.


packages/eve/src/runtime/sandbox/keys.ts
```diff
@@ -56,28 +62,83 @@ export async function createRuntimeSandboxTemplateKey(input: {
...
+): string | null {
+  if (parts.versionHash === null) {
     return null;
   }
 
-  const metadata = await loadCompileMetadataForKeys(input.compiledArtifactsSource);
-  const scope = await resolveRuntimeSandboxScope({
-    backendName: input.backendName,
-    compiledArtifactsSource: input.compiledArtifactsSource,
-    scopeKind: input.templatePlan.kind === "source-graph" ? "deployment" : "stable",
```
ruiconti, line 67: This drops the `deploymentId` from the hash and always uses `projectId` as `scope`.

AndrewBarba (approved):

PR #569, opened 2026-07-07
fix(eve): bubble subagent authorization events up to the parent session

### LINKED ISSUE #568: Subagent auth requests are not bubbled up to the parent session
## Summary

When a subagent's tool run requires authorization (e.g. an OAuth flow for a connection), the resulting `authorization.required` event is never forwarded to the parent session or the user-facing channel. The child session parks with a pending authorization and hangs indefinitely. Approval and other input requests from subagents bubble up correctly; auth requests do not.

Reported by a customer as a blocker for using subagents.

## Reproduction

1. Give a subagent a tool/connection that triggers `requestAuthorization`.
2. Have the parent agent invoke the subagent.
3. The subagent's tool returns an auth signal; the child emits `authorization.required` and parks.
4. The event is written only to the child's own stream — nothing reaches the parent or the channel. The user never sees an auth prompt and the run never completes.

In contrast, a tool approval or `ask_question` from the same subagent is shown to the user as expected.

## Cause

Input requests bubble up because `SUBAGENT_ADAPTER` declares an `"input.requested"` handler (`src/execution/subagent-adapter.ts`) that forwards a `subagent-input-request` hook payload to the parent, where the HITL proxy (`src/execution/subagent-hitl-proxy.ts`) re-emits it on the parent's stream.

Auth uses a separate event type, `authorization.required` (`src/harness/tool-loop.ts`). The subagent adapter has no handler for it, so `callAdapterEventHandler` (`src/channel/adapter.ts`) passes it through to the child's stream, which no interactive channel consumes. The `HookPayload` union (`src/channel/types.ts`) has no auth variant, so there is currently no way to carry the request to the parent.

## Proposed fix

Mirror the existing HITL proxy path:

- Add an `"authorization.required"` handler to `SUBAGENT_ADAPTER` that forwards the challenge to the parent via `resumeHook`.
- Add a corresponding `HookPayload` variant and a branch in the parent's turn-workflow inbox loop.
- Re-emit the auth request on the parent's stream so the channel renders it, and route the completed authorization back down to the parked child (analogous to `routeDeliverToChildren`).

Auth requests should propagate through nested subagent chains the same way HITL requests already do.

blurrah:
## Summary

Fixes #568.

A subagent whose connection needed interactive OAuth parked silently: the `authorization.required` event was written only to the child's unconsumed stream, so the sign-in link never reached the user and the run hung. Approval and other input requests already bubble up through the HITL proxy; authorization events did not.

This mirrors the **upward half** of the HITL pipeline — and only that half:

- `SUBAGENT_ADAPTER` gains `authorization.required` / `authorization.completed` handlers that forward the child's event verbatim to the parent's turn inbox as a new runtime-internal `subagent-authorization-event` hook payload.
- The parent's inbox loop dispatches it to a new `runProxyAuthorizationEventStep`, which re-emits the event through the parent's channel adapter (so Slack, the TUI, etc. render the sign-in challenge) and persists adapter-state mutations across the step boundary.
- **No downward routing**: the challenge's webhook URL targets the child's own `${childSessionId}:auth` hook, so the OAuth callback resumes the child directly. No routing map, no delivery request through the parent's channel.
- **No turn epilogue**: unlike proxied HITL, the parent turn is not waiting on its own channel — it's still awaiting the subagent result, and the child resumes autonomously after the callback. Pinned with a test.
- Nested delegation chains forward recursively for free: each middle hop's channel adapter is itself the subagent adapter, so the re-emit triggers the next hop up.

`authorization.completed` is bubbled too so channels can resolve their pending-auth UI (e.g. Slack edits its "waiting for sign-in" message).

The step lives in a new `subagent-auth-proxy.ts` module (sibling of `subagent-hitl-proxy.ts`) rather than `workflow-steps.ts`, which is at the 700-line structural cap.

## Tradeoff noted

Re-emitting verbatim means Slack's pending-auth message is keyed by bare connection name: a parent and child authorizing the same connection dedupe to one public message. That's semantically right — credentials are scoped (connection, principal) at the provider, one sign-in satisfies both — with a minor edge that a completion can clear a same-name pending message from another session early.

## Testing

- Unit: adapter forwarding + failure breadcrumbs (`subagent-adapter.test.ts`), proxy step verbatim re-emit / state persistence / no-epilogue (`subagent-auth-proxy.test.ts`), inbox branch asserting no `turn-delivery-request` is issued (`turn-workflow.test.ts`)
- Integration: adapter-state lifecycle across a serialize/rehydrate boundary and framework-registry rehydration for nested hops (`subagent-auth-proxy.integration.test.ts`)
- `pnpm typecheck`, `pnpm lint`, `pnpm guard:invariants`, `pnpm docs:check`, full unit + integration tiers all pass

## Docs & changeset

- `docs/subagents.mdx`: paragraph on how subagent HITL and authorization surface on the parent channel
- `docs/connections/overview.mdx`: note that subagent auth surfaces on the root session's channel
- Patch changeset

### DIFF
--- a/.changeset/subagent-auth-bubbling.md
+++ b/.changeset/subagent-auth-bubbling.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Bubble subagent `authorization.required` and `authorization.completed` events up to the parent session's channel. Previously a subagent whose connection needed interactive OAuth parked silently — the sign-in link was written only to the child's unconsumed stream and the run hung. The events now proxy up through each delegation hop so the user can complete the sign-in; the OAuth callback still resumes the subagent directly.
--- a/docs/connections/overview.mdx
+++ b/docs/connections/overview.mdx
@@ -136,7 +136,7 @@ export default defineMcpClientConnection({
 });
 ```
 
-`"linear/myagent"` is the UID you chose when registering the Connect client. `connect("linear/myagent")` is shorthand for a user-scoped interactive OAuth connection: eve resolves a token for the active user before each tool call, emits `authorization.required` when that user has not authorized yet, and resumes the parked turn after the callback completes.
+`"linear/myagent"` is the UID you chose when registering the Connect client. `connect("linear/myagent")` is shorthand for a user-scoped interactive OAuth connection: eve resolves a token for the active user before each tool call, emits `authorization.required` when that user has not authorized yet, and resumes the parked turn after the callback completes. When the connection is used inside a [subagent](/docs/subagents), the authorization events surface on the root session's channel and the callback resumes the subagent directly.
 
 That means the channel that creates or continues the session must authenticate a real user. For a web app, configure `agent/channels/eve.ts` so your app session maps to `principalType: "user"`; for platform channels, use the built-in channel auth that maps the sender to a user principal. If no authenticated user is attached to the session, the first user-scoped connection call fails with `reason: "principal_required"`.
 
--- a/docs/subagents.mdx
+++ b/docs/subagents.mdx
@@ -109,7 +109,9 @@ Because the name lives in the same runtime tool namespace as authored tools, a s
 
 Do not rely on subagent delegation by itself as an approval boundary. Put sensitive tools behind `approval`, connection approval, route/session authorization, or other controls wherever those tools can be called.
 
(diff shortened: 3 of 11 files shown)


blurrah:
Fixed by #603

PR #573, opened 2026-07-07
feat(eve): cancellation - settle cancelled turns through the turn workflow

AndrewBarba:
### Description

Layer 1 of turn cancellation (#483), building on the layer-0 harness abort propagation (#494): turns are now cancellable at the workflow layer.

Each turn-owned turn claims a durable **session-scoped** cancel hook (`{sessionId}:cancel`) and a durable `AbortController` whose signal is serialized into every `turnStep`. The token is stable for the session's lifetime, so a cancel trigger derives it from the session id alone — no per-turn token discovery. The payload's optional `turnId` guard scopes the cancel to the turn the caller observed (mismatches are consumed as benign no-ops), so a cancel racing a turn boundary can never kill a turn the caller never saw. Resuming the hook mid-turn aborts in-flight model, tool, and subagent-wait work in real time, and the turn settles as a new `turn.cancelled` stream event followed by `session.waiting` (stream version 19) — never as a failure. A cancelled turn:

- keeps whatever it streamed before the abort, while durable history keeps only what had already settled,
- discards pending runtime-action state and descendant HITL routing so nothing re-dispatches or routes to orphaned children,
- leaves the session parked and ready for its next message,
- and treats a late or duplicate cancel as a benign no-op.

Key design decisions:

- **Driver-negotiated, parkable-only.** The pinned driver body advertises `cancelledTurnSettle`; the turn workflow registers the cancel hook only when it is set and the session can park (conversation mode or an anchored continuation token — delegated subagents always have one). Drivers pinned before this PR get no cancel hook, and unparkable root task runs can never settle a cancel as `session.failed`.
- **Session-scoped token, payload-guarded turn.** Each turn run claims `{sessionId}:cancel` at turn start and disposes it before publishing its turn result, so the next turn's claim never races the prior run's teardown; a stale claim from a crashed turn degrades that turn to uncancellable, never to a failure. This shape is enabled by the upstream hook-reuse and dispose-race fixes (workflow#2779, workflow#2808).
- **Replay-deterministic abort.** The abort fires in the cancel-hook read continuation, keyed to the `hook_received` journal event — never to live signal state or a promise-race winner.
- **Cancellation is a result, not an error.** The aborted `turnStep` returns a pure `cancelled` marker, so the runtime never classifies the abort as a step failure or retries it.
- **Driver-owned epilogue.** `turn.cancelled` → `session.waiting` and session persistence run in a single driver-owned step (`settleCancelledTurnStep`), single-flighted against duplicate step dispatch; at-least-once under crash retry, like every stream emission. The epilogue runs the authored-hook pipeline. When a descendant HITL wait already streamed the turn's waiting boundary, the settle emits nothing extra.
- **Teardown that lets runs finish.** Turn runs dispose their hooks without closing iterators holding pending durable reads, so every settled turn run reaches `run_completed` and the world sweeps its hooks; the driver defers each turn's control-hook disposal by one turn so a late duplicate of the turn's final control send still lands on a live hook.

`turn.cancelled` is a public channel event (typed handler maps for every built-in channel), `TurnCancelledStreamEvent` is exported from `eve/client`, and the default message reducer finalizes partially streamed messages on cancellation.

No public trigger ships here: the HTTP cancellation API is layer 2 (tests resume the hook directly). Design notes live in `research/turn-cancellation-turn-workflow.md`.

- `pnpm fmt`, `pnpm lint`, `pnpm typecheck`, `pnpm guard:invariants`, `pnpm docs:check`
- `pnpm test:unit` (4701 passed) and `pnpm test:integration` (434 passed)
- Integration suite `execution/turn-cancellation.integration.test.ts`: mid-tool cancel with a matching `turnId` guard + normal follow-up turn, a stale-guard cancel consumed as a no-op with the turn still cancellable afterwards, cancel during an in-flight subagent wait with no re-dispatch, cancel during a descendant HITL wait (single waiting boundary, proxy map cleared, child never re-woken), late/duplicate cancel no-ops, a no-retry canary over the turn run's event log, and a regression test that settled turn runs complete so the world sweeps their hooks (including the session cancel hook)
- Unit tests: `createTurnCancellationControl` (token derivation, claim conflict → uncancellable, guard match/skip, idempotent disposal), the `turnWorkflow` cancelled arm and capability/parkability gates (session-scoped token, dispose-before-result ordering), dispatch-input capability stamping, the driver's cancelled-park settle path, deferred control-hook disposal, `activeTurnId`, `emitCancelledTurn`, `createTurnCancelledEvent`, the `turn.cancelled` channel-event wiring, reducer finalization of partial messages, and the stream-version pin

### PR Checklist


AndrewBarba:
**Status: blocked on upstream — converting to draft.**

The `e2e-local (agent-workflow-stress)` failure is real and caused by this PR: threading the durable `AbortSignal` into `turnStep` costs ~0.5 s of fixed latency per step on world-local, and the abort-stream tail reader it opens polls with a full `readdir` of the world-global chunks directory — so a long session compounds from ~0.9 s/turn into a 240 s replay timeout around turn 40. The Vercel world (service-backed streams) is unaffected, which is why `e2e-vercel (agent-workflow-stress)` passed.

Bisection (in-process bench, 40 sequential turns):

| variant | per-turn floor | turn 40 |
|---|---|---|
| cancel hook + controller + signal-in-step (this PR) | ~910 ms | ~1075 ms |
| cancel hook only | ~375 ms | ~476 ms |
| controller created, signal **not** serialized into the step | ~390 ms | ~485 ms |
| neither (≈ main) | ~385 ms | ~501 ms |

The cancel hook and the controller are free; **serializing the signal into the step is 100% of the regression.**

Filed upstream with standalone repros:
- vercel/workflow#2795 — ~0.5 s fixed latency per signal-bearing step invocation (world-local)
- vercel/workflow#2797 — stream tail readers poll with O(world-chunks) `readdir` per 100 ms tick (20k chunk files → 24–38 s per step)

Everything else in this PR is validated and unaffected. Plan: hold for the upstream fixes (the team turned vercel/workflow#2777/#2778 around within days via vercel/workflow#2779); if they stall, fall back to shipping cancellation without the per-step signal (immediate during runtime-action waits, next-step-boundary otherwise) and add real-time mid-step abort back once the primitive is cheap.

AndrewBarba:
**Unblocked — upstream fixes landed and validated.**

[workflow#2795](https://github.com/vercel/workflow/issues/2795) and [workflow#2797](https://github.com/vercel/workflow/issues/2797) were both fixed by [workflow#2807](https://github.com/vercel/workflow/pull/2807) (`executeStep` now cancels abort-stream readers after user code; chunk storage sharded per-stream), shipped in `@workflow/core@5.0.0-beta.28` / `@workflow/world-local@5.0.0-beta.24`, which #612 vendored into main.

Rebased onto main (resolving the #603 proxy-step extraction and taking **stream version 19** — main claimed 18 for attachment parts in #574) and re-benched on the fixed runtime:

| bench (world-local, 40 sequential turns) | before | after |
|---|---|---|
| per-turn floor with signal-in-step | ~910ms, compounding to replay timeout | **~380–460ms, flat** |
| no-signal baseline (bisection) | ~385ms | ~385ms |

Only residual drift is the same ~+4.6ms/turn session-history growth the baseline shows. Local gates green: unit, integration (incl. the 3-test cancellation suite), typecheck, lint, guard, docs:check.

Marking ready for review. The `agent-workflow-stress` e2e on this PR is the official re-validation of the failure mode from run 28877952878.

AndrewBarba:
**Traced the residual local-world drift — found and fixed a real leak** (6ba9a9e0).

## Root cause
`runTurnOwnedWorkflow`'s teardown called `iterator.return()` on the cancel hook's async iterator. That iterator always has an in-flight durable read (the abort continuation), and an async generator only honors `return()` after its pending `await` settles — which never happens for a turn that wasn't cancelled. Every turn run therefore suspended forever in its `finally`:

- no `run_completed` → the world's terminal-run sweep (`deleteAllHooksForRun`) never fired
- **+1 permanently-running run and +3 leaked hooks per turn** (inbox, cancel, durable-abort)
- world-local's `getByToken` is an O(live hooks) directory scan per resume → growing per-turn cost

Bench (40 turns, world-local): live hooks **124 and climbing** before → **flat at 7** after; every settled turn run now journals `hook_disposed ×3` + `run_completed`. Fix is dispose-only teardown (the runtime's sanctioned dispose-with-outstanding-read pattern, same as the driver's delivery hook). New regression test fails pre-fix, passes post-fix.

## What remains (shared with main, upstream)
A/B benches: main slope 2.8ms/turn, this branch ~4.5–5ms/turn — both linear, neither compounding. The shared term: world-local stores entities in flat global dirs and `paginatedFileSystemQuery` re-`readdir`s the entire `events/` dir per query (O(world events), prefix-filtered in memory) on every wake. This branch journals ~21% more events/turn and ~1.5× the wakes (cancel-hook create + dispose cycles), hence the steeper-but-linear slope. Same class as workflow#2797 (fixed for stream chunks by sharding); candidate upstream report for the entity stores.

Two follow-ups worth tracking outside this PR: (1) upstream issue for the O(world) events-dir listing; (2) `workflow-entry.ts` session-end teardown closes the auth iterator the same way — latent on main, only bites at session end.

AndrewBarba:
All seven findings addressed in 71ebcad0 (rebased onto main first). Point by point:

**[P1] Negotiate cancellation separately from turn inbox — fixed.** The pinned driver body now passes `driverCancelledTurnSettle` through `dispatchAndAwaitTurn` → `dispatchTurnStep` → `createTurnWorkflowInput`, which stamps `driverCapabilities.cancelledTurnSettle`. The turn workflow registers the cancel hook only when it is set. Since the flag originates in driver-body code, drivers pinned before this PR never advertise it and their turns get no cancel hook — a layer-2 trigger resolves to "not cancellable" instead of an ignored `park + cancelled`. Unit-tested both ways.

**[P1] Task cancellation can end as `session.failed` — fixed.** Cancel-hook registration is additionally gated on parkability (`mode === "conversation" || continuationToken !== ""`). Unparkable sessions (root task runs, pre-anchor sessions) simply aren't cancellable in layer 1, so the cancel-park path can never reach the no-token throw. Delegated subagents keep their hooks — they are always born with a minted continuation token. Task-mode terminal semantics are deferred to layer 2's trigger design.

**[P1] Subagent HITL boundary — fixed, both halves.** `settleCancelledTurnStep` now (a) skips the epilogue emission when the persisted state is already between-turns with live proxy entries (the proxy epilogue streamed the boundary; re-emitting would fabricate a turn id and duplicate `session.waiting`), and (b) clears the proxy input-request map, so later HITL answers stay with the parent. New integration test covers the full flow: child `ask_question`, cancel mid-HITL, single waiting boundary, follow-up answer reaches the parent, child never starts a second turn.

**[P1] The epilogue is not exactly-once — contract weakened, as suggested.** You're right that the single-flight only covers duplicate dispatch on one instance. That's intentional parity with the rest of the system: every stream emission in eve (including a retried `turnStep` re-streaming its whole turn) is at-least-once under crash retry. Durable dedup here would exceed the guarantees of everything around it. All "exactly-once" claims are rewritten to "single-flighted against duplicate dispatch; at-least-once under crash retry, like every stream emission." The duplicate-dispatch trigger itself is fixed upstream by vercel/workflow#2848, after which the single-flight can be deleted.

**[P2] Cancellation events bypass authored hooks — fixed.** The settle emit now runs inside `withContextScope` and dispatches `dispatchStreamEventHooks` for both epilogue events, same pipeline as `turnStep`. (The subagent proxy step has the same gap on main for proxied events — out of scope here, worth a follow-up issue.)

**[P2] Reducer ignores `turn.cancelled` — fixed.** The default reducer treats it as a turn boundary: message status → `complete` and streaming text parts finalized to `done`. Reducer test added.

**[P2] Public surfaces — fixed.** `turn.cancelled` added to the `ChannelEvents` handler map + runtime wiring list, to every built-in channel's typed event map (Slack, Discord, Linear, GitHub, Telegram, Teams, Twilio; chat-sdk/eve derive from `ChannelEvents`), and `TurnCancelledStreamEvent` is exported from `eve/client`. Wiring test added.

Also rebased onto latest main (was 2 commits behind) so checks run against the current base.

AndrewBarba:
**Status update: upstream fixes landed; cancel token redesigned to session-scoped.**

Two developments since the last status:

**1. All four upstream findings are fixed and vendored.** [workflow#2848](https://github.com/vercel/workflow/pull/2848) fixed #2780 (inline step message ownership) and [workflow#2808](https://github.com/vercel/workflow/pull/2808) closed #2781's acceptance window, joining the earlier #2779/#2807 fixes — all present in the runtime this branch vendors (core beta.31, world-local beta.27). One correction to my earlier point-by-point: I said the settle single-flight "can be deleted" once #2848 lands. It landed, but the single-flight stays as cheap insurance for the queue's residual at-least-once envelope (its comment now says exactly that); deferred control-hook disposal likewise stays until hosted-world parity for #2808 is confirmed. Details in the research doc's updated "Runtime findings" section (c08a2d6b).

**2. The cancel token is now session-scoped: `{sessionId}:cancel` with an optional `turnId` payload guard** (6d416d1c). The per-turn shape (`{completionToken}:cancel`) had no reasonable discovery path — a layer-2 route holds a session id, not a turn-control sequence, and nothing persists that mapping outside the driver's workflow body. Rather than build an index or push turn tokens onto clients, the token is stable per session (derivable from the session id alone) and turn-scoping moved into the payload: a `turnId` guard mismatch is consumed as a benign no-op, so a cancel racing a turn boundary can never kill a turn the caller never saw. Each turn run claims the token at start (a stale claim from a crashed turn degrades to uncancellable, never a failure) and disposes it before publishing its turn result, so consecutive claims never race. This shape is exactly what #2779 + #2808 made safe. The #2848/#2808 fixes also relax layer 2's trigger contract: no cancel single-flighting needed — "already resumed/disposed" is success.

New coverage: unit suite for `createTurnCancellationControl` (claim conflict, guard match/skip, idempotent disposal), dispose-before-result ordering, and a stale-guard integration test proving the turn stays cancellable after a mismatched cancel. All CI green (the two e2e failures on d12d28cd were environment flakes — `ECONNRESET` on a stream read and a sandbox timing gate — both passed on rerun with no code change).

packages/eve/src/execution/turn-workflow.ts
```diff
@@ -151,20 +197,26 @@ async function runTurnOwnedWorkflow(input: TurnWorkflowInput): Promise<void> {
...
   readonly bufferedDeliveries: DeliverHookPayload[];
+  readonly cancellation: TurnCancellationControl | undefined;
   readonly cursor: TurnExecutionCursor;
   readonly inboxToken: string;
   readonly initialResults: readonly RuntimeActionResult[];
   readonly iterator: AsyncIterator<TurnInboxPayload>;
   readonly nextDeliveryRequestId: () => string;
   readonly pendingActionKeys: readonly string[];
-}): Promise<readonly RuntimeActionResult[]> {
+}): Promise<readonly RuntimeActionResult[] | "cancelled"> {
```
ctgowrie, line 234: What's the reason to not expand the RuntimeActionResult type

packages/eve/src/execution/turn-workflow.ts
```diff
@@ -72,10 +81,39 @@ async function runTurnOwnedWorkflow(input: TurnWorkflowInput): Promise<void> {
...
+    }
+
     while (true) {
-      const result = await turnStep(cursor.createStepInput(nextStepInput));
+      const result = await turnStep(cursor.createStepInput(nextStepInput, cancellation?.signal));
+
+      if (result.action === "cancelled") {
+        // The epilogue runs in the driver (`settleCancelledTurnStep`), not
+        // as a step in this run, where queued cancel wakes could
+        // re-dispatch it. The `canPark` gate below is bypassed on purpose.
```
ctgowrie: This comment doesnt help me fully understand why the canPark gate is bypassed

packages/eve/src/execution/turn-workflow.ts
```diff
@@ -72,10 +81,39 @@ async function runTurnOwnedWorkflow(input: TurnWorkflowInput): Promise<void> {
...
     }
 
+    // Claimed after the inbox claim so a losing duplicate run never
+    // contends for the session cancel token. Gated on the pinned driver
+    // settling `park + cancelled` and on the session being parkable —
+    // cancellation settles as a park, and an unparkable session would
+    // fail instead.
+    if (
+      input.driverCapabilities?.cancelledTurnSettle === true &&
+      (input.mode === "conversation" || input.stepInput.sessionState.continuationToken !== "")
```
ctgowrie: This statement is confusing. Perhaps we should standardize the definition as parkable somehow?

packages/eve/src/execution/turn-workflow.ts
AndrewBarba: Kept the sentinel since `RuntimeActionResult` is the schema-validated wire union and cancellation never leaves the workflow — added a comment explaining this in 2ac75520.

packages/eve/src/execution/turn-workflow.ts
AndrewBarba: Rewrote the comment in 2ac75520 — the gate rejects model-authored task-mode waits, but a cancelled turn parks by design and its parkability was already checked at hook claim.

packages/eve/src/execution/turn-workflow.ts
AndrewBarba: Extracted it into a named `canSettleCancelledTurnAsPark` helper with a comment spelling out both arms in 2ac75520.

packages/eve/src/execution/next-driver-action.ts
```diff
@@ -30,6 +30,13 @@ export type NextDriverAction =
       readonly sessionState: DurableSessionState;
       readonly serializedContext: Record<string, unknown>;
       readonly authorizationNames?: readonly string[];
+      /**
+       * Set when the parked turn was cancelled: the driver runs
+       * `settleCancelledTurnStep` before the normal park playbook. An
+       * optional field rather than a new arm so pinned drivers keep
+       * working.
+       */
+      readonly cancelled?: true;
```
ruiconti, line 39: Why is `cancelled` a field in `kind:"park"`? I'd expect this to be a final payload that's either in `done` but since it has significantly semantic differences than it, a separate `kind:"cancelled"`

AndrewBarba: I don't agree with this. A canceled turn can still perform a lot of work and that work is saved on when you resume. The Boolean indicates it was canceled but marking the entire is canceled feels pretty wrong.

ruiconti: was not arguing to drop `sessionState` and `serializedContext`, but rather that `park !== cancelled`, but I did trace and the resume path is similar for both `park` and `cancelled` so the hierarchy makes sense to me now.

packages/eve/src/execution/settle-cancelled-turn-step.ts
```diff
@@ -0,0 +1,107 @@
...
+    durable: durableSession,
+    turnAgent: bundle.turnAgent,
+  });
+
+  let emissionState = getHarnessEmissionState(durableSession.state);
+  // A descendant HITL wait already streamed this turn's waiting boundary
+  // (the proxy epilogue clears the turn id); re-emitting would fabricate
+  // a turn id and duplicate the boundary.
+  const alreadyEpilogued =
+    emissionState.turnId === "" && hasProxyInputRequests(durableSession.state);
```
ruiconti, line 66: Could use `isHarnessBetweenTurns` (instead of `emissionState.turnId === ""`)

packages/eve/src/execution/settle-cancelled-turn-step.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  if (!alreadyEpilogued) {
```
ruiconti, line 68: If the current turn is waiting on HITL and a cancellation happens, does this check prevent from the stream emitting the `cancelled` event? because it doesn't run `emitCancelledTurn`?

packages/eve/src/execution/workflow-steps.ts
```diff
@@ -297,66 +305,81 @@ export async function turnStep(rawInput: TurnStepInput): Promise<DurableStepResu
...
-            stepIndex: emissionState.stepIndex,
-            turnId: emissionState.turnId,
-          }),
-        );
+  let stepResult: StepResult;
+  try {
+    // A signal already aborted at entry (cancellation during an in-line
+    // runtime-action wait) must settle before the park-resume stages run,
+    // or the pending batch would re-park and later re-dispatch.
+    throwIfTurnAborted(input.abortSignal);
```
ruiconti, line 313: Any reason we just check `if (isAborted(input.abortSignal)) return { action: "cancelled", ... }` instead of the implicit goto via exception raising?

packages/eve/src/execution/turn-cancellation-control.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+  // Hook promises and iterators share one durable cursor. Create the
+  // iterator before claiming so conflict replay is consumed by
+  // getConflict(), not a later iterator read.
+  const iterator = hook[Symbol.asyncIterator]();
+
+  try {
+    await claimHookOwnership(hook);
+  } catch (error) {
+    if (isHookConflictError(error)) return undefined;
```
ruiconti, line 62: Wonder if there's a way to advertise this upstream.. I can see that it could be relevant for clients to know when a turn is not cancelable e.g `turn.cancelable` or via an event?

ruiconti: Can be deferred to a follow-up though

ruiconti (approved):

PR #481, opened 2026-07-02
feat(eve): project structured file parts on message.received

NicoMoli:
## Summary

The `message.received` stream event now carries structured `parts` (text plus file/image attachment metadata) alongside the flattened `message` summary. The default message reducer projects attachments as `file` message parts, so chat UIs can render user-attached files and images instead of parsing the `[file: …]` placeholder text.

- Bumps `EVE_MESSAGE_STREAM_VERSION` `16 → 17`.
- Attachment **bytes** and internal **sandbox paths** are never projected.
- A `url` is included only when the attachment is a client-resolvable `http(s)`/`data:` URL.

This PR should address this issue:
https://github.com/vercel/eve/issues/286

## Testing

- Unit coverage for `createMessageReceivedEvent` parts projection (inline bytes, URL parts, `eve-url:` refs, `data:` URLs, `eve-sandbox:` refs, opaque base64, image normalization).
- Reducer tests projecting structured file parts and the no-`parts` fallback.
- E2E `image-attachment.eval.ts` now asserts `message.received` carries a structured `image/png` file part.

PR #574, opened 2026-07-07
feat(eve): message received - project attachment parts

### LINKED ISSUE #286: eve drops file/image parts from the transcript projection

Hit a wall rendering file uploads in our chat UI and I think it's a gap in eve, not us.

When a user sends a turn with attachments, eve forwards the file parts to the model fine. But the only thing clients get back is message.received, and that event flattens everything to a string — createMessageReceivedEvent just stores summarizeUserContent(message), so a file part becomes [file: name (type)] placeholder text. The structured part is gone by the time it hits the wire.

So defaultMessageReducer only ever produces a text part for the user message (there's no file in EveMessagePart), and a custom reducer can't help because the data's already been dropped upstream. To show the file the user attached, we end up keeping our own upload catalog on the side, correlating it back to the eve turnId, and regex-stripping the [file: …] text out of the bubble. Works, but it's a hack around missing data.

### Proposed solution

Would it be reasonable to carry the non-text parts through on message.received? Something additive like data.attachments: { filename?, mediaType, url }[] plus a text-only data.text, keeping data.message as-is so nothing breaks, and a file variant on EveMessagePart that the default reducer projects. eve wouldn't store bytes — just pass through whatever the client authored (data URL or remote URL).

Mostly want to know if the lossy projection was deliberate beyond "keep the event a simple string," and whether you'd want the authored data passed through verbatim or just metadata.

### Alternatives considered

_No response_

AndrewBarba:
## Summary

Implements an internal fix for #286, using #481 as reference without merging the external branch directly. `message.received` now keeps the existing flattened `data.message` summary and adds structured `data.parts` for text plus attachment metadata, so clients can render user attachments without parsing placeholder text.

The default message reducer projects confirmed attachments as `file` message parts, and the web chat / framework sample UIs render those file parts. Raw bytes and internal sandbox paths stay off the stream; `url` is included only for client-resolvable `http(s)` and `data:` URLs.

Co-authored implementation credit: Nicolas Molina <n_m09@hotmail.com>.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/protocol/message.test.ts src/client/message-reducer.test.ts test/dev-send-message.test.ts`
- `pnpm --filter eve run check:web-template`
- `pnpm fmt`
- `pnpm lint` (completed with existing warnings in `apps/docs/scripts/render-eve-5.ts`)
- `pnpm typecheck`
- `pnpm docs:check`

## Notes

The exact focused-test command from the plan used repo-root paths under `pnpm --filter eve exec`, which runs from `packages/eve`; I reran the same files with package-relative paths. E2E was not run locally, per plan.

Fixes #286.
References #481.


packages/eve/src/protocol/message.ts
```diff
@@ -156,16 +162,41 @@ export interface TurnStartedStreamEvent {
...
+      readonly text: string;
+      readonly type: "text";
+    }
+  | {
+      readonly filename?: string;
+      readonly mediaType: string;
+      readonly size?: number;
+      readonly type: "file";
+      readonly url?: string;
+    };
```
ruiconti: Does aisdk expose these types? Would be good to consolidate

AndrewBarba: narrowed it a bit

ruiconti (approved):

NicoMoli:
Thanks for the credits! happy to see contribution on this project :)

Issue #577, opened 2026-07-07
feat(agent): dynamic model resolution via defineDynamic


AndrewBarba:
## Problem

Most agent capabilities in eve support dynamic, per-session/per-turn resolution through `defineDynamic` — tools, skills, and instructions can all be resolved at runtime with session, channel, and conversation context. The agent's **model** is the glaring omission: `defineAgent({ model })` is evaluated once at compile time and frozen into the manifest.

There is no way to:

- route cheap/latency-sensitive channels (e.g. Slack) to a smaller model and keep a frontier model elsewhere;
- pick a model per tenant/plan from `ctx.session.auth`;
- escalate to a larger-context model when the conversation grows;
- A/B models per session.

## Proposal

Let the existing `model` field accept the established `defineDynamic` sentinel:

```ts
export default defineAgent({
  model: defineDynamic({
    events: {
      "session.started": () => "anthropic/claude-sonnet-5",
      "turn.started": (event, ctx) =>
        ctx.channel.kind === "slack" ? "anthropic/claude-haiku-4.5" : null,
    },
  }),
});
```

- Events: `session.started` + `turn.started` (model stable within a turn, matching dynamic instructions/skills).
- Handlers return a gateway model id string, a live AI SDK `LanguageModel`, a `defineModel({ model, contextWindowTokens?, providerOptions? })` entry, or `null` (keep current scope choice).
- The harness already re-resolves the model once per step (`resolveModel(session.agent.modelReference)`), so the runtime seam exists.

A research plan will follow in `research/`.

PR #578, opened 2026-07-07
research: dynamic model resolution plan

AndrewBarba:
## Summary

Adds `research/dynamic-model.md`, an implementation plan for issue #577: letting the agent `model` field accept the established `defineDynamic` sentinel so the model can be resolved per session/turn with session, channel, and conversation context.

Highlights:

- `model: LanguageModel | defineDynamic({ events })` — reuses the existing dynamic-capabilities authoring pattern; static form unchanged.
- Handlers resolve on `session.started` / `turn.started` (model stable within a turn, matching dynamic instructions/skills) and return a gateway id string, a live AI SDK `LanguageModel`, a `defineModel({ model, contextWindowTokens?, providerOptions? })` entry, or `null`.
- Serializable results pin durably for their scope (replay-safe, A/B stays stable mid-scope); live-instance results re-invoke the resolver after replay with a documented determinism constraint, mirroring the dynamic-tool inline-`execute` rule.
- Rides the existing seams: per-step `resolveModel(session.agent.modelReference)` in the tool-loop, the source-backed agent-config re-import, and the dispatch → durable-context pipeline dynamic instructions use.
- Compile-time surface: `ModelRouting` gains `{ kind: "dynamic" }`; credential checks treat it as gateway-recommended; TUI `/model` reports it cannot rewrite a dynamic config.

Closes nothing yet — plan only. Implementation will follow as a separate PR per the delivery section.

Refs #577

## Verification

- `pnpm guard:invariants` passes (research frontmatter: `issue`, `status`, `last_updated`).
- Docs-only change: no changeset needed.

### DIFF
--- a/research/dynamic-model.md
+++ b/research/dynamic-model.md
@@ -1,0 +1,162 @@
+---
+issue: https://github.com/vercel/eve/issues/577
+status: proposed
+last_updated: "2026-07-07"
+---
+
+# Dynamic model resolution
+
+## Summary
+
+Tools, skills, and instructions can all be resolved dynamically per session or turn through
+`defineDynamic`, with session, channel, and conversation context. The agent's model cannot: the
+`model` field on `defineAgent` is evaluated once at compile time and frozen into the manifest. That
+blocks routing cheap channels to a smaller model, picking a model per tenant from session auth,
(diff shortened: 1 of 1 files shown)


ctgowrie (commented):
LGTM

ruiconti (approved):

PR #580, opened 2026-07-07
Add phase to eve hero

matiasngf:
## Summary

Integrates the new [`phase`](https://www.npmjs.com/package/phase) animation library (vercel-labs) into the Eve hero shader to manage the animation loop lifecycle. **vgpu keeps owning all WebGPU rendering; phase owns lifecycle/visibility signals.** The hero's rAF loop now fully stops when the hero is out of the viewport or the tab is hidden, and resumes cleanly when it becomes visible again.

## What changed

- **`index.tsx`** — wires `createLifecycle` (framework-free) inside the existing client effect: composes IntersectionObserver on the stable hero container + document visibility + reduced-motion. `onPhaseChange`: `active → drawLoop.start()`, otherwise `drawLoop.stop()`. Migrates reduced-motion handling to `usePrefersReducedMotion` from `phase/react`.
- **`runtime/frame-loop.ts`** — `createDrawLoop` now returns `{ start, stop, dispose }` with idempotent rAF gating. Frame delta is clamped to 50 ms and the clock resets on `start()`, so resuming after a long pause cannot decay-wipe the persisted GPU paint texture. Adds dev-only QA probes: `window.__eveHeroFrameLoop` and `data-hero-*` attributes on the canvas (throttled, stripped from production by the `NODE_ENV` gate).
- **`hooks.ts`** — removes the bespoke reduced-motion hook (now provided by `phase/react`).
- **`scripts/guard-invariants.mjs`** — new rule 34: no `phase` imports allowed inside the GPU/shader boundary (`render/`, `shaders/`, `scripts/eve-render/`). Validated with a synthetic violation.
- **Dependency** — `phase@0.0.1` pinned via pnpm catalog, with `minimumReleaseAgeExclude` entry and `allowBuilds.phase: false` (its `lefthook` postinstall is disallowed).

## StrictMode / `GPUDevice.lost` fix

React StrictMode double-mounts the component in dev. Our own teardown calls `device.destroy()`, which resolves the `GPUDevice.lost` promise with `reason: "destroyed"`. The lost handler treated that as a fatal device loss, re-entered the dispose path, and stopped the lifecycle after ~2 frames. The handler now ignores `lost` events caused by intentional destroys/cancellation, and the lifecycle/runtime cleanup is reachable from the effect cleanup so no orphaned listeners survive fast unmounts.

## Validation

- `pnpm typecheck`, `pnpm lint`, `pnpm guard:invariants` — all green
- Offline Docker render baselines **byte-identical** (rendering path untouched):
  - dark imprint-1: `0e96449327f59e9f1124cd3ee2d4de2448c8c531da088e305e0e96b2311fde13`
  - dark glass imprint-0: `6eb5f6e59943cd4a19e37ed22e759610909d5d9d825372531710e20cc8a02bb9`
- Browser QA with WebGPU (Chromium + swiftshader): activation verified (`heroPhase="active"`, frames increasing, canvas revealed). Scroll pause/resume automated validation was limited by headless IntersectionObserver callbacks after WebGPU warmup; verified manually.

## phase utils adopted

Final branch state uses `phase` for lifecycle/scheduling/observers/signals only; WebGPU rendering stays inside the existing Eve renderer boundary.

- **`phase/react` `useLoop`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx`: attaches to the hero container, owns the page-wide rAF via phase's shared scheduler, calls `drawLoopRef.current?.step(frame.time)` in `runtime/frame-loop.ts`, and auto-pauses on visibility plus `reducedMotion: "pause"`.
- **`phase/react` `useSize`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx`: uses phase's shared `ResizeObserver` for the hero container size, updates `canvasLayoutRef`, feeds canvas backing-store sizing, and invalidates the pointer rect cache when layout changes.
- **`phase/react` `useMediaQuery`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx` and `hooks.ts`: reads `(pointer: coarse)` for pointer interaction mode via refs into `pointer-input`, and reads `(prefers-color-scheme: dark)` for `useResolvedTheme`; `hooks.ts` keeps its `MutationObserver` so explicit `class` / `data-theme` changes still win over the media query.
- **`phase/react` `usePrefersReducedMotion` + core `prefersReducedMotion`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx`: combines the reactive reduced-motion gate with a synchronous `prefersReducedMotionSync()` re-check immediately before WebGPU initialization.
- **core `createDevicePixelRatio`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx`: event-driven DPR watcher replaces per-frame `window.devicePixelRatio` reads, updates `devicePixelRatioRef`, and invalidates the pointer rect cache for DPR / monitor changes.
- **`phase/react` `useStableCallback`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx`: keeps stable callback identities for `onCanvasRevealed`, `onFallback`, and `onFatalError` while still reading the latest refs/state.
- **`phase/ease` `lerp`, `clamp`, `clamp01`, `easeInOutCubic`** — `apps/docs/app/[lang]/(home)/components/eve-logo-shader/runtime/frame-loop.ts`: replaces local math helpers for frame interpolation and transition clamping; `clamp01(...)` remains wrapped around `lerp` progress values to preserve the previous clamped interpolation semantics.

Architecture boundary: `phase` is limited to lifecycle/scheduling/observers/math utilities in orchestration files; `eve-logo-shader/render/**`, `eve-logo-shader/shaders/**`, and `apps/docs/scripts/eve-render/**` remain phase-free, enforced by `pnpm guard:invariants` rule 34.


apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx
```diff
@@ -109,7 +114,7 @@ export function EveLogoShader({ audience = "humans" }: { audience?: InstallAudie
     resetCanvasVisibility(canvas);
     setRevealed(false);
 
-    if (prefersReducedMotion !== false) {
+    if (prefersReducedMotion) {
```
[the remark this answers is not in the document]
Comment (matiasngf) [reply apps/docs/app/[lang]/(home)/components/eve-logo-shader/index.tsx]: Confirmed — good catch. `phase/react`'s `usePrefersReducedMotion` initializes with `useState(false)` and only syncs the real `matchMedia` value inside a `useEffect`, so with the hook alone the WebGPU init effect ran one render before the preference resolved (the old local hook's `null` state guarded exactly that window).

Fixed in 6107938: the effect now re-checks the preference synchronously via `prefersReducedMotion()` from `phase` core before any GPU setup (`prefersReducedMotion || prefersReducedMotionSync()`), bailing to the static fallback like the pre-migration behavior. The reactive hook is kept for runtime toggles.

AndrewBarba (approved):

apps/docs/app/[lang]/(home)/components/eve-logo-shader/hooks.ts
```diff
@@ -1,53 +1,40 @@
...
   if (root.classList.contains("light") || root.dataset.theme === "light") return "light";
-  return window.matchMedia("(prefers-color-scheme: dark)").matches ? "dark" : "light";
+  return prefersDarkTheme ? "dark" : "light";
 }
 
-export function useResolvedTheme() {
+export function useResolvedTheme(prefersDarkTheme: boolean) {
   const [theme, setTheme] = useState<"light" | "dark">("light");
+  const prefersDarkThemeRef = useRef(prefersDarkTheme);
+  prefersDarkThemeRef.current = prefersDarkTheme;
```
Comment (rvnewolf) [inline apps/docs/app/[lang]/(home)/components/eve-logo-shader/hooks.ts]: suggestion: can do useSyncedRef from phase to one-line this!

Comment (matiasngf) [reply apps/docs/app/[lang]/(home)/components/eve-logo-shader/hooks.ts]: Applied in a830221615812a72d99a82c1fdaf2536dc83ff2f — replaced the manual latest-value ref in `useResolvedTheme` with `useSyncedRef`, and applied the same pattern to the Eve hero coarse pointer latest-value ref.

rvnewolf (approved):
LGTM! Nice work!

AndrewBarba (approved):

PR #585, opened 2026-07-08
Report Eve issue events through Workflow

JohnPhamous:
- Why: issue grouping should be owned by the Vercel platform so grouping can evolve without forcing Eve framework releases.
- What: removes `$eve.issue` and framework-side fingerprint state.
- What: sends issue-source stream events to Workflow’s observability reporter before serialization.
- Keeps the path best-effort and backwards-compatible: older Workflow cores simply no-op.


packages/eve/src/execution/eve-workflow-attributes.ts
```diff
@@ -239,8 +247,38 @@ export function buildSubagentRootAttributes(input: {
...
+    "$eve.failed_turn_count": summary.failedTurnCount,
+    "$eve.issue_count": summary.issueCount,
+    "$eve.last_issue_at": summary.lastIssueAt,
+    "$eve.last_issue_code": summary.lastIssueCode,
+    "$eve.last_issue_source": summary.lastIssueSource,
+    "$eve.last_issue_tool": summary.lastIssueTool,
+    "$eve.last_issue_tool_call_id": summary.lastIssueToolCallId,
+    "$eve.last_issue_turn_id": summary.lastIssueTurnId,
+    "$eve.last_issue_type": summary.lastIssueType,
+    "$eve.rejected_action_count": summary.rejectedActionCount,
```
ruiconti: How expensive would it be to derive this data (eg platform-side) instead of expecting the client to relay? Worries:
- tagged data drifts from the actual source of truth eg workflow state (active/completed) diverging from the tagged attribute
- client needs to be aware of versioning and account for schema mismatches/backcompat

For instance,
- `session_status` can be derived from workflow run state
- `error_count`, `last_*` and `failed_*` can be derived from the event stream

JohnPhamous: removed, ty!

ruiconti (approved):

Issue #529, opened 2026-07-03
Linear channel: approving a HITL tool call permanently kills the session — channel context is appended after the approval response, so `collectToolApprovals` never sees it (dangling `tool_use` → Anthropic 400)


Lukavyi:
### Link to a minimal reproduction

Any Linear-channel agent with one approval-gated tool reproduces it deterministically (steps below) — the trigger is the channel's per-prompt context, not app code.

### Steps to reproduce

1. Agent with an authored tool gated by `approval: always()`, an Anthropic model (direct `@ai-sdk/anthropic` or an Anthropic-compatible endpoint), and the built-in Linear channel (`eve/channels/linear`).
2. From Linear, delegate an issue so the model calls the gated tool → the elicitation "Approve tool call: <tool> / Yes / No" is posted, turn parks.
3. Reply "Yes" in the Agent Session.
4. Next model call 400s:
   ```
   AI_APICallError: messages.N: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_… Each `tool_use` block must have a corresponding `tool_result` block in the next message.
   ```
5. Every subsequent prompt rebuilds the same corrupted history — the session is permanently dead.

### Current vs. expected behavior

**Root cause (traced through `dist/src/harness/tool-loop.js`, `dist/src/harness/input-requests.js`, `ai@7.0.14`, `@ai-sdk/anthropic@4.0.7`):**

1. The Linear channel sends `context` (the `<linear_context>` block + `previousComments`) on **every** prompt, including the one that answers an approval (`dispatchAgentSession` → `send({ context: [formatLinearContextBlock(t), …] })`).
2. `resolvePendingInput` correctly resolves "Yes" and builds `[...history, assistant(tool_use + tool-approval-request), tool(tool-approval-response)]`.
3. `executeStepBody` then appends `S.input.context` as `user` messages **after** that tool message, so the prompt no longer *ends* with the approval response.
4. The AI SDK's `collectToolApprovals` only inspects `messages.at(-1)`. It now sees a `user` message → no approvals collected → **the approved tool never executes** and no `tool_result` is produced.
5. Client-side validation doesn't catch it (`convertToLanguageModelPrompt` excuses dangling `tool_use` ids that have an approval response anywhere in history), and `@ai-sdk/anthropic` silently `continue`s over `tool-approval-response` parts — so Anthropic receives a dangling `tool_use` and rejects the request.

[...]

PR #588, opened 2026-07-08
fix(eve): persist HITL approval results in session history

### LINKED ISSUE #236: needsApproval on an authored (local) executable tool crashes on approve-resume with the OpenAI Responses provider: `No tool output found for function call`
## Summary

An **authored, executable** tool with `needsApproval: always()` parks correctly for approval (`input.requested` → `session.waiting`), but on the **resume turn after the user approves**, the next model call is rejected by the OpenAI **Responses** API:

```
AI_APICallError: No tool output found for function call call_<id>
```

The approved tool's `execute()` never runs (the durable step stays `waiting_approval`), and `turn.failed` → `session.failed`. So an approval-gated executable tool cannot complete a single approve→execute cycle on the OpenAI Responses provider.

This is **distinct from #203/#204** (`defineClientTool`): that is about *client-resolved, executeless* HITL tools producing a **duplicate** `tool_result`, verified on Anthropic via Gateway. This is an approval-gated **executable** tool producing a **missing** result on the **OpenAI Responses** provider. #203's discussion asserts the approval-gated-executable path works — but the cited eval uses Anthropic; on OpenAI Responses it fails.

## Environment

- `eve` **0.13.1** (also repro'd on 0.11.5)
- `ai` **7.0.0-beta.178**, `@ai-sdk/openai` **4.0.0-beta.74**
- Model: **`openai/gpt-5.5`** (defaults to the Responses API)
- Next.js 16 (App Router), `withEve`, `useEveAgent` over same-origin `/eve/v1/*`
- Default harness; approval answered via structured `inputResponses` (`optionId: 'approve'`), freeform disabled during the gate

## Reproduction

1. Author a tool that has **both** `execute` and `needsApproval: always()` (`eve/tools/approval`):
   ```ts
   export default defineTool({
     description: 'Do a gated side effect.',
     inputSchema: z.object({ /* … */ }),
     needsApproval: always(),
     async execute(input, ctx) { /* real side effect */ return { success: true } },
   })
   ```
2. Have the agent call it. The harness parks it for approval.
3. Approve via the client (`agent.send({ inputResponses: [{ requestId, optionId: 'approve' }] })`).
4. The resume turn fails on the first model call.

## Observed event trace

```
input.requested            ← harness parks the gated tool call
turn.completed
session.waiting            ← parked ✅
[user approves]
turn.started               ← resume accepted ✅
step.started
step.failed                ← MODEL_CALL_FAILED: AI_APICallError
                             "No tool output found for function call call_<id>"
turn.failed
session.failed
```

## Root-cause analysis (evidence)

1. The error string `No tool output found for function call` is **not** present in `@ai-sdk/openai` — it is OpenAI's **server** rejecting a request whose input contains a `function_call` with **no** matching `function_call_output`.
2. On approve, the harness records the approval and emits a `tool-approval-response` part for the gated call, but the call's **`tool-result` is never produced before the model is re-invoked** (the executor hasn't run — the step stays `waiting_approval`).
3. The `ai` package **strips `tool-approval-response` parts for non-`providerExecuted` (local) tools** before sending to the provider — i.e. for a local tool the approval-response is *not* the provider-level resolution; the **executor's `tool-result` is expected** instead.
4. The OpenAI provider converters confirm neither transport pairs an approval-only local call:
   - **Responses** converter maps `tool-approval-response` → `mcp_approval_response` keyed by `approval_request_id` (MCP tools only). A regular authored function tool has no `approval_request_id`, so the `function_call` is left **without** a `function_call_output`.
   - **Chat** converter **skips** `tool-approval-response` entirely (`if (toolResponse.type === 'tool-approval-response') continue;`), leaving the assistant `tool_calls` unpaired too.

Net: on the resume turn the reconstructed prompt carries the gated tool's `function_call`/`tool_call` with **no result**, because the executor hasn't run and the approval-response isn't a provider-level result for a local tool → the OpenAI Responses API 400s.

## Expected

Per the contract described in #203 ("approval allows execution; the executor produces the result"), after the user approves a **local executable** tool the harness should **run `execute()` and supply its `tool-result` for that call id before the next model call** (so the `function_call` is paired). Today, on the OpenAI Responses provider, the model is re-invoked with the call still unpaired.

## Workaround

Drop `needsApproval` on the side-effect tools and gate via the built-in `ask_question` (Confirm/Cancel) instead — the answer is paired as a real `tool-result`, so resume succeeds. This trades the framework-enforced gate for an instruction-enforced one.

### LINKED ISSUE #460: Approved HITL tool's execution result missing from history — "tool_use ids were found without tool_result blocks"
### eve version
`eve@0.17.1`

### Node.js version
`v24.18.0`

### Where does the bug occur?
- Production (`eve start` / deployed, on Vercel)

### Describe the bug
An approval-gated (`approval: once()`) authored tool with a real `execute` sometimes has its side effect run successfully, but the resulting `tool_result` never makes it into session history before the harness sends the next model call — leaving that `tool_use` id dangling. The next request to Anthropic is rejected with:

```
messages.10: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01KM8aeG4pT8SNYub2Cbgj4Q. Each `tool_use` block must have a corresponding `tool_result` block in the next message.
```

This is not a one-off: it reproduced twice independently, in the same app, on two different turns of two different conversations, always with the same shape:

1. The model calls an approval-gated tool (`hubspot_create_object`, `approval: once()`, with a real async `execute` that PATCHes/POSTs to an external API).
2. The user approves the tool call (`Responded: Yes` in the UI).
3. The tool's real side effect completes — confirmed both times by checking the external system directly (a HubSpot company record was created both times, with the exact properties from the approved call's input).
4. A second call to the same tool name (a different object, e.g. a HubSpot deal or contact — auto-approved via `once()`'s "already approved this session" logic) completes and renders normally in the UI.
5. The *next* turn's model call to Anthropic 400s with the error above, naming the **first, human-approved** call's `tool_use` id as the one missing a `tool_result`. From that point on, every subsequent message in the session fails identically, since eve keeps resending the same corrupted history.

I instrumented nothing in the built output for this one — this is from production usage of a Next.js app using `@eve/react`'s chat hook against eve's session/stream API. I confirmed there's no custom code in the app that constructs or replays the Anthropic message array; the app only renders/forwards eve's own session/stream events, so the corrupted history is assembled entirely inside eve's harness.

I traced through `node_modules/eve/dist/src/harness/{tool-loop,input-requests,input-extraction}.js` and `node_modules/ai/dist/index.js` (`ai@7.0.4`) trying to find the exact defect:

- Ruled out: the tool's `execute` throwing (it didn't — the HubSpot property-name validation the tool does locally didn't reject either payload).
- Ruled out: an `approvalKey` collision between the two same-tool-name calls causing them to be matched to the wrong approval response — each `tool_use` gets an independent approval request keyed by `toolCallId`/`approvalId` in the AI SDK (`resolveToolApproval`, `collectToolApprovals`), and eve's `approvalKey` only affects what gets recorded into the session's `approvedTools` set for future `once()` checks, not response matching.
- Still open: something in the approval-resume path (`resolvePendingInput` in `input-requests.js`, and/or `handleStepResult` in `tool-loop.js`) appears to let the harness proceed to build the next model request before the approved tool's actual (async, network-bound) execution result is durably appended to session history — i.e. the tool call resolves and its side effect fires, but the harness doesn't wait for/persist that specific `tool_result` before continuing the loop.

This looks related to the general fragility around HITL approval-resume + multi-round tool history reconstruction already visible in #203 (duplicate `tool_result` for client-resolved tools on resume) and #236 (approve-resume crash on OpenAI: "No tool output found for function call"), but distinct from both — here the tool is a normal `execute`-based approval-gated tool (not client-resolved), and the failure is a *missing* result rather than a duplicate one.

### Steps to reproduce
1. `npx eve@latest init repro`
2. `agent/agent.ts`: `export default defineAgent({ model: anthropic("claude-sonnet-4-6") })` (direct `@ai-sdk/anthropic` provider, not the AI Gateway model-reference string).
3. Add an authored tool with `approval: once()` and a real async `execute` that does non-trivial I/O (e.g. an `await fetch(...)` to an external API — the delay seems relevant, since a fast/instant `execute` may not reproduce it).
4. Give the agent a skill/instruction that calls this tool twice with different inputs in one user turn (e.g. "create object A, then create object B" using the same tool).
5. Approve the first (human-gated) call when prompted.
6. Watch the second, auto-approved (`once()`-satisfied) call complete, then send any further message in the same session.
7. The next model call to Anthropic 400s with `tool_use ids were found without tool_result blocks immediately after: <id>`, where `<id>` is the *first, human-approved* call's id — even though its side effect actually ran.

### Expected behavior
After a human approves a HITL-gated tool call, eve should durably persist that specific tool call's `tool_result` into session history before advancing the turn/loop, so every `tool_use` block always has a matching `tool_result` in the next Anthropic request — regardless of how long the approved tool's `execute` takes to resolve, and regardless of whether other tool calls in the same turn complete first.

### LINKED ISSUE #533: Anthropic provider: pending approval-parked tool calls break every subsequent model call (tool_use without adjacent tool_result → 400 → session.failed)
Title: Anthropic provider: pending approval-parked tool calls break every subsequent model call (tool_use without adjacent tool_result → 400 → session.failed)

## Summary

On an Anthropic model, any model call whose replayed history contains a **pending approval-parked tool call** is rejected by the Anthropic Messages API, and the session dies terminally. Approval-gated tools (`approval: always()`) are effectively unusable on Anthropic backends: the owner clicking **approve** is itself what kills the session.

OpenAI backends tolerate the same replayed shape, which is presumably why this hasn't surfaced on the default path before — but `eve init` now scaffolds `anthropic/claude-sonnet-5` (0.17.2), so new users hit this as soon as they add an approval-gated tool.

## Versions

- eve `0.18.1` (also inspected 0.18.2/0.19.0 changelogs — the 0.19.0 "tool execution failures now return failed tool results" fix is adjacent but covers execution failures, not approval parks)
- `@ai-sdk/anthropic` `4.0.7`, `ai` `7.0.2`
- model `anthropic/claude-sonnet-5`

## Reproduction

1. Agent with a dynamic tool whose approval resolves to `always()`.
2. Send a message that makes the model call that tool → the call parks, `input.requested` (confirmation) is emitted, turn completes, session waits. Durable history now holds the assistant `tool-call` part with **no** tool result — by design since 0.16.1 (pending parks survive unrelated follow-ups).
3. Either **respond to the approval** or send any input response that starts a new turn.
4. The next model call replays history verbatim and Anthropic rejects it:

```
AI_APICallError: messages.20: `tool_use` ids were found without `tool_result` blocks
immediately after: toolu_01XAi9GCiC1bN4CfpoUJhdNw. Each `tool_use` block must have
a corresponding `tool_result` block in the next message.
```

5. `step.failed` → `turn.failed` → **`session.failed`** (`MODEL_CALL_FAILED`). Observed live twice in one afternoon, on two different parked tools; event sequence from our durable log:

```
actions.requested (send_customer_email)   ← parks
input.requested   kinds=confirmation
turn.completed / session.waiting
turn.started                              ← the approval response
step.started
step.failed  MODEL_CALL_FAILED (the 400 above)
turn.failed
session.failed                            ← terminal; client starts a fresh session
```

## Expected

The Anthropic prompt conversion should sanitize pending tool calls the way the OpenAI path evidently does (0.15.1/0.15.2 fixed the equivalent "No tool output found for function call" poisoning there) — e.g. inject a synthetic "pending approval" tool result for any call id with no result yet, or defer emitting the tool-call part until resolution.

## Workaround we're shipping

A `transformParams` middleware via `wrapLanguageModel`: walk the prompt; for every assistant `tool-call` part whose id has no `tool-result` anywhere later, inject `{type: "tool-result", output: {type: "error-text", value: "Not executed yet — waiting for the owner to approve or decline this action."}}` immediately after its message. Outbound-only, so the repair stops applying once the real result lands in history. This unblocks approvals on Anthropic fully in our testing (approve resumes execute correctly; unrelated messages queue without killing the session). Happy to share the implementation or test cases if useful.

ruiconti:
## What

- Persist AI SDK's call-wide `responseMessages` after a streamed call.
- Keep `fullStream` consumption in `emission.ts` for live eve events. Stop serializing successful tool results there.
- Add provider evals for #236, #460, and #533 plus a real AI SDK contract test.

## Concrete execution

`history` below means durable `session.history`. `messages` is the array passed to one `agent.stream()` call.

```ts
turn0: callProvider() -> requestApproval()                     // no tool execution
turn1: collectApproval() -> executeToolCall() -> callProvider() // tool executes once
turn2: callProvider()                                          // no tool execution
```

```ts
const user = (content) => ({ role: "user", content })
const assistant = (...content) => ({ role: "assistant", content })
const tool = (...content) => ({ role: "tool", content })
const providerUser = (text) => user([{ type: "text", text }])

const U0 = user("Run guardedEcho with the approved note.")

const CALL = {
  type: "tool-call",
  toolCallId: "call-1",
  toolName: "guardedEcho",
  input: { note: "approved" },
}

const APPROVAL_REQUEST = {
  type: "tool-approval-request",
  approvalId: "approval-1", // AI SDK generates the real ID
  toolCallId: "call-1",
}

const APPROVAL_RESPONSE = {
  type: "tool-approval-response",
  approvalId: "approval-1",
  approved: true,
}

const RAW_RESULT = {
  type: "tool-result",
  toolCallId: "call-1",
  toolName: "guardedEcho",
  input: { note: "approved" },
  output: { echoed: "approved" },
  dynamic: false,
}

const MODEL_RESULT = {
  type: "tool-result",
  toolCallId: "call-1",
  toolName: "guardedEcho",
  output: { type: "text", value: "canonical:approved" },
}

const ANSWER_1 = assistant({
  type: "text",
  text: "Approved result received.",
})
```

### Turn 0: the model asks to run the tool

```ts
async function turn0() {
  let history = []
  let pendingInputBatch
  let executeCount = 0

  const messages = [...history, U0]

  // Inside agent.stream(): AI SDK calls the provider first.
  const providerMessages = [providerUser(U0.content)]
  const providerStep = await callProvider(providerMessages)
  // providerStep contains CALL

  // AI SDK applies the approval policy to CALL.
  // It emits APPROVAL_REQUEST and does NOT call executeToolCall().
  const responseMessages = [assistant(CALL, APPROVAL_REQUEST)]
  const fullStream = [
    CALL,
    {
      type: "tool-approval-request",
      approvalId: "approval-1",
      toolCall: CALL,
    },
  ]
  executeCount === 0

  // eve parks. The unresolved assistant response stays in the pending batch.
  history = messages
  pendingInputBatch = {
    requests: [{
      requestId: "approval-1",
      prompt: "Approve tool call: guardedEcho",
      options: [
        { id: "approve", label: "Yes" },
        { id: "deny", label: "No" },
      ],
      action: {
        kind: "tool-call",
        callId: "call-1",
        toolName: "guardedEcho",
        input: { note: "approved" },
      },
    }],
    responseMessages,
  }

  return { history, pendingInputBatch, executeCount, fullStream }
}
```

### Turn 1: the human approves, then AI SDK executes the tool

```ts
async function turn1({ history, pendingInputBatch }) {
  const inputResponses = [
    { requestId: "approval-1", optionId: "approve" },
  ]

  // eve resolves the pending input. This records permission, not execution.
  const messages = [
    ...history,
    ...pendingInputBatch.responseMessages,
    tool(APPROVAL_RESPONSE),
  ]
  pendingInputBatch = undefined
  let executeCount = 0

  // agent.stream({ messages }) starts here.
  // AI SDK performs this block BEFORE it calls the provider.
  const { approvedToolApprovals } = collectToolApprovals(messages)
  // approvedToolApprovals = [{ toolCall: CALL, approvalResponse: APPROVAL_RESPONSE }]

  const execution = await executeToolCall(CALL) // <-- THE TOOL EXECUTES HERE
  // execution.output = RAW_RESULT
  executeCount += 1

  const rawResult = execution.output
  const modelResult = await createToolModelOutput(rawResult)
  // createToolModelOutput() invokes guardedEcho.toModelOutput().
  // modelResult = MODEL_RESULT

  const initialResponseMessages = [tool(modelResult)]

  // AI SDK removes local approval-control parts, then adds the result.
  const providerMessages = [
    providerUser(U0.content),
    assistant(CALL),
    tool(MODEL_RESULT),
  ]

  // The provider sees CALL followed by MODEL_RESULT.
  const providerStep = await callProvider(providerMessages)
  // providerStep.response.messages = [ANSWER_1]
  const steps = [providerStep]

  // RAW_RESULT is emitted before the provider step's stream events.
  const fullStream = [RAW_RESULT, ...providerStep.fullStream]

  // AI SDK builds this implicitly. The final StepResult alone has only ANSWER_1.
  const responseMessages = [
    ...initialResponseMessages,
    ...steps.flatMap((step) => step.response.messages),
  ]
  // responseMessages = [tool(MODEL_RESULT), ANSWER_1]

  // eve persists the complete call output after agent.stream() finishes.
  history = [...messages, ...responseMessages]
  // history = [
  //   U0,
  //   assistant(CALL, APPROVAL_REQUEST),
  //   tool(APPROVAL_RESPONSE),
  //   tool(MODEL_RESULT),
  //   ANSWER_1,
  // ]

  return { history, pendingInputBatch, executeCount, fullStream }
}
```

`role: "tool"` on `tool(APPROVAL_RESPONSE)` is only the AI SDK message envelope. The tool has not run at that point. Execution happens at the marked `await executeToolCall(CALL)` line.

### Turn 2: a later user message replays the completed call

```ts
async function turn2({ history, executeCount }) {
  const messages = [...history, user("What happened?")]

  // No approval sits at the tail, so AI SDK executes no tool in this turn.
  collectToolApprovals(messages).approvedToolApprovals.length === 0
  executeCount === 1

  // AI SDK strips APPROVAL_REQUEST and APPROVAL_RESPONSE.
  // CALL and MODEL_RESULT remain paired.
  const providerMessages = [
    providerUser(U0.content),
    assistant(CALL),
    tool(MODEL_RESULT),
    ANSWER_1,
    providerUser("What happened?"),
  ]

  await callProvider(providerMessages)
}
```

## What changed in eve

```ts
function emitLiveEvents(streamResult) {
  // Unchanged: emission.ts consumes RAW_RESULT and emits action.result.
  return consume(streamResult.fullStream)
}

function oldMessagesToPersist(streamResult) {
  // Removed: eve rebuilt a ModelMessage from RAW_RESULT.
  const eveSerializedResult = tool({
    ...MODEL_RESULT,
    output: { type: "json", value: { echoed: "approved" } },
  })
  return [eveSerializedResult, ...streamResult.steps[0].response.messages]
}

async function newMessagesToPersist(streamResult) {
  // Added: AI SDK already applied toModelOutput and included pre-model results.
  return await streamResult.responseMessages
}
```

The persistence timing is unchanged. AI SDK supplies `MODEL_RESULT` to the current provider call before the provider runs. eve persists that result after the call so later turns can replay it.

## Failure branches

```ts
async function failure529({ history, pendingInputBatch }) {
  const messages = [
    ...history,
    ...pendingInputBatch.responseMessages,
    tool(APPROVAL_RESPONSE),
    user("<linear_context>...</linear_context>"),
  ]

  // AI SDK 7.0.0 only collects approvals when messages.at(-1).role === "tool".
  const { approvedToolApprovals } = collectToolApprovals(messages)
  approvedToolApprovals.length === 0

  // executeToolCall(CALL) IS NEVER CALLED.
  const initialResponseMessages = []

  const providerMessages = [
    providerUser(U0.content),
    assistant(CALL),
    providerUser("<linear_context>...</linear_context>"),
  ]

  await callProvider(providerMessages) // fails in turn 1: CALL has no result
}

async function failureMissingDurableResult(turn1State) {
  // Turn 1 succeeded. The tool executed and its provider prompt contained MODEL_RESULT.
  turn1State.executeCount === 1

  // Buggy persistence drops initialResponseMessages and saves only the provider step.
  const badHistory = [
    U0,
    assistant(CALL, APPROVAL_REQUEST),
    tool(APPROVAL_RESPONSE),
    ANSWER_1,
  ]

  const providerMessages = [
    providerUser(U0.content),
    assistant(CALL),
    ANSWER_1,
    providerUser("What happened?"),
  ]

  await callProvider(providerMessages) // fails in turn 2: CALL has no result
}
```

#529 is `failure529()` and remains deferred to [vercel/ai#17033](https://github.com/vercel/ai/issues/17033). The missing durable-result branch is the boundary this PR hardens. The old manual serializer also persisted raw JSON instead of the `toModelOutput` text sent to the current model.

The [AI SDK integration test](https://github.com/vercel/eve/blob/9319f4bfe9051425a135aaf0156b00213a123141/packages/eve/src/harness/ai-sdk-approval-resume.integration.test.ts#L48-L183) pins the execution order and response collections. The [eve persistence unit](https://github.com/vercel/eve/blob/9319f4bfe9051425a135aaf0156b00213a123141/packages/eve/src/harness/tool-loop.test.ts#L4829-L4883) pins the final history.

## Issue split

The #236, #460, and #533 evals passed before the production change, so they are regression coverage rather than red-to-green evidence. #529 remains upstream.

Closes #236
Closes #460
Closes #533


packages/eve/src/harness/emission.ts
ruiconti, line 571: Below here is the handling for `tool_result` events that did _not_ have a matching `tool_call` in current stream.

ctgowrie (approved):

packages/eve/src/harness/tool-loop.ts
```diff
@@ -900,51 +902,22 @@ export function createToolLoopHarness(config: ToolLoopHarnessConfig): StepFn {
...
+            invalidInputToolCallIds,
+            responseMessages: appendMissingToolResultMessages({
+              append: trailingInlineToolResultParts,
+              responseMessages: accumulatedResponseMessages,
+            }),
+            stepResult,
+            toolResults: [...toolResultsByCallId.values()],
+          });
         }
         await agent.generate({ abortSignal: config.abortSignal, messages: callMessages });
```
AndrewBarba: **[P2] Persist `generate()` response messages too**

`handleEvent` is optional, so this branch remains reachable. AI SDK exposes approved local tool results produced before the model step through the call-wide `GenerateTextResult.responseMessages`, but this discards the generate result and persists only `hooks.stepResult`, whose response contains the final-step messages. A no-event harness can therefore still leave a dangling tool call in durable history and fail on the next provider replay. Please capture `agent.generate()` and apply the same accumulated-message and empty-response handling as the streaming branch, or make streaming mandatory.

[reviewed with gpt-5.6-sol]

zico-io:
Tested this branch (`rui/hitl-approval-red-evals`) built from source and installed into a real app on `anthropic/claude-sonnet-5` via the AI Gateway (local `eve start`). Confirmed the fix is actually live — `inlineAuthorizationResults` is present in the compiled bundle after a clean rebuild.

**The inline path this PR's contract test covers is fixed, but the out-of-band approval-resume flow from #533 still reproduces identically.** By "out-of-band" I mean: the tool call parks at `input.requested`, and the approval is delivered in a *separate* follow-up request (`inputResponses: [{ requestId, optionId: "approve" }]`) on the same session — i.e. a human clicking approve later, not an approval supplied inline with the triggering message.

Repro (same trivial `approval: always()` tool as my comment on #533):
1. `POST /eve/v1/session` with a message that makes the model call the gated tool → parks, `input.requested`, session waits.
2. Answer on the same session with `inputResponses: [{ requestId, optionId: "approve" }]` (no inline message).
3. On resume:

```
turn.failed: MODEL_CALL_FAILED
[eve:harness.tool-loop] AI Gateway rejected the model request before the agent produced a response
  messages.2: `tool_use` ids were found without `tool_result` blocks immediately after: toolu_01... 
```

So the approved tool never executes on the out-of-band path. The fix here seems scoped to `inlineAuthorizationResults`; the park → separate-answer → resume path (which is what #533 is about, and what a real approval UI does) still assembles the resume model call with the parked `tool_use` unpaired.

Caveat: this is a from-source `pnpm build` of the branch, so if there's a build nuance I've missed I'm glad to hear it — but the fix symbol is in the running bundle and it's a clean rebuild.

PR #602, opened 2026-07-08
fix(subagents): proxy authorization events

iroiro147:
### Description

Fixes #568.

Subagent sessions already proxied `input.requested` events to the parent turn, but `authorization.required` stayed on the child stream. This adds subagent authorization hook payloads and forwards child `authorization.required` / `authorization.completed` events through the parent turn inbox so user-facing channels can render and clear connection auth prompts.

The proxied required event preserves the child callback URL, so the existing connection callback route resumes the child session that owns the pending authorization.

### How did you test your changes?

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/subagent-adapter.test.ts src/execution/turn-workflow.test.ts src/execution/workflow-steps.test.ts src/internal/structural-tests/file-length.test.ts`
- `pnpm --filter eve run typecheck`
- `pnpm exec oxfmt --check packages/eve/src/channel/types.ts packages/eve/src/execution/subagent-adapter.ts packages/eve/src/execution/subagent-hitl-proxy.ts packages/eve/src/execution/subagent-auth-proxy-step.ts packages/eve/src/execution/turn-workflow.ts packages/eve/src/execution/workflow-steps.ts packages/eve/src/execution/subagent-adapter.test.ts packages/eve/src/execution/turn-workflow.test.ts packages/eve/src/execution/workflow-steps.test.ts`
- `pnpm exec oxlint packages/eve/src/channel/types.ts packages/eve/src/execution/subagent-adapter.ts packages/eve/src/execution/subagent-hitl-proxy.ts packages/eve/src/execution/subagent-auth-proxy-step.ts packages/eve/src/execution/turn-workflow.ts packages/eve/src/execution/workflow-steps.ts packages/eve/src/execution/subagent-adapter.test.ts packages/eve/src/execution/turn-workflow.test.ts packages/eve/src/execution/workflow-steps.test.ts`
- `pnpm changeset status --since origin/main`
- `npx -y -p node@24 -p pnpm@11.7.0 -c 'node -v && pnpm --filter eve run test:unit'` (Node v24.18.0; 430 files passed, 4533 passed, 1 skipped)
- `npx -y -p node@24 -p pnpm@11.7.0 -c 'node -v && pnpm --filter eve run build:js && pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/workflow-entry.integration.test.ts'` (Node v24.18.0; 1 file passed, 8 passed)
- `git diff --check`

### PR Checklist

[...]

PR #603, opened 2026-07-08
fix(eve): subagents - proxy authorization events

### LINKED ISSUE #568: Subagent auth requests are not bubbled up to the parent session

AndrewBarba:
## Summary

- forward `authorization.required` and `authorization.completed` from delegated children through each active parent channel
- preserve parent adapter state and continuation-token changes across the durable proxy boundary
- keep OAuth callbacks scoped to the child session while allowing nested subagent chains to surface auth at the root channel
- document the behavior and include a patch changeset

## Root cause

The framework-owned subagent adapter forwarded `input.requested` events to the active parent turn, but it had no equivalent path for authorization lifecycle events. A child could therefore park on interactive authorization while its challenge remained visible only on the unconsumed child stream.

## User impact

Authorization prompts raised by local subagents now reach the user-facing parent channel, and completion events update or clear channel UI before the child result returns. No public API or authorization event wire shape changes.

## Validation

- `pnpm test:unit` — 4,534 passed, 1 skipped
- `pnpm test:integration` — 386 passed
- `pnpm typecheck`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`

## End-to-end coverage

This PR does not add a fixture-owned e2e eval for delegated authorization. The existing eval driver waits for a public turn boundary, but this proxy intentionally keeps the parent turn active while the child waits for OAuth completion, so the driver cannot currently observe and complete the in-flight authorization flow.

A follow-up PR will add the missing eval capability for observing in-flight events and completing authorization callbacks, then extend `agent-subagents-hitl` to cover `authorization.required`, `authorization.completed`, and eventual child completion.

The signed commit includes co-author trailers for the contributors behind #519, #569, and #602 in appreciation of their prior work.

Fixes #568

### DIFF
--- a/.changeset/subagents-forward-authorization.md
+++ b/.changeset/subagents-forward-authorization.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Surface authorization prompts and completion updates from local subagents on the parent channel, including through nested delegation chains, while keeping the authorization callback scoped to the child session.
--- a/apps/frameworks/nuxt/package.json
+++ b/apps/frameworks/nuxt/package.json
@@ -17,6 +17,8 @@
     "zod": "catalog:"
   },
   "devDependencies": {
-    "@types/node": "catalog:"
+    "@types/node": "catalog:",
+    "typescript": "6.0.3",
+    "vue-tsc": "3.3.6"
   }
 }
--- a/docs/connections/overview.mdx
+++ b/docs/connections/overview.mdx
@@ -138,6 +138,8 @@ export default defineMcpClientConnection({
 
 `"linear/myagent"` is the UID you chose when registering the Connect client. `connect("linear/myagent")` is shorthand for a user-scoped interactive OAuth connection: eve resolves a token for the active user before each tool call, emits `authorization.required` when that user has not authorized yet, and resumes the parked turn after the callback completes.
 
+When a local subagent needs interactive authorization, eve surfaces the authorization lifecycle on the root session's channel, including through nested subagent chains. The challenge still points to the child session's callback, so completing it resumes the child directly while the parent continues waiting for its result.
+
 That means the channel that creates or continues the session must authenticate a real user. For a web app, configure `agent/channels/eve.ts` so your app session maps to `principalType: "user"`; for platform channels, use the built-in channel auth that maps the sender to a user principal. If no authenticated user is attached to the session, the first user-scoped connection call fails with `reason: "principal_required"`.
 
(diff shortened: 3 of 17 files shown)


AndrewBarba:
Thank you @vanessaxteo, @blurrah, and @iroiro147 for the work you put into #519, #569, and #602. Each PR helped confirm the problem and explore the intended behavior for propagating subagent authorization events.

Although the final implementation was developed independently, we added all three of you as co-authors on the final commit to recognize your contributions. We really appreciate the time and care you invested here!

allenzhou101 (approved):

ruiconti (approved):

PR #607, opened 2026-07-08
feat(sandbox): forward just-bash SandboxOptions through justbash(), enable python/js-exec

### LINKED ISSUE #431: justbash() backend can't enable just-bash's bundled Python (capability passthrough)

arimxyer:
## What

Forward just-bash's `SandboxOptions` through the `justbash()` backend (via `JustBashSandboxCreateOptions` → `Sandbox.create(...)`), and enable its bundled interpreters:

- **Execution controls** the backend previously hardcoded or omitted: `timeoutMs`, `maxCallDepth`, `maxCommandCount`, `maxLoopIterations`, `defenseInDepth`.
- **Bundled-interpreter capability flags** `python` (stdlib-only CPython via WebAssembly) and `javascript` (js-exec via QuickJS), both off by default.

`justbash({ python: true })` now runs `python3` inside the sandbox. **Closes #431.**

## Why

`justbash()` built its interpreter with `Sandbox.create({ cwd, env, fs, network })` and exposed only `autoInstall`, so an author couldn't set execution limits, an overall timeout, defense-in-depth, or — the driver behind #431 — enable `python3` / js-exec. The capability flags live on just-bash's `BashOptions`, and just-bash's `Sandbox.create` (the call eve uses) didn't forward them; that upstream gap was fixed in **vercel-labs/just-bash#284**, shipped in **just-bash 3.1.0**.

## Requires `just-bash@^3.1.0` (peer floor raised from `^3.0.0`)

This is load-bearing, not housekeeping: `python`/`javascript` only reach the interpreter through `Sandbox.create` as of 3.1.0. On older just-bash they are **silently dropped** — `python: true` would no-op rather than error — so the version floor is the only guard against that silent failure.

just-bash is added to `minimumReleaseAgeExclude` so the 3.1.0 bump resolves without waiting out the 2-day quarantine (it's a first-party vercel-labs package; the lockfile change is scoped to just-bash's own subtree). Happy to drop that and let 3.1.0 age out instead if you'd prefer.

## Scope: why not `commands` / `customCommands` / `fetch`

#431 asks for python3 + js-exec, which this delivers. The other three just-bash capability flags are intentionally **not** surfaced: their just-bash types (`CommandName[]`, `CustomCommand`, `SecureFetch`) can't be honestly restated as primitives, and exposing them would put just-bash's types on eve's *public* API — the same reason `defenseInDepth` is narrowed to a plain `boolean` here. They can follow later if there's demand.

## Backward compatibility

No behavior change when the new fields are omitted — each forwards as `undefined`, i.e. just-bash's existing default (network still defaults to full internet access exactly as before).

## Tests

- `just-bash-capability-options.integration.test.ts` (mock): forwarding assertions extended to `python`/`javascript`, plus the default-`undefined` case.
- `just-bash.scenario.test.ts` (real just-bash): two-sided e2e for each interpreter — `justbash({ python: true })` runs `python3 -c 'print(1 + 1)'` and `justbash({ javascript: true })` runs `js-exec -c 'console.log(1 + 1)'`, both → exit 0, stdout `2`; without the flag each command is not-found (exit 127).

## Validation

`turbo run typecheck --filter=eve`, `oxlint`, the mock integration test, and the full just-bash scenario suite (21/21) all pass locally against just-bash 3.1.0. A `minor` changeset is included.

## Supersedes #446

#446 forwarded only the already-supported limits/timeout fields and deferred the capability flags until just-bash#284 shipped. Now that 3.1.0 is out, this PR does the whole thing (limits + python/js-exec) off fresh `main`; I'll close #446 in favor of it.

### DIFF
--- a/.changeset/justbash-forward-sandbox-options.md
+++ b/.changeset/justbash-forward-sandbox-options.md
@@ -1,0 +1,14 @@
+---
+"eve": minor
+---
+
+Forward just-bash `SandboxOptions` through the `justbash()` backend via `JustBashSandboxCreateOptions`, and enable its bundled interpreters.
+
+- Execution controls the backend previously hardcoded or omitted: `timeoutMs`, `maxCallDepth`, `maxCommandCount`, `maxLoopIterations`, and `defenseInDepth`.
+- Bundled-interpreter capability flags `python` (stdlib-only CPython via WebAssembly) and `javascript` (js-exec via QuickJS), both off by default. Passing `justbash({ python: true })` now runs `python3` in the sandbox, closing #431.
+
+Each field is optional and falls back to just-bash's own default when omitted, so existing apps are unaffected.
+
+**Requires `just-bash@^3.1.0`** (the peer floor is raised from `^3.0.0`). This is load-bearing, not housekeeping: `python`/`javascript` only reach the interpreter through `Sandbox.create` as of just-bash 3.1.0 (vercel-labs/just-bash#284). On older just-bash they are silently dropped — `python: true` would no-op rather than error — so the version floor is the only guard against that silent failure.
+
+The remaining just-bash capability flags (`commands`, `customCommands`, `fetch`) are intentionally not surfaced: their just-bash types cannot be honestly restated as primitives, and exposing them would put just-bash's types on eve's public API (the same reason `defenseInDepth` is narrowed to a plain boolean).
--- a/packages/eve/package.json
+++ b/packages/eve/package.json
@@ -330,7 +330,7 @@
     "gray-matter": "4.0.3",
     "jose": "6.2.3",
     "jsonc-parser": "3.3.1",
-    "just-bash": "3.0.1",
+    "just-bash": "3.1.0",
     "microsandbox": "0.5.5",
     "next": "catalog:",
     "picocolors": "catalog:",
@@ -349,7 +349,7 @@
     "@opentelemetry/api": "^1.0.0",
     "ai": "catalog:",
     "braintrust": "^3.0.0",
-    "just-bash": "^3.0.0",
+    "just-bash": "^3.1.0",
     "microsandbox": "^0.5.0"
   },
   "peerDependenciesMeta": {
--- a/packages/eve/src/execution/sandbox/bindings/just-bash-capability-options.integration.test.ts
+++ b/packages/eve/src/execution/sandbox/bindings/just-bash-capability-options.integration.test.ts
@@ -1,0 +1,108 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { createJustBashSandboxBackend } from "#execution/sandbox/bindings/just-bash.js";
+import { useTemporaryDirectories } from "#internal/testing/use-temporary-app-roots.js";
+
+// Capture every `Sandbox.create(options)` call so we can assert which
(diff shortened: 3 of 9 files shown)


komly:
This still affects eve 0.54.0 and current `main` (`df62d7361afd7e4ee01b57ed8ce7e74ae9fe40ca`). With Node 24.21.0 and just-bash 3.1.0, a real backend check returned exit 127 for both `python3` and `js-exec` during prewarm bootstrap and live execution despite explicit `{ python: true, javascript: true }`. The public `JustBashSandboxCreateOptions` does not expose `python` or `javascript`, and `createBashSandbox()` does not pass either option to `Sandbox.create()`.

The change in #607 covers both paths that matter: template prewarming (including authored bootstrap commands) and live sandbox creation. That overlaps the needed fix; no separate issue is necessary. The PR currently has merge conflicts with `main`, which now also supports `customCommands` and custom filesystems for live handles.

When updating the PR, please retain a real interpreter check for each capability during both prewarm bootstrap and live execution. This catches a missing prewarm forwarding path even if the live-session checks pass.

PR #608, opened 2026-07-08
feat(eve): forward end-user identity across remote agent hops

### LINKED ISSUE #604: Forward end-user identity across `defineRemoteAgent` hops
## Problem

A `defineRemoteAgent` hop drops the caller principal. The create-session POST body carries only `{ callback, message, mode, outputSchema }`, and `OutboundAuthFn` is a zero-arg header factory, so the only identity that can cross the hop is deployment-level trust (e.g. `vercelOidc()`). The receiving agent authenticates the *calling app* (`principalType: "runtime"`/`"service"`), not the end user.

This breaks per-user workloads on the remote deployment — most notably per-user Vercel Connect: `resolveConnectionPrincipal` requires `principalType === "user"` and fails with `principal_required` when the session principal is the calling service.

Local subagents already forward identity natively (`auth` + `initiatorAuth` on `RunInput` via `buildSubagentRunInput`); the `remote-agent-call` branch does not, and there is no supported seam to inject it (hooks are observe-only, inbound context accessors are package-private).

Reported by an enterprise customer splitting a router agent (Slack/admin channels) from a `site-ops` agent deployment that needs each end user to OAuth their own Datadog/Railway/Vercel/GitHub via Connect.

## Proposal

Mirror the local subagent semantics across the remote hop with an explicit opt-in on both sides:

- **Sender:** `defineRemoteAgent({ forwardAuth: true })` serializes the parent turn's `auth.current` / `auth.initiator` into a new strict-validated `forwardedAuth` field on the create-session body. Additionally, `OutboundAuthFn` gains a context argument carrying the session auth so custom schemes can mint per-user credentials at dispatch time.
- **Receiver:** `eveChannel({ forwardedAuth: AuthFn | AuthFn[] })` gates which *transport-authenticated* callers may assert a forwarded principal (reusing the existing `AuthFn` / `routeAuth` primitive, e.g. `vercelOidc({ subjects: [vercelSubject(...)] })`). When accepted, the forwarded identity becomes the session principal (`auth.current` / `auth.initiator`), so downstream Connect, local subagents, and further remote hops work unchanged. The transport caller is preserved for audit as an `eve:forwarded-by` attribute.

Only principal metadata is forwarded — never tokens or credentials. Trust model is "trusted forwarder": the transport auth authenticates the asserting deployment; the receiver explicitly names which callers it trusts.

Research plan: `research/remote-agent-auth-forwarding.md` (PR to follow).

AndrewBarba:
## Summary

A `defineRemoteAgent` hop drops the caller principal: the receiving deployment authenticates the calling *app*, never the end user, which breaks per-user workloads split across deployments — most directly per-user Vercel Connect (`principal_required`). This PR adds explicit, opt-in auth forwarding on both sides of the hop, plus the research plan that specifies it (`research/remote-agent-auth-forwarding.md`, status: implemented).

Closes #604. (Implementation reviewed and merged into this branch via #1128.)

## Design

**Sender — `defineRemoteAgent({ forwardAuth: true })`**
- Dispatch serializes the parent turn's `auth` / `initiatorAuth` into a `forwardedAuth: { current, initiator? }` field on the create-session body. Omitted when the turn has no auth; only principal metadata crosses the wire — never tokens or credentials.
- The sender requires the receiver's `forwardedAuth: "accepted"` acknowledgment on the 202 response. A pre-forwarding receiver silently drops unknown body fields and would run the session as the calling service — the exact silent downgrade this design rejects — so a missing ack fails the dispatch after best-effort cancelling the already-started orphan session via the existing cancel path.
- The flag rides the module-backed runtime definition next to `auth` / `headers`; the compiled manifest node is unchanged.

**Receiver — `eveChannel({ acceptForwardedAuth })`**
- A predicate over the *verified transport principal* (the route-auth result): it authorizes who may forward, not what is forwarded. Because it runs on the already-authenticated principal, the Vercel OIDC current-project bypass can't leak into forwarder trust — the author must match explicitly (e.g. `({ subject }) => subject === vercelSubject({ teamSlug, projectName })`).
- The forwarded payload is validated by a new strict wire schema (`channel/forwarded-auth.ts`, beside `session-callback.ts`) that keeps `authenticator` / `principalType` open — a Slack-authenticated user (`authenticator: "slack-webhook"`) is the flagship case.
- Accepted forwarding replaces the session principal (`session.auth.current` / `.initiator` seed `RunInput.auth` / `.initiatorAuth`), so `resolveConnectionPrincipal` sees a `user` principal, local subagents inherit it, and further `forwardAuth` hops chain it. The verified transport caller is stamped as the `eve:forwarded-by` attribute (receiver-written, overwrites any sender-supplied value, stamped before `onMessage`); it never affects Connect token-cache keying.
- Fail loud, never fall back silently: field without the option → 403, predicate false → 403, malformed → 400, predicate throws → 500, `forwardedAuth` on the continue route → 400. The framework default channel never accepts forwarded auth.

## Tests

- **Unit**: wire schema matrix (open authenticator/principalType, strict keys, attribute shapes), stamping/overwrite, dispatch body construction (forwarded / initiator-omitted / null-auth-omitted / flag-unset), missing-ack failure + orphan cancel, full receiver gate matrix on the create and continue routes.
- **Integration**: create route → `buildRunContext` → forwarded principal seeds `AuthKey` / `InitiatorAuthKey` and resolves a `user` Connect principal (`principal_required` control without forwarding; `principalKey` unaffected by the audit attribute).
- **Scenario (real HTTP)**: a receiver dev server with `acceptForwardedAuth` configured, driven by the real sender dispatch function — trusted forwarder accepted + acknowledged with the receiver's `onMessage` asserting the stamped forwarded principal became the effective caller; untrusted forwarder → 403.

## Notes

- Patch changeset (both options are additive; no public API breaks).
- Docs updated in the same PR: `docs/guides/remote-agents.md` (sender) and `docs/guides/auth-and-route-protection.md` (receiver + trust-model warning).
- `pnpm fmt`, `lint`, `typecheck`, `guard:invariants`, `docs:check`, `test:unit`, `test:integration`, and the new scenario all pass locally.


AndrewBarba:
Pushed a hardening pass after an end-to-end review of the plan (54318138):

1. **Version-skew silent downgrade closed.** A pre-forwarding receiver silently ignores unknown create-body fields, which would run the session as the calling service — the exact failure this design rejects. The receiver now acknowledges acceptance (`forwardedAuth: "accepted"` on the 202), and the sender fails the dispatch when it forwarded a principal and the ack is missing.
2. **Gate runs in strict mode.** The Vercel OIDC always-on current-project bypass does not apply inside `forwardedAuth` — otherwise any same-project caller (including preview deployments) could assert arbitrary principals against production Connect grants.
3. **Audit trail is receiver-written.** `eve:forwarded-by` is always overwritten from the verified transport principal; a forwarder cannot falsify it. Multi-hop chains record the most recent hop only.
4. **Changeset corrected to minor** — the `OutboundAuthFn` signature change is a public API break.
5. **Deliver-route parity cut to out-of-scope.** Remote dispatch is create-only; nothing in eve would send it today.

ruiconti (approved):

ruiconti:
Heads up — this branch picked up a series of review-driven changes, and the PR description predates them (it still names `forwardAuth` / `acceptForwardedAuth` and a response-ack contract that no longer exists). Where the branch actually is:

## API renames

All pre-merge, which is what makes the wire-field rename free — after release it would be a cross-deployment skew event.

- `defineRemoteAgent({ forwardAuth })` → **`forwardPrincipal`**. Next to the `auth: OutboundAuthFn` credential hook, `forwardAuth: true` read as "forward my credentials" — the one thing the design forbids. What crosses the wire is principal metadata, so the name says principal.
- Wire body field: `forwardedAuth` → **`forwardedPrincipal`**.
- `eveChannel({ acceptForwardedAuth })` → **`trustedForwarders`**. It's a policy fact — the set of forwarders this deployment trusts — so it takes the declarative-noun grammar of its precedents (CORS `allowedOrigins`, Express `trust proxy`, nginx `set_real_ip_from`) and matches the channel's other noun options. The predicate receives the *forwarder* (the verified transport principal), never the asserted identity.

## Dropped the response acknowledgment

The `forwardedAuth: "accepted"` ack and its best-effort orphan-cancel are gone. Forwarding only works once the receiver's author configures `trustedForwarders` — which requires a forwarding-aware eve — so a stale receiver is always an incomplete setup, surfaced during enablement as `principal_required` at per-user Connect. Permanent wire surface to defend a transitional, pre-1.0 skew window wasn't justified. Rejections are unchanged: an unconfigured receiver or refused forwarder still 403s and fails the dispatch. Docs carry an upgrade-both-sides caveat; the research plan records the ack as considered-and-dropped.

## Receiver gate consolidated

`channel/forwarded-principal.ts` now owns the whole gate: `resolveForwardedPrincipal` parses, authorizes, stamps `eve:forwarded-by`, and defaults the initiator, returning the resolved effective principals (`{ accepted, auth, initiatorAuth }`). The create handler keeps a single call; the continue route's create-only rejection moved into `parseContinueBody`.

## E2E coverage

`agent-subagents` gains a real-hop eval: `remote-loopback` is a `defineRemoteAgent({ forwardPrincipal: true })` pointing back at the same deployment (`VERCEL_URL` on Vercel, the dev server's self-published origin locally), gated by the fixture channel's `trustedForwarders`. The eval runs a two-turn parent as two *different* users, then requires the child's whoami marker to read `current=user:e2e-user-2 initiator=user:e2e-user forwarded-by=router-app` — a string only reachable if both principals cross the hop independently, the gate accepts the forwarder, and the replacement reaches the child runtime. Green on `e2e-local` and `e2e-vercel`, both model families.

Two things the eval flushed out, fixed on the branch:

- Remote dispatch forwarded a model-passed `outputSchema: {}` verbatim, flipping the child into structured-output mode and discarding its text reply; local subagent dispatch already dropped empty schemas. Both paths now share `normalizeRequestedOutputSchema` (patch changeset).
- The client's per-turn `authorization` override (which the eval's second-user turn relies on for deployed runs) had no contract test; one now pins it on both the create and stream requests, with ambient Vercel OIDC riding its own trusted-OIDC header.

The branch is rebased on main (post session-reset #1139). One design question is still open and worth a maintainer opinion: when a custom `onMessage` overrides `auth` on an accepted forwarded create, should the asserted initiator stand (initiator = session origin, `onMessage` projects only the current turn) or should the override win for both? Currently the wiring gives the former; we'd like to make it a stated rule with a test either way.

I'll update the PR description to match — shout if you'd rather keep any of the original design points in it.

PR #613, opened 2026-07-08
fix(eve): harness - add bounded durable task retries

AndrewBarba:
## Summary

- preserve structured provider metadata from streamed model errors and classify `overloaded_error` as transient
- retry classified transient failures in process with at most three fresh model/hook attempts
- rethrow other recoverable task-mode errors so Workflow retries the step from the last committed session snapshot
- keep terminal errors, exhausted transient attempts, and exhausted empty-response recovery as one structured failed task result rather than stacking retry budgets
- document both retry boundaries and include a patch changeset

## Root cause

Provider failures can arrive as plain objects after a response stream has started. Normalization retained the message but discarded the provider `type`, so `overloaded_error` missed the existing transient classifier. Reusing the first attempt's hooks could also let stale partial output settle a later retry.

The harness now retains the raw provider payload as the normalized error's cause and rebuilds attempt-local agent/hooks for each transient attempt. Completed earlier steps and tool results stay in durable session history; only the interrupted, uncommitted model call is repeated.

Task mode also needs a fallback for recoverable errors that are not classified for the in-process retry loop. Those errors now rethrow to Workflow, which retries from the committed snapshot. This is deliberately narrower than #615's blanket rethrow: errors that already consumed the three-attempt transient budget or the dedicated empty-response reissue return one failed task result and do not receive a second Workflow retry budget.

## Coverage

- stream-error normalization retains raw provider discriminators
- partial streamed overload retries with fresh hooks, preserves prior history, and emits one completed step
- three exhausted classified transient attempts produce one structured failed task result
- generic recoverable task errors rethrow without terminal or waiting events
- real Workflow retries from the committed snapshot and succeeds on attempt 2
- persistent recoverable failure exhausts Workflow's default retries on attempt 4 and invokes the delegated-parent terminal handoff once
- existing empty-response and terminal configuration behavior remains bounded

## Validation

- `pnpm test:unit` (4,538 passed, 1 skipped)
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/task-model-retry.integration.test.ts src/harness/tool-loop-stream-retry.integration.test.ts`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm guard:invariants`
- `pnpm docs:check`


AndrewBarba:
CI follow-up after the latest push:

- `test-unit` now passes; `harness/emission.ts` is 694 lines.
- The previously failing Vercel `agent-subagents` and `agent-tools-hitl` jobs now pass.
- `e2e-vercel (agent-tools)` remains red in `web-search-narration-ordering`, `web-search-parallel`, and `web-search`. The uploaded eval artifacts show every failed `web_search` action received `Insufficient credit in account` from Parallel, so this is external credential/billing state rather than the retry-policy change. No fixture or product-code change is warranted in this PR.

Failing external job: https://github.com/vercel/eve/actions/runs/28963747565/job/85941555873

AndrewBarba:
The local `agent-tools` job has now finished with the identical external failure: 18 evals passed and the same three `web_search` evals failed because Parallel returned `Insufficient credit in account`. Job: https://github.com/vercel/eve/actions/runs/28963747675/job/85941544829

ruiconti (approved):

PR #662, opened 2026-07-09
fix(eve): SSRF-guard outbound fetches through a safe-fetch wrapper

dnukumamras:
## What

Adds SSRF hardening and controls to eve's outbound HTTP. Requests whose host comes from author, tenant, or model input could otherwise be steered at an internal-only service, a cloud metadata endpoint (`169.254.169.254`), or a host that resolves into a private range — and several of these requests carry the connection's credentials.

This routes those requests through a new `#shared/safe-fetch.js` wrapper (zero new runtime deps — `node:dns` + `node:net` + global `fetch`, reusing the existing `network-address.ts` IP blocklist).

Every hop, `safeFetch` / `assertUrlSafeToFetch`:
- pins `https` (plain `http` only on loopback, or with `allowHttp` for the web-fetch tool),
- DNS-resolves the host and rejects any private / link-local / carrier-grade-NAT / reserved address (incl. cloud metadata), catching hostnames that *resolve* private, not just IP literals (including IPv4 tunneled through IPv4-compatible / mapped / SIIT / 6to4 / NAT64 IPv6 forms),
- follows redirects manually and re-validates each hop, dropping `Authorization`/`Cookie` on cross-origin redirects,
- bounds the request with a timeout and a response-size cap that stay in force across the caller's body read.

## Guarded call sites

- OpenAPI spec fetch + credentialed operation calls
- `web_fetch` tool (model-supplied URL)
- remote-agent (subagent) dispatch
- OIDC discovery + a `jwks_uri` preflight (the `jwks_uri` is read from the fetched discovery doc)
- MCP connections (URL preflight; `@ai-sdk/mcp` owns the socket)
- session callbacks (DNS check at the async POST site)

Plus CI guard **rule 36**: bare `fetch(` is forbidden in the network-facing directories (`runtime/connections/`, `execution/`, `runtime/governance/auth/`) — it must go through the wrapper.

## Behavior changes (pre-1.0)

- Session callbacks and non-loopback OpenAPI/callback URLs now require `https`.
- OpenAPI spec and operation response bodies are capped (10 MB default).

## Not included / deferred

- Connect-time IP pinning (DNS-rebinding TOCTOU): needs a custom undici dispatcher = a runtime dep, which conflicts with eve's `nitro`-only rule. The resolve-then-check preflight blocks the common cases; the sub-second rebind window is a documented residual.
- `allowLoopback` production-hardening switch is present but not yet wired to a deploy-time setting.
- Request-header sanitization and opaque errors.

## Test plan

- `packages/eve/src/shared/safe-fetch.test.ts` (new): reserved-IP + DNS-resolves-private rejection, loopback toggle, protocol pin, redirect revalidation, cross-origin credential drop, timeout (incl. across the body read), size cap.
- `network-address.ts` unit tests for the IPv6-tunneled IPv4 encodings.
- Migrated call-site tests updated for the new call convention; `pnpm typecheck`, `pnpm guard:invariants`, `pnpm docs:check`, and unit tests pass.


packages/eve/src/shared/network-address.ts
```diff
@@ -27,7 +27,14 @@ reservedRanges.addAddress("::", "ipv6"); // unspecified
 reservedRanges.addSubnet("fc00::", 7, "ipv6"); // unique-local
 reservedRanges.addSubnet("fe80::", 10, "ipv6"); // link-local
```
caleb-vercel, line 28: IPv6 blocklist looks good for the common cases (`fc00::/7`, `fe80::/10`, `::`), and dotted-decimal `::ffff:x.x.x.x` is handled via `normalizeAddress` unwrapping. A few alternate IPv6 encodings of private/metadata IPv4 can still slip through though:

| Form | Encodes | Blocked today? |
|---|---|---|
| `https://[::10.0.0.1]/` → `::a00:1` | RFC1918 private | No |
| `https://[::169.254.169.254]/` → `::a9fe:a9fe` | link-local / metadata | No |
| `::ffff:0:0a00:0001` | `10.0.0.1` (hex mapped) | No |
| `2002:c0a8:0101::` | `192.168.1.1` (6to4) | No |
| `::ffff:0:7f00:1` | `127.0.0.1` (hex mapped) | No — also bypasses `allowLoopback: false` |

Root cause: `normalizeAddress` only unwraps `::ffff:` when the tail is dotted-decimal IPv4 (`isIP(candidate) === 4`), so IPv4-compatible literals and hex-form mapped addresses stay as opaque IPv6 through `BlockList.check`.

Not a local-dev concern (`localhost` / `127.0.0.1` still work), but worth noting for dual-stack SSRF. Possible follow-ups: normalize all IPv6 literals down to embedded IPv4 before checking, and/or extend the blocklist with `2002::/16` and `64:ff9b::/96`.

caleb-vercel:
I think you'll need someone from the core Eve team to approve

packages/eve/src/shared/safe-fetch.ts
```diff
@@ -0,0 +1,417 @@
...
+        currentHeaders.delete("authorization");
+        currentHeaders.delete("cookie");
+      }
+
+      currentUrl = nextUrl;
+    }
+  } finally {
+    clearTimeout(timer);
+    if (externalSignal !== undefined) {
+      externalSignal.removeEventListener("abort", onExternalAbort);
```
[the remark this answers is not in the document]
dnukumamras: Fixed in 28b435b9. `safeFetch` now defers its timeout/abort teardown to the returned body's lifetime via a wrapping stream (cleanup on read-to-completion, error, or cancel), so `timeoutMs` and the caller `signal` keep bounding and cancelling the body read. The timer is `unref`'d for the rare never-read case. Added a test that a hung body still trips the timeout after the response is returned.

packages/eve/src/shared/network-address.ts
dnukumamras: Fixed in 28b435b9. `normalizeAddress` now parses the IPv6 to its 16 bytes and extracts the embedded IPv4 for IPv4-mapped, SIIT-translated, IPv4-compatible (`[::169.254.169.254]`), 6to4, and NAT64 forms, so they hit the IPv4 reserved ranges; `::`/`::1` are preserved. Added tests covering each encoding plus the loopback `::ffff:0:7f00:1` → 127.0.0.1 case.

PR #663, opened 2026-07-09
feat(eve): expose trigger_id and forward foreign view_submissions on the Slack channel

allenzhou101:
## Why

A Slack channel using `onInteraction` today can handle button clicks, but cannot run the standard click → modal → submit pattern:

1. The click's `trigger_id` (required for `views.open`, valid ~3 seconds) is parsed off the payload but never exposed on `SlackInteractionAction`.
2. Any `view_submission` whose `callback_id` isn't the framework's HITL freeform modal is acked and **dropped** in `handleViewSubmission` — a channel's own modal submit can never reach app code.

Concrete consumer: e0's ambient review cards (internal-agents) want 👍/👎 verdict buttons that apply immediately, then open an optional "why?" modal whose comment feeds goal reflection. The verdict works today; the modal is impossible without these two hooks.

## What

- `SlackInteractionAction.triggerId?` — populated in both `parseBlockActionsPayload` branches (raw payload `trigger_id`, shared-parser `body.triggerId`). Doc note warns about the ~3s validity, i.e. call `views.open` inline, not behind queued work.
- `SlackChannelConfig.onViewSubmission?(view: SlackViewSubmission)` — non-HITL view submissions are forwarded (under `waitUntil`, after the empty-200 ack that closes the modal) instead of dropped. `SlackViewSubmission` carries `callbackId`, `privateMetadata`, flattened `values`, the submitting `user`, and `teamId`. Slack sends no channel/thread on view submissions, so the doc directs hosts to stash routing context in `private_metadata` when opening the modal.
- HITL routing is untouched: the framework's own freeform modal still resolves parked requests first, and only foreign callback ids are forwarded.

## Verification

`pnpm typecheck` · slack channel unit suite 258/258 (5 new: trigger_id through both parser branches; foreign-callback forwarding with values/metadata/user; HITL callback never forwarded; no-handler ack). Changeset: minor.

Note: committed with `--no-verify` — the worktree's `simple-git-hooks` install fails on git-worktree `.git` files (ENOTDIR); `pnpm fmt` and `oxlint` were run manually with no diff on the changed files.

### DIFF
--- a/.changeset/slack-interaction-modals.md
+++ b/.changeset/slack-interaction-modals.md
@@ -1,5 +1,5 @@
 ---
-"eve": minor
+"eve": patch
 ---
 
 Slack channels can now open and receive their own modals. `SlackInteractionAction` gains `triggerId` (the payload's `trigger_id`, valid ~3 seconds, required for `views.open`), and a new optional `onViewSubmission` hook receives any `view_submission` whose `callback_id` is not the framework's HITL freeform modal — previously those were acked and dropped. Slack sends no channel or thread on view submissions, so stash routing context in `private_metadata` when opening the modal and read it back from the forwarded `SlackViewSubmission`.
--- a/.changeset/slack-interaction-modals.md
+++ b/.changeset/slack-interaction-modals.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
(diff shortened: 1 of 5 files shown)


pango07:
bump 👍

PR #667, opened 2026-07-09
fix(eve): forward delegated subagent stream events

### LINKED ISSUE #666: Feature request: forward task-mode subagent stream events to the parent session's stream
## Use case

A chat product where the root agent delegates a turn to a specialized subagent (task mode). The client consumes the **parent** session's stream (`GET /eve/v1/session/:id/stream`). Today the delegated child's text never streams to that client — the parent's wire shows `action.started {subagent-call}`, then silence while the parent workflow is suspended, then `action.result {subagent-result}` with the complete buffered answer. To get streaming UX, the root has to re-generate/relay the child's answer — paying a second model generation and defeating the point of delegating (e.g. binding the subagent to a different concurrency/priority lane).

## What we found in eve@0.12.0

- A delegated subagent runs as its **own workflow run** (`execution/subagent-tool.js`, `mode: "task"`, own sessionId/runId, `subagent:<uuid>` continuation token) with a parent link.
- Only two things cross the parent↔child boundary, both **durable**: HITL `input.requested` forwarding via `resumeHook`, and the final result via `notifyDelegatedParentStep`.
- Deltas are emitted per-session into that session's **own** run log (`harness/emission.js` → `createMessageAppendedEvent`), and the public stream is exactly one run's durable readable (`getSession(id).getEventStream()` → `getRun(runId).getReadable()`). There is no cross-run event bus and no aggregation by root session.

So there is no seam for an application (or a `patchedDependencies` patch) to surface child deltas on the parent stream without either per-delta durable workflow resumes (pathological) or a live side-channel that breaks the durable journal's replay/resume-cursor semantics.

## Proposal

First-class **task-mode subagent stream forwarding**: while a delegated child runs, emit its stream content into the **parent run's durable journal** as a distinguishable event kind, e.g.

```
subagent.message.appended { callId, subagentName, delta }
```

(optionally opt-in per subagent or per delegation call, e.g. `forwardStream: true`), with the same durability/replay semantics as the parent's own `message.appended` — so `?startIndex`-style resume and replay-from-journal consumers stay deterministic. The final `subagent-result` stays as-is; consumers that ignore the new event kind see today's behavior unchanged.

This would let products bind subagents to their own model/priority lanes AND keep token-streaming UX, without the root re-generating the answer.

Happy to provide more detail from our investigation if useful.

iroiro147:
## Summary

Delegated task-mode subagents now forward their child session stream events into the parent session stream as durable `subagent.event` entries. The child stream still receives its original events, and the final `subagent-result` notification shape remains unchanged.

Closes #666.

## Verification

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/workflow-steps.test.ts src/execution/workflow-entry.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/durable-session-migrations/turn-workflow.test.ts src/execution/turn-execution-cursor.test.ts src/execution/subagent-tool.test.ts src/execution/subagent-adapter.test.ts src/protocol/message.test.ts`
- `pnpm --filter eve run test:unit`
- `pnpm exec oxlint packages/eve/src/channel/types.ts packages/eve/src/protocol/message.ts packages/eve/src/execution/dispatch-runtime-actions-step.ts packages/eve/src/execution/durable-session-migrations/turn-workflow.ts packages/eve/src/execution/turn-dispatch.ts packages/eve/src/execution/turn-execution-cursor.ts packages/eve/src/execution/turn-workflow.ts packages/eve/src/execution/workflow-entry.ts packages/eve/src/execution/workflow-runtime.ts packages/eve/src/execution/workflow-steps.ts packages/eve/src/execution/subagent-stream-forwarding.ts packages/eve/src/execution/workflow-entry.test.ts packages/eve/src/execution/workflow-steps.test.ts`
- `pnpm guard:invariants`


iroiro147:
Rebased onto latest `main`. Resolved 5 conflicts:
- `dispatch-runtime-actions-step.ts`: merged `forwardedSubagentStream` into the new try/catch error-handling block
- `turn-execution-cursor.ts`: kept both the upstream `abortSignal` parameter and the PR's mutable stepInput pattern
- `workflow-entry.ts` (3 conflicts): re-added `deliveryHook`, `mode`, `parentWritable`, and `forwardedSubagentStream` to all three `runTurn` call sites that upstream had simplified
- `workflow-runtime.ts`: kept the PR's expanded workflow input with `forwardedSubagentStream`
- `workflow-steps.ts`: kept the PR's `writeSessionEvent` call with forwarded subagent stream

iroiro147:
Superseded by #1725 (rebased onto latest main; resolved upstream renames: session-delivery-hook→session-command-inbox, Runtime.run→createSession, stampMessageStreamEvent map, session.continuation.rekey). All 594 execution unit tests + tsc --noEmit green on the v2 branch.

PR #669, opened 2026-07-09
fix(eve): subagents - restrict recursive delegation to root

AndrewBarba:
## Summary

- expose the built-in `agent` delegation tool only on the root runtime node and root session
- keep declared and remote subagents available at any accumulated depth, including a declared subagent named `agent`
- reject stale recursive child calls before starting another durable workflow
- remove the obsolete `limits.maxSubagentDepth` authoring API, compiled field, inherited run limit, and durable session state
- update the subagent docs, changeset, and CI eval to describe and enforce the new boundary
- include the pending docs renderer and glTF formatting cleanup

## Why this change

The old implementation treated every delegated agent capability as the same kind of recursion. The framework injected the built-in `agent` tool into every runtime node, so a root copy, a recursively called copy, and every explicitly declared specialist could all create another copy of themselves. `maxSubagentDepth` was the only brake.

That design created several footguns:

- **Bounded depth did not bound fan-out.** Each model response could issue several `agent` calls in parallel, so a small depth limit could still produce an exponentially growing tree of durable sessions.
- **One slow leaf parked the whole ancestor chain.** Parents wait for every child in a requested batch. A stalled leaf could leave several layers of parents waiting and make nominally bounded recursion look indefinite.
- **Safety depended on propagated metadata.** Correct suppression relied on `subagentDepth` and `subagentMaxDepth` surviving session creation, durable projection, hydration, and child dispatch. Missing or stale lineage state could advertise recursion again.
- **The depth guard blocked the wrong agents.** The same generic check hid and rejected declared local and remote subagents, even though the authored local graph is finite and those calls are not implicit self-recursion.
- **Declared specialists received a surprising capability.** A narrowly scoped subagent implicitly gained the ability to clone itself, share its tools, and expand work without that capability appearing in its directory.
- **A configurable limit implied unsupported semantics.** After making recursive delegation root-only, values above one could no longer change valid behavior, leaving `maxSubagentDepth` as a misleading dead API.

## Old vs new behavior

| Caller or capability | Before | After |
| --- | --- | --- |
| Root agent | Receives the built-in `agent` tool | Receives the built-in `agent` tool |
| Child created by the built-in `agent` tool | Receives `agent` again until the configured depth cap | Never receives the recursive `agent` tool |
| Explicitly declared local subagent | Automatically receives its own built-in recursive `agent` tool | Never receives the built-in recursive tool |
| Declared nested subagent call | Hidden and rejected when accumulated session depth reaches the global cap | Remains callable; the authored filesystem graph defines the boundary |
| Remote agent call | Hidden and rejected at the same global depth cap | Remains callable regardless of local accumulated depth |
| Explicit subagent named `agent` | Could be confused conceptually with the built-in tool | Remains callable and is distinguished through the runtime registry, not by name alone |
| Stale or forced recursive action from a child | Allowed until the depth cap, then returned `SUBAGENT_DEPTH_LIMIT_REACHED` | Rejected immediately with `RECURSIVE_AGENT_ROOT_ONLY` |
| `limits.maxSubagentDepth` | Public option, defaulting to 3 and propagated through manifests and durable sessions | Removed; old authored configs and compiled manifests are rejected |
| `Workflow` | Root-only with its independent `maxSubagents` call budget | Unchanged |

## New safety boundary

Recursive delegation is now exactly one root-to-copy level and is enforced in three places:

1. **Runtime graph construction:** only the root node receives the built-in `agent` tool. Declared subagent nodes never contain it.
2. **Model advertising:** delegated sessions hide the recursive tool even when they execute a copy of the root node. Declared and remote delegation tools stay visible.
3. **Execution backstop:** stale durable state or a forced recursive action from a child is rejected before another child workflow starts.

`subagentDepth` remains as lineage and observability state, but it no longer controls capability. The structural root-only invariant is the guardrail.

## Breaking change

`limits.maxSubagentDepth` has been removed from the public authoring type, compiled manifest, inherited run limits, and durable session projection. Recursive delegation is always one root-to-copy level; declared subagent chains remain unrestricted by accumulated session depth.

The changeset is `minor` because this removes a public pre-1.0 API.

## Testing

- `pnpm test:unit` — 4,627 passed, 1 skipped across 439 files
- `pnpm typecheck`
- `pnpm lint`
- `pnpm docs:check`
- `pnpm guard:invariants`
- `pnpm --filter agent-subagents typecheck`

Coverage includes root-only model exposure, declared subagents at deep accumulated depths, an explicit subagent named `agent`, execution-time rejection of stale recursion, strict rejection of the removed authored option, and strict rejection of old compiled manifests.


ctgowrie (approved):

docs/concepts/default-harness.md
```diff
@@ -1,9 +1,9 @@
 ---
 title: "The Harness"
-description: "The out-of-the-box eve agent loop and the built-in tools every agent ships with, plus how to override or disable them."
+description: "The out-of-the-box eve agent loop and its built-in tools, including root-only and conditional capabilities."
```
ctgowrie: not sure this is important thing to highlight in desc

ctgowrie (approved):

PR #685, opened 2026-07-10
feat(eve): support file parts in tool model output

### LINKED ISSUE #456: Allow authored tools to return file/image parts to the model

Eve supports file parts in user messages, so a user can send an image/PDF/etc. to the model:
  ```ts
  [
    { type: "text", text: "Describe this image." },
    { type: "file", data: imageBytesOrDataUrl, mediaType: "image/png" }
  ]
```
But authored tool outputs are currently model-facing as text/json only. The docs for `toModelOutput` describe only:
```
  { type: "text", value: string }
  { type: "json", value: JsonValue }
```
and tool outputs must be JSON-serializable.

This makes it hard for a tool that produces an image, for example a Playwright screenshot, to pass that image as an actual vision input to the next model step. The tool can return a path, metadata, OCR, or JSON diagnostics, but the primary model cannot inspect the pixels directly.

### Proposed solution

Allow `toModelOutput` to return model content parts, or introduce a new model-facing output shape that maps to AI SDK content parts.

Example:
```ts
  toModelOutput(output) {
    return [
      {
        type: "text",
        text: "Screenshot captured. Inspect the attached image before deciding whether to upload it.",
      },
      {
        type: "file",
        mediaType: "image/png",
        data: output.imageBytes,
        filename: "contacts.png",
      },
    ];
  }
```
Alternative wrapper shape:
```ts
  toModelOutput(output) {
    return {
      type: "content",
      value: [
        { type: "text", text: "Screenshot captured." },
        { type: "file", mediaType: "image/png", data: output.imageBytes },
      ],
    };
  }
```

For larger or durable artifacts, Eve could support sandbox/artifact references that the runtime resolves into file parts before the next model call:
```ts
  {
    type: "file",
    mediaType: "image/png",
    data: {
      type: "sandbox-ref",
      path: "/workspace/screenshots/contacts.png"
    }
  }
```
This could be additive:

- keep existing text and json `toModelOutput` behavior unchanged;
- only tools that opt into rich model output return file parts;
- channel/UI output can continue receiving a JSON-safe tool result;
- model-facing output and channel-facing output can stay separate, as `toModelOutput` already only affects what the model sees.

### Alternatives considered

_No response_

BlackishGreen33:
Closes #456.

### Description

Authored tools could return text or JSON to the model, but they could not return files or images. This meant a tool could create an image, while the model still had no way to inspect it.

This PR adds a `content` output type with text and file parts. File parts use `{ type: "data", data: string }`; the string must contain base64-encoded data. Eve validates the fields and base64 syntax, then copies each part into a new, clean object before passing it to the AI SDK. URLs, data URLs, provider references, raw bytes, and sandbox references remain out of scope.

The original `execute` return value is still used for `action.result`, channels, and hooks. Only the value sent back to the model is normalized.

The tools documentation now includes an image example and explains the supported file source. The existing `agent-tools` model-output eval now returns a small PNG and checks that the model can identify it while `action.result` keeps the original tool result.

Fork pull requests cannot access the repository secrets required by the model-backed E2E, Vercel E2E, and TUI jobs. Those jobs now use the repository's existing fork guard pattern and report as skipped for fork pull requests. Same-repository pull requests, main pushes, and manual workflow runs still run the full checks.

Passed with Node.js v24.15.0:

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/tools.test.ts` (49 passed)
- `pnpm typecheck`
- `pnpm lint`
- `pnpm exec oxfmt --check` on the changed files
- `pnpm check:deps`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm test:unit`
- `pnpm test:integration`
- `pnpm build`
- `pnpm test:scenario`
- `git diff --check`

I also attempted `pnpm exec eve eval --strict dynamic-tools/model-output`. It stopped before the model or tool call because this environment does not have `AI_GATEWAY_API_KEY` or `VERCEL_OIDC_TOKEN`. The PR is draft until that eval can run with provider credentials.

A patch changeset is included for the published `eve` package.

### PR Checklist


0x962:
Tested this branch with our browser agent (Anthropic models via AI Gateway). Small images work end to end: the model sees the pixels, and the image is still visible in later turns of the same session.

One real problem: large images get destroyed before the model sees them. Compaction estimates tokens with `JSON.stringify(value).length / 4`, so a normal screenshot (~450KB+ of base64) looks like hundreds of thousands of tokens and triggers compaction, which replaces the tool result with a stub like `Tool <name> returned object(...)`. The tests don't catch this because the fixture image is only 232 chars. Two possible fixes: count file parts at a fixed cost in `estimateTokens`, or have compaction keep a filename+mediaType stub like it already does for message-level file parts.

Also small: `isBase64Data("")` returns true, so an empty file part passes validation and fails later at the provider.

Happy to help get this landed — we have a real use case waiting on it.

Issue #512, opened 2026-07-02
Decouple the agent loop from durable-execution orchestration behind a `Loop` interface


ruiconti:
The agent loop — deliver input, call the model, execute tools, wait for humans, repeat — is not written anywhere as a loop. It is smeared across four layers, each hard-coupled to Workflow DevKit primitives (`"use workflow"`, `"use step"`, `createHook`, `getWritable`, `start`):

- `workflowEntry` — pinned session driver; owns the delivery and auth hooks and dispatches turns via the `NextDriverAction` closed contract
- `turnWorkflow` — child run on latest; turn inbox hook, `TurnExecutionCursor`, in-line runtime-action waits
- `turnStep` — one `"use step"` that ferries `serializedContext` + `DurableSessionState` blobs across every boundary
- `tool-loop.ts` — 2,248 lines; model call *and* tool execution inside AI SDK `ToolLoopAgent`, parking by returning `next: null` with stop reasons encoded as flags in `session.state`

Control flow is inverted: the harness returns instead of awaiting, and higher layers decode state flags (`pendingRuntimeActionKeys`, `hasPendingInputBatch`, `hasPendingAuthorization`) to pick the next dispatch. Every stop reason grew its own hook protocol. Reading "what happens on a turn" requires holding five protocols in your head, and the loop cannot be tested as a loop without the workflow bundler.

## Proposal

Write the loop once as plain, readable async code (`runSessionLoop`, `runTurnLoop`) over two seams:

- **`Loop`** — the substrate primitives inherent to any loop, analogous to `workflow`'s `World`: `runTurn(input)` (per-turn child run on the latest deployment), `mailbox(token)` (one parking primitive with `receive`/`rekey`/`dispose`, replacing the delivery/auth/inbox/control hook zoo), and `emit(event)` (the single event-stream write path).
- **Effect interfaces** (`TurnEffects`, `SessionEffects`) — everything side-effecting (`callModel`, `executeTool`, `adapterDeliver`, `dispatchSubagent`, ...) as ordinary injected dependencies. Durability is a wiring concern invisible to the loop: the workflow composition root binds each member to a `"use step"` function, so effects are memoized and replay-safe by construction; the memory composition executes directly and makes the whole loop unit-testable without the bundler.

[...]

PR #687, opened 2026-07-10
refactor(eve): prototype decoupled loop backends

ruiconti:
## What

- Replace the generic `Loop.spawn()` proposal with eve-owned `runSession` and `runTurn` programs over a closed `LoopBackend`.
- Exercise the same programs through inline, Workflow DevKit, and Temporal adapters.
- Add shared conformance cases and a dated evidence record for the claims each adapter proves.
- Keep the prototype under `internal/testing`; no production runtime path selects these adapters.

## Why

The current loop mixes domain transitions with Workflow mechanics across `workflowEntry`, `turnWorkflow`, `turnStep`, and `tool-loop`. Those transitions cannot be exercised independently of the engine today.

The generic `spawn(): Promise<O>` proposal also hides child identity until completion, treats session and turn children alike, and has no checkpoint acknowledgement. A closed execution port makes those semantics explicit without adding Hook, Signal, Step, or Activity terms to the programs.

Inline is the non-durable reference. Workflow runs the port on eve's Workflow engine in the real local World. Temporal tests the same boundary in a second durable engine. Remaining production gates include backend selection, codecs, cancellation, recovery, delivery rekeying, and atomic terminal publication.

## Evidence

All three adapters pass the same nine-case conformance suite. The Workflow adapter runs against the real local World. The Temporal adapter runs against a real local server and Worker and records Signal, Activity, child Workflow, and acknowledgement ordering in history.

Exact receipts and limitations are recorded in the [prototype evidence](https://github.com/vercel/eve/blob/c989f91e2a6dfac8d2766e30badef08fd3ccac0c/research/loop-interface-prototype-results.md).

## Scope

This draft is stacked on `rui/loop-interface-research` and contains one commit.

The package change is five exact Temporal 1.20.1 development dependencies plus their generated lockfile graph. Published runtime code, compiler output, and backend selection are unchanged.

Relates to #512.


ruiconti:
Superseded by the restack: the loop programs, port, and Workflow adapter moved into #513 (with the simplified port surface); the inline and Temporal backends moved into #700. Branch kept for reference.

PR #700, opened 2026-07-10
feat(eve): inline and Temporal loop backends, runtime-overhead benchmark

ruiconti:
Part of #512, stacked on #513 (base: `rui/loop-interface-research`).

## What

Two pieces on top of the loop port landed in #513:

**Inline and Temporal backends** (`internal/testing/loop-prototype/inline`, `.../temporal`). The inline runtime is the reference non-durable interpreter; the Temporal runtime runs the same programs over Activities, Signals, and Child Workflows against a real local server and Worker. Both consume `LoopBackend` unchanged and pass the shared nine-test conformance suite, which is the portability claim of the research plan. This brings the exact `@temporalio` 1.20.1 dev dependencies (+1,180 lockfile lines) and the pnpm allowance for the `@swc/core` install script.

**Runtime-overhead benchmark** (`internal/loop-benchmark`, `apps/fixtures/loop-backend-benchmark`). A correctness-gated benchmark for end-to-end agent latency and the runtime layers inside one turn, comparing inline, Workflow DevKit, and Temporal without changing the authoring API. Includes the decoupled Workflow runtime extraction (`src/execution` → `src/internal/workflow`) the benchmark needs to instrument layer boundaries.

## Review notes

- The adapters and the benchmark are separate commits; the benchmark commits are unchanged from the previous revision of this branch, rebased onto the restacked base.
- Temporal scenario tests need a local server and run in CI; the inline conformance runs in the unit tier.


ruiconti:
Closing: this branch's content has been redistributed. The inline and Temporal prototype adapters moved into #513 (cherry-picked, byte-identical). The runnable runtimes were pruned of their benchmark apparatus and renamed to internal/loops with an eve dev/start --loop flag on rui/loops. The benchmark client tooling (runner, driver, report, result artifacts) is archived on rui/loop-benchmark-wip in case the latency comparison is picked back up.

PR #704, opened 2026-07-11
fix(slack): only treat the agent's own bot as the last-agent-reply boundary

### LINKED ISSUE #694: Slack threadContext "last-agent-reply" treats any bot's message as the agent's own, silently truncating context in multi-bot threads

_(the failing case below reproduces this at the unit level against `parseThreadMessage`/`applySinceBoundary`)_

### Steps to reproduce

1. Create an agent with a Slack channel using `threadContext: { since: "last-agent-reply" }`.
2. In a Slack thread where a **second Slack app/bot** also posts (any workflow bot, another agent, an integration), have: a human message → the agent's reply → more human messages → a message from the other bot → a human message.
3. @mention the agent.

Unit-level reproduction of the same defect:

```ts
// parseThreadMessage marks ANY bot-authored message as the agent's own:
const otherBot = parseThreadMessage({ bot_id: "B_OTHER_APP", ts: "4", text: "…" }, "1");
otherBot.isMe; // => true, but this is a different app

// so applySinceBoundary("last-agent-reply") cuts at the other bot's message:
// [human(1), me(2), human(3), otherBot(4), human(5)] → returns [5]
// expected (per JSDoc "last agent-authored reply") → [3, 4, 5]
```

Current: the injected thread context contains only messages after the **other bot's** last message. If the other bot posted immediately before the mention, the agent receives **zero** thread context. No warning is emitted, so the model cannot tell "nothing in the thread" from "context dropped."

Expected: per the `ThreadContextSince` JSDoc ("messages after the last **agent-authored** reply"), the boundary should be the agent's own last reply; other apps' messages are thread content, not boundaries.

Root cause: `parseThreadMessage` (`packages/eve/src/public/channels/slack/api.ts:568`) computes `isMe: botId !== undefined` — presence of *any* `bot_id` — and `applySinceBoundary` (`packages/eve/src/public/channels/slack/thread.ts:85-88`) uses `findLastIndex(m => m.isMe)` for `"last-agent-reply"`. The channel never resolves its own bot identity (no `auth.test` call; `SlackChannelCredentials` carries only the token), so `isMe` has nothing to compare against. Fix direction: resolve and cache the app's own `user_id`/`bot_id` via `auth.test` with the existing `botToken` and compare per message; a message from a different `bot_id` must not match. Threads where agents get mentioned alongside other bots are exactly where thread context matters most.

### eve version

0.16.2 (defect still present on `main`: `packages/eve/src/public/channels/slack/api.ts:568`)

### Where does the bug occur?

Production (`eve start` / deployed) and local dev — any environment using the Slack channel.

### Build and runtime logs

No error output — the truncation is silent (context is filtered after a successful `conversations.replies` fetch). That silence is part of the report: a debug log when the boundary drops messages would make this diagnosable.

serhiizghama:
Fixes #694.

`threadContext: { since: "last-agent-reply" }` was cutting the boundary at *any* bot's last message, not the agent's own. `parseThreadMessage` set `isMe: botId !== undefined`, so in a thread where a second Slack app also posts (another agent, a workflow bot, an integration), the context got sliced at that other bot instead of the agent's reply — and if the other bot posted right before the mention, the model received zero thread context, silently.

The channel never knew its own identity (`SlackChannelCredentials` only carries a token), so `isMe` had nothing to compare against. This resolves our own `bot_id` via `auth.test` — once per binding, in parallel with `conversations.replies`, best-effort — and marks a message as `isMe` only when its `bot_id` matches ours. If `auth.test` fails we fall back to treating nothing as agent-authored, so the boundary widens (full prior thread) rather than truncating.

Added a refresh test asserting only the agent's own bot counts as `isMe` while another bot in the same thread does not; reverting the comparison fails it.

### DIFF
--- a/.changeset/slack-thread-last-agent-reply-own-bot.md
+++ b/.changeset/slack-thread-last-agent-reply-own-bot.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix Slack `threadContext: { since: "last-agent-reply" }` treating every bot's message as the agent's own. In a thread where another Slack app also posts, the boundary was cut at the other bot's last message, silently dropping context (and sometimes leaving the agent with none). The channel now resolves its own bot id via `auth.test` and only marks the agent's own replies as the boundary.
--- a/packages/eve/src/public/channels/slack/api.test.ts
+++ b/packages/eve/src/public/channels/slack/api.test.ts
@@ -40,6 +40,11 @@ function buildFetchMock(): { fetch: ReturnType<typeof vi.fn>; calls: FetchCall[]
         { headers: { "content-type": "application/json" } },
       );
     }
+    if (url === "https://slack.com/api/auth.test") {
+      return new Response(JSON.stringify({ ok: true, bot_id: "B01", user_id: "U_BOT" }), {
+        headers: { "content-type": "application/json" },
+      });
+    }
     if (url === "https://slack.com/api/conversations.replies") {
       return new Response(
         JSON.stringify({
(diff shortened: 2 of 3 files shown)


serhiizghama:
Closing, #694 got resolved separately.

PR #706, opened 2026-07-11
fix(slack): collect thread attachments across recent messages instead of only the newest

### LINKED ISSUE #705: Slack channel: images/files posted earlier in a thread are silently dropped — attachment lookback only ever inspects one message

https://github.com/shiminshen/eve/tree/repro/slack-thread-attachment-lookback — one commit adding a failing unit test to the package's own `attachments.test.ts`, following the file's existing fixtures. No Slack workspace needed:

```
pnpm install
cd packages/eve && pnpm run test:unit -- src/public/channels/slack/attachments.test.ts
# ✗ looks past a newer text-only message to find files earlier in the thread
#   AssertionError: expected [] to have a length of 1
```

### Steps to reproduce

Slack-level symptom:

1. In a Slack thread, post an image (message A).
2. Post a follow-up text-only message in the same thread (message B).
3. Mention the agent in the thread with a question about the image (the mention itself has no attachment).
4. The turn reaches the model with **zero file parts** — the agent answers as if no image was ever shared.

The thread-lookback loop in `collectInboundFileParts` (`packages/eve/src/public/channels/slack/attachments.ts:103-113`) is written as a backwards scan over `thread.recentMessages`, but the loop body ends with:

```ts
if (parts.length > 0) return parts;
return [];
```

— an unconditional return on the **first** non-bot message inspected, so the loop never advances. Consequences:

- if the newest non-bot message has no files, all earlier attachments in the thread are dropped;
- files spread across multiple thread messages are never combined;
- the drop is silent — no `log.warn`, unlike the policy-violation paths in `collectSlackFileParts`.

Expected: the lookback collects files across recent thread messages (newest first, bounded — we run a cap of 10 in production via a patch-package patch). If "only the most recent message that has files" is the intended semantics instead, the dangling loop shape and the silent drop still look unintended.

Happy to send a PR (fix + changeset, DCO-signed; the failing test is already on the repro branch) once maintainers confirm which semantics you want.

### eve version

eve@0.22.3 (verified still present in the 0.22.5 dist and on `main` at HEAD)

### Environment

```
OS: macOS 15 (arm64) / Vercel Fluid Compute
Node: v24.16.0
Package manager: npm (consumer) / pnpm@11.7.0 (this repo)
```

Slack channel via `defineSlackChannel` with `threadContext: { since: "thread-root" }`.

### Where does the bug occur?

Local dev (`eve dev`), Production (deployed)

### Build and runtime logs

No error or warning is emitted — this code path silently returns `[]`, which is part of the report. With `EVE_LOG_LEVEL=debug` the turn proceeds normally, just with no file parts attached.

### Additional context

Found running an internal knowledge agent on the Slack channel in production. #298's attachment-lookback rework ("reuse fetched thread replies for attachment lookback") kept this single-message behavior. Related but distinct: #399 / #276 are about base64 file-part failures, not about which thread messages are scanned.

shiminshen:
### Description

Fixes #705.

`collectInboundFileParts` scans `thread.recentMessages` backwards, but the loop body ends with an unconditional return on the first non-bot message it inspects — the loop never advances. When a user posts an image in a thread and then mentions the bot in a text-only follow-up, the newest non-bot message has no files, so everything earlier is silently dropped and the model sees zero file parts. This is the exact case the function's own doc comment says the lookback exists to cover ("a user dropped a file in the thread first, then mentioned the bot in a follow-up").

The fix collects attachments across non-bot messages, newest first, capped at 10 file parts to keep model input bounded on long threads. If you'd rather keep single-message semantics ("most recent message that has files") instead of bounded accumulation, I'm happy to adjust — flagged the same question in #705.

We've been running this change in production for an internal Slack knowledge agent (as a patch-package patch of the dist) since eve 0.17.

- `pnpm run test:unit` in `packages/eve` — 445 files, 4683 passed (adds a regression test that fails on `main`, plus accumulation and cap cases)
- `pnpm run typecheck` — clean
- `pnpm lint` — clean
- Production use via the equivalent dist patch in our Slack deployment

### PR Checklist

- [x] I linked an issue with prior discussion confirming this change is wanted (#705 — filed with a failing-test reproduction branch; no maintainer reply yet, the one open design question is flagged there and above)
- [x] I added tests and documentation where relevant (JSDoc updated; no public API change)

### DIFF
--- a/.changeset/slack-thread-attachment-lookback.md
+++ b/.changeset/slack-thread-attachment-lookback.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix the Slack thread attachment lookback only ever inspecting the newest non-bot message. When a user posted an image in a thread and then mentioned the agent in a text-only follow-up, the image was silently dropped and the model saw no file parts. The lookback now collects attachments from recent non-bot messages across the thread, newest first, capped at 10 file parts.
--- a/packages/eve/src/public/channels/slack/attachments.test.ts
+++ b/packages/eve/src/public/channels/slack/attachments.test.ts
@@ -360,6 +360,38 @@ describe("collectInboundFileParts", () => {
     expect((parts[0]!.data as URL).href).toBe("https://files.slack.com/a/b/from-user.csv");
   });
 
+  it("looks past a newer text-only message to find files earlier in the thread", async () => {
+    const refresh = vi.fn().mockResolvedValue(undefined);
+    const thread = makeSlackThread({
+      refresh,
+      recentMessages: [
+        {
+          isMe: false,
+          raw: {
+            files: [
+              {
+                id: "F500",
+                name: "earlier.png",
+                mimetype: "image/png",
+                url_private: "https://files.slack.com/a/b/earlier.png",
+              },
+            ],
+          },
+        },
+        { isMe: false, raw: {} },
+      ],
+    });
+
+    const parts = await collectInboundFileParts({
+      mention: emptyMention,
+      thread,
+      policy: DEFAULT_UPLOAD_POLICY,
+    });
+
+    expect(parts).toHaveLength(1);
+    expect((parts[0]!.data as URL).href).toBe("https://files.slack.com/a/b/earlier.png");
+  });
+
   it.each([
     ["'disabled' literal", DISABLED_POLICY],
     ["maxBytes: 0", ZERO_BYTES_POLICY],
--- a/packages/eve/src/public/channels/slack/attachments.test.ts
+++ b/packages/eve/src/public/channels/slack/attachments.test.ts
@@ -392,6 +392,76 @@ describe("collectInboundFileParts", () => {
     expect((parts[0]!.data as URL).href).toBe("https://files.slack.com/a/b/earlier.png");
   });
 
+  it("collects files from multiple thread messages, newest first", async () => {
+    const refresh = vi.fn().mockResolvedValue(undefined);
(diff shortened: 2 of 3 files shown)


shiminshen:
One note on the `THREAD_LOOKBACK_MAX_FILES = 10` bound: it's intentionally a hardcoded backstop to keep this diff minimal. If you'd prefer it configurable, I'm happy to lift it into `UploadPolicy` (e.g. an optional `maxFiles`, defaulting to 10, alongside `maxBytes`) — either in this PR or as a follow-up, whichever you prefer.

shiminshen:
Rebased onto current `main`. Worth noting what changed underneath: `main` has since improved this lookback to skip messages the app itself authored (the `isMe` classification), which fixes the case where the bot's own reply shadowed a user's file. The underlying bug in #705 is still there though — the loop stops at the first non-bot message and returns unconditionally, so a file posted two messages back is still dropped when the message in between has no attachments. This PR still applies; it changes that early return into a bounded walk across the thread.

The only conflict was the doc comment, and I merged both descriptions — the wording from `main` about `isMe` is more precise than mine, so I kept it.

Verified on the rebased branch: `packages/eve` typecheck clean, unit suite green at 6845 passed / 1 skipped across 651 files. DCO sign-offs are intact on both commits.

The open question from the original description still stands, if you have a view: `THREAD_LOOKBACK_MAX_FILES` is a hardcoded backstop of 10 to keep the diff small. Happy to lift it into `UploadPolicy` if you'd rather it be configurable.

PR #709, opened 2026-07-11
fix(eve): preserve schedule delivery provenance

### LINKED ISSUE #708: Stamp schedule-dispatched runs with a $eve.schedule attribution attribute
## Problem

The only durable trace a schedule execution leaves is `$eve.trigger = 'schedule'` on workflow runs — nothing records **which** schedule fired:

- run history cannot be aggregated per schedule ("did `daily` run last night? what's its success rate?" is unanswerable);
- handler-form schedules that dispatch through `receive(channel, ...)` produce sessions whose `$eve.trigger` is the channel kind, not `schedule`, so filtering by trigger misses them entirely — and that dispatch path is a first-class pattern (#104);
- observability tooling is left with trigger-based heuristics and cannot attribute runs precisely.

## Proposal

`ScheduleDispatcher` is the single choke point for all schedule dispatch (markdown via `runMarkdown`; handler-form `receive()` runs in its context too — see the analysis in #387). Stamp the runs it dispatches with:

- `$eve.schedule` = the schedule id (optionally `$eve.schedule_cron` as well);
- sessions spawned in a schedule context carry the attribute, propagated the same way as `$eve.root` / `$eve.parent`;
- `$eve.trigger` semantics stay unchanged — channel-dispatched sessions keep the channel kind; this only adds provenance.

Pure metadata addition, no behavior change, backward compatible. With it, `attributes ->> '$eve.schedule' = <id>` yields any schedule's complete run history regardless of how the handler dispatched.

warmwind:
### Summary

Closes #708. A production `@v` schedule launched background work under its owner's Slack credentials and posted the normal launch acknowledgement. The schedule was running eve 0.45.0, which includes the schedule exception from #2406, but that exception inferred provenance from the active app principal; replacing it with user credentials made the turn look user-initiated.

Schedule dispatch now stamps provenance independently from auth. Sessions created inside the dispatch retain the schedule id while continuing to use the authored user principal for authorization. Background-task delivery resolves through one policy shared by prompting and empty-response recovery, so scheduled launches stay silent, pending wakes stay silent, settled wakes report, and ordinary user launches still acknowledge.

Before → after: schedule origin was inferred from `AuthKey` → schedule origin is durable session metadata; instruction selection and silent-retry eligibility were separate predicates → one delivery policy owns both.

`$eve.schedule` also exposes the same provenance for workflow attribution while `$eve.trigger` remains the target channel. This preserves the original purpose and Fable-authored lineage of #709.

### Scope

This covers sessions created by a schedule dispatch, including handler sessions started asynchronously through `waitUntil` with user credentials. Delivering a schedule turn into an already-existing conversation still needs per-delivery provenance and a session-inbox wire migration; this PR does not overload auth or initiator identity to approximate that case.

### Validation

Reproduced the custom-user-auth schedule path with a deterministic background-tool eval. The launch and pending wake complete with `message: null`, and only the settled wake reports. The complete unit tier passes with 7,211 tests; the focused eval passes all 14 gates.

### Checklist

### Diff size

**Docs** — 4 files · `+10 / -0`

Documents custom-auth schedule semantics, background-launch delivery, workflow attribution, and the patch release note.

**Implementation** — 6 files · `+93 / -36`

Adds durable schedule provenance and a pure delivery-policy module. Removing the nested instruction/phase predicates accounts for most deletions.

**Tests** — 9 files · `+390 / -3`

Unit coverage pins provenance, workflow attributes, policy precedence, retry behavior, and later human turns. The deterministic fixture contributes most lines because it exercises schedule dispatch, custom auth, background execution, pending silence, and settled reporting end to end.


ruiconti:
Superseded by #2648. The replacement carries this PR’s schedule-provenance design forward on a clean current-main branch, adds the background-task delivery fix and deterministic regression coverage, and credits Oscar Jiang and Claude Fable 5 as co-authors.

Issue #727, opened 2026-07-12
Expose a finite session event snapshot API for server-side projections


thomaskwon416:
## Problem

Eve hooks are documented as the extension point for persisting session/message events to an application database, but the current public APIs do not expose a finite, cursor-bearing server-side read of a session's durable event log.

In `eve@0.22.5`:

- `message.completed` / `session.waiting` hooks receive the durable event and `HookContext`, but neither carries the event's remote `streamIndex`.
- `ClientSession.stream({ startIndex })` is a live async stream. It correctly remains open after `session.waiting` because the durable parent accepts future turns.
- Therefore an awaited boundary hook cannot notify an application that then opens `session.stream()` and waits to project the suffix: the callback never returns, the hook never settles, and the next queued turn cannot start.
- Browser `onEvent` / `onFinish` persistence is useful for UI state, but cannot be the correctness boundary for a server-owned product projection when the browser refreshes or disappears.

Observed local timeline:

1. `message.completed` is visible to the browser.
2. A downstream model/tool fetch times out and Eve correctly retries the durable step.
3. Eve emits `turn.completed` and `session.waiting` on the retry.
4. The application callback opens `session.stream()` from its stored cursor.
5. The stream waits for future parent events, so the callback and `session.waiting` hook remain pending; the product turn stays `running`, and the queued follow-up remains unattached.

This is not a request to make live streams close at `session.waiting`; that would break continuation semantics.

## Requested public capability

Please expose a finite snapshot/replay operation on `ClientSession` (name flexible), for example:

```ts
const { events, state } = await session.snapshot({ startIndex });
// events: durable events from startIndex through a tail fixed at invocation
// state.streamIndex: the exclusive cursor immediately after that captured tail
```

Required semantics:

- Captures a finite tail at invocation and never waits for future events.
- Returns ordered public `HandleMessageStreamEvent[]` plus the resulting `SessionState`/exclusive cursor.
- Uses the existing Eve session authorization/continuation contract.
- Works while the durable parent is `running`, `waiting`, completed, or failed.
- Empty suffix is a successful idempotent result with an unchanged cursor.
- Concurrent appends after the captured tail are excluded and available to the next snapshot.
- Does not require consumers to call raw HTTP routes, parse NDJSON, count chunks, or mirror Eve internals.

An equivalent hook API that supplies the exact durable cursor after the current event could also solve incremental projection, but a finite snapshot is more generally useful for bounded recovery and idempotent server-side projection.

## Suggested coverage

[...]

Issue #724, opened 2026-07-12
Expose a typed retryable readiness error for route auth


thomaskwon416:
## Problem

`eveChannel()` authorizers sometimes need to distinguish an authenticated caller whose authorization dependency is not queryable *yet* from an unauthenticated or forbidden caller.

A concrete case is owner-authorized initial session streaming:

1. `POST /eve/v1/session` authenticates and returns a new session ID.
2. The client immediately opens `GET /eve/v1/session/:sessionId/stream`.
3. An application-owned, awaited `session.started` projection may not yet have made the session-to-owner mapping queryable.
4. The caller is valid, but returning `403` permanently fails the native client. The native stream opener already treats `425` as bounded/retryable.

Today the public `AuthFn` contract only documents principal/null plus `UnauthenticatedError` and `ForbiddenError`. `routeAuth` internally preserves errors carrying a `Response`, but application code must rely on that private duck-typed behavior to emit `425`.

## Proposed invariant

**An authenticated route dependency that is provably pending may fail with one public, typed retryable readiness error. Unauthorized callers must continue to receive 401/403. Application code must not copy Eve routes, proxy streams, or implement retry sleeps.**

## Proposed API

Export a public typed error from `eve/channels/auth`, name to maintainer preference, for example:

```ts
throw new SessionNotReadyError();
```

The error should own a framework-shaped response:

- HTTP `425 Too Early`
- `cache-control: no-store`
- stable machine code such as `session_not_ready`
- optional bounded `Retry-After`, if desired

## Scope

- Public error class and types.
- `routeAuth` tests proving exact response preservation and auth-walk termination.
- Native client coverage confirming 425 retry remains bounded and abort-aware.
- Auth guide documentation and patch changeset.

No new transport or retry implementation is required; the native client already retries 425.

## Alternatives considered

Exposing target-session initiator/current auth at route time would be stronger but substantially larger: the current channel session handle exposes only ID, continuation token, and event stream, while `session.started` does not expose `SessionAuthContext`. A typed readiness error is the smallest public contract that closes the gap without leaking session auth or adding a durable metadata index.

## Consumer

FreshlakeAI/livetaxbot TGL-493/TGL-494. The consumer has adversarial coverage ensuring 425 is returned only for an exact owned, running, unattached reservation; missing, stale, cross-user, cross-chat, or differently attached records remain 403.

thomaskwon416:
[...]

PR #734, opened 2026-07-13
Add finite session lifecycle primitives

thomaskwon416:
## What changed

- add a public finite `ClientSession.snapshot()` API that returns ordered events plus the resulting session state/cursor
- add a typed public `SessionNotReadyError` (HTTP 425, no-store) and bounded client retry for a just-created session
- make continuation sends resume-only so they never create a replacement parent
- expose the raw client continuation token to session handles and lifecycle hooks without application namespace parsing

## Why

Server-authoritative applications need to project a finite event suffix and authorize the short create-to-addressable window without copying Eve routes, adopting a live stream inside hooks, or keeping duplicate continuation-token authorities. The continuation client must also fail rather than silently fork a new parent.

Addresses #724 and #727.

## Validation

- focused Eve unit suites: 88/88
- hook lifecycle integration: 1/1 on Windows
- TypeScript: green on Node 24 under WSL
- production build and package: green
- `git diff --check`: green

The broader Windows unit run passes changed suites; unrelated existing path/bin tests remain platform-specific. Direct Linux integration config currently hits the repository's existing `#internal/workflow-bundle/workflow-builders.js` resolution failure, while the focused Windows integration tracer passes.

### DIFF
--- a/.changeset/clean-finite-sessions.md
+++ b/.changeset/clean-finite-sessions.md
@@ -4,4 +4,6 @@
 
 Add a finite `ClientSession.snapshot()` API for durable event reconciliation
 and a public `SessionNotReadyError` auth contract that maps transient session
-readiness to a stable, non-cacheable HTTP 425 response.
+readiness to a stable, non-cacheable HTTP 425 response. Eve continuation
+requests are now resume-only and bounded-retry readiness races instead of
+silently starting a duplicate session.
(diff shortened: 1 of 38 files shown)


De-ASI-INTERFACE:
Richard patterson

On Sun, Jul 12, 2026 at 11:37 PM thomaskwon416 <
***@***.***>
wrote:

PR #611, opened 2026-07-08
fix(eve): protect nitro-referenced dev snapshots

iroiro147:
## Summary

- protect stale dev-runtime snapshots that are still referenced by generated `.eve/nitro` artifacts
- keep existing pointer and active workflow-data protections intact while allowing unrelated stale snapshots to prune
- add regression coverage for an aggressive prune that preserves a Nitro-referenced snapshot

Refs #534.

## Tests

- `pnpm install` (completed; hook installer printed an `ENOTDIR` warning because this checkout is a Git worktree)
- `pnpm --filter eve run test:integration -- src/internal/nitro/dev-runtime-artifacts.integration.test.ts -t 'generated Nitro files'` (50 files passed, 386 tests passed)
- `pnpm --filter eve run typecheck`
- `pnpm guard:invariants`
- `pnpm exec oxfmt --check packages/eve/src/internal/nitro/dev-runtime-artifacts.ts packages/eve/src/internal/nitro/dev-runtime-artifacts.integration.test.ts .changeset/dev-runtime-nitro-prune-protection.md`
- `pnpm exec oxlint packages/eve/src/internal/nitro/dev-runtime-artifacts.ts packages/eve/src/internal/nitro/dev-runtime-artifacts.integration.test.ts`
- `git diff --check`
- `pnpm changeset status --since origin/main`

PR #763, opened 2026-07-14
fix(eve): retain Nitro-referenced dev snapshots

### LINKED ISSUE #534: `eve dev` nitro modules pin the boot-time dev-runtime snapshot; pruner GCs it → every route 500s with UNRESOLVED_IMPORT
### eve version

`eve@0.19.0`

### Node.js version

`v24.x` (macOS, bun workspaces monorepo)

### Where does the bug occur?

Local dev (`eve dev --no-ui`)

### Describe the bug

The nitro dev server's generated modules pin the **absolute path of the boot-time dev-runtime snapshot**, but the background snapshot pruner does not protect that snapshot. Once enough rebuilds happen (newer snapshots push it out of the `retainCount: 5` / 15-minute window), the pruner deletes it and **every HTTP route on the dev server fails with a 500 build error** until the process is restarted. A restart only helps for another ~15 minutes if rebuild traffic continues.

Two on-disk/virtual references pin the boot snapshot for the life of the process:

1. `#nitro/virtual/plugins` imports `…/.eve/dev-runtime/snapshots/<boot-id>/source/<app>/.eve/compile/compiled-artifacts-instrumentation.mjs`
2. The generated `.eve/nitro/workflow/steps.mjs` imports `…/snapshots/<boot-id>/source/<app>/.eve/compile/compiled-artifacts-bootstrap.mjs`

Neither is regenerated when a rebuild publishes a new snapshot and moves `.eve/dev-runtime/current.json`, and neither is covered by the pruner's protected-path collection. `pruneDevelopmentRuntimeArtifactsSnapshots` protects the pointer's current snapshot and snapshot refs found in `.workflow-data` (the #172 → #181/#187 fix), but not refs in the generated `.eve/nitro/**` files. This is the same bug class as #172, resurfacing through a reference the fix didn't cover.

Every request then fails with:

```
Build failed with 2 errors:

[UNRESOLVED_IMPORT] Could not resolve '…/.eve/dev-runtime/snapshots/<boot-id>/source/apps/agent/.eve/compile/compiled-artifacts-instrumentation.mjs' in #nitro/virtual/plugins
  Help: '#nitro/virtual/plugins' is imported by:
        - #nitro/virtual/app
        - …/nitro/dist/runtime/internal/app.mjs
        - …/nitro/dist/presets/_nitro/runtime/nitro-dev.mjs

[UNLOADABLE_DEPENDENCY] Could not load .eve/dev-runtime/snapshots/<boot-id>/source/apps/agent/.eve/compile/compiled-artifacts-bootstrap.mjs
  ╭─[ .eve/nitro/workflow/steps.mjs:3:8 ]
  No such file or directory (os error 2)
```

This takes down custom channel webhook routes, sessions, everything — the authored code never runs, so the 500 is easy to misattribute to application logic (in our case we initially chased a webhook handler bug).

### Reproduction

1. `eve dev --no-ui`, note the snapshot id referenced by `.eve/nitro/workflow/steps.mjs` (equals the boot snapshot).
2. Cause rebuilds so newer snapshots are published and `current.json` moves on (source touches or session-client `POST /eve/v1/dev/runtime-artifacts/rebuild?force=1` — anything that calls `prepareApplicationHost`).
3. Wait until the boot snapshot ages out of the 15-minute window / newest-5 set; the background pruner removes it.
4. Any request now 500s with the build error above.

Deterministic shortcut: call `pruneDevelopmentRuntimeArtifactsSnapshots({ appRoot, retainCount: 0, recentWindowMs: 0, now: Date.now() })` after step 2 — the boot snapshot is deleted while still referenced.

### Suggested fix (verified locally)

Extend the pruner's protected-path scan to the generated nitro directory. We're running this as a local patch and it resolves the issue (pinned snapshot survives an aggressive prune, routes stay up):

```diff
 async function collectWorkflowDataSnapshotPaths(e) {
   let t = join(e.appRoot, `.workflow-data`), n = new Set;
-  return await collectSnapshotPathsFromDirectory({ directory: t, snapshotPaths: n, snapshotsDirectory: e.snapshotsDirectory }), [...n];
+  return await collectSnapshotPathsFromDirectory({ directory: t, snapshotPaths: n, snapshotsDirectory: e.snapshotsDirectory }),
+    await collectSnapshotPathsFromDirectory({ directory: join(e.appRoot, `.eve`, `nitro`), snapshotPaths: n, snapshotsDirectory: e.snapshotsDirectory }),
+    [...n];
 }
```

(Alternatively/in addition: re-point the generated `.eve/nitro/**` modules and the `#nitro/virtual/plugins` import at the new snapshot whenever one is published, so the dev server doesn't depend on the boot snapshot at all.)

### Related observation

While the dev server sits in the failed-build state, every incoming request (including health checks) retries the full bundle, spamming `PLUGIN_TIMINGS` warnings for `eve-authored-relative-extension-resolver` every few seconds and burning CPU/memory — which accelerates the snapshot churn that triggers the bug in the first place.

Refs: #172, #181, #187

ruiconti:
## What

- treat generated `.eve/nitro` files as live snapshot references
- keep the existing bounded, best-effort file scan
- prove that pruning removes an unreferenced snapshot but preserves both the active pointer and the Nitro boot revision

## Failure boundary

The dev snapshot pruner protected the active pointer and `.workflow-data` references, but generated Nitro modules can still import the snapshot used at server boot. Once that revision aged out, cleanup deleted code the running server still referenced and later requests failed with `Cannot find module`.

Snapshot reachability now includes `.eve/nitro` without changing retention policy or the 1 MiB per-file scan bound.

Closes #534. Prior implementation: #611.

### DIFF
--- a/.changeset/protect-nitro-dev-snapshots.md
+++ b/.changeset/protect-nitro-dev-snapshots.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+The development runtime now keeps snapshots referenced by generated Nitro files, preventing active development servers from losing their boot-time modules during snapshot cleanup.
--- a/packages/eve/src/internal/nitro/dev-runtime-artifacts.integration.test.ts
+++ b/packages/eve/src/internal/nitro/dev-runtime-artifacts.integration.test.ts
@@ -423,6 +423,58 @@ describe("development runtime artifact snapshots", () => {
     expect(existsSync(staleSnapshotRoot)).toBe(false);
   });
 
+  it("preserves snapshots referenced by generated Nitro files", async () => {
+    const appRoot = await createScratchDirectory("eve-dev-runtime-artifacts-prune-nitro-");
+    const snapshotsRoot = join(appRoot, ".eve", "dev-runtime", "snapshots");
+    const activeSnapshotRoot = join(snapshotsRoot, "active");
+    const nitroReferencedSnapshotRoot = join(snapshotsRoot, "nitro-referenced");
+    const staleSnapshotRoot = join(snapshotsRoot, "stale");
+    const oldSnapshotTime = new Date(1_000);
+
+    for (const snapshotRoot of [
+      activeSnapshotRoot,
+      nitroReferencedSnapshotRoot,
+      staleSnapshotRoot,
+    ]) {
+      await mkdir(snapshotRoot, { recursive: true });
+      await writeFile(join(snapshotRoot, "marker.txt"), snapshotRoot);
+      await utimes(snapshotRoot, oldSnapshotTime, oldSnapshotTime);
+    }
+
+    await activateDevelopmentRuntimeArtifactsSnapshot({
+      appRoot,
+      snapshot: {
+        runtimeAppRoot: join(activeSnapshotRoot, "source", "app"),
+        snapshotRoot: activeSnapshotRoot,
+        snapshotSourceRoot: join(activeSnapshotRoot, "source"),
+      },
+    });
+    await mkdir(join(appRoot, ".eve", "nitro", "workflow"), { recursive: true });
+    await writeFile(
+      join(appRoot, ".eve", "nitro", "workflow", "steps.mjs"),
+      `import "${join(
+        nitroReferencedSnapshotRoot,
+        "source",
+        "app",
+        ".eve",
+        "compile",
+        "compiled-artifacts-bootstrap.mjs",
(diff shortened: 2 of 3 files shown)


ctgowrie (commented):
Believe this is fixed by https://github.com/vercel/eve/pull/722?

ruiconti:
It is fixed by #722.

PR #597, opened 2026-07-08
fix(eve): merge system instructions before model calls

iroiro147:
## Summary
- merge assembled system instructions into a single system message before constructing the model call
- keep session history, dynamic instruction lifecycle, and array-form user content unchanged
- preserve Anthropic direct-cache provider options by applying the cache marker before merging

Fixes #596.

## Tests
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/tool-loop.test.ts` (148 passed)
- `pnpm --filter eve run typecheck`
- `pnpm --filter eve exec oxlint src/harness/tool-loop.ts src/harness/tool-loop.test.ts src/harness/instrumentation-runtime-context.ts src/public/instrumentation/index.ts`
- `pnpm exec oxfmt --check packages/eve/src/harness/tool-loop.ts packages/eve/src/harness/tool-loop.test.ts packages/eve/src/harness/instrumentation-runtime-context.ts packages/eve/src/public/instrumentation/index.ts .changeset/merge-system-instructions.md`
- `pnpm guard:invariants`
- `pnpm changeset status --since origin/main`
- `git diff --check`

PR #765, opened 2026-07-14
fix(eve): merge system instructions

### LINKED ISSUE #596: Multi-system-message prompt assembly breaks Fireworks (opaque 500), merge system messages at the model-call boundary
### TL;DR

**Current output** (what reaches the provider when an agent has `instructions` + one `defineInstructions`/`defineDynamic`, and the user message carries a file part):

```jsonc
[
  { "role": "system", "content": "<agent instructions>" },
  { "role": "system", "content": "<dynamic instructions>" },
  {
    "role": "user",
    "content": [
      { "type": "text", "text": "What's my P&L for 2025?" },
      { "type": "file", "filename": "receipt.jpg", "mediaType": "image/jpeg", "...": "..." }
    ]
  }
]
// Fireworks → HTTP 500 (opaque), turn parks with MODEL_CALL_FAILED
```

**Expected output** (system messages merged at the model-call boundary; history/lifecycle unchanged):

```jsonc
[
  { "role": "system", "content": "<agent instructions>\n\n<dynamic instructions>" },
  {
    "role": "user",
    "content": [
      { "type": "text", "text": "What's my P&L for 2025?" },
      { "type": "file", "filename": "receipt.jpg", "mediaType": "image/jpeg", "...": "..." }
    ]
  }
]
// Fireworks → 200
```

eve assembles the model-call `instructions` as an array of system messages: `prepareModelCallInput` (harness/tool-loop) builds `[agent.system, ...dynamic instruction messages, skill announcements, delivery note]`. Any agent that combines `defineAgent({ instructions })` with one `defineInstructions`/`defineDynamic` therefore sends **two or more system-role messages** on every call.

Fireworks' OpenAI-compatible chat completions endpoint (observed on `accounts/fireworks/models/qwen3p7-plus`, eve 0.21.1 via `@ai-sdk/fireworks`) returns an opaque HTTP 500 when a request combines **2+ system messages with array-form user content**. Each half of the pattern works alone: one system message + array user content is fine, multiple system messages + plain string user content is fine. The intersection fails deterministically. We've verified this is a Fireworks server-side bug and reported it to them separately — but eve emits the triggering pattern by design, and nothing about the semantics requires N separate system messages.

This is about to hit more people: eve 0.21's native attachment lane produces array-form user content whenever a message carries file parts, so "agent with dynamic instructions + user sends an attachment" = every turn fails with `MODEL_CALL_FAILED` on Fireworks-hosted models.

**Repro sketch:**
1. `defineAgent({ model: fireworks('accounts/fireworks/models/qwen3p7-plus'), instructions })` plus any `defineInstructions`/`defineDynamic` (→ 2 system messages).
2. Send a turn whose user message has array content (e.g. a file part).
3. Provider returns 500; turn parks with `MODEL_CALL_FAILED`.

**Workaround we're running:** a `LanguageModelMiddleware` whose `transformParams` concatenates all system messages into one (`join('\n\n')`). No behavioral difference observed — which supports merging being semantically safe.

### Proposed solution

Merge the assembled system messages into a **single** system message at the model-call boundary (`prepareModelCallInput`), keeping them separate in session history and in the dynamic-instruction lifecycle. For OpenAI-compatible APIs the concatenation is semantically equivalent; Anthropic handles `system` separately anyway. The `applySystemCacheBreakpoint` logic would apply to the merged message. If unconditional merging feels too opinionated, an agent-level opt-in (e.g. `mergeSystemMessages: true`) would do — but defensive-by-default seems right given providers demonstrably choke on the pattern.

ruiconti:
## What

- collapse base, retry, and dynamic system instructions into one AI SDK system message
- preserve instruction order and provider options
- retain the Anthropic system cache breakpoint on the merged message
- expose the normalized shape to instrumentation

## Failure boundary

A base agent prompt plus a durable or dynamic system instruction reached the provider as two system messages. Fireworks rejects that request shape, so otherwise valid turns failed before generation. Array-form user content made the reported case easy to reproduce but is not the source of the invalid system-message count.

The Anthropic marker is applied before the merge and retained on the combined message. A direct-provider regression test pins that path. Plain single-string prompts retain their existing shape.

Closes #596. Prior implementation: #597.

### DIFF
--- a/.changeset/merge-system-instructions.md
+++ b/.changeset/merge-system-instructions.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Model calls now merge multiple system instructions into one message, avoiding provider failures when dynamic instructions are combined with array-form user content.
--- a/packages/eve/src/harness/instrumentation-runtime-context.ts
+++ b/packages/eve/src/harness/instrumentation-runtime-context.ts
@@ -32,7 +32,7 @@ export interface BuildTelemetryRuntimeContextInput {
   readonly emissionState: HarnessEmissionState;
   readonly environment: string;
   readonly modelInput: {
-    readonly instructions: string | readonly SystemModelMessage[] | undefined;
+    readonly instructions: string | SystemModelMessage | undefined;
     readonly messages: readonly ModelMessage[];
   };
   readonly session: HarnessSession;
--- a/packages/eve/src/harness/tool-loop.test.ts
+++ b/packages/eve/src/harness/tool-loop.test.ts
@@ -7,6 +7,7 @@ import {
   ToolLoopAgent,
   type UserContent,
 } from "ai";
+import { MockLanguageModelV3 } from "ai/test";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
 import { ContextContainer, contextStorage } from "#context/container.js";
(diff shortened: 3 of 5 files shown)


ctgowrie (commented):
Should this be upstreamed into ai sdk?

AndrewBarba (approved):

iroiro147:
Thanks for landing this @ruiconti.

One note on attribution: #597 was open from 8 Jul (I closed it on 30 Jul) carrying the same fix — collapsing base, retry, and dynamic system instructions into a single system message, since providers that reject a multi-system-message request shape were failing otherwise valid turns. GitHub cross-references the two PRs. If #597 informed this work, a credit line would be appreciated; if you arrived at it independently, no concern at all.

PR #770, opened 2026-07-14
fix: Escape unsafe JS characters in channel routes

odaysec:
Updated dispatchChannelWebSocketRequest and dispatchChannelRequest to escape unsafe JavaScript code characters in routeKey.

### Description
Using string concatenation to construct JavaScript code can be error-prone, or in the worst case, enable code injection if an input is constructed by an attacker.

The constructs a function that assigns the number 42 to the property `key` on an object `obj`. However, if `key` contains `</script>`, then the generated code will break out of a `</script>` if inserted into a `</script>` tag.

### PR Checklist


odaysec (commented):
Use a dedicated escaping helper for values inserted into generated JavaScript module source. Keep `JSON.stringify(...)` (to preserve string-literal semantics) and then post-process to escape dangerous characters (`<`, `>`, `/`, backslashes, line separators, control chars) so generated module text cannot be broken by special sequences.

Best fix in this file:
1. Add a local `charMap` and `escapeUnsafeJsCodeChars` helper near the existing constants/helpers.
2. Replace direct `JSON.stringify(routeKey)` interpolations in generated export lines with `escapeUnsafeJsCodeChars(JSON.stringify(routeKey))`.
3. Keep behavior unchanged otherwise (same values, same routing logic, same module generation layout).

PR #780, opened 2026-07-14
test(dev): pin the public dev-server contract for the stock-Nitro migration

ctgowrie:
## Why

We want to stop owning a bespoke dev listener/worker transport and align `eve dev` onto the stock Nitro dev server, without giving up authored HMR or in-flight Workflow turns finishing on the code that started them. `research/dev-stock-nitro-alignment.md` (added here) lays out the architecture, the two deliberately relaxed invariants, and the five-stage delivery plan.

This is **stage A**: no runtime changes. It re-expresses every transport-level assertion the migration will delete as a public-surface contract test, so later stages can remove `dev-worker-server` and friends while provably preserving behavior.

## New contract tests (all pass against the current stack)

- Chaos: a worker crash settles an open channel stream within a bounded deadline, and the server recovers to a completed streamed turn.
- Chaos: shutdown stays bounded and graceful (beats the harness SIGKILL fallback) while a stream is open.
- Chaos: authored routes observe the socket peer as `requestIp` even when the request carries forged worker metadata and `x-forwarded-for`.
- Lifecycle: a kept-alive connection serves a second request on the same socket across a structural worker replacement.

## Parity rules for the rest of the stack

- `test/scenarios/dev-server*.ts` stay green and unchanged through every stage — with one deliberate exception in stage C (instrumentation pinning across structural edits), flagged in the plan.
- Implementation-level suites (`dev-worker-server.scenario`, metadata tests) are deleted only in the stage that deletes their subject.

## Stack

Stacked on #778. Next stages: B (server-side generation selection), C (active-worker turns), D (parent-private World RPC), E (stock Nitro dev server + drained replacement, with the env-runner drain offered upstream).

### DIFF
--- a/packages/eve/test/scenarios/dev-server-chaos.scenario.test.ts
+++ b/packages/eve/test/scenarios/dev-server-chaos.scenario.test.ts
@@ -19,6 +19,7 @@ import {
   startEveDev,
   wait,
   waitForCondition,
+  withinDeadline,
   type RunningEveDev,
 } from "./dev-server-harness.js";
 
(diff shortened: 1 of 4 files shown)


ctgowrie:
Superseded: the stock-Nitro route will be a parallel stack off latest main, not stacked on the current dev-lifecycle stack. The contract tests and research doc here get ported into that stack phase by phase.

PR #782, opened 2026-07-14
feat(dev): drain retired dev workers behind the stock Nitro contract

ctgowrie:
## Why

First implementation stage of the stock-Nitro dev-server route (parallel to #760–#778, branched off #759 as the shared foundation). Stock Nitro's dev server already ready-gates reloads and gates admissions — but `RunnerManager.reload` terminates the previous worker as soon as the next one attaches, resetting every admitted response and socket. That termination is the only gap between stock Nitro and eve's no-reset requirement.

## What changed

- `DrainedNitroDevServer`: keeps the stock dev-server contract (`dev:start`/`dev:reload`/`dev:error` hooks, ready-gated replacement, crash restart) and adds **drained replacement** — the retired worker keeps serving the responses and websockets it already admitted (unbounded; a streaming turn can hold one for minutes) and is terminated once its last exchange settles, via parent-side exchange accounting. No worker-side protocol needed.
- Requests are served by the active worker *throughout* rebuilds (no stall), a failed candidate stays invisible behind the previous worker, and shutdown terminates immediately so open streams cannot block exit.
- Workers run on the vendored `env-runner` node-worker — the same runner stock Nitro uses. The class is deliberately shaped to be deleted once equivalent drain semantics land upstream in Nitro/env-runner (upstream PR to follow).

## Tests

Real-server contract suite (`dev-server-drain.scenario.test.ts`): an admitted stream keeps flowing on the retired worker across a replacement; a kept-alive connection serves across a replacement on the same socket; crash restart; bounded shutdown with an open stream; socket-derived client address despite forged forwarding headers. Class-level suite covers failed-candidate invisibility, drain-on-last-exchange, crash restart, failed-upgrade socket destruction, and idempotent close. The pre-existing dev-server scenario passes unchanged.

## Next stages

P3: transactional rebuild coordinator driving stock reloads. P4: Workflow World in the CLI process with latest-per-turn selection. Plan and test-parity matrix tracked in `research/dev-stock-nitro-alignment.md` (uncommitted).

### DIFF
--- a/.changeset/drained-dev-worker-replacement.md
+++ b/.changeset/drained-dev-worker-replacement.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Replace `eve dev` workers gracefully: the retired worker keeps serving the responses and sockets it already admitted while new requests go to the ready replacement, a failed replacement leaves the previous worker serving, and shutdown stays bounded with streams open.
--- a/packages/eve/package.json
+++ b/packages/eve/package.json
@@ -337,6 +337,7 @@
     "chokidar": "5.0.0",
     "commander": "14.0.3",
     "emulate": "0.6.0",
+    "env-runner": "0.1.12",
     "eventsource-parser": "3.0.8",
     "experimental-ai-sdk-code-mode": "1.0.16",
     "gray-matter": "4.0.3",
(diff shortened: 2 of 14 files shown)


ctgowrie:
Folded into #787 — the rebuild coordinator (originally here as a follow-up) is what gives the drained server its final shape, so the two ship as one PR. The intermediate hook-driven wiring this PR carried was deleted by that reshaping and had no standalone value.

PR #783, opened 2026-07-14
fix(eve): preserve Next.js middleware with generated services

### LINKED ISSUE #693: withEve Vercel Build Output service routes (0.22.2+) break deploys for Next 16 apps using proxy.ts — "Mapping /_middleware not found"
**Summary:** Since eve 0.22.2 ("Update `withEve()` to generate Vercel Build Output service routes"), Vercel deployments of a Next.js 16 app embedding eve via `withEve` fail at the "Deploying outputs" step with:

```
Mapping /_middleware not found. Maybe you provided a wrong middlewarePath?
```

Pinning `eve@0.22.1` (same app, same config) deploys green. 0.22.3 and 0.22.4 fail identically to 0.22.2's mechanism.

**Environment**
- eve 0.22.4 (also reproduced conceptually on 0.22.2+ — see bisect below)
- Next.js 16.2.10 (Turbopack), embedded via `withEve(withSentryConfig(nextConfig, …))`
- The app uses Next 16's **`src/proxy.ts`** (the renamed middleware); the build's route summary prints `ƒ Proxy (Middleware)`
- Vercel build machine, `vercel build` → `pnpm run build` → `next build`

**What happens on 0.22.2+**

During `next build` on Vercel, eve builds its runtime into the Build Output (visible in the build log: nitro `preset: vercel` writing `.vercel/output/functions/__server.func` plus a second build into `.eve/nitro-output/flow`), and `withEve` writes a `.vercel/output/config.json` declaring an eve service:

```json
{
  "routes": [
    { "destination": { "service": "eve", "type": "service" }, "src": "^/eve/v1/(.*)$" }
  ],
  "version": 3,
  "services": {
    "eve": {
      "buildCommand": "node 'node_modules/eve/bin/eve.js' build",
      "framework": "eve",
      "routes": [ { "src": "^/eve/v1/(.*)$", "transforms": [ { "args": "/eve/v1/$1", "op": "set", "type": "request.path" } ] } ],
      "root": "."
    }
  }
}
```

The deployment is then rejected at output-mapping validation with the `/_middleware` error above — the merged Build Output references a middleware mapping that doesn't line up with what Next 16 emits for `proxy.ts`.

**Bisect (local repro)**

`rm -rf .vercel && VERCEL=1 pnpm build` (i.e. `next build` with the env gate eve checks), then check whether `.vercel/output` was written:

| eve version | `.vercel/output` after `next build` |
|---|---|
| 0.18.1 | not written (matches production, which deploys fine) |
| 0.21.0 | not written |
| **0.22.1** | **not written — deploys green on Vercel** |
| 0.22.4 | `config.json` above written (on the Vercel platform, plus the nitro function output) → deploy fails |

**Build log excerpt (0.22.4, failing deploy)**

```
11:55:31.144 [success] [nitro] Generated public .vercel/output/static
11:55:31.147 [info] [nitro:vercel] Using `nodejs24.x` runtime.
...
11:55:32.520   ├─ .vercel/output/functions/__server.func/_libs/eve.mjs (3.35 MB) (825 kB gzip)
...
11:55:34.393 [BUILD] built output at /vercel/path0/.vercel/output
11:55:34.563 Build Completed in /vercel/output [3m]
11:55:37.102 Deploying outputs...
11:55:37.257 Mapping /_middleware not found. Maybe you provided a wrong middlewarePath?
```

**Workaround:** pin `eve@0.22.1` exactly.

Happy to provide the full build log or test candidate fixes on our deployment.

AndrewBarba:
## What

- build generated, co-located eve services under `.eve/vercel-services/<name>` instead of the Next.js output directory
- pass the host and service output directories through one internal build contract
- copy the host middleware function into the generated service output before Vercel collects it
- cover the emitted artifacts with a local `vercel build` scenario and the collector boundary with a real source deployment test

## Why

A Next.js source deployment with `proxy.ts` completed its build and then failed while Vercel collected the outputs:

```
Mapping /_middleware not found. Maybe you provided a wrong middlewarePath?
```

The root Next.js output was present, including `config.json` with `middlewarePath: "/_middleware"` and `functions/_middleware.func`. That is not enough. Vercel validates the middleware route against each generated service builder's function map. The generated eve service did not contain `_middleware`, so collection failed.

`withEve()` now keeps the eve build in its own output directory and gives `eve build` the host output location as part of the same internal contract. After eve assembles its service output, it copies the opaque middleware function into that output. Next.js still owns the root config and function. The service copy exists so Vercel can deserialize and validate the generated service.

The regression scenario requires all three artifacts: the root middleware route, the root middleware function, and the generated service middleware function. The Vercel test then source-deploys the packed package, which is the boundary that produced the original error.

Fixes #693


ruiconti:
Verified this against the concrete #693 shape: Next.js 16.2.10 with `src/proxy.ts`, an authored `agent/agent.ts`, and a prebuilt Vercel deployment.

Before this patch, the assembled root config referenced `/_middleware` after the root `functions/` directory had been displaced, and upload failed at `Deploying outputs...` with `Mapping /_middleware not found`.

Using a packed package from this PR:

- `vercel build` preserves `.vercel/output/functions/_middleware.func`
- eve is isolated at `.vercel/output/services/eve/functions/__server.func`
- the generated service entrypoint is `agent/agent.ts`
- the service retains the `^/eve/v1/(.*)$` route
- a fresh `vercel deploy --prebuilt` passed `Deploying outputs...` and reached `READY`

Targeted local checks also pass:

- `src/public/next/index.integration.test.ts`: 11/11
- `src/internal/nitro/host/build-application.scenario.test.ts`: 10/10

The initial Windows integration failure was not an assertion failure: 431 tests passed and Vitest reported one worker process exiting unexpectedly. The rerun passed.

ruiconti (approved):

PR #792, opened 2026-07-14
test(dev): pin the public dev-server contract with bun and chaos coverage

ctgowrie:
Third PR of the stock-Nitro dev-server stack, on top of #790 → #787 (followed by #798).

## What

Pins the rebuilt server to its public contract and failure behavior:

- **Contract suite:** kept-alive connections across structural worker replacement, dev schedule dispatch through the active generation, npm-installed hoisted dependency layouts, rebuild transactionality, and generation isolation.
- **Chaos suite:** authored edit storms under a continuous health probe, broken candidates, concurrent forced rebuilds, worker crashes during rebuilds and streams, aborted streams, bounded shutdown, and forged-header handling.
- **Bun suite:** a bun-installed app served by the Node runtime and a fast explanatory error when the CLI itself is run under bun. CI installs bun; the cases skip when it is absent locally.
- Exchange-release coverage for aborted uploads and cancelled retired-worker streams.

The schedule scenario validates the generation-pinned dispatch fix owned by #787; this PR no longer carries duplicate implementation or a duplicate changeset.

## Validation

The contract, chaos, bun, drain, unit, and integration suites pass locally. CI provides the official end-to-end coverage.

### DIFF
--- a/packages/eve/src/internal/testing/scenario-app.ts
+++ b/packages/eve/src/internal/testing/scenario-app.ts
@@ -8,9 +8,11 @@ import {
   rm,
   writeFile,
 } from "node:fs/promises";
+import { execFile } from "node:child_process";
 import { createRequire } from "node:module";
 import { tmpdir } from "node:os";
 import { basename, dirname, join } from "node:path";
+import { promisify } from "node:util";
 
 import { afterEach } from "vitest";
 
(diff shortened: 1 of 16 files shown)


packages/eve/src/internal/testing/scenario-app.ts
```diff
@@ -8,9 +8,11 @@ import {
   rm,
   writeFile,
 } from "node:fs/promises";
+import { execFile } from "node:child_process";
```
De-ASI-INTERFACE, line 11: I’m

De-ASI-INTERFACE (approved):

AndrewBarba (approved):

PR #794, opened 2026-07-14
fix(eve): streaming - expose waiting continuation tokens

AndrewBarba:
## Summary

- include the current channel-local continuation token in session.waiting events and adopt it in the TypeScript client
- allow negative stream start indexes, including -1 for the latest event, without corrupting absolute client cursors
- document tail lookups and add a patch changeset

## Testing

- pnpm test:unit
- pnpm typecheck
- pnpm lint
- pnpm fmt
- pnpm docs:check
- pnpm guard:invariants


packages/eve/src/harness/emission.ts
```diff
@@ -273,6 +273,7 @@ export async function emitTurnEpilogue(
   emitFn: HarnessEmitFn,
   state: HarnessEmissionState,
```
ruiconti, line 274: Would it make sense to enrich `HarnessEmissionState` with `continuationToken`?

ruiconti (approved):

packages/eve/src/harness/emission.ts
AndrewBarba: I considered that, but I think keeping the token on HarnessSession is safer. HarnessEmissionState is persisted emission lifecycle state, while the continuation token already lives on the session and may be re-keyed by channel handlers; duplicating it in emission state would create a potentially stale second source of truth. I kept the terminal helpers explicit instead. After syncing main, the new cancellation-settlement path now also passes enrichedSession.continuationToken directly.

PR #806, opened 2026-07-15
fix(eve): store local Workflow data under .eve

ctgowrie:
## Summary

- Resolve parent-owned local Workflow World data through one eve-owned `.eve/.workflow-data` path.
- Use that resolver for both `eve dev` and the production local-World startup path.
- Update generated ignore guidance, documentation, and assertions without a legacy fallback.

## Validation

- `pnpm --filter eve build`
- Focused unit, integration, and CLI scenario suites
- `pnpm docs:check`
- `pnpm guard:invariants`


AndrewBarba (approved):
Great change

Issue #216, opened 2026-06-23
Support session reset and scoped cancellation for long-lived channels


AndrewBarba:
## Summary

Identity-based channels such as Telegram private chats and Twilio conversations reuse one continuation token indefinitely, causing a single durable session to accumulate history and state forever.

Add an explicit session-reset capability alongside active-turn cancellation. Telegram and Twilio should support `/new` by default, while custom channels, Slack handlers, and session event subscribers should be able to invoke the same behavior through public APIs. Channel authors may explicitly configure alternative command names.

## Required behavior

- expose one authenticated `POST /eve/v1/session/:sessionId/cancel` route with a required `scope: "turn" | "session"`
- preserve resumable session behavior for `scope: "turn"`
- make `scope: "session"` terminate the entry workflow, active turn, and delegated local or remote descendants
- make bare `/new` silently end the current session without creating an empty replacement
- treat trailing command content as the first turn of a fresh session
- expose clean runtime, `defineChannel`, built-in channel-handler, and event-subscriber APIs
- update the TypeScript client with first-class active-turn and whole-session cancellation
- give evals first-class cancellation controls that can drive and assert expected cancellation outcomes
- validate cancellation capabilities against the session id and handle cancellation races without allowing the old workflow to reclaim its continuation token
- add a dedicated e2e fixture whose custom `defineChannel` exercises both turn and session cancellation using only public channel APIs

## Acceptance criteria

- Telegram and Twilio can start a clean session under the same channel identity
- `/new` is the only built-in reset command enabled by default
- turn cancellation remains resumable while session cancellation is terminal
- the eve channel has one scoped cancellation endpoint with no implicit scope
- TypeScript client and eval APIs support both cancellation lifecycles without confusing local request aborts with server cancellation
- custom channels and handlers can reproduce the built-in reset behavior
- the dedicated custom-channel fixture proves turn cancellation resumes the same session and session cancellation creates a fresh session under the reused identity
- intentional resets do not surface as ordinary agent failures

andrewgreenh:
Would also be cool if we could reset or clear sessions via the vercel -> observability -> Agent runs UI.
I'm running into the same problem with the teams setup.

PR #817, opened 2026-07-15
feat(eve): cancellation layer 2 - HTTP cancel-turn trigger route (+ channel-ops research)

AndrewBarba:
## Summary

Continues the turn-cancellation stack (#483) after layer 1 (#573).

This PR implements layer 2 by adding the first production trigger for the existing turn-cancellation hook:

- `POST /eve/v1/session/:sessionId/cancel` on `eveChannel()`.
- Public `Agent.cancelTurn(...)` and runtime `Runtime.cancelTurn(...)` operations.
- Unit, integration, and hosted e2e coverage.
- User-facing documentation and a patch changeset.

It also adds the channel-ops superseding plan for follow-up work. The proposed `send(payload, { supersede: "turn" })` and route-author `cancelTurn` APIs are research-only in this PR.

## HTTP API

`POST /eve/v1/session/:sessionId/cancel` uses the eve channel's existing `routeAuth` policy. The body is optional:

```json
{ "turnId": "turn_0" }
```

Omitting `turnId` targets whichever turn currently owns the session's cancellation hook. Supplying it scopes the request to the turn the caller observed; a mismatch is accepted and consumed as a no-op so a stale request cannot cancel a newer turn.

| Response | Meaning |
| --- | --- |
| `202 { ok, sessionId, status: "cancelling" }` | A cancellation hook accepted the request. This is acceptance, not settlement. |
| `202 { ok, sessionId, status: "no_active_turn" }` | No resumable cancellation target exists, including an unknown, settled, duplicate, or uncancellable task-mode session. |
| `400` | The body is malformed or `turnId` is invalid. |
| `401/403` | Authentication rejected the request. |
| `500 { ok: false, error, errorId }` | An unexpected runtime failure occurred. |

Both `202` statuses are successful, making late and duplicate stop requests safe. Successful responses are non-cacheable and include the eve session-id header.

## Cancellation semantics

- Confirm an effective cancellation on the stream as `turn.cancelled` followed by `session.waiting`; cancellation does not emit a failure event.
- The session remains reusable and accepts the next message normally.
- Partial model output remains on the durable event stream. Durable model history keeps only content that had already settled.
- Already-executed tool and channel side effects remain; cancellation is not rollback.
- Client `stop()` and `AbortSignal` only detach the in-flight client stream. They do not stop server-side work; callers must use the cancel route for that.

## Implementation

- Adds eve-owned `CancelTurnInput` and `CancelTurnResult` types and exposes cancellation through `Agent` and `Runtime`.
- The eve route delegates `Agent.cancelTurn` to `Runtime.cancelTurn`, which resumes the stable `{sessionId}:cancel` hook with the optional turn guard.
- Missing hooks or owner runs, expired runs, and terminal-state conflicts normalize to `"no_active_turn"`. Unexpected failures propagate to the route, which logs them once and returns an `errorId`.
- Keeps `sessionCancelHookToken` and `TurnCancelPayload` in a leaf module so HTTP route bundles do not import workflow-VM modules.
- Adds the protocol route constant and encoded path helper.

## Coverage and documentation

- Route and runtime unit tests cover auth ordering, body validation, guard forwarding, result mapping, headers, inactive targets, and unexpected failures.
- World-local integration covers cancellation through the real route, missing and duplicate targets, stream settlement, tool abortion, stale guards, and a successful follow-up turn.
- The `agent-cancellation` fixture exercises the route against local and hosted Vercel worlds with a hanging tool bound to `ctx.abortSignal`, then verifies a normal follow-up.
- Eval regression coverage pins continuation-token recovery when sending a follow-up on an attached session.
- Docs cover the route contract, cancellation/history semantics, attached eval sessions, and the distinction between client stream abortion and server-side turn cancellation.

## Research follow-up

- `research/turn-cancellation-trigger-surface.md` documents layer 2 and is implemented by this PR.
- `research/turn-cancellation-channel-ops.md` proposes race-safe cancel-before-deliver superseding for authored channels. Its implementation remains a separate PR.

Refs #483, #216.


packages/eve/src/execution/request-turn-cancellation.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function requestTurnCancellation(input: {
+  readonly sessionId: string;
+  readonly turnId?: string;
+}): Promise<TurnCancellationRequestStatus> {
+  const payload: TurnCancelPayload = input.turnId === undefined ? {} : { turnId: input.turnId };
+
+  try {
+    await resumeHook(sessionCancelHookToken(input.sessionId), payload);
+    return "cancelling";
+  } catch (error) {
```
ctgowrie: Can we make this an action on the agent the way we have `deliver`? Instead of an arbitrary hook token

AndrewBarba: Done — `cancelTurn` is now an agent action backed by the runtime, and the hook helper is gone. Thanks!

ctgowrie (approved):

PR #823, opened 2026-07-15
fix(eve): retry + wait during prewarm step if sandbox is mid-snapshot

benpankow:
### Description

Simultaneous builds (e.g. multiple deploys triggered off of the same branch) resulted in build failures when the prewarm/snapshot step triggered while another sandbox snapshot was in progress. This raised a 422 (`sandbox_snapshotting`) which wasn't specially handled, leading to a build failure.

This PR introduces a check for the `sandbox_snapshotting` 422 and probes until the snapshot completes (or 120s has elapsed) instead of failing the build entirely. Adds a unit test for this case which failed before the change was impl.

```bash
  src/execution/sandbox/bindings/vercel.test.ts \
  src/execution/sandbox/bindings/vercel-errors.test.ts
```

### PR Checklist

### DIFF
--- a/.changeset/retry-sandbox-snapshotting-prewarm.md
+++ b/.changeset/retry-sandbox-snapshotting-prewarm.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fixed an intermittent build failure when two deployments build the same app at the same time. Builds now wait for the in-progress sandbox snapshot to finish instead of failing.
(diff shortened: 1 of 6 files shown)


AndrewBarba (approved):

benpankow:
Closing because the sandbox and provider-session lifecycle has since been redesigned, including in #3271, and this PR targets an error path that no longer exists on current `main`. If the snapshot race reappears in the current architecture, it should be addressed there rather than reviving this branch.

PR #832, opened 2026-07-15
feat(eve): cancellation - cascade to descendant turns

AndrewBarba:
## Summary

- record adopted local and remote child session IDs in pending runtime-action batches
- cascade cancelled parent turns through local cancellation hooks and the authenticated remote cancel endpoint
- preserve the dispatch adoption barrier and retry newly registered or transiently unavailable descendants 12 times at 250 ms
- contain descendant cancellation failures so the parent still settles as `turn.cancelled` → `session.waiting`
- add authenticated `ClientSession.cancel()` plus eval `start()`, `watchTurn()`, typed `waitForEvent()`, cancellation, and single-stream result settlement
- retain subagent call/session identities in eval facts and rewrite both cancellation evals without raw HTTP or NDJSON handling
- document the layer-3 design and new client/eval controls, and include a patch changeset

## Testing

- [x] full unit suite (4,814 passed, 1 skipped)
- [x] cancellation integration suite (7 passed)
- [x] authenticated two-server remote cancellation scenario
- [x] cancellation fixture typecheck/build
- [x] `pnpm fmt`
- [x] `pnpm lint`
- [x] `pnpm typecheck`
- [x] `pnpm guard:invariants`
- [x] `pnpm docs:check`
- [x] `pnpm build`
- [ ] cancellation e2e fixture (CI only)

Part of #483.


packages/eve/src/harness/runtime-actions.ts
```diff
@@ -136,9 +135,16 @@ export function recordPendingSubagentChildToken(input: {
...
+    childContinuationTokens:
+      input.childContinuationToken === undefined
+        ? batch.childContinuationTokens
+        : {
+            ...batch.childContinuationTokens,
+            [input.callId]: input.childContinuationToken,
+          },
+    childSessionIds: {
+      ...batch.childSessionIds,
+      [input.callId]: input.childSessionId,
```
ctgowrie: This logic feels clunky, why is the continuationToken now optional? Generally the logic now smells

AndrewBarba: You are right. Remote children have no continuation token, but the original shared input weakened the local invariant. Fixed with one discriminated local/remote adoption input in 0e4cc6a1; local requires both identities.

packages/eve/src/client/session.ts
```diff
@@ -374,3 +426,13 @@ function createHandleMessageBody(input: {
 
   return body;
 }
+
+function readCancelStatus(value: unknown): CancelSessionResult["status"] | undefined {
+  if (typeof value !== "object" || value === null) return undefined;
+  const status = (value as { readonly status?: unknown }).status;
+  return status === "cancelling" || status === "no_active_turn" ? status : undefined;
+}
```
ruiconti: Nit: more of a high-level suggestion but the public HTTP API should expose its schema in a way that all routes can be parsed against that schema, ideally via a validator-backed one e.g `zod`. Otherwise changes to `/cancel` now can silently drift.

packages/eve/src/client/session.ts
```diff
@@ -374,3 +426,13 @@ function createHandleMessageBody(input: {
 
   return body;
 }
+
+function readCancelStatus(value: unknown): CancelSessionResult["status"] | undefined {
+  if (typeof value !== "object" || value === null) return undefined;
+  const status = (value as { readonly status?: unknown }).status;
+  return status === "cancelling" || status === "no_active_turn" ? status : undefined;
```
ruiconti: Use same verbiage as the internal loop signal? e.g, `cancel-accepted`?

packages/eve/src/execution/cancel-descendant-turns-step.ts
```diff
@@ -0,0 +1,134 @@
...
+      childSessionId: input.childSessionId,
+      subagentName: input.action.subagentName,
+    });
+  }
+}
+
+async function cancelRemoteDescendant(input: {
+  readonly action: RuntimeRemoteAgentCallActionRequest;
+  readonly childSessionId: string;
+  readonly remoteRegistry: Promise<RuntimeSubagentRegistry["subagentsByNodeId"]> | undefined;
```
ruiconti: Wouldn't expect `remoteRegistry` to be undefined here - checking for its presence should belong to its parent.

packages/eve/src/execution/cancel-descendant-turns-step.ts
```diff
@@ -0,0 +1,134 @@
...
+    const remote = resolveRemoteAgentForAction({
+      nodeId: input.action.nodeId,
+      remoteAgentName: input.action.remoteAgentName,
+      registry,
+    });
+
+    await requestCancellationWithRetry({
+      request: () => cancelRemoteAgentTurn({ remote, sessionId: input.childSessionId }),
+      shouldRetryError: isRetryableRemoteAgentCancelError,
+    });
```
ruiconti, line 107: Nit: at some point we should decompose this
 
```suggestion
    await withRetry(() => {
      const result = await cancelRemoteAgentTurn({ remote, sessionId: input.childSessionId });
      if (result.status === "cancelling" || attempt === CANCEL_ATTEMPTS) return true;
      throw new RetryException()
      }, {
      shouldRetryError: isRetryableRemoteAgentCancelError,
    });
```

packages/eve/src/harness/runtime-actions.ts
```diff
@@ -117,14 +117,11 @@ export function setPendingRuntimeActionBatch(input: {
...
 }
 
-/**
- * Records the child continuation token for a dispatched subagent-call
- * so {@link resolvePendingRuntimeActions} can clear proxy-input
- * entries when the child finishes.
- */
-export function recordPendingSubagentChildToken(input: {
+/** Records one successfully dispatched local child's durable identities. */
+export function recordPendingLocalSubagentChild(input: {
```
ruiconti: Why has this been split up by local and remote?

e2e/fixtures/agent-cancellation/evals/cancellation/cancel-subagent.eval.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const child = t.target.watchTurn(called.data.childSessionId);
+    await child.waitForEvent("actions.requested", {
+      data: {
+        actions: (actions) =>
+          actions.some(
+            (action) => action.kind === "tool-call" && action.toolName === "wait-for-cancellation",
+          ),
+      },
+    });
```
ruiconti, line 22: If I'm reading this right, this is testing HITL which we should have. I'd expect evals to also exercise

1. Mid-tool subagent interrupt eg after a marked tool execution had begun and before marked tool execution had ended to assert that cancellation properly aborts it
2. Racing subagent interrupts - similar as above but N_SUBAGENT>1 (this is probably the more generic one)
3. **Racing both local and remote subagent interrupts**
4. **Core failure mode: local fails to respond to cancellation and/or remote fails to respond to cancellation but parent still aborts**

Thinking about this while writing, believe that 3. is the more general case and exercising it would give enough confidence on the signal propagation.

packages/eve/src/execution/cancel-descendant-turns-step.ts
AndrewBarba: Moved that guarantee to the parent in 2d249c5f. The remote helper now requires the registry and is only invoked after the parent establishes remote children exist.

packages/eve/src/execution/cancel-descendant-turns-step.ts
AndrewBarba: Agreed. There is no shared retry primitive yet, so I kept the cancellation-specific loop here; I would extract it when a second caller establishes the common API.

packages/eve/src/client/session.ts
AndrewBarba: Addressed for cancel in 2d249c5f: client and remote callers parse one shared eve-owned Zod response schema, and the route is typed against the same contract.

packages/eve/src/harness/runtime-actions.ts
AndrewBarba: Consolidated in 0e4cc6a1: one recordPendingSubagentChild function takes a discriminated identity. Local adoption requires token plus session ID; remote adoption requires the session ID.

e2e/fixtures/agent-cancellation/evals/cancellation/cancel-subagent.eval.ts
AndrewBarba: This eval is mid-tool rather than HITL: it waits for actions.requested, then cancellation aborts the running tool signal. In 0e4cc6a1 I also expanded the real two-server scenario to start local and authenticated remote sleepers together, wait for both tools, then verify both child boundaries and the parent boundary. The existing unit suite covers descendant failure containment.

packages/eve/src/client/session.ts
AndrewBarba: Renamed the accepted cancellation result across layer 2 and layer 3 in 2d249c5f. The status is now accepted; no_active_turn remains unchanged.

packages/eve/src/execution/cancel-descendant-turns-step.ts
```diff
@@ -0,0 +1,132 @@
...
+  if (remoteChildren.length > 0) {
+    const remoteRegistry = deserializeContext(input.serializedContext).then(
+      (ctx) => ctx.require(BundleKey).subagentRegistry.subagentsByNodeId,
+    );
+    cancellations.push(
+      ...remoteChildren.map(({ action, childSessionId }) =>
+        cancelRemoteDescendant({ action, childSessionId, remoteRegistry }),
+      ),
+    );
+  }
```
ruiconti: Splitting the local and remote loops here is weird as deserializingContext shouldn't be that expensive?

ruiconti (approved):

packages/eve/src/execution/cancel-descendant-turns-step.ts
AndrewBarba: Agreed. Simplified to one child pass in 8e41e2db. Remote context is resolved lazily once, so local cancellation remains independent of remote-context failures.

PR #834, opened 2026-07-15
feat(eve): channels - add turn steering and cancellation

### LINKED ISSUE #483: No way to cancel an in-flight turn: stop() only detaches the client; the model call keeps running (and billing)

AndrewBarba:
### Description

Closes #483.

Adds a channel-author turn control surface:

- send options now accept `turnPolicy: "queue" | "steer"`, with `queue` as the default
- `steer` durably redirects the active logical turn at its next safe boundary and requeues input if settlement wins the race
- route handlers now receive continuation-token-addressed `cancelTurn` with an optional `turnId` guard
- Chat SDK forwarding, HITL behavior, descendant routing, continuation-token rekeying, and durable workflow compatibility are covered
- channel docs, the research decision, and a patch changeset are included

### Authoring example

```ts
import { defineChannel, POST } from "eve/channels";

export default defineChannel({
  routes: [
    POST("/messages", async (req, { send }) => {
      const body = (await req.json()) as { message: string; threadId: string };
      const session = await send(body.message, {
        auth: null,
        continuationToken: body.threadId,
        turnPolicy: "steer",
      });

      return Response.json({ sessionId: session.id });
    }),
    POST("/threads/:threadId/cancel", async (_req, { cancelTurn, params }) => {
      const result = await cancelTurn({
        continuationToken: params.threadId,
        // turnId: observedTurnId, // Optional guard against a stale cancel.
      });

      return Response.json(result);
    }),
  ],
});
```

Omit `turnPolicy` (or pass `"queue"`) when the message should wait for the active turn to settle and start the next turn.

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test:unit`
- `pnpm test:integration`
- `pnpm guard:invariants`
- `pnpm docs:check`
- targeted workflow integration coverage for in-flight steering and token-addressed cancellation

### PR Checklist


packages/eve/src/execution/durable-session-migrations/turn-workflow.ts
```diff
@@ -22,6 +22,8 @@ export const TURN_WORKFLOW_INPUT_VERSION = 1;
 export interface TurnStepInput {
   /** Cancellation signal forwarded into the turn step. */
   readonly abortSignal?: AbortSignal;
```
ruiconti, line 25: Nit: Now that we're introducing another `AbortSignal`, I'd rename the above to `cancelSignal`.

packages/eve/src/execution/turn-control-receiver.ts
```diff
@@ -17,6 +18,8 @@ export class TurnControlReceiver {
   private readonly controlIterator: AsyncIterator<TurnControlPayload>;
   private readonly deliveryHook: SessionDeliveryHook;
   private pendingControl: Promise<IteratorResult<TurnControlPayload>> | null = null;
+  private steeringSequence = 0;
+  private steeringToken: string | undefined;
```
ruiconti, line 29: Don't love that we're storing the `steeringToken` in a class instead of propagating it `this.readTerminalControl`. Makes it hard to reason about control flow

packages/eve/src/execution/durable-session-migrations/turn-workflow.ts
AndrewBarba: I think internally abort is still the right word?

packages/eve/src/execution/turn-workflow.ts
```diff
@@ -281,6 +325,23 @@ async function waitForRuntimeActionResults(input: {
       }
       return "cancelled";
     }
+    if (next === "steer") {
```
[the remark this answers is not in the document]
AndrewBarba: Fixed in a6f69669. `waitForRuntimeActionResults` now retains one pending inbox read when steering wins and clears it only after the inbox read settles. I also added a deterministic regression test where steering wins before the runtime-action result arrives and assert that `iterator.next()` is called once.

packages/eve/src/execution/turn-control-receiver.ts
AndrewBarba: Addressed in a6f69669. The steering token is no longer mutable receiver state: `turn-steering-ready` transitions into an explicit steerable phase, and that phase passes `steeringToken` directly into `forwardSteering`.

PR #840, opened 2026-07-15
fix(eve): include per-agent route prefix in framework callback URLs

### LINKED ISSUE #839: Connector-auth and remote-subagent callback URLs 404 in multi-agent mode (per-agent route prefix dropped)
### Summary

In multi-agent mode (`withEve({ agents })`), each named agent mounts **only** under
`/eve/agents/<name>/eve/v1/*` — there is no bare `/eve/v1/*` mount. But two framework-minted
callback URLs are built from the deployment origin plus a bare `/eve/v1/...` path, dropping
the per-agent prefix, so they 404:

- **A — connection callback (OAuth).** `getHookUrl` (`packages/eve/src/harness/authorization.ts`)
  mints `<origin>/eve/v1/connections/<name>/callback/<token>`. After the user consents, the
  IdP redirects the browser to a path nothing serves → 404 → the parked turn never resumes →
  the "Connect &lt;provider&gt;" card re-prompts in a loop.
- **B — remote-subagent session callback.** `startRemoteAgentSession`
  (`packages/eve/src/execution/remote-agent-dispatch.ts`) mints `<origin>/eve/v1/callback/<token>`
  via `createWorkflowCallbackUrl`. The remote subagent completes, then its terminal result POST
  404s → retried 3× → `run_failed` → the parent turn hangs forever.

Single/unnamed agents are unaffected: the single root IS mounted at the bare `/eve/v1/*`, so the
prefix-less callback matches.

Root cause is shared: the running function does not know its own public prefix — the Vercel
route applies a `request.path` transform to `/eve/v1/$1` before the request reaches the service,
so callbacks minted mid-turn cannot reconstruct `/eve/agents/<name>`.

No standalone repro repo — this is a structural bug found by code inspection and confirmed in
production. It reproduces on any deployed `withEve({ agents })` app where a named agent either
declares an interactive-auth connector (A) or delegates to a `defineRemoteAgent` (B). A fix PR
accompanies this issue.

### Steps to reproduce

1. Deploy an eve app with `withEve({ agents: { researcher: './agents/researcher', ... } })`.
2. **A:** give `researcher` a connector that requires interactive authorization; start a turn
   that triggers `authorization.required`, then complete the provider consent.
   **B:** have a parent agent delegate to `researcher` via `defineRemoteAgent`; let the remote
   subagent finish.
3. Observe the callback 404.

- **A — current:** browser lands on `<origin>/eve/v1/connections/<name>/callback/<token>` → 404;
  turn stays parked; connect card re-prompts. **Expected:** lands on
  `<origin>/eve/agents/researcher/eve/v1/connections/<name>/callback/<token>` and the turn resumes.
- **B — current:** result callback POSTs to `<origin>/eve/v1/callback/<token>` → 404 → 3 retries →
  `run_failed`; parent hangs. **Expected:** POSTs to
  `<origin>/eve/agents/researcher/eve/v1/callback/<token>` and the parent turn continues.

### eve version

eve@0.24.3 through HEAD — structural since named multi-agent roots shipped (#655). Not a regression.

### Environment

```shell
Deployed on Vercel (multi-service Build Output). Reproduces on the deployed agent, not local single-agent dev.
```

### Where does the bug occur?

### Build and runtime logs

```shell
# Site B (remote-subagent session callback):
FatalError: Step "step//eve@0.24.3//fireSessionCallbackStep" failed after 3 retries: Session callback failed with HTTP 404.
[eve:execution.session-callback] failed to post session callback … HTTP 404
# lifecycle: subagent step_completed at +1.6s, then hook_received → step_retrying ×3 → step_failed → run_failed
```

### Additional context

Both mint sites share one root cause and one fix: teach the framework its per-agent public route
prefix (inject it into the workflow function's runtime environment at build) and prepend it to
framework-owned callback paths at mint time. This also covers nested-subagent connects that an
app-side redirect shim cannot resolve.

Two related issues in the same class are intentionally **out of scope** here and tracked separately:
the remote create-session URL (`new URL` drops the base path) and the hosted-service file-tracing
gap (`#public/...` specifiers not followed).

IKatsuba:
Closes #839.

## Problem

In multi-agent mode (`withEve({ agents })`) each named agent mounts **only** under
`/eve/agents/<name>/eve/v1/*`. Two framework-minted callback URLs were built from the bare
deployment origin, dropping the per-agent prefix, so they 404:

- **Connection callback (OAuth)** — `getHookUrl` minted `<origin>/eve/v1/connections/<name>/callback/<token>`.
  After consent the browser landed on a path nothing serves → parked turn never resumed → connect
  card re-prompted in a loop.
- **Remote-subagent session callback** — `startRemoteAgentSession` minted `<origin>/eve/v1/callback/<token>`.
  The subagent finished, then its result POST 404'd → 3 retries → `run_failed` → parent turn hung.

Both share one root cause: the Vercel route strips the prefix (`request.path` → `/eve/v1/$1`) before
the request reaches the function, so callbacks minted mid-turn cannot reconstruct `/eve/agents/<name>`.

## Fix

Carry the per-agent public prefix into the workflow function's runtime environment at build, then
prepend it to framework callback **paths** at mint time.

- **`protocol/public-route-prefix.ts`** (new) — `resolveEvePublicRoutePrefix()` reads
  `EVE_PUBLIC_ROUTE_PREFIX`; `prefixFrameworkCallbackPath(path)` prepends it.
- **`normalizeEveVercelFunctionOutput`** bakes `environment.EVE_PUBLIC_ROUTE_PREFIX = servicePrefix`
  into eve-owned function `.vc-config.json` files. The value is the already-resolved `servicePrefix`
  (`build-application.ts`), so no new build-command export is needed. Runs only for named agents;
  single/unnamed output is byte-identical.
- **Mint sites** — `getHookUrl` (connection callback) and `startRemoteAgentSession` (session callback)
  prepend the prefix to the callback path.

The path must be prefixed at mint time only — never in the shared `#protocol/routes.js` builders,
which also register the **receiving** routes that run post-transform against the bare `/eve/v1/*`
path.

## Scope

Two related bugs in the same class are intentionally out of scope and tracked separately: the remote
create-session URL (`new URL` drops the base path) and the hosted-service file-tracing gap.

## Testing

- Unit: `public-route-prefix` helper; `createWorkflowCallbackUrl` preserves a prefixed absolute path.
- Integration: `getHookUrl` mints a prefixed URL when `EVE_PUBLIC_ROUTE_PREFIX` is set.
- Scenario: `normalizeEveVercelFunctionOutput` injects the env for a named agent and leaves single-agent
  output untouched.
- `pnpm fmt`, `pnpm lint`, `pnpm typecheck`, `pnpm guard:invariants` pass.

## Open question

`EVE_PUBLIC_ROUTE_PREFIX` is injected only into Vercel build output. If multi-agent `eve dev` also
mounts under `/eve/agents/<name>`, the dev server would need the same env; if dev serves a single
agent (no prefix), nothing to do. Flagging for reviewer input.

### DIFF
--- a/.changeset/multi-agent-callback-prefix.md
+++ b/.changeset/multi-agent-callback-prefix.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix connector authorization and remote-subagent result callbacks 404ing in multi-agent mode (`withEve({ agents })`). Framework-minted callback URLs now include the agent's public route prefix (`/eve/agents/<name>`), so after connecting a provider or when a remote subagent finishes, the parked turn resumes instead of the connect card re-prompting in a loop or the parent turn hanging.
--- a/packages/eve/src/channel/session-callback.test.ts
+++ b/packages/eve/src/channel/session-callback.test.ts
@@ -1,0 +1,47 @@
+import { describe, expect, it } from "vitest";
+
+import { parseSessionCallback } from "#channel/session-callback.js";
+
+const BASE_CALLBACK = {
+  callId: "call-remote",
+  subagentName: "research",
+  token: "eve:parent-token",
+};
+
+describe("parseSessionCallback", () => {
+  it("accepts a bare single-agent callback url", () => {
+    const result = parseSessionCallback({
+      ...BASE_CALLBACK,
+      url: "https://caller.example.com/eve/v1/callback/eve%3Aparent-token",
+    });
+
+    expect(result.ok).toBe(true);
+  });
+
+  it("accepts a callback url carrying a per-agent public route prefix (multi-agent mode)", () => {
+    const result = parseSessionCallback({
+      ...BASE_CALLBACK,
+      url: "https://caller.example.com/eve/agents/researcher/eve/v1/callback/eve%3Aparent-token",
+    });
+
+    expect(result.ok).toBe(true);
+  });
+
+  it("rejects a callback url whose token segment does not match the token", () => {
+    const result = parseSessionCallback({
+      ...BASE_CALLBACK,
+      url: "https://caller.example.com/eve/agents/researcher/eve/v1/callback/someone-else",
+    });
+
+    expect(result.ok).toBe(false);
+  });
+
+  it("rejects a private/reserved callback host", () => {
+    const result = parseSessionCallback({
+      ...BASE_CALLBACK,
+      url: "http://169.254.169.254/eve/v1/callback/eve%3Aparent-token",
+    });
+
+    expect(result.ok).toBe(false);
+  });
+});
--- a/packages/eve/src/channel/session-callback.ts
+++ b/packages/eve/src/channel/session-callback.ts
@@ -70,12 +70,16 @@ export function parseSessionCallback(value: unknown): SessionCallbackParseResult
 }
 
 function readCallbackUrlToken(url: URL): string | null {
+  // In multi-agent mode the minted callback carries the agent's public route
+  // prefix (`/eve/agents/<name>/eve/v1/callback/<token>`), so match the callback
+  // segment anywhere in the path rather than anchoring it to the origin.
   const tokenPrefix = createEveCallbackRoutePath("");
-  if (!url.pathname.startsWith(tokenPrefix)) {
+  const markerIndex = url.pathname.lastIndexOf(tokenPrefix);
+  if (markerIndex === -1) {
     return null;
   }
 
-  const encodedToken = url.pathname.slice(tokenPrefix.length);
+  const encodedToken = url.pathname.slice(markerIndex + tokenPrefix.length);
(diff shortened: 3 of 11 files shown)


IKatsuba:
Good catch from the automated review — confirmed and fixed in 5467bc9.

`readCallbackUrlToken` (session-callback schema) anchored the token match to the origin via `startsWith("/eve/v1/callback/")`, so once the minted callback carries the agent prefix (`/eve/agents/<name>/eve/v1/callback/<token>`) it returned `null` and `parseSessionCallback` rejected the dispatch — the child would 400 on create-session and delegating to a named remote agent would break. Now it matches the callback segment anywhere in the path, so both bare single-agent and prefixed multi-agent URLs validate. Added unit coverage in `channel/session-callback.test.ts`.

Issue #843, opened 2026-07-15
Design structured local diagnostics for authoring failures


ruiconti:
## Problem\n\nLocal authoring failures currently reach users through several unrelated surfaces: compiler diagnostics, dev-server stderr, harness logs, stream failure events, and Workflow run state. Upstream errors can expose generated paths or full stack objects, while handled failures may be logged more than once. Fixing individual messages does not establish a stable contract for future failures.\n\n## Research scope\n\nDefine an eve-owned diagnostics model for local authoring that:\n\n- preserves the original failure and stable identifiers for detailed inspection;\n- renders one terse, actionable stderr message at the user-facing boundary;\n- assigns ownership before a session exists and after a session ID is known;\n- normalizes known compiler, authored-module, model, tool, workflow, and client failures without string-rewriting unknown diagnostics;\n- deduplicates projections of the same root failure;\n- specifies where details live and how local session inspection reaches them.\n\n## Non-goals\n\nThis issue does not authorize a case-specific suppression for the instrumentation warning or an implementation of the diagnostics system. The first deliverable is a scoped research plan.

PR #844, opened 2026-07-15
feat(eve): define structured local diagnostics

ruiconti:
## What

- Define an eve-owned local diagnostic contract across compile, dev, tool/model, Workflow, and TUI boundaries.
- Specify a stable diagnostic catalog, one-line stderr, application/session ownership, bounded local detail storage, and diagnostic-ID deduplication.
- Capture instrumentation and handled-tool failures as acceptance cases without implementing case-specific fixes.

## Why

Current boundaries independently render upstream warnings, structured failures, and stacks. The plan separates failure outcomes from diagnostics and logs so one root incident stays terse in stderr while retaining local evidence.

Tracks #843.

### DIFF
--- a/research/structured-local-diagnostics.md
+++ b/research/structured-local-diagnostics.md
@@ -1,0 +1,192 @@
+---
+issue: https://github.com/vercel/eve/issues/843
+status: proposed
+last_updated: "2026-07-15"
+---
+
+# Structured local diagnostics
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


ruiconti:
Closing: the dev-diagnostics slice of this plan is now implemented in #881 (local per-process diagnostic log, concise stderr with stored detail). The broader diagnostic catalog / dedup contract can come back as a follow-up plan once that lands.

PR #850, opened 2026-07-16
feat(eve): improve the local authoring loop

ruiconti:
## What

- Record an automatic local diagnostic log for every root session by projecting Workflow lifecycle events and eve's durable session-event stream, plus process and sandbox output that has no session-event equivalent.
- Add `eve logs`, `eve logs <session-id>`, and `eve logs ls` for following the most recently used log, selecting an exact session, and listing local logs.
- Start `eve dev` directly after a successful `eve init`. Coding-agent sessions receive the build handoff instead of spawning a nested coding-agent REPL.
- Recognize inherited `AI_GATEWAY_API_KEY` and `VERCEL_OIDC_TOKEN` credentials, order model search results newest-first without hiding fast variants, and point the coding-agent handoff at eve's versioned skill.

## Automatic session logs

The previous inspection command read session metadata only after a run. It missed the stdout, errors, tool calls, timing, and durable events needed to understand the run itself.

`eve dev` now creates `.eve/logs/<session-id>.log` while the session runs. The CLI parent owns the file writer because it already owns the local Workflow World and survives Nitro worker reloads.

For model calls, tools, errors, usage, and timing, the recorder replays the session's default Workflow stream and keeps its reader open for new events. It derives time to first output, step duration, and tool duration from the durable event timestamps. Committed Workflow lifecycle events are reconciled from the World's event log. Stable completion markers let a restarted parent fill gaps without duplicating entries or maintaining a second event journal.

Process writes and sandbox backend output do not have a session-event equivalent. Workers send only those session-attributed output events through the authenticated development route. Strict Zod schemas validate that transport boundary. Sandbox backends publish output where they receive the native process events, so logging does not replace or wrap the public `SandboxSession` and `SandboxProcess` handles.

`eve logs` prints the last 10 lines for the most recently updated Workflow session with a log, then follows new writes. A session ID selects one file, `--no-follow` prints once, and `eve logs ls` lists available logs. Recording is on by default and can be disabled with `EVE_SESSION_LOGS=0` before `eve dev` starts. Log files request owner-only modes on POSIX systems; Windows access follows the log directory's ACL. Logs may contain unredacted runtime data.

## One post-init path

`eve init` now continues into `eve dev` in a human terminal. It no longer detects installed coding-agent CLIs, owns their invocation syntax, or opens a second agent session. When a coding agent invokes init, eve returns the existing build guide to that agent, including the local log commands used to inspect a representative run.

## Model setup

The model setup flow treats inherited Gateway credentials as configured before checking env files. Model search uses Gateway release metadata within each provider, so newer matching models appear first.

## Scope

TUI rendering changes are split into #851. The log recorder and commands in this PR cover local development only; they do not query deployed or remote sessions. This includes a patch changeset because the published `eve` package gains new CLI behavior and changes the post-init flow.


ruiconti:
Closing in favor of the split TUI work: #880 (/model controls), #881 (local diagnostics log), #882 (TUI rendering). The session-log projection, `eve logs`, and the init→dev handoff from this branch aren't carried over; if we still want them, they should come back as their own focused PRs on top of #881's sink.

PR #851, opened 2026-07-16
fix(eve): stabilize terminal rendering

ruiconti:
## What

- Align the live prompt with the submitted user message by removing the extra leading cell from prompt rows.
- Rebuild and replay committed transcript rows after a terminal resize so previous output wraps at the new width without duplicating the live prompt.
- Strip complete ANSI CSI sequences before filtering remaining terminal controls, preventing printable CSI suffixes from leaking into captured text.

The live input and submitted message previously used different left offsets. Resize only repainted the existing row model, so committed output kept its old wrapping. Terminal control filtering could also expose text such as `1m` and `22m` after removing the escape byte without consuming the full sequence.

These fixes stay at the renderer and terminal-text boundaries that own those behaviors. The init and local authoring changes remain in #850.

### DIFF
--- a/.changeset/tidy-terminal-rendering.md
+++ b/.changeset/tidy-terminal-rendering.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Keep the dev TUI prompt aligned with submitted messages, reflow the transcript after terminal resize, and strip complete ANSI sequences from captured output.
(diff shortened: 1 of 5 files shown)


ruiconti:
Closing in favor of #882, which reworks the same rendering surface (markdown, tool blocks, pulse). The terminal-text CSI-stripping and prompt-alignment fixes here aren't fully carried over — worth re-raising as a small focused PR if they still reproduce after #882.

PR #853, opened 2026-07-16
docs(frontend): document resumable session replay caveats

### LINKED ISSUE #729: docs: Resumable Sessions recipe needs the exact-mirror caveat + a documented per-event identity (turnId/sequence, verbatim replay) for client reconciliation
## Summary

`docs/guides/frontend/overview.mdx` §"Resumable sessions" tells database-backed chat apps to persist events per-arrival with `onEvent`, snapshot in `onFinish`, and "reconnect an interrupted in-flight turn with `session.stream({ startIndex: savedEvents.length })`". That recipe is correct **only while the saved log is an exact 1:1 mirror of the events received** — an invariant the docs never state, and one that real apps can't guarantee (a single failed persistence call breaks it permanently, because the count-as-cursor is then wrong forever and every reconnect replays events the app already saved).

We root-caused a production runaway to exactly this in an eve-chat-template app (full report with deterministic repro and a working fix: vercel-labs/eve-chat-template#23): failed snapshot → stale cursor → whole-log replay per send → unbounded duplication. Suggested doc additions below; all of them document guarantees eve already provides but only in `.d.ts` comments today.

## Suggested additions to the Resumable Sessions guide

1. **State the invariant and the failure mode.** `startIndex: savedEvents.length` assumes lossless, duplicate-free persistence. Say so, and say what happens otherwise (replays events the client already consumed — the client must reconcile idempotently, mirroring the durability contract's "an interrupted step re-runs, so make non-idempotent side effects idempotent", which currently reads as server-side-only guidance).
2. **Document the per-event identity clients should reconcile on.** Every turn-scoped content event carries `data.turnId` + `data.sequence`, and replays serve recorded events verbatim — "Replays must preserve the original `meta.at`… replay never invents new timestamps" lives only in `protocol/message.d.ts` doc comments. That contract is exactly what a client needs for safe dedup/reconciliation, and it deserves a docs surface. (Caveat worth documenting: `subagent.completed` carries `callId` but no `turnId`; `session.waiting` carries neither and is content-identical across turns.)
3. **Warn about the cumulative-delta footprint.** `message.appended`/`reasoning.appended` carry cumulative text, so a naively persisted raw event log is O(len²) in reply length per turn. Reasoning models emit ~10× the deltas of models with summarized/empty reasoning, which is how this crosses real-world transport limits (e.g. Next Server Actions' 1 MB default) — worth a sentence where the recipe says "persist stream events as they arrive."

## Related framework gap (optional, separate concern)

The Workflow SDK that backs eve's streams supports tail-relative reads precisely to avoid replaying "potentially thousands of chunks", but eve's client `stream({ startIndex })` only accepts an absolute index. For long reasoning-model sessions, exposing a tail-relative or server-authoritative resume would remove the client's cursor-bookkeeping burden entirely.

Happy to provide the full cited analysis or a docs PR.

yoshibase:
## Summary

Documents the resumable-sessions recipe caveats requested in #729:

- **Exact-mirror invariant** — `startIndex: savedEvents.length` only works when persistence is lossless and duplicate-free; cursor drift replays already-consumed events.
- **Event identity** — reconcile replays on `data.turnId` + `data.sequence`; notes exceptions for `subagent.completed` and `session.waiting`.
- **Cumulative deltas** — warns that naively persisting every `message.appended` / `reasoning.appended` event is O(len²) per turn.

Fixes #729

## Test plan

- [x] `node ./scripts/check-docs.mjs` passes
- [x] `node ./scripts/check-doc-snippets.mjs` passes
- [x] `node ./scripts/check-doc-mdx.mjs` passes

### DIFF
--- a/docs/guides/frontend/overview.mdx
+++ b/docs/guides/frontend/overview.mdx
@@ -259,6 +259,25 @@ For multiple chat threads, keep one saved event log and session cursor per threa
 
 If the user can refresh or navigate immediately after pressing send, create your app-level chat row and store the pending user message before calling `send()`. After the request starts, persist the session state as soon as it contains a `sessionId`, then reconnect an interrupted in-flight turn with `session.stream({ startIndex: savedEvents.length })` from the lower-level client.
 
+### Exact-mirror invariant
+
+`startIndex: savedEvents.length` assumes your saved log is an **exact 1:1 mirror** of the events your client received: lossless, in order, and without duplicates. The stream cursor is an absolute event count, not a logical message id.
+
+If one persistence call fails after you already advanced your UI, or you save a subset of events, the cursor drifts. The next reconnect replays events your app already consumed. Treat persistence as part of the durability contract: make client-side storage idempotent the same way server-side tool steps must be idempotent when a workflow step re-runs.
+
+### Event identity for reconciliation
+
+Turn-scoped content events carry `data.turnId` and `data.sequence`. Replays serve recorded events verbatim — eve preserves the original `meta.at` timestamp on replay rather than inventing new ones. Use `(turnId, sequence)` (and `type` when needed) to deduplicate replays safely.
+
+Caveats when building a dedup key:
+
+- `subagent.completed` carries `callId` but no `turnId`.
+- `session.waiting` carries neither `turnId` nor `sequence`, and is content-identical across turns (it only carries the next `continuationToken`).
+
+### Cumulative delta storage
+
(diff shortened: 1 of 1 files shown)


yoshibase:
Docs-only PR — DCO passes. Happy to revise wording if anything reads unclear.

yoshibase:
Thanks for flagging this. The reducer changes were unrelated branch pollution. I rebuilt the branch from current `main`; the PR is now a single documentation file and no longer changes `upsertRun` or its regression test.

PR #864, opened 2026-07-16
feat(eve): web search - exa provider

### LINKED ISSUE #688: Add Exa as a web search provider

Eve includes a built-in `web_search` tool. When using models with native web search, such as Anthropic, Google, OpenAI, or xAI models, Eve can use the provider's own search tool.

For models without native web search, developers still need a high-quality search backend. Developers who specifically want Exa web search currently need to replace or override Eve's built-in `web_search` tool themselves.

[Exa is now available through Vercel AI Gateway](https://vercel.com/docs/ai-gateway/models-and-providers/web-search#using-exa-search) via `gateway.tools.exaSearch()`, so Eve should be able to use Exa without requiring developers to integrate the Exa API, implement a custom tool, or bring an Exa API key.

### Proposed solution

Add `gateway.tools.exaSearch()` as a web search backend for AI Gateway models that do not provide their own native web search tool.

### Alternatives considered

Developers can already replace or override Eve's `web_search` tool with a custom Exa integration. However, this requires maintaining a separate tool implementation despite Exa already being available as a provider-managed search tool through AI Gateway. Integrating with Exa directly may also require developers to bring and manage a separate Exa API key.

AndrewBarba:
### Description

Closes #688.

Adds agent-level provider selection for eve's built-in `web_search` tool while keeping Parallel as the default for AI Gateway models.

Agents can opt into Exa with:

```ts
// agent/tools/web_search.ts
import { webSearch } from "eve/tools";

export default webSearch({ provider: "exa" });
```

`provider` accepts `"parallel"` or `"exa"`. Direct/BYO OpenAI, Anthropic, and Google models continue to use their native web search implementations, and a user-authored `web_search` tool still takes precedence.

For Exa, eve caps returned highlights at 1,000 characters per result while retaining up to 10 results, keeping multi-search turns bounded. The Exa E2E eval runs only in real-model suites and requests eight calls in one parallel batch, matching Gateway's bounded server-side tool loop.

This also adds manifest/compiler support, public API and compatibility coverage, documentation, and a changeset.

```sh
pnpm fmt
pnpm lint
pnpm typecheck
pnpm guard:invariants
pnpm docs:check
pnpm --filter eve exec vitest run \
  --config vitest.unit.config.ts \
  src/harness/provider-tools.test.ts \
  src/harness/tools.test.ts \
  src/public/tools/web-search.test.ts \
  src/compiler/normalize-manifest.test.ts \
  src/compiler/manifest.test.ts
```

The focused unit suite passes 75/75. Exa behavior was also validated against artifacts from the real-model CI E2E suite.

### PR Checklist


ruiconti (approved):

kesku:
Hi! I authored the PR this is based on and dug into the CI failures.

From what I can tell the Exa integration itself looks fine. The failures seem to come from model adherence rather than the implementation?

One eval had the model emit 9 `web_search` calls instead of the requested 10 (but all 9 completed successfully!). Another had the parent model invent an `outputSchema` for a subagent that wasn't asked for structured output, so the child returned `SUBAGENT_EXECUTION_FAILED` when it responded with prose.

The remaining HITL failure was an eval expectation issue I think, the prompt never instructed the model to repeat the selected color

PR #845, opened 2026-07-15
feat(eve): reconcile Vercel framework preset on channels add web

OwenKephart:
### Description

In the flow where you: 1. `eve init` with no channels 2. `eve channels add web` to add a next.js project into the scaffolded repo, you end up hitting an error on deployment (No Output Directory named ".output" found).

This is because the linked Vercel project 

### How did you test your changes?

Added tests, and ran through the `eve init` / `eve channels add web` flow manually, confirmed deployment was successful.

### PR Checklist

- [ ] I linked an issue with prior discussion confirming this change is wanted
- [ ] I ran the relevant checks from `CONTRIBUTING.md`
- [ ] I added tests and documentation where relevant
- [ ] I added a changeset if this touches the published `eve` package
- [ ] DCO sign-off passes for every commit (`git commit --signoff`)

PR #868, opened 2026-07-16
fix(eve): update `/channels` flow to correctly reconcile Vercel project settings Signed-off-by: owenkephart <owen.kephart@vercel.com>

OwenKephart:
### Description

Follow-on to: https://github.com/vercel/eve/pull/845, fixes the path for the `/channels` flow within the TUI and centralizes the logic in the deploy box to prevent similar drift in the future

### PR Checklist


OwenKephart:
* **#868** <a href="https://app.graphite.com/github/pr/vercel/eve/868?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/eve/868?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/eve/src/setup/boxes/deploy-project.ts
```diff
@@ -130,6 +141,16 @@ export function deployProject(
...
+      // the host framework on disk before deploying. Best-effort and a no-op for a
+      // plain agent or an already-correct preset.
+      await deps.syncHostFrameworkPreset(
+        options.prompter,
+        projectPath,
+        onOutput,
+        signal ? { signal } : {},
+      );
+
       if (!state.deploymentDependenciesInstalled) {
```
ruiconti, line 156: Wonder if this should happen right after we're making changes that are known to cause the drift vs pre-deploy. Otherwise, the following scenario would still be broken

```
eve dev
# adds a channel via /channels
vercel deploy # doesn't run `/deploy`
```

ruiconti (approved):

packages/eve/src/setup/boxes/deploy-project.ts
OwenKephart: so this is a very valid concern, however i think mutating project settings outside of an explicit deploy step is too scary, as it means users in local dev that do not intend to deploy could end up messing with a real deployed app

PR #870, opened 2026-07-16
feat(eve): add channel cancellation api

### LINKED ISSUE #483: No way to cancel an in-flight turn: stop() only detaches the client; the model call keeps running (and billing)

AndrewBarba:
Implements [research/turn-cancellation-channel-ops.md](https://github.com/vercel/eve/blob/barba/channel-cancel-helper/research/turn-cancellation-channel-ops.md).

Closes #483

## What

Custom channel authors can now cancel a session's in-flight turn through two public surfaces:

- **`cancel({ continuationToken, turnId? })`** on `RouteHandlerArgs` (HTTP and WebSocket routes) — addressed by the same channel-local continuation token used with `send()`. Never starts a session, sends input, or clears history.
- **`Session.cancel({ turnId? })`** on the handles returned by `send()` and `getSession()` — session-id-addressed.
- **`ClientSession.cancel({ turnId? })`** — the TypeScript client forwards the same optional stale-request guard to the cancel route (previously it could not send one).

Both delegate to the existing cancellation primitive and reuse the existing `CancelTurnResult` (`"accepted" | "no_active_turn"`; both statuses are successful, and confirmation is `turn.cancelled` → `session.waiting` on the event stream).

## How

Token resolution is a delivery-free Workflow hook-by-token read wrapped in a new eve-owned `Runtime.resolveSession(continuationToken)` method (`hook.runId` is the session id). Because it is a pure read, nothing is consumed, adapter delivery and the input queue are untouched, and no new hook payload kind or durable driver/turn protocol change is required. An unknown or retired token maps to `no_active_turn`; other lookup failures propagate and never fall back to session creation. The already-landed cancellation protocol and events are unchanged.

## Testing

- **Unit**: `channel/cancel.test.ts` (namespacing, guard forwarding, unresolved-token short-circuit, status passthrough, error propagation with no run/deliver fallback), `Session.cancel` cases in `channel/session.test.ts`, `resolveSession` cases in `workflow-runtime.test.ts`.
- **Integration**: new case in `turn-cancellation.integration.test.ts` exercising real hook-by-token resolution against the workflow world — unknown token, mid-tool cancel with the boundary sequence, benign `Session.cancel()` duplicate after sweep, then normal resume.
- **E2E**: new `threads` channel in the `agent-cancellation` fixture with `/threads/:threadId/messages` and `/threads/:threadId/stop` routes using only public APIs, plus `cancel-channel-route.eval.ts` — hangs mid-tool, stops via the helper, asserts `turn.cancelled` → `session.waiting` with zero failure events, and proves the same thread resumes the same session.

`fmt`, `lint`, `guard:invariants`, `docs:check`, workspace `typecheck`, the full unit tier, and the turn-cancellation integration file all pass locally.

## Docs

- `docs/channels/custom.mdx`: `cancel` in the route-helpers list, new "Cancel a turn" section, WS helper list updated.
- `docs/concepts/sessions-runs-and-streaming.md`: pointer from the HTTP cancel contract to the channel surfaces.
- Patch changeset included.


ctgowrie (approved):

De-ASI-INTERFACE:
⚠️ Architectural Feedback: ⁠feat(eve): add channel cancellation api⁠

1. API Surface Bloat and Tight Coupling
Exposing cancellation across three separate public surfaces (⁠RouteHandlerArgs⁠, ⁠Session⁠, and ⁠ClientSession⁠) introduces significant API fragmentation. Instead of isolating cancellation logic within a unified context or manager, this approach requires custom channel authors to understand and integrate three different interfaces. This increases the long-term maintenance overhead and elevates the likelihood of upstream integration breaking changes.

De-ASI-INTERFACE:
2. Race Conditions and Concurrency Risks in Transport Delegation
The ⁠RouteHandlerArgs⁠ surface attempts to abstract cancellation over both HTTP and WebSocket routes using a channel-local ⁠continuationToken⁠.
 State Synchronization: In flight, a turn executed over a persistent WebSocket connection decoupled from an asynchronous HTTP POST cancellation request introduces a distributed synchronization boundary.
 Locking Mechanisms: Without explicit transactional locks or atomic state checks at the core primitive level, a cancellation request can easily race against a rapidly finalizing turn, leading to unhandled edge cases where a turn is marked ⁠"accepted"⁠ for cancellation after execution has already committed downstream.

De-ASI-INTERFACE:
3. Stale-Request Guard Vulnerabilities
The introduction of an optional stale-request guard forwarded by the TypeScript client (⁠ClientSession.cancel⁠) presents architectural risks:
 Determinism: Making the guard optional means the system fails to strictly guarantee client-server state consistency.
 Network Partitions: If a client experiences a temporary network partition or drops the guard parameter, the backend primitive may process a stale or out-of-order cancellation payload. This can result in orphaned operations, silently leaked server-side routines, or corrupted session history.

De-ASI-INTERFACE:
4. Insufficient Granularity in Result Types
Reusing the coarse ⁠CancelTurnResult⁠ (⁠"accepted" | "no_active_turn"⁠) across all three surfaces lacks the fidelity required for robust error handling.
 ⁠"no_active_turn"⁠ is ambiguous; it fails to differentiate between a turn that already completed successfully, a turn that never existed due to an invalid ⁠turnId⁠, or a turn that was already terminated by a prior request.
 A robust implementation demands granular variants (e.g., ⁠TurnAlreadyCompleted⁠, ⁠TokenMismatch⁠, ⁠CancellationPending⁠) to allow callers to safely resolve client-side state machine transitions.

De-ASI-INTERFACE:
No

PR #896, opened 2026-07-17
feat(eve): export SessionNotReadyError for route auth readiness

### LINKED ISSUE #724: Expose a typed retryable readiness error for route auth

yoshibase:
## Summary

Closes #724.

Route auth had no public way to signal "caller is authenticated, but this session isn't readable yet." Apps had to duck-type a private `Response` to get HTTP 425, which the native stream opener already retries.

This adds:

- `createReadinessResponse()` — framework-shaped `425` with `cache-control: no-store`, code `session_not_ready`, optional `Retry-After`
- `SessionNotReadyError` — throw from an `AuthFn`; `routeAuth` returns the response and stops the walk (same pattern as `ForbiddenError`)
- Unit tests for the new error, `routeAuth` preservation, and `openStreamBody` 425 retry / abort behavior
- Auth guide + changeset

No new transport or client retry logic — the client already treats 425 as retryable.

## Test plan

```bash
cd packages/eve
CI=true pnpm run build:compiled
CI=true pnpm exec vitest run --config vitest.unit.config.ts \
  src/public/channels/auth.test.ts \
  src/client/open-stream.test.ts
node ../../scripts/check-docs.mjs
```

- [x] Auth tests cover `createReadinessResponse`, `SessionNotReadyError`, and `routeAuth` 425 termination
- [x] `open-stream.test.ts` covers 425 retry, non-retryable 403, and abort-before-fetch
- [x] `docs:check` passes

### DIFF
--- a/.changeset/session-not-ready-auth-error.md
+++ b/.changeset/session-not-ready-auth-error.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Export `SessionNotReadyError` and `createReadinessResponse` from `eve/channels/auth` so route auth can return a typed, retryable `425` readiness failure.
--- a/docs/guides/auth-and-route-protection.md
+++ b/docs/guides/auth-and-route-protection.md
@@ -71,13 +71,17 @@ Put your own providers ahead of the catch-all helpers. `localDev()` is the final
 To reject with a precise status instead of skipping, throw:
 
 ```ts
-import { ForbiddenError, UnauthenticatedError } from "eve/channels/auth";
+import { ForbiddenError, SessionNotReadyError, UnauthenticatedError } from "eve/channels/auth";
 
 throw new UnauthenticatedError({
   code: "authentication_required",
   message: "Sign in to continue.",
 }); // 401
 throw new ForbiddenError({ message: "Not allowed on this workspace." }); // 403
+throw new SessionNotReadyError({
+  message: "Session projection is still catching up.",
+  retryAfterSeconds: 1,
+}); // 425 — native clients retry this status
 ```
 
 Any other thrown error follows the normal channel failure path. When building a custom channel on `defineChannel`, call `routeAuth(request, auth)` from `eve/channels/auth` to reuse the same walk semantics.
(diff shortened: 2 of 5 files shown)


yoshibase:
CI is green on my side (DCO, Socket, Agent Review). Happy to adjust if anything looks off.

yoshibase:
Rebased onto latest `main` (was 51 commits behind). Resolved a small import conflict in `auth.test.ts` (`UnauthorizedChallenge` + `SessionNotReadyError`).

Local verification after rebase:
- `auth.test.ts` + `open-stream.test.ts`: 99 passed
- `node ./scripts/check-docs.mjs`: 77 files validated

Commit is DCO-signed with no tooling co-authors. Happy to adjust if anything looks off.

PR #878, opened 2026-07-16
chore(eve): e2e - add agent-browser extension fixture

AndrewBarba:
## Summary

- add a standalone e2e fixture mounting the published @agent-browser/eve extension
- preinstall agent-browser and Chromium through the extension sandbox helpers
- exercise navigation, accessibility snapshots, ref-based clicking, DOM evaluation, and cleanup against a deterministic inline page
- pin the integration packages and admit only those packages through the workspace release-age policy

## Validation

- pnpm install --frozen-lockfile
- pnpm --filter eve run build
- pnpm --filter agent-browser-extension run build
- pnpm --filter agent-browser-extension exec tsc --noEmit
- pnpm --filter agent-browser-extension exec eve eval --list
- pnpm exec oxfmt --check e2e/fixtures/agent-browser-extension
- pnpm exec oxlint e2e/fixtures/agent-browser-extension
- pnpm guard:invariants
- pnpm guard:fixtures

The model-backed browser eval is intentionally left to the repository e2e CI matrix.

PR #898, opened 2026-07-17
fix(eve): resolve source-backed extension dependencies from the package's real location

ctgowrie:
## Problem

An extension installed with a registry-style store layout fails in the dev runtime (`eve dev` / `eve eval`). pnpm's virtual-store layout puts the extension's dependencies next to the package copy in the store, not inside it — they only resolve from the package's real location. The compiled module map imports extension modules by literal path through the consumer's `node_modules` symlink, and the dev-generation snapshot relocates every importer path while reaching the package only through that symlink — so bare imports like `import "zod"` can't resolve: rolldown marks them external (`UNRESOLVED_IMPORT`) and the materialized generation crashes at import time with `ERR_MODULE_NOT_FOUND`.

Workspace-linked extension fixtures hid this because pnpm links a workspace package's dependencies physically inside the package directory, so bare imports resolve from anywhere. This is why the existing `extensions` fixture passed while a registry install of the same shape (e.g. `@agent-browser/eve`, #878) failed.

## Fix

`packages/eve/src/internal/authored-package-boundary.ts`: when default resolution of an ordinary package import fails, the generation package boundary retries from the importer's realpath, so the dependency resolves and inlines from where it actually lives. This restores standard resolver semantics (Node, webpack, esbuild, and rolldown all realpath resolved modules by default) only where eve's literal-path module ids bypass them, without churning module identity. Prod (`eve build`) was never affected because it bundles in place.

## E2E coverage

Pushed first to prove the failure in CI (see comments for the failing-run evidence):

- `gizmo-extension` and a new `gadget-extension` are buildable extension packages (`eve extension build`) in published form: `exports` → `dist/`, `files` whitelist, pinned deps. No committed dist — the e2e workflows run the extension builds, then re-pack the `file:`-protocol store copies (a plain reinstall short-circuits and leaves them stale).
- New `dist-extensions` fixture consumes both via `file:`, which makes pnpm copy them into the virtual store like registry packages (dependencies as store siblings) without publishing anything.
- Coverage matrix: gizmo is workspace-mounted in `extensions` **and** store-installed in `dist-extensions` (both layouts); gadget is store-installed only; toolkit stays workspace-only.
- Scenario test reproduces the store layout + literal symlink path import; it fails without the fix and passes with it.

## Verification

- CI on the fixture-only commit (4a26455): `dist-extensions` failed both e2e workflows with `UNRESOLVED_IMPORT: zod/slugify` + `ERR_MODULE_NOT_FOUND`.
- CI on the fix commit (27b9464): `dist-extensions` and `extensions` pass on both model jobs in e2e-local and e2e-vercel.


ctgowrie:
CI confirms the fixture reproduces the failure (run 29579444403, both model matrix jobs):

```
node_modules/gadget-extension/extension/tools/gadget_echo.ts (1:20) [UNRESOLVED_IMPORT] Could not resolve 'slugify' in node_modules/gadget-extension/extension/tools/gadget_echo.ts
node_modules/gadget-extension/extension/tools/gadget_echo.ts (3:18) [UNRESOLVED_IMPORT] Could not resolve 'zod' in node_modules/gadget-extension/extension/tools/gadget_echo.ts
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'slugify' imported from .../.eve/dev-runtime/snapshots/<id>/source/e2e/fixtures/registry-extension/.eve/compile/authored-modules/<hash>.mjs
```

Same failure chain as `@agent-browser/eve` on #878: the generation snapshot relocates the store-installed extension's source, the authored-module compile leaves its bare imports unresolved, and the compiled module crashes at import time.

ctgowrie:
Fix verified in CI on 27b9464:

- `e2e-local (dist-extensions)` and `e2e-vercel (dist-extensions)` pass on both model jobs (failed with `UNRESOLVED_IMPORT`/`ERR_MODULE_NOT_FOUND` on the previous commit).
- `e2e-local (extensions)` and `e2e-vercel (extensions)` stay green with gizmo consumed through its built dist in the workspace layout.
- The two remaining matrix failures were unrelated model flakes (`agent-tools` web-search eval hit a terminated model call on openai; `agent-subagents` dynamic-workflow called subagents 4× instead of 2×) — re-ran those jobs.

AndrewBarba (approved):

PR #904, opened 2026-07-17
feat(eve): extension build contribution manifest and beta peer range

ctgowrie:
First slice of source-free extension distribution (splitting up `cgowrie/source-free-extensions`) — the **producer side only**. Consumers keep recompiling extension source; nothing reads the new artifact yet.

## What's in this PR

**`eve extension build` stamps a contribution manifest.** The build now writes `dist/_ext-manifest.json`: the extension's compiled contribution metadata (base names, authored logical paths, descriptions, JSON schemas) plus two compatibility stamps:

- `eveVersion` — the eve the extension was built with.
- `capabilityVersions` — a hand-maintained contract version per capability the extension uses (`EXTENSION_CAPABILITY_VERSIONS`). `config` and `state` always stamp (the mount factory can bind config and any module may scope state); the rest stamp only when the extension ships one.

Producing the manifest compiles the extension's contributions (loading its modules), so `tryReadExtensionBuildConfig` now collects declared dependency names and keeps them external during that load.

**`eve extension init` scaffolds the peer range as `>=<installed-version> <1`.** The installed eve becomes the beta extension's compatibility floor while allowing later pre-1.0 releases (`formatExtensionEvePeerSpecifier`); non-registry specifiers pass through for local/tarball scaffolds.

**Refactor:** `composeManifestContributions` splits into `produceBaseContributions` (base names, no rebase) + `namespaceContributions`, so the build can serialize base-named contributions via the new `produceExtensionArtifactContributions`. Consumer composition behavior is unchanged (existing normalize-extension tests untouched and green).

**e2e inclusivity:** gizmo/gadget fixtures adopt the scaffolded peer-range shape, so the existing e2e matrix (workspace + store-installed consumption, from #898) now exercises the artifact-emitting build and a real peer range end to end.

## Not in this PR (later slices)

Source-free discovery/composition (consuming `_ext-manifest.json` without source), the graph-bundled dist + declaration emit, dist-only fixtures/staging, and the consumer-side compatibility checks that read these stamps.

## Verification

- New unit tests: `extension-artifact.test.ts` (round-trip, malformed input, version-skew error), `extension-capabilities.test.ts`; new scenario assertion for the stamped manifest contents.
- `mounted-extension-installed` scenario ships the full recursive `dist/` (now including the manifest) and stays green — consumers still compile source.
- Full unit (4866) + integration (478) suites, typecheck, lint, guard, docs:check pass.
- Local e2e boot-checks: `dist-extensions` and `extensions` fixtures build with the artifact-emitting producer, boot, and `eve build` passes.

### DIFF
--- a/.changeset/eager-planets-carry.md
+++ b/.changeset/eager-planets-carry.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+`eve extension build` now writes a `dist/_ext-manifest.json` contribution manifest stamped with the building eve version and a contract version for each capability the extension uses. `eve extension init` scaffolds the eve peer range as `>=<installed-version> <1` so beta extensions declare their compatibility floor.
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -79,7 +79,7 @@ Point `eve.extension` at the source directory and run `eve extension build` (wir
   "type": "module",
   "eve": { "extension": "./extension" },
   "files": ["extension", "dist"],
-  "peerDependencies": { "eve": "^x" },
+  "peerDependencies": { "eve": ">=0.24.6 <1" },
   "dependencies": { "zod": "^3" },
   "scripts": { "build": "eve extension build", "prepare": "eve extension build" },
 }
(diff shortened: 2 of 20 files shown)


ctgowrie:
CI is fully green: extensions/dist-extensions pass on both models in both e2e workflows against the manifest-emitting build and the new peer ranges. The two flaky evals that needed reruns (`agent-tools` web-search, `agent-tools-hitl` stale-ask-question-selection-idle closedQA judge on openai-sol) are unrelated fixtures untouched by this PR — the hitl judge flaked twice before passing and may deserve a stabilization pass separately.

PR #915, opened 2026-07-17
fix(eve): (server) guarantee session event streams end in a terminal boundary event

OwenKephart:
Server half of the stream-liveness fix (client half: #919).

A workflow run's durable event log can settle without a session boundary event when the run is terminated outside eve's own code paths — platform-level cancellation, run expiry, or the terminal-failure emitter itself dying. Consumers then can't tell "turn still running" from "run is gone" by transport behavior alone.

- **Terminal framing**: served event streams now append a synthesized `session.failed` / `session.completed` (derived from `getRun(id).status`) when a settled run's log ends without a boundary event. Implemented as a small `TransformStream.flush` wrapper (`execution/terminal-framing.ts`) on both stream-serving paths.
- **App function** **`maxDuration: "max"`**: the nitro `__server.func` that serves `GET /eve/v1/session/:id/stream` previously ran at the platform default (~300s), so every open stream was severed every ~5 minutes while the flow function ran at `"max"`. Patched post-build, same mechanism as the flow function.

Together with #919 this establishes the stream contract: _a settled run's stream always ends in a boundary event; everything else is a transport cut the client reconnects through._

### DIFF
--- a/.changeset/terminal-stream-framing.md
+++ b/.changeset/terminal-stream-framing.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Session event streams are now guaranteed to end in a terminal boundary event: when a settled workflow run's durable log ends without one (for example after a platform-level cancellation), the stream appends a synthesized `session.failed` or `session.completed` derived from the run status. The deployed app function's `maxDuration` is also raised to `"max"`, so long event streams are cut at the plan's maximum window instead of the platform default.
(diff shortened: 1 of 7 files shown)


OwenKephart:
* **#919** <a href="https://app.graphite.com/github/pr/vercel/eve/919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#915** <a href="https://app.graphite.com/github/pr/vercel/eve/915?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/eve/915?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #917, opened 2026-07-17
fix(eve): advertise the configured auth scheme in routeAuth

### LINKED ISSUE #909: httpBasic() failures advertise WWW-Authenticate: Bearer instead of Basic
## Summary

A route protected only by `httpBasic()` returns a `401` with:

```http
WWW-Authenticate: Bearer
```

even though the route accepts HTTP Basic credentials, not a bearer token.

This prevents HTTP clients, proxies, and application UIs from discovering the authentication scheme from the standard challenge, and a direct browser navigation cannot present its normal Basic authentication prompt.

## Reproduction

Using `eve@0.24.6`:

```ts
import { httpBasic, routeAuth } from "eve/channels/auth";

const response = await routeAuth(
  new Request("https://agent.example/eve/v1/session"),
  httpBasic({ username: "eve", password: "eve" }),
);

console.log({
  status: response.status,
  challenge: response.headers.get("www-authenticate"),
  body: await response.json(),
});
```

Actual result:

```json
{
  "status": 401,
  "challenge": "Bearer",
  "body": {
    "code": "unauthorized",
    "error": "Authorization is required for this route.",
    "ok": false
  }
}
```

The same behavior is observable through an `eveChannel` configured with only:

```ts
eveChannel({
  auth: [httpBasic({ username: "eve", password: "eve" })],
});
```

## Expected behavior

The failure should advertise the scheme the route actually accepts, for example:

```http
WWW-Authenticate: Basic realm="agent", charset="UTF-8"
```

For an auth array with multiple supported schemes, the response should ideally advertise all applicable challenges. At minimum, a Bearer challenge should not be emitted when no bearer-compatible auth strategy is configured.

## Current cause

`httpBasic()` returns `null` when verification fails. After every auth function returns `null`, `routeAuth()` creates the generic unauthorized response with a hard-coded Bearer challenge:

```ts
return createUnauthorizedResponse({
  challenges: [{ scheme: "Bearer" }],
});
```

## Possible direction

One option is to let auth strategies declare their challenge metadata and have `routeAuth()` collect the challenges for strategies that did not authenticate the request. That would preserve fallback auth arrays without making a failed `httpBasic()` terminate the auth walk.

It would also be useful to cover:

- `httpBasic()` alone advertises Basic;
- bearer/OIDC strategies advertise Bearer;
- mixed Basic + Bearer arrays advertise both;
- successful authentication includes no challenge;
- custom `UnauthenticatedError` responses continue to take precedence.

## Version

Observed with `eve@0.24.6`.

NicoMoli:
## Summary

Routes protected only by httpBasic() returned 401 with a hard-coded WWW-Authenticate: Bearer, so browsers never showed their Basic-auth prompt and clients could not discover the scheme.

Auth strategies now declare their challenge metadata via withAuthChallenges(), and routeAuth() collects the declared challenges from the configured entries on an exhausted walk (deduped, order preserved), falling back to Bearer only when none declare a scheme. httpBasic() gains an optional realm option.

Fixes #909

### PR Checklist


AndrewBarba:
Re-homed as #1001 on an in-repository branch so the full CI suite can run.

Thank you for the contribution, @NicoMoli — the replacement commit preserves your work and credits you as a co-author.

NicoMoli:
> Re-homed as #1001 on an in-repository branch so the full CI suite can run.
> 
> Thank you for the contribution, @NicoMoli — the replacement commit preserves your work and credits you as a co-author.

Thanks for the support and the time for review this PR @AndrewBarba!

PR #919, opened 2026-07-17
fix(eve): follow durable event streams across connection cuts

OwenKephart:
### Description

High-level changes here:

- Refactors stream following logic to be shared between send() / stream() (previously these treated stream interruptions subtly differently depending on if it was a clean EOF vs. a disconnect error)
- Removes the `maxReconnectAttempts` option from the public API, as the reconnection logic was changed to be more predictable / stable, described below
    - note that there are a wide variety of potentially-useful-to-configure parameters beyond just the raw count of reconnection attempts (e.g. backoff params). my take is that we should either expose all of them or none of them, and exposing none is fine for now, given reasonable defaults.

**Previous State of the World**

- the max duration of the stream-serving function in an eve deployment is set to the default value (300s), whereas turn workflows are allowed to last as long as the plan's maximum (300-800s).
- this means that, during a long running (>300s) turn, if you are following the stream (using either send() or stream()), it is expected that there will be EOFs / disconnect errors that we will need to handle / interpret as something other than "the external work is done". Other infra / transient issues can also cause these issues, not just standard function shutdowns, but just noting that these cuts are _guaranteed_ to occur once you hit a certain length.
- the previous logic was to have a "reconnect budget" (defaulting to 3) that would be eaten up as these issues occurred. once the budget was depleted, we would completely lose the session from the tailer, and present this to the user (in the TUI, at least) that the session was completely lost
    - notably, this budget was never reset during the turn. i.e. if we had an interrupt at the beginning of a turn, then reconnected, then got successful messages after that, then disconnected again, etc., these would all drain the reconnection budget even though the underlying workflow was clearly still alive
- in reality, work was still ongoing (if slow), but the user was left in a frustrating / somewhat unrecoverable state

**New State of the World**

- When the stream follower hits a 'cut' (EOF/steam disconnect error), it will reconnect, but it will reset its reconnection budget if it gets a new event (which proves that the underlying workflow is still alive)
- It will only truly end tracking if it hits 5 reconnect errors in a row without receiving a new event
  - if the underlying workflow run is truly dead but with no boundary event, the fn will return a clean EOF ~instantly, meaning given the exp backoff it's around 8 seconds of delay to fully give up on the stream. note that this should be a quite rare case (workflow dying with no terminal event), so this seems acceptable
- If this happens but the stream is actually alive (which should be rarer now), in the dev TUI, we give the user a more informative message and default to reconnecting to the in-progress session on the next message rather than forcing them to abandon the session entirely

### PR Checklist

### DIFF
--- a/.changeset/follow-until-boundary.md
+++ b/.changeset/follow-until-boundary.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
+The client now reconnects durable event streams from their last cursor, so long turns continue across transient connection cuts without replaying events. Stream retries are now managed internally, interrupted sessions remain resumable, and `maxReconnectAttempts` has been removed from `ClientOptions`, `EveAgentStoreInit`, and the React, Svelte, and Vue `UseEveAgentOptions` APIs.
--- a/docs/concepts/sessions-runs-and-streaming.md
+++ b/docs/concepts/sessions-runs-and-streaming.md
@@ -110,7 +110,7 @@ Custom channel routes request the same cancellation without knowing the session 
 
 ## Reconnect and rewind
 
-The stream is durable. Every event is recorded before a step completes, so the whole stream is replayable. A nonnegative `startIndex` is an absolute event count: use it to pick up where you dropped off or pass `0` to rewind to the start.
+The stream is durable. Every event is recorded before a step completes, so consumers can reconnect from their cursor when an HTTP connection ends. A nonnegative `startIndex` is an absolute event count: use it to pick up where you dropped off or pass `0` to rewind to the start.
 
 ```bash
 curl "http://127.0.0.1:3000/eve/v1/session/<sessionId>/stream?startIndex=<count>"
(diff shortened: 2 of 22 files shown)


OwenKephart:
* **#936** <a href="https://app.graphite.com/github/pr/vercel/eve/936?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#919** <a href="https://app.graphite.com/github/pr/vercel/eve/919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/eve/919?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changeset/follow-until-boundary.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": minor
```
AndrewBarba, line 2: fine to make this patch

AndrewBarba (approved):

PR #928, opened 2026-07-18
feat(eve): add onEvent hook to slackChannel for arbitrary Slack events

bensabic:
`slackChannel` previously only surfaced Slack events it had dedicated pipelines for — `app_mention` via `onAppMention`, IM messages via `onDirectMessage`, and `block_actions` via `onInteraction`. Every other Events API callback (`reaction_added`, `channel_created`, `team_join`, …) was silently dropped at the webhook.

This adds an `onEvent(ctx, event)` hook that observes custom Events API callbacks — event types with no dedicated handler:

```ts
import { slackChannel } from "eve/channels/slack";

export default slackChannel({
  onEvent(ctx, event) {
    if (event.type === "reaction_added" && event.event.reaction === "white_check_mark") {
      // ctx.thread is bound to the reacted message's channel/thread
      return ctx.thread.post("Marking this as resolved.");
    }
  },
});
```

Events are normalized into a channel-owned `SlackEvent` shape (also exported):

```ts
interface SlackEvent {
  type: string;                        // inner event type, e.g. "reaction_added"
  event: Record<string, unknown>;      // inner `event` object, verbatim
  teamId: string | undefined;
  eventId: string | undefined;         // envelope event_id (used for dedup)
  eventTime: number | undefined;
  channelId: string | undefined;       // derived from event.channel or event.item.channel
  threadTs: string | undefined;        // thread_ts ?? ts ?? item.ts
}
```

## Motivation

The Events API is much larger than the three surfaces the channel dispatches on. Agents commonly want to react to reactions (approval/resolution flows), membership changes, or channel lifecycle events without eve needing a bespoke pipeline for each. `onEvent` is the escape hatch: subscribe the Slack app to the event type, handle it in one place, and use `ctx.slack.request(...)` (or `ctx.thread` when the event carries a channel) for side effects.

## How it works

- The vendored webhook parser already preserves unknown event types as `kind: "unsupported"` with the full envelope intact, so no vendored code changed. A new `slackEventFromWebhookPayload` normalizer in `inbound.ts` extracts the inner event and envelope fields.
- **Events with a dedicated hook never reach `onEvent`**: `app_mention` and IM `message` events route exclusively to `onAppMention` / `onDirectMessage`, so the two surfaces can't double up on the same event. Non-IM channel `message` events (no dedicated handler) do reach it.
- `onEvent` runs on the inbound webhook side under `waitUntil()`, after `event_id` dedup, so Slack gets its `200 OK` immediately. Handler errors are caught and logged and never affect the ACK or the mention/DM pipelines.
- `ctx` is a standard pre-dispatch `SlackContext` rebuilt best-effort from the event's channel and thread; events with no channel still get the raw `ctx.slack.request(...)` escape hatch.

## Effect on current behavior

- **No `onEvent` configured:** the inbound path is unchanged — unknown events are dropped exactly as before.
- **`onEvent` configured:** it is a pure observer for uncovered event types. It cannot dispatch a turn.
- Interactivity payloads (`block_actions`, view submissions) are not Events API events and continue to route to `onInteraction` / HITL only.
- Event-id dedup now also covers observed events, so Slack redeliveries don't re-invoke the hook.

Includes docs (`docs/channels/slack.mdx` dispatch section with an example), unit tests covering the pipeline (including that covered events do NOT reach `onEvent`), the `SlackEvent` export, and a patch changeset.


ruiconti:
This has been merged in https://github.com/vercel/eve/pull/1044.

PR #939, opened 2026-07-19
feat(eve): identify eve on AI Gateway, Workflow, and Sandbox requests

ruiconti:
Vercel services can't tell eve traffic from any other AI SDK app today. This appends an `eve/<installed version>` User-Agent product token to every service request eve makes, so receivers can attribute traffic to eve and its version — segmenting by framework during incidents and, later, controlling traffic keyed on the same token.

## What

- `internal/user-agent.ts`: builds the `eve/<version>` product token, appends it to a `Headers` without touching existing products (idempotent, space-joined per the RFC 9110 UA grammar — not `Headers.append()`, whose comma-combination breaks whitespace-delimited parsers), and exposes `withPackageUserAgent(fetch)`, the generic UA-decorating fetch wrapper both the Gateway and Sandbox transports build on.
- `internal/gateway.ts` owns the Gateway transport surface: the `/v1/models` and `/v1/models/catalog` URLs, `getVercelGatewayFetch()` (a UA-decorated fetch, mirroring the Sandbox binding's `getVercelSandboxFetch`), and `resolveProviderHeaders(model)` — headers for Gateway-routed models (bare ids and `gateway.*` instances), `undefined` for direct providers.
- AI Gateway call sites: tool-loop generation and compaction, autoeval judge calls, model catalog, setup catalog, CLI model lookup, API-key validation — all through the two helpers above; no call site does header plumbing or URL assembly.
- Workflow: generated world-vercel startup code passes the token via `createWorld({ headers })`, with the per-world wiring dispatched in one place (`resolveWorkflowWorldWiring`).
- Sandbox: the SDK fetch decoration reuses `withPackageUserAgent`. A separate commit fixes a real gap here: `Sandbox.get()` lost the custom fetch on the credential-discovery failure path.

## Wire format

Captured against a local server through the real `ai`/`@ai-sdk/gateway` packages — existing products stay intact:

```
eve/0.24.6 ai/7.0.26 ai-sdk/provider-utils/5.0.9 runtime/node.js/26
```

## The Workflow leg is inert until a vendored bump

The vendored `@workflow/world-vercel@5.0.0-beta.31` overwrites caller User-Agents, so the header this PR passes is dropped on the wire today (requests degrade to current behavior — nothing fails). vercel/workflow#2998 fixes the append; a follow-up bump of the vendored package lights this leg up.

## Compaction attribution fix

Gateway attribution headers for compaction were previously computed from the main model; they now derive from the compaction model itself, since the two can differ.


packages/eve/src/compiler/model-catalog.ts
```diff
@@ -213,7 +212,7 @@ type CompiledRuntimeModelCatalogCache = z.infer<typeof compiledRuntimeModelCatal
 async function fetchAndPersistModelCatalog(
   appRoot: string,
 ): Promise<CompiledRuntimeModelCatalogCache> {
-  const response = await fetch(AI_GATEWAY_MODELS_CATALOG_URL);
+  const response = await getVercelGatewayFetch()(AI_GATEWAY_MODELS_CATALOG_URL);
```
ctgowrie: Why does this return a function instead of just making the fetch?

packages/eve/src/internal/user-agent.ts
```diff
@@ -0,0 +1,44 @@
+import { resolveInstalledPackageInfo } from "#internal/application/package.js";
+
+/** Product token for the installed package, such as `package-name/1.2.3`. */
+export function buildPackageUserAgent(): string {
+  const { name, version } = resolveInstalledPackageInfo();
+  return `${name}/${version}`;
```
ctgowrie, line 6: Should we include agent name as well?

packages/eve/src/compiler/model-catalog.ts
ruiconti: Was following the existing constructor the precedence -- `getVercelSandboxFetch()` -- but agree that's not required

packages/eve/src/internal/user-agent.ts
ruiconti: We can infer project via deploymentId which is already part of User-Agent. Unless you were thinking of some other use for it

AndrewBarba (approved):

PR #947, opened 2026-07-19
fix(eve): provide auth context to dynamic tools

### LINKED ISSUE #946: Dynamic tools receive a ToolContext without ctx.getToken / ctx.requireAuth

I do not have a standalone public reproduction repository yet. The issue can be reproduced with this single dynamic tool in a new eve project:

```ts
// agent/tools/oauth_probe.ts
import { defineDynamic, defineTool } from "eve/tools";
import { z } from "zod";

const auth = {
  principalType: "app" as const,
  async getToken() {
    return { token: "test-token" };
  },
};

  events: {
    "session.started": () => ({
      oauth_probe: defineTool({
        description: "Return whether inline tool auth is available.",
        inputSchema: z.object({}),
        execute: async (_input, ctx) => {
          const { token } = await ctx.getToken(auth);
          return { ok: token === "test-token" };
        },
      }),
    }),
  },
});
```

The same `defineTool` works when exported as a static authored tool. It fails when returned from `defineDynamic`.

### Steps to reproduce

1. Create an eve project using `eve@0.25.1`.
2. Add the dynamic tool above.
3. Start the agent and ask it to call `oauth_probe`.
4. Observe that the tool fails before the provider's `getToken` method runs.
5. Optionally add `approval: always()`; after approval/resume, the same failure occurs.

Expected: the `ctx` passed to a dynamic tool's `execute` satisfies the documented `ToolContext`, including `ctx.getToken(provider)` and `ctx.requireAuth(provider)`, just like a static authored tool.

Actual: `ctx.getToken` and `ctx.requireAuth` are absent at runtime, so the first token lookup throws:

```
TypeError: ctx.getToken is not a function
```

This also occurs in a production Slack approval flow using a user-scoped Vercel Connect provider. The approval succeeds, but the resumed dynamic tool fails before OAuth or the downstream API is reached.

### eve version

`eve@0.25.1`

### Environment

```shell
OS: Vercel production runtime (also source-reproducible from the dynamic executor)
Node: v24
Package manager: pnpm@11
@vercel/connect: 0.4.0
```

### Where does the bug occur?

- Production (`eve start` / deployed)
- The same context construction is used by the local dynamic-tool runtime path.

### Deployment

Private Vercel deployment; no public deployment URL is available.

### Build and runtime logs

```shell
[eve:harness.tool-loop] tool execution failed {
  toolName: 'calendar_create_event',
  error: {
    message: 'TypeError: ctx.getToken is not a function',
    name: 'TypeError',
    detail: 'TypeError: ctx.getToken is not a function\n' +
      '    at __eve_dynamic_exec_27 (.../__eve_nitro_handler__.mjs:453:30)\n' +
      '    at execute (.../_libs/eve.mjs:80907:23)\n' +
      '    at Object.execute (.../_libs/eve.mjs:84113:17)'
  }
}
```

### Additional context

The public types and auth documentation say tool executors receive a `ToolContext` with provider-scoped auth accessors. Static authored tools are wrapped with `createToolExecuteWithAuth` in `packages/eve/src/execution/node-step.ts`.

However, both dynamic-tool execution paths currently construct only a base context:

- Live dynamic tools: `toHarnessToolDefinition` calls `buildBaseToolContext`.
- Durable/replayed dynamic tools: `replayDynamicSessionTools` also calls `buildBaseToolContext`.

Current source:
https://github.com/vercel/eve/blob/main/packages/eve/src/context/dynamic-tool-lifecycle.ts

`buildBaseToolContext` does not add `getToken` or `requireAuth`:
https://github.com/vercel/eve/blob/main/packages/eve/src/context/build-base-tool-context.ts

This appears to be a missing auth wrapper on dynamic tools rather than a provider, principal, approval, or downstream API problem. It may need coverage for both the live dynamic path and the durable replay path after an approval/authorization resume.

Thank you!

ahwei:
## Summary

- route live dynamic-tool execution through the same auth-capable wrapper used by static authored tools
- preserve `ctx.getToken()` and `ctx.requireAuth()` when session/turn-scoped dynamic tools are replayed from durable metadata
- align the public dynamic-tool executor context type with `ToolContext`
- document the auth behavior and add a patch changeset

## Root cause

Dynamic tools were wrapped with `buildBaseToolContext`, which only provides the base session/tool fields. Static authored tools use `createToolExecuteWithAuth`, which augments that context with inline provider auth. The live dynamic path and both durable replay paths therefore omitted `getToken` and `requireAuth` at runtime.

## Validation

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm test`
- focused dynamic lifecycle regression suite: 41 tests passed

Fixes #946

### DIFF
--- a/.changeset/bright-mice-resolve.md
+++ b/.changeset/bright-mice-resolve.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Provide dynamic tools with the full auth-capable `ToolContext`, including `ctx.getToken(provider)` and `ctx.requireAuth(provider)`, across live execution and durable replay.
--- a/docs/guides/dynamic-capabilities.md
+++ b/docs/guides/dynamic-capabilities.md
@@ -25,6 +25,8 @@ their `defineDynamic` export is a build error.
 
 Pass `defineDynamic` an `events` object whose handlers return either a single `defineTool(...)`, a `Record<string, defineTool(...)>`, or `null` for no tools. Wrap every entry in `defineTool()`. The wrapper stamps them so their `execute` functions survive workflow step boundaries.
 
+Dynamic tool executors receive the same `ToolContext` as static authored tools, including inline provider auth through `ctx.getToken(provider)` and `ctx.requireAuth(provider)`.
+
 The example below builds one tool per warehouse table. A map return names each tool by its bare key, so the model sees `orders`, `users`, and so on.
 
 ```ts title="agent/tools/query.ts"
--- a/packages/eve/src/context/build-dynamic-tools.ts
+++ b/packages/eve/src/context/build-dynamic-tools.ts
@@ -6,7 +6,7 @@ import {
   LiveStepToolsKey,
 } from "#context/keys.js";
 import type { DurableDynamicToolMetadata } from "#context/keys.js";
-import { buildBaseToolContext } from "#context/build-base-tool-context.js";
+import { createToolExecuteWithAuth } from "#execution/tool-auth.js";
 import { createLogger } from "#internal/logging.js";
 import type { ApprovalContext, ApprovalStatus } from "#public/definitions/approval.js";
 import { toInputSchema, toOutputSchema } from "#shared/tool-schema.js";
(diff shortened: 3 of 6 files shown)


ahwei:
Superseded by #954 with the conventional `fix/dynamic-tool-auth` branch name. The commit and PR contents are unchanged.

PR #973, opened 2026-07-20
fix(eve): Fix incorrect overwrite of bot text when a single turn emits many messages

benpankow:
### Description

Addresses #436.

It's possible for a single turn to emit multiple messages, such as when an OAuth is triggered (which emits text -> triggers OAuth popup -> emits more text).

This PR adds a new resolver constructed slightly differently. It ensures that if the in-progress run is completed, a new text part will be added for a subsequent run rather than replacing the current part in-place. 

New unit test highlighting the single-turn, multiple-message case.

### DIFF
--- a/packages/eve/src/client/message-action-parts.ts
+++ b/packages/eve/src/client/message-action-parts.ts
@@ -1,0 +1,156 @@
+import type { RuntimeActionRequest, RuntimeActionResult } from "#runtime/actions/types.js";
+import type { InputRequest } from "#runtime/input/types.js";
+import type {
+  EveDynamicToolPart,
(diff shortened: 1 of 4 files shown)


packages/eve/src/client/message-action-parts.ts
```diff
@@ -0,0 +1,156 @@
+import type { RuntimeActionRequest, RuntimeActionResult } from "#runtime/actions/types.js";
```
benpankow, line 1: refactor, moved out of message-reducer.ts to avoid 700 line limit

AndrewBarba (approved):

PR #984, opened 2026-07-20
docs(eve): runtime - propose pluggable Loop architecture

AndrewBarba:
## Summary

- add a full research proposal for a pluggable eve execution Loop
- define data-only Loop, kernel transition, checkpoint, identity, delivery, event, and effect contracts
- describe WorkflowLoop, ProcessLoop, and future Temporal boundaries
- provide migration gates, deletion opportunities, conformance coverage, and acceptance criteria

## Why

The current `Runtime` facade is not a complete substitution boundary. Workflow-specific hooks,
streams, child dispatch, callbacks, cancellation, checkpoint resolution, dynamic-tool machinery,
telemetry, and build integration escape through execution and runtime layers. A second backend
would therefore need to reproduce Workflow behavior instead of implementing a small eve-owned
contract.

The proposal makes Workflow one Loop implementation, separates the engine-neutral agent kernel,
and uses a continuous `ProcessLoop` as the first portability proof.

## Impact

This is a research-only change and does not alter runtime behavior. It proposes breaking semantic
changes for a future implementation, notably pinning the agent revision per session and returning
`busy` for delivery during an active turn instead of timing-dependent coalescing.

## Validation

- `pnpm guard:invariants`
- `pnpm exec oxfmt --check research/pluggable-loop-execution.md`
- `git diff --check`

PR #983, opened 2026-07-20
research(loop-boundary): review execution/harness/runtime boundaries toward one pluggable Loop

AndrewBarba:
### Description

Adds `research/loop-boundary.md`: a thorough review of eve's execution, harness, and runtime boundaries, and a proposed target architecture in which Vercel Workflow becomes one implementation of an eve-owned **Loop** seam — replaceable by a single-process continuous loop or another durable engine without touching channels, the harness, or the definition layer.

The doc covers:

- **Findings** — verified boundary violations (`export *` engine re-export, `resumeHook` calls from `runtime/` routes, two large dependency cycles), leaky contracts (park vocabulary split across three layers, engine value types in shared step signatures, engine-minted session identity), and removable code (legacy pinned-driver path, dual runtime-action dispatch arms, duplicated park stores).
- **Target design** — a four-method `Loop` contract (`start / signal / resolveWaiter / events`) with an eve-owned error taxonomy, a three-point durability contract (park-boundary state, token addressing, append-only event log), harness-owned park reasons in `StepNext`, and per-implementation driver control flow guarded by a shared conformance suite.
- **Plan** — subtractive first: Phase 0 deletions, contract tightening, seam hygiene, then quarantining all workflow mechanics under `loop/vercel-workflow/` alongside a new in-process Loop (which also unlocks scenario tests without the workflow bundler).

Frontmatter `issue` is `TBD` pending an issue for this proposal; will update before marking ready for review.

### How did you test your changes?

- `pnpm guard:invariants` (includes rule 32 research frontmatter validation)
- `pnpm fmt`

### PR Checklist

- [ ] I linked an issue with prior discussion confirming this change is wanted
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [ ] I added tests and documentation where relevant (research doc only)
- [ ] I added a changeset if this touches the published `eve` package (docs-only, not needed)
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

Issue #982, opened 2026-07-20
Anchored-channel follow-up delivered mid-turn forks a new session instead of resuming


ruiconti:
## Symptom

A follow-up message posted to an anchored channel thread while the session's turn is still running is delivered to a **new session** instead of resuming the anchored one. The new run then dies with:

```
HookConflictError: Hook token "anchored:thread:<id>" is already in use (run "wrun_<original>")
    at claimHookOwnership → rekey → runDriverLoop → workflowEntry
```

Observed as a deterministic `test-tui` failure (`tui-connection-auth-user`) on PR #881 after the fixture model changed, and reproduced locally.

## Root cause

The session's delivery hook is a **single rotating slot**:

1. At session start, `workflow-entry.ts` (~L196) rekeys the delivery hook to `sessionState.continuationToken` — the anchored thread token. Resume works.
2. **During the turn**, in-turn continuations rotate that one slot: `turn-control-receiver.ts` calls `deliveryHook.rekey(payload.continuationToken)` (L89/L106/L121). During a connection-auth dance the hook becomes the auth continuation (`wrun_…:auth`), and `session-delivery-hook.ts#rekey` (~L179) claims the candidate and **disposes the previous hook** — the thread anchor becomes unreachable.
3. A channel delivery to the thread token in this window → `resumeHook` → `HookNotFoundError` → `RuntimeNoActiveSessionError` → the channel route starts a **new session**.
4. Both runs later claim the anchor (original at park, new at entry); the loser dies with `HookConflictError`.

## Why it surfaced now

The exposure window is (hook rotation → park). `tui-connection-auth-user` posts its follow-up immediately after the tool-result marker; swapping the fixture model (opus → gpt-5.6-luna@medium, PR #880's commit) surfaced the marker earlier in the turn, landing the follow-up inside the window every time. Bisect-verified: passes at the pre-#880 commit, fails at it, passes after reverting the fixture (`8c161c54`).

**This is not test-only**: any user replying quickly in a Slack/anchored thread mid-turn can fork their session.

## Repro

```sh
# with the fixture on gpt-5.6-luna@medium (pre-8c161c54)
cd packages/eve && pnpm run build:js
PORT=3210 node test/tui-client/tui-connection-auth-user.ts
```

## Candidate directions

[...]

Issue #867, opened 2026-07-16
Support steering active channel turns with replacement input


AndrewBarba:
## Summary

Chat-style channels need to apply corrective or additional input while an agent is still working. Add an explicit delivery policy that lets channel authors either queue input for the next turn or steer the active logical turn.

## Public API

```ts
await send(message, {
  auth,
  continuationToken: threadId,
  turnPolicy: "steer",
});
```

`turnPolicy` is `"queue" | "steer"` and defaults to `"queue"`. The same option flows through Chat SDK sends and low-level delivery. Built-in channels remain queue-only unless they explicitly opt in.

## Required behavior

- `"steer"` applies input to the active conversation turn at its next safe step boundary
- successfully steered input keeps the same turn ID and emits no cancellation boundary
- a parked or missing session treats steering as ordinary new-turn input
- a terminal race preserves unapplied input for the next turn in durable admission order
- steering is exactly once and multiple steering deliveries are applied in admission order
- pinned workflows without steering capability safely degrade to queue
- active task-mode sessions reject steering until they have a durable successor design
- structured input responses and delegated-child routing retain their existing addressing semantics
- cancellation remains a separate stop-without-replacement operation

## Architecture constraints

- the session driver remains the sole durable owner of admitted deliveries
- the active turn receives a single-flight lease and returns a consumed, released, or remainder disposition
- only the session driver converts unapplied steering into queued input
- internal soft notification may stop obsolete work, but steering does not use hard turn-cancellation settlement
- durable command state and live runtime signals remain separate across Workflow replay and version skew

## Acceptance criteria

- custom channel authors can select `turnPolicy: "steer"` without using internal workflow APIs
- queue remains the default and existing channel behavior is unchanged
- applied steering reaches the active harness step under the existing turn
- unapplied steering is never dropped, duplicated, or reordered by acknowledgement timing
- cancellation and steering races have deterministic, durable outcomes
- unit, integration, scenario, and custom-channel e2e coverage exercise active steering, fallback, replay, ordering, input requests, descendants, and pinned-version compatibility

## Plan

The implementation plan lives in `research/channel-turn-steering.md`.

PR #988, opened 2026-07-20
docs(eve): runtime - consolidate pluggable Loop architecture

AndrewBarba:
## Summary

- replace the two competing Loop research drafts with one issue-backed normative plan
- define shared eve session and turn programs over an internal pluggable `LoopBackend`
- preserve current latest-turn execution and Workflow World configuration
- incorporate atomic admission, durable queue and steer ownership, address lifetime, effect safety, build isolation, migration, and conformance requirements

Part of #512.

Supersedes #983 and #984. Builds on the executable evidence in #513 and #700 and aligns the boundary with the delivery requirements in #867 and #982.

## Validation

- `pnpm exec oxfmt --check research/pluggable-loop-execution.md`
- `pnpm guard:invariants`
- `git diff --cached --check`

Research-only change; no changeset required.

### DIFF
--- a/research/pluggable-loop-execution.md
+++ b/research/pluggable-loop-execution.md
@@ -1,0 +1,576 @@
+---
+issue: https://github.com/vercel/eve/issues/512
+status: proposed
+last_updated: "2026-07-20"
+---
+
+# Pluggable loop execution
+
+## Decision
+
+eve will run sessions through one shared session program and one shared turn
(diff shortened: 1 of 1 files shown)


dcosson:
Excited to see this! Stuff like steering messages and async background tasks & sub-agents would be great to see added to Eve, looks like this is a step in that direction.

Are you open to community support with an implementation, or still just weighing ideas?

PR #999, opened 2026-07-20
feat(eve): Add ability to launch workflow sdk devserver alongside eve TUI

benpankow:
### Description

Adds a `--workflow-ui` flag to `npx eve dev` which spins up the Workflow SDK observability UI at `localhost:3456` configured for eve's use-case (`.eve/.workflow-data` as the local data dir etc)

Screenshot 2026-07-20 at 4 09 30 PM

Tested locally, small unit test

### DIFF
--- a/packages/eve/src/cli/run-types.ts
+++ b/packages/eve/src/cli/run-types.ts
@@ -1,0 +1,91 @@
+import type { BuildHost } from "#cli/commands/build.js";
+import type { DevelopmentRequestHeaders } from "#cli/dev/url-target.js";
+import type { WorkflowWebUiHandle } from "#cli/dev/workflow-web-ui.js";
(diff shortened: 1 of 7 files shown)


packages/eve/src/cli/run-types.ts
```diff
@@ -0,0 +1,91 @@
+import type { BuildHost } from "#cli/commands/build.js";
```
benpankow, line 1: line limit refactor

benpankow:
Closing as an old exploration. The `eve dev` and local observability architecture has changed substantially since this was opened, so this Workflow SDK devserver integration is no longer a viable merge target. We can revisit the workflow UI use case with a fresh design if it becomes a priority.

PR #1010, opened 2026-07-21
feat(eve): selectable loop runtimes behind eve dev --loop

ruiconti:
Runnable follow-up to #513: the three loop implementations from the research become selectable runtimes that serve real sessions.

## What

- `packages/eve/src/internal/loops/` — inline, Temporal, and Workflow DevKit loop runtimes implementing eve's `Runtime` contract, plus the in-memory event log, delivery parsing, and Temporal address bookkeeping they share.
- `eve dev --loop <inline|workflow|temporal>` and the same flag on `eve start` (env: `EVE_LOOP`). Unset, eve keeps the production Workflow runtime — selection is wired per request in `runtime-stack.ts`, with guards refusing inline and Temporal inside Vercel Functions.
- Execution-layer seams the runtimes need: `createSessionOperation`/`executeTurnStepOperation` extracted as engine-neutral operations, and a `writeEveAttributes` injection so the tool loop no longer hard-codes the Workflow-only attribute emitter.
- Temporal dev-server observability: `EVE_LOOP_TEMPORAL_DB` persists the server to SQLite, `EVE_LOOP_TEMPORAL_UI_PORT` serves the Temporal Web UI — both off by default. A dev recompile swaps the process-global Temporal runtime instead of erroring; `@temporalio/*` is externalized in the dist build because bundling the CJS SDK into eve's ESM tree breaks at import.

This started as the #700 benchmark: the runtimes are the same code with the telemetry layer deleted (recorders, sample IDs, record stores, ~35 instrumented call sites) and the module renamed out of benchmark vocabulary. The benchmark client tooling is archived on `rui/loop-benchmark-wip`.

## Why this shape

The research (#513) proves the program/adapter split on scripted scenarios; this is the smallest step that makes the boundary runnable against real agents. Selection stays an internal, env-gated experiment rather than public API: the runtimes reject what they don't support (subagents, approvals, callbacks on inline; the same plus context/schemas on Temporal and Workflow) instead of silently degrading.

## Base and review scope — draft until forward-ported

The branch predates main's last months of `execution/` work (cancellation cascade, stale-HITL delivery, ToolSchema), so it is conflicting with main and CI will not run until it is rebased — that forward-port is the next piece of work and needs per-feature decisions, not a mechanical rebase. The diff also includes #513's research doc and prototype lineage because the branch forked before that PR was rebased; review those in #513. The reviewable content here is `internal/loops/`, the `runtime-stack`/`channel-dispatch` wiring, the CLI flag, and the execution seams.

Verified on this base: full unit (4,426), integration (383), and scenario (incl. a real local Temporal server) tiers; live `eve dev --loop` smoke returns the correct runtime identity for all three (`inline_…`, `wrun_…`, `eve-loop-…`).


ruiconti:
Closing: wrong shape. This adds parallel runtimes beside the production loop instead of implementing the LoopBackend/TurnDependencies contract from #513 and replacing the existing implementation. The branch stays as a lab reference; the real change gets built on the research contract against current main.

PR #1042, opened 2026-07-21
fix(eve): dynamic tools - preserve toModelOutput replay

### LINKED ISSUE #1039: Dynamic tools lose toModelOutput on replayed durable steps — model receives the raw tool output
## Summary

A dynamic session tool — one returned from a `defineDynamic` `session.started` resolver — loses its `toModelOutput` hook on every durable step after the one where it was resolved. Each tool call is a step boundary, so the step where the model reads the tool result and narrates is always a replayed step. In practice the hook never runs where it matters: the model receives the raw `execute` output. Static (module) tools keep the hook throughout.

`toModelOutput` is the documented seam for shaping the model's view of a tool result (full object to channel handlers, shaped view to the model), so any redaction built on it silently stops working after the resolution step.

Confirmed in `0.24.0` and `0.27.0`.

**Repro:** https://github.com/msywulak/eve-tomodeloutput-repro — deterministic, driven by a scripted in-process `LanguageModel` (no API keys). `npm install && npm run repro` prints the verdict; exit code 0 means reproduced.

## What the repro shows

Two tools with identical bodies and identical `toModelOutput` hooks; only the registration differs. Both `execute` to `{ name: "Ada", secret: "TOP-SECRET-VALUE" }` and both define `toModelOutput` returning `{ type: "text", value: "Profile loaded for Ada (secret redacted)." }`. The scripted model appends every provider-bound request to a capture log.

Tool-result message sent to the model on the post-tool-call step, from that log:

`fetch_profile_static` (module tool) — hook ran:

```json
{
  "type": "tool-result",
  "toolCallId": "call-1",
  "toolName": "fetch_profile_static",
  "output": {
    "type": "text",
    "value": "Profile loaded for Ada (secret redacted)."
  }
}
```

`fetch_profile_dynamic` (defineDynamic, inline execute per the dynamic-capabilities docs) — hook gone:

```json
{
  "type": "tool-result",
  "toolCallId": "call-3",
  "toolName": "fetch_profile_dynamic",
  "output": {
    "type": "json",
    "value": {
      "name": "Ada",
      "secret": "TOP-SECRET-VALUE"
    }
  }
}
```

## Mechanism

Verified in the published `0.27.0` dist (`dist/src/context/dynamic-tool-lifecycle.js` and `build-dynamic-tools.js`):

- Fresh resolution (`toHarnessToolDefinition`, the step where `session.started` fires) carries the hook: `...t.toModelOutput===void 0?{}:{toModelOutput:t.toModelOutput}`.
- The resolved tool is persisted into durable session metadata as plain JSON, which strips functions. Two of them get survival machinery: the bundler transform registers each dynamic tool's `execute` in the step-function registry (`executeStepFnName` + `closureVars`), and since #275 the `approval` callback is registered the same way (`approvalStepFnName`).
- The replay rebuild restores `description`, `execute` (registry lookup), `inputSchema`, `approval` (registry lookup via `buildReplayedApproval`), and `outputSchema`. `toModelOutput` has no registry entry, so the rebuilt tool omits it.

```bash
grep -o 'toModelOutput[^,}]*' node_modules/eve/dist/src/context/dynamic-tool-lifecycle.js
# present in toHarnessToolDefinition (fresh resolution);
# zero hits in build-dynamic-tools.js (the replay rebuild)
```

## Suggested fix

#275 ("preserve approval replay") already established the pattern for exactly this bug on the `approval` hook. Give `toModelOutput` the same treatment: register it at resolution time, store a `toModelOutputStepFnName` in the durable metadata, and rebind it in the replay rebuild with a `buildReplayedToModelOutput` mirroring `buildReplayedApproval`.

## Background

Found while building an agent that redacts tool output. A dynamic tool's `toModelOutput` was the redaction seam, and the model verifiably read the raw output on the narration step. Environment: Node 24, eve `0.24.0` and `0.27.0`.

msywulak:
## Summary

Preserves `toModelOutput` when session- and turn-scoped dynamic tools are replayed from durable metadata, so the model keeps receiving the shaped view of a tool result on every step.

Fixes #1039.

## Root cause

Dynamic tool replay metadata kept the execute step function (and, since #275, the approval callback), but never retained `toModelOutput`. Each tool call is a step boundary, so the step where the model reads a tool result is always a replayed step — a dynamic tool's `toModelOutput` effectively never ran where it mattered, and the model received the raw `execute` output. Redaction built on the hook silently stopped working after the resolution step. Static tools were unaffected.

## Changes

- Register each dynamic tool's `toModelOutput` in the step-function registry at resolution time and store a `toModelOutputStepFnName` in `DurableDynamicToolMetadata`.
- Rebind the hook in the replay rebuild with `buildReplayedToModelOutput`, mirroring `buildReplayedApproval` from #275.
- Fail closed when the registered hook cannot be recovered on replay: withhold the raw output and give the model a short placeholder. This matches the fail-closed choice #275 made for approval — the hook exists to shape or redact what the model sees, so passing the raw output through would silently defeat it. Happy to switch to warn-and-pass-through if you'd rather degrade the other way; it's a one-line change and the design question is flagged here on purpose.
- Update the stale `approval` doc comment on `DynamicToolEntry`, which predated #275 and still said session/turn-scoped tools cannot carry a function across replay.
- Three unit regressions: replaying the hook from session-scoped metadata, withholding raw output when the registration is missing, and leaving the hook absent when the entry omits it.
- A patch changeset.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/context/dynamic-tool-lifecycle.test.ts` — 44 passed
- `pnpm lint`, `pnpm fmt`, `pnpm typecheck`, `pnpm guard:invariants`
- End to end against the repro from #1039 (https://github.com/msywulak/eve-tomodeloutput-repro): on published `0.27.0` the dynamic twin's tool result reaches the model as raw JSON containing the planted secret. With this branch packed (`pnpm --filter eve pack`) and installed into the repro app, both twins arrive redacted:

```json
{
  "type": "tool-result",
  "toolName": "fetch_profile_dynamic",
  "output": {
    "type": "text",
    "value": "Profile loaded for Ada (secret redacted)."
  }
}
```

One note for anyone re-running that comparison: delete the app's `.eve/` directory between version swaps. The compiled dev runtime caches framework code and can mask the change.


msywulak:
Flagging #1152 as a sibling of this bug — different hook, same seam.

That issue reports the `approval` callback hitting its fail-closed fallback `(() => "user-approval")` on a cold serverless invocation, so an authored no-approval tool starts demanding approval. The root cause is that the step registry is per-process: `globalThis[Symbol.for("@workflow/core//registeredSteps")]` starts empty in a fresh process, so `lookupStepFunction` returns null until the bundler transform re-registers.

That's the same condition `buildReplayedToModelOutput` keys off in this PR. So worth stating plainly: once this merges, a cold-process replay will show the placeholder in place of the real shaped output, not just when a registration is genuinely broken. The fail-closed direction still matches #275. What #1152 adds is that the trigger for the fallback is too eager — a same-revision cold start skips the refresh that would have repopulated the registry.

I don't think this blocks the PR. Restoring `toModelOutput` to parity with `execute` and `approval` is a real gap on its own, and #1152 is a second reporter hitting the same seam from the approval side. But the two together point at the registry-replay path wanting a look as a whole: both hooks lean on a fallback that can't tell "registration missing because the process is cold" from "registration missing because it's broken."

PR #1044, opened 2026-07-21
feat(eve): slack - add generic event handler

AndrewBarba:
Adds `slackChannel({ onEvent })` — a generic fallback handler for signed Slack Events API callbacks (`reaction_added`, `team_join`, `channel_created`, …) that the specialized mention/DM handlers don't model. The handler receives the raw event and owns control flow: it can start zero, one, or many agent turns.

```ts
import { slackChannel } from "eve/channels/slack";

export default slackChannel({
  async onEvent({ receive }, event) {
    if (event.type !== "reaction_added" || event.reaction !== "eyes") return;

    const item = event.item as { channel: string; ts: string };
    await receive({
      message: "Someone reacted with :eyes: — investigate this thread and respond.",
      target: { channelId: item.channel, threadTs: item.ts },
      auth: null,
    });
  },
});
```

### Handler context

| Surface | Purpose |
| --- | --- |
| `ctx.receive(options)` | The schedule API's `receive(slack, options)` pre-bound to this channel. Each call starts a turn and returns the session; the callback's team id is seeded into session state. |
| `ctx.slack` | Workspace-scoped handle: `teamId` plus the raw `request(operation, body)` Web API escape hatch. |
| `ctx.envelope` | Full delivery envelope (`team_id`, `event_id`, `event_time`, …). |
| `ctx.waitUntil(promise)` | Keeps deliberately detached work alive after the webhook ack. |

### Semantics

- **Precedence** — authored `onAppMention` / `onDirectMessage` always win for their event types. When the specific handler is absent, `onEvent` replaces the built-in mention/DM default and also receives bot-authored/subtype DMs the specialized filter rejects. Without `onEvent`, behavior is unchanged.
- **Dedup** — generic events participate in the existing `event_id` dedup.
- **Errors** — handler failures are caught, logged, and acked without starting a turn.
- **Scope** — JSON `event_callback` deliveries only; URL verification, slash commands, and interactive payloads never reach it.

### Internals

The proactive receive path is shared between the channel's `receive` hook and `ctx.receive` (`receiveOnSlack`), and `SlackEventReceiveOptions` aliases the cross-channel `CrossChannelReceiveOptions<SlackReceiveTarget>` contract so the two surfaces can't drift. Includes docs (`docs/channels/slack.mdx`) and a patch changeset.

### Validation

- `pnpm fmt`, `pnpm lint`, `pnpm typecheck`, `pnpm guard:invariants`, `pnpm docs:check`
- Slack channel unit tests: 267 passed (15 files); full unit suite green (`pnpm test:unit`)


packages/eve/src/public/channels/slack/slackChannel.ts
```diff
@@ -688,59 +753,109 @@ async function handleEventPost(input: {
...
+  if (envelope === null) return new Response("ok");
+
+  let dispatch: (() => Promise<void>) | null = null;
+  if (payload.kind === "app_mention" || payload.kind === "direct_message") {
+    const kind = payload.kind;
+    const message = slackMessageFromWebhookPayload(payload);
+    const authored = kind === "app_mention" ? config.onAppMention : config.onDirectMessage;
+    const handler =
+      authored ??
+      (config.onEvent === undefined
```
benpankow: definitely nitpick but this ternary/control flow was tough to parse for me - wonder if the default on app mention / default DM handlers should live below the `onEvent` logic below so the code structure matches the fall-through ordering?


1) defined app mention/DM hook - handled here
2) defined `onEvent` hook - not handled here, handled below
3) no custom hook - handled here

benpankow (commented):
this is sweet, very powerful

packages/eve/src/public/channels/slack/slackChannel.ts
AndrewBarba: yes good catch thank you

AndrewBarba: Restructured in 96f52d2d — the nested ternary is gone. Selection now reads in fall-through order: (1) authored mention/DM handler, (2) authored `onEvent`, (3) built-in defaults applied last via `dispatch ??= builtinDefault`.

benpankow (approved):
🙏

Issue #223, opened 2026-06-23
Slack bots should not require mentions for every response


seflless:
When a user @mentions the bot in a channel, the bot replies in a thread. After that:

1. Follow-up messages in the thread should trigger the bot without another @mention
2. Multiple people should be able to participate in the thread naturally
3. The bot should stay engaged for the life of the thread

Currently every message in the thread requires a fresh @mention to get a response.

This is how Anthropic's Claude Tag works — one @mention starts the conversation, then the thread flows like a group chat.

<img width="75%"  alt="Claude Tag" src="https://github.com/user-attachments/assets/f3e4e73e-e873-48ad-aab5-2a286e9365f3" />

Here by contrast is the out of the box eve slack agent works, there was no reply to my second/third message, but once I used a mention I received a reply.

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/b4fc4b3e-1844-44aa-8f00-1401cc833ac6" />

seflless:
I patched eve in my project so I could get follow up messages without mentions, but then hit another UX issue: messages sent while the agent is taking a turn should steer the agent. This is what people would expect. Without it long running turns from agents will often waste effort that could have been saved if it was steered instead. This wastes tokens and user time waiting for work that is wrong or ineffective.

harrybradshaw:
Looks like a duplicate to the features requested in https://github.com/vercel/eve/issues/235 ?

berenar:
Agree

seflless:
I read through: https://github.com/vercel/eve/issues/235, it's covers my main issue. Though making it so you can steer eve turns versus queue messages is not included in #235, unless I'm misreading it. Feel free to close this issue though, I can follow up about steering support at some point. I'm not working on it right now, so too lazy to write a good issue.

benpankow:
Hi all, #1046 introduces a generic `onMessage` handler which gives complete control over which messages to respond to (e.g. only mentions and messages in a thread the bot has been tagged into). It'll go out with the next release along with #1044 which is an even lower-level hook into the entire Slack event stream. Hopefully this addresses your use-case, but feel free to reopen if you run into functionality gaps.

PR #1046, opened 2026-07-21
feat(eve): Add `onMessage` hook w/ `isSubscribed` utility

benpankow:
## Description

Reimplementation of #400, addresses #223.

Adds a generic Slack `onMessage` hook which gives users more granualar control over bot replies to messages:
| Incoming event         | Handler order                                                       |
| ---------------------- | ------------------------------------------------------------------- |
| App mention            | `onAppMention` → `onMessage` → `onEvent` → built-in mention default |
| Direct message         | `onDirectMessage` → `onMessage` → `onEvent` → built-in DM default   |
| Other Slack message    | `onMessage` → `onEvent` → ignore                                    |
| Other Events API event | `onEvent` → ignore                                                   |

In particular, introduces `ctx.isSubscribed()` helper which can be used to toggle behavior based on whether the message is in a thread the bot is tagged into already.

 ```ts
   export default slackChannel({
     async onMessage(ctx, message) {
       if (message.author?.isBot) return null;
       return ctx.isBotMentioned() || (await ctx.isSubscribed())
         ? { auth: null }
         : null;
     },
   });
 ```

Screenshot 2026-07-21 at 1 32 26 PM

ctx.isBotMentioned() identifies explicit mentions, while ctx.isSubscribed() checks whether the message’s thread has an active eve session. This avoids fetching and scanning Slack thread history. If that session terminates, the bot must be mentioned again.

Slack requires the corresponding message events and scopes; the docs include Vercel Connect setup instructions.

 ## Test Plan

Unit tests; tested live with a deployed agent.

### DIFF
--- a/packages/eve/src/public/channels/slack/slackChannel.ts
+++ b/packages/eve/src/public/channels/slack/slackChannel.ts
@@ -30,7 +30,9 @@ import {
   defaultOnDirectMessage,
 } from "#public/channels/slack/defaults.js";
 import {
+  parseMessageEvent,
   type SlackEvent,
+  slackEventBotUserId,
   type SlackEventEnvelope,
   type SlackInboundContext,
   type SlackMessage,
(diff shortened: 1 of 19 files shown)


packages/eve/src/public/channels/slack/slackChannel.ts
```diff
@@ -756,7 +778,37 @@ async function handleEventPost(input: {
...
+          credentials: config.credentials,
+          handler: defaultOnAppMention,
+          kind: "thread_reply",
+          message: genericChannelMessage,
+          send: input.send,
+          threadContext: config.threadContext,
+          uploadPolicy: input.uploadPolicy,
+        });
+    }
+  }
```
ruiconti: This makes the agent engage and reply to every inbound message on the thread even though the message may not be directed at the agent, for instance, which is not a very good UX.

The requirement is narrow, however, and not domain specific (eg different than ambient mode, which awaking rules are highly context-dependent) - it just needs to know whether, on the current thread, it needs to determine if there is a question or message directed at it.

benpankow: Definitely think it would be useful to have some sort of overridable "identify if I'm being asked a question" classifier in the loop that a user could customize.

I do think "respond to everything in the thread" is not an unreasonable pattern for single-player agent interactions though, could envision some users would want that. Maybe basic heuristic of "is the user @-mentioning someone else in the response?"

AndrewBarba (approved):

packages/eve/src/public/channels/slack/slackChannel.ts
Phelickz: @ruiconti @benpankow 

Following up on the "when to respond" thread here — as a downstream consumer we just adopted onMessage/isSubscribed() in 0.27.1 and hit exactly this: in multi-person channel threads the agent replies to every message once the thread is subscribed, including cross-talk not directed at it.

Before we build our own gating, is there intent to ship something native — e.g. an overridable "is this directed at me / is this a question?" classifier step in the onMessage path (per @benpankow's suggestion), or a built-in heuristic like "suppress when the message @-mentions a different user" (@ruiconti's framing)?

If it's on the roadmap we'll wait / align with the eventual hook shape; if not, we'll implement it in our own onMessage. Just want to avoid building something you're about to replace.

PR #1041, opened 2026-07-21
fix(eve): add nf3 hook for recursive inclusion of `#...` imports

cmpadden:
Closes #842.

## Problem

Hosted builds trace packages listed in `agent.build.externalDependencies`. When an external package such as `@vercel/connect` imports `eve/connections` or `eve/channels/auth`, Nitro/nf3 discovers eve transitively and copies it into the function's `node_modules`.

nf3's vendored resolver cannot follow eve's embedded `#*.js` package-import wildcard. It emits only the public export entrypoints and silently omits their internal closure, so the function crashes at startup:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
'/var/task/node_modules/eve/dist/src/public/connections/errors.js'
imported from /var/task/node_modules/eve/dist/src/public/connections/index.js'
```

## Reproduction

I reproduced this on both `eve@0.24.3` and the latest published `eve@0.27.0` using a Next.js `withEve({ agents })` app. The affected agent configured:

```ts
build: { externalDependencies: ["@vercel/connect"] }
```

and an MCP connection importing `@vercel/connect/eve` and `eve/connections`. A second agent without configured externals served as the control.

The research agent failed with the exact missing `connections/errors.js` error; the control agent was healthy. A local hosted build independently emitted only public eve entrypoints and failed when importing the generated function entrypoint.

## Fix

Add an eve-owned nf3 `traceResult` hook. When nf3 traces an eve public entrypoint, the hook resolves eve's package-internal `#...` imports to `dist/src/...` and recursively adds their static internal and relative-import closure before nf3 copies trace files.

## Validation

I packed this exact branch as a local eve tarball and deployed it through the same failing multi-agent reproduction without publishing it. The previously failing research info route now succeeds and reports the authored Connect connection; the control agent remains healthy. The local generated function includes `connections/errors.js` and imports successfully.

Added integration coverage for a traced `eve/connections` entrypoint and scenario coverage that verifies hosted Nitro passes the eve trace hook to nf3.

## Testing

- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/internal/nitro/host/eve-package-trace-hook.integration.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/internal/nitro/host/create-application-nitro.scenario.test.ts`
- `pnpm --filter eve typecheck`
- targeted `oxfmt` and `oxlint`
- `pnpm guard:invariants`

Issue #842, opened 2026-07-15
Hosted service output ships incomplete node_modules/eve (internal #public/... specifiers not traced) → ERR_MODULE_NOT_FOUND


IKatsuba:
### Summary

A deployed eve service function ships an **incomplete** `node_modules/eve`: eve traces
`build.externalDependencies` into the hosted service output, but its file tracer cannot follow eve's
own package-internal `#public/...` import specifiers. So the function ships `node_modules/eve` with only
the ~4 export-map entry files (out of ~1300) — the rest of the closure is missing. The first runtime
import of e.g. `eve/connections` then throws `ERR_MODULE_NOT_FOUND` (`connections/errors.js` missing) and
crashes the function.

This is distinct from the `/eve/agents/<name>` prefix bugs (#839): it is a packaging/tracing gap, not a
routing one.

### Link to a minimal reproduction

No standalone repro repo — structural bug found while diagnosing a production crash (it took down a
deployed Slack channel right after a multi-agent cutover). Reproduces on any hosted service whose bundled
`node_modules/eve` is produced by the current tracer and then imports `eve/connections` (or another
`#public/...`-backed subpath) at runtime.

### Steps to reproduce

1. Deploy an eve app that produces a hosted service function bundling `node_modules/eve` (e.g. a
   `@vercel/connect`-using agent, or any multi-agent service output).
2. Trigger a runtime path that imports `eve/connections` / `eve/channels/auth`.
3. The function crashes on the first such import.

### Current vs. expected behavior

- **Current:** the deployed function's `node_modules/eve` contains only the export-map entry files; the
  first `eve/connections` import throws `ERR_MODULE_NOT_FOUND` and the function crashes.
- **Expected:** eve traces its own internal `#public/...` closure into the hosted service output, so all
  reachable modules ship and runtime imports resolve.

### eve version

eve@0.24.3 (prod), unchanged through 0.24.4 / HEAD. The 0.24.1 "external dependency closures" fix covers
eve **dev** generations only; the hosted service output still ships the byte-identical `#public/...`-importing
entry files with the closure missing.

### Environment

```shell
Deployed on Vercel (hosted service function output). Not observed in local single-package dev.
```

### Where does the bug occur?

Build (`eve build`), Production (`eve start` / deployed)

### Build and runtime logs

```shell
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../node_modules/eve/dist/connections/errors.js'
[...]

PR #1048, opened 2026-07-22
fix(eve): resolve internal imports during hosted tracing

cmpadden:
Alternative implementation to #1041 for #842.

## Problem

When a configured hosted external dependency imports an eve public subpath, Nitro/nf3 traces the external eve package into the function. NFT does not resolve eve's embedded `#*.js` package-import wildcard, so it includes the public entrypoint but omits eve-internal files that entrypoint imports. The deployed function then fails with `ERR_MODULE_NOT_FOUND`.

## Read-time approach

Provide NFT a virtual `readFile` implementation through Nitro's `traceOpts`.

For files inside the eve package NFT actually traced, the reader presents `#...` package imports as equivalent relative imports. NFT can then use its normal parser and graph traversal to discover the complete runtime closure. This only changes the source NFT analyzes; nf3 still copies the original package files unchanged.

Compared with #1041's post-trace closure hook, this approach leaves parsing and recursive graph traversal entirely to NFT.

## Validation

- Regression scenario traces a fixture external dependency that imports `eve/connections` and verifies its internal closure is copied.
- Integration coverage verifies aliases are virtually presented to NFT as relative imports from a separate installed eve package.
- Direct nf3 experiment traced the expected closure and successfully loaded both `eve/connections` and `eve/channels/auth`.

## Testing

- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/internal/nitro/host/eve-package-trace-options.integration.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts test/scenarios/app-runtime-dependencies.scenario.test.ts -t "traces hosted external dependencies configured in agent.ts"`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/internal/nitro/host/create-application-nitro.scenario.test.ts`
- targeted `oxfmt` and `oxlint`
- `pnpm guard:invariants`

### DIFF
--- a/.changeset/trace-eve-hosted-closure.md
+++ b/.changeset/trace-eve-hosted-closure.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Prevent hosted functions from shipping an incomplete eve package when an external dependency imports an eve public subpath. Transitively traced eve packages now include their complete runtime closure instead of failing with `ERR_MODULE_NOT_FOUND`.
--- a/packages/eve/src/internal/nitro/host/create-application-nitro.scenario.test.ts
+++ b/packages/eve/src/internal/nitro/host/create-application-nitro.scenario.test.ts
@@ -546,6 +546,9 @@ describe("application Nitro creation", () => {
 
     for (const call of createNitroMock.mock.calls.slice(0, 3)) {
       const traceDeps = call[0].traceDeps;
+      expect(call[0].traceOpts).toEqual({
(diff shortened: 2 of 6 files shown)


cmpadden:
[exsolve@1.1.1](https://github.com/unjs/exsolve/releases/tag/v1.1.1) and [nf3@0.3.22](https://github.com/unjs/nf3/releases/tag/v0.3.23) released with fix.

PR #1053, opened 2026-07-22
fix(eve): use native elicitations for the Linear channel

OwenKephart:
### Description

This simplifies / improves the Linear channel implementation of user prompts in two ways:

- It now uses native elicitations in Linear, meaning the user is presented with option buttons in the Linear UI, rather than just freeform text
- We now no longer render base64 encoded text on every user-facing prompt

The second point is of course the more important one. Previously, we attempted to directly generate structured input_request responses through the Linear channel. This is challenging because this requires specific request_id metadata. The previous implementation attempted to stash the required metadata in hidden blocks (`<!-- (content) -->`), which could then be read back when the user responded and transformed into a structured input response. However, these are not actually hidden in the linear UI, meaning the user is presented with confusing text.

Linear does not currently appear to have a good way of stashing metadata in an invisible way, but thankfully it's not actually a requirement to pass the structured responses to eve. There's already built-in matching logic for plain text, meaning the linear agent can certainly just send back a standard text response and have it get matched as expected. Using this gets rid of a lot of unnecessary code.

unit, live demo

before:

after:

### PR Checklist

### DIFF
--- a/packages/eve/src/public/channels/linear/api.ts
+++ b/packages/eve/src/public/channels/linear/api.ts
@@ -88,6 +88,8 @@ export interface LinearAgentActivityRecord {
     __typename?: string;
   };
   id: string;
+  signal?: string | null;
+  signalMetadata?: JsonObject | null;
   updatedAt?: string;
 }
 
@@ -306,6 +308,8 @@ export async function listLinearAgentSessionActivities(input: {
           activities(last: $last) {
             nodes {
               id
+              signal
+              signalMetadata
               updatedAt
               content {
                 __typename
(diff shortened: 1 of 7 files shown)


OwenKephart:
* **#1053** <a href="https://app.graphite.com/github/pr/vercel/eve/1053?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/vercel/eve/1053?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ruiconti (approved):

PR #1109, opened 2026-07-23
feat(eve): slack - add thread participant routing and self-loop guard

AndrewBarba:
## Summary

- add `ctx.thread.listParticipants()` to return unique human Slack participants in first-appearance order
- prevent messages authored by the installed Slack app from reaching message hooks and triggering self-reply loops
- preserve `onMessage` delivery for messages authored by other bots
- document participant-based routing and self-message filtering

## Root cause

Once `message.channels` and `message.im` event delivery is enabled, Slack also sends events for messages posted by the installed app. eve previously forwarded those events to authored message hooks, allowing an agent response to recursively trigger another turn.

The inbound Slack boundary now identifies the installed app by its authorized bot user ID, with the event app ID as a fallback when Slack omits `user`, and drops only those self-authored messages.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack` (277 tests)
- `pnpm docs:check`
- targeted `oxlint` for the changed Slack source and tests
- `git diff --check`

### DIFF
--- a/.changeset/fuzzy-bots-listen.md
+++ b/.changeset/fuzzy-bots-listen.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Added `thread.listParticipants()` for Slack thread routing. Messages posted by the installed Slack app are now ignored before reaching message hooks to prevent self-reply loops.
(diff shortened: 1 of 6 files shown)


AndrewBarba:
Superseded by the three focused follow-ups: #1110 (thread participants), #1111 (self-reply loop guard), and #1112 (agent reply identity).

PR #1113, opened 2026-07-23
fix(eve): slack - coalesce thread refreshes

AndrewBarba:
## Summary

- share one `conversations.replies` request across overlapping `thread.refresh()` calls
- atomically replace `recentMessages` after a successful refresh
- preserve the last successful snapshot when a later refresh fails
- let thread-context loading reuse messages another helper already loaded in the same binding
- keep sequential explicit `thread.refresh()` calls fresh

## Root cause

Slack thread consumers independently called `thread.refresh()`. A handler that loaded thread data before dispatch could therefore trigger another identical request while building `threadContext`. Concurrent refreshes also cleared and appended to the same array independently, allowing duplicate messages.

The binding now coalesces only in-flight refreshes. Internal context loading reuses an existing snapshot, while a later explicit refresh still performs a new request.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack` (280 tests)
- `pnpm docs:check`
- targeted `oxlint`
- `git diff --check`


packages/eve/src/public/channels/slack/api.ts
```diff
@@ -385,8 +387,42 @@ export function buildSlackBinding(input: {
...
+  function refreshThread(): Promise<void> {
+    if (refreshInFlight !== undefined) return refreshInFlight;
+    const refresh = refreshMessages().finally(() => {
+      if (refreshInFlight === refresh) {
+        refreshInFlight = undefined;
+      }
+    });
+    refreshInFlight = refresh;
+    return refresh;
+  }
```
ruiconti: This is a bit leaky. I would expect this to be in `refreshMessages` so that there's no way for an direct (ie that side-steps the coalescing) refreshMessages call

ruiconti (approved):

packages/eve/src/public/channels/slack/api.ts
AndrewBarba: Good call — collapsed into a single `refreshMessages` with the fetch scoped inside the coalescing check in fdc7e571, so an uncoalesced call is no longer possible.

PR #1118, opened 2026-07-23
feat(eve): slack - expose cancellation in handler contexts

AndrewBarba:
## Summary

- expose thread-bound `ctx.cancel({ turnId? })` helpers to Slack message hooks and custom interactions
- expose target-addressed cancellation to `onEvent`
- document debounce-style replacement delivery and publish the new context types

## Why

Cancellation is available to raw `defineChannel` routes, but the Slack channel did not forward that route helper into its authored inbound contexts. As a result, Slack handlers could not cancel an in-flight turn before delivering a newer message or implement a custom Stop interaction.

This maps Slack channel/thread identities back to the existing channel-local continuation token and delegates to the standard cancellation path.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack/slackChannel.test.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm fmt --check`
- `pnpm docs:check`
- `pnpm guard:invariants`


docs/channels/slack.mdx
```diff
@@ -104,6 +104,25 @@ async onMessage(ctx, message) {
 
 Like `threadContext`, this helper calls `conversations.replies` and requires the matching Slack history scope. It observes at most the first 50 messages of the thread, and a failed fetch is logged and swallowed, so the returned list may be empty or stale — treat an unexpected empty list as "don't route" rather than "no participants".
 
+#### Cancel and replace in-flight work
+
+Message hooks (`onMessage`, `onAppMention`, and `onDirectMessage`) receive a thread-bound `ctx.cancel({ turnId? })` helper. Call it before returning `{ auth }` to stop the current turn and queue the new Slack message as replacement input, producing a debounce-like experience:
```
benpankow, line 109: is the agent aware of the previous turn input if it's canceled? e.g. with the below code, if I send another message, does the agent completely ignore the previous one, or does the input text accumulate?

benpankow (approved):
🙌

docs/channels/slack.mdx
AndrewBarba: the input text accumulates its already in the messages array

PR #1127, opened 2026-07-23
fix(eve): build - analyze dynamic tool captures from syntax

### LINKED ISSUE #1086: Build transform captures closure variables from object keys and string literals, breaking dynamic tool resolvers at runtime
### Summary

The build transform for dynamic tools (`defineDynamic` under `agent/tools/`) computes each lifted `execute`'s closure captures from a token-level scan of the function source. Identifiers that appear **only as object-literal keys** or **only inside string literals** are captured as closure variables. When the same name happens to be declared later in the resolver (or is the resolver's own `tools` accumulator), the generated capture object references a binding that is not in scope (or captures the accumulator itself), and the resolver throws at `session.started` — which silently disables **every** tool from that resolver:

```
[eve:dynamic-tools] Dynamic tool resolver (session.started) threw — skipping. { error: 'url is not defined' }
```

Reproduced on `eve@0.24.6` and `eve@0.27.0` (build path only, `eve build` / Vercel Build Output — `eve dev` is unaffected, so the failure first appears in production).

### Minimal repro

`agent/tools/probe.ts`:

```ts
import { defineDynamic, defineTool } from "eve/tools";
import { z } from "zod";

  events: {
    "session.started": async () => {
      const tools: Record<string, ReturnType<typeof defineTool>> = {};

      tools.probe_alpha = defineTool({
        description: "probe alpha",
        inputSchema: z.object({ q: z.string() }),
        async execute({ q }) {
          // `url` appears ONLY as an object key; `tools` ONLY inside a string.
          return { url: `/x/${q}`, note: "answer from app tools only" };
        },
      });

      {
        const url = "https://example.com";
        const cred = { a: 1 };
        tools.probe_beta = defineTool({
          description: "probe beta",
          inputSchema: z.object({}),
          async execute() {
            return { target: url, cred };
          },
        });
      }

      return tools;
    },
  },
});
```

Run `eve build`, then inspect `.output/server/index.mjs`:

```
probe_alpha captures: tools, url     // expected: (none)
probe_beta  captures: url, cred      // correct
```

`probe_alpha`'s capture object `{ tools, url }` is evaluated at resolver time where `url` (block-scoped below) is not in scope → ReferenceError → the resolver is skipped and all its tools disappear.

### Expected

Free-variable analysis should be scope-aware: object keys and string contents are not variable references. `probe_alpha` should capture nothing.

### Impact

Hit this in production: a legitimately-named `const url` in one tool block plus an unrelated tool whose *result object* had a `url:` key made the entire tool surface of our channel bot vanish on deploy, with no build-time diagnostic. Also observed a captured *function* helper (shared between executes), which serializes into `__closureVars` but cannot be reconstructed on replay — a build-time error for non-serializable captures would help there too.

### Environment

eve 0.24.6 and 0.27.0 · Node 24.18.0 · darwin-arm64 build host · Vercel Build Output deploy target

AndrewBarba:
Closes #1086.

## What changed

- replace source-text regex matching with AST-based identifier-reference collection for lifted dynamic tool functions
- exclude object keys, non-computed member names, binding positions, labels, and TypeScript-only syntax from closure captures
- preserve real computed/reference expressions and existing inner-scope shadowing behavior
- add an end-to-end transform regression matching the reported multi-tool TDZ failure

## Why

Scanning the function body as text treated names inside object keys and string literals as free variables. A later block-scoped declaration with the same name then appeared in the generated capture object before initialization, throwing during `session.started` and silently removing every tool from the resolver.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/internal/workflow-bundle/dynamic-tool-transform.test.ts`
- `pnpm fmt`
- `pnpm lint`

### DIFF
--- a/packages/eve/src/internal/workflow-bundle/dynamic-tool-ast-references.ts
+++ b/packages/eve/src/internal/workflow-bundle/dynamic-tool-ast-references.ts
@@ -1,0 +1,140 @@
+/** Rolldown AST subset consumed by the dynamic-tool transform. */
+export type DynamicToolAstNode = {
+  argument?: DynamicToolAstNode | null;
+  arguments?: DynamicToolAstNode[];
+  async?: boolean;
+  body?:
+    | DynamicToolAstNode
+    | DynamicToolAstNode[]
+    | { body?: DynamicToolAstNode[]; type?: string; start?: number; end?: number };
+  callee?: DynamicToolAstNode;
+  computed?: boolean;
+  declaration?: DynamicToolAstNode | null;
+  declarations?: DynamicToolAstNode[];
+  end?: number;
+  expression?: DynamicToolAstNode | null;
+  id?: { name?: string; start?: number; end?: number } | null;
+  init?: DynamicToolAstNode | null;
+  key?: DynamicToolAstNode | null;
+  kind?: string;
+  left?: DynamicToolAstNode | null;
+  method?: boolean;
+  name?: string;
+  params?: DynamicToolAstNode[];
+  properties?: DynamicToolAstNode[];
+  right?: DynamicToolAstNode | null;
+  start?: number;
+  type?: string;
+  value?: DynamicToolAstNode | unknown;
+};
+
+/**
+ * Collects identifiers used as runtime references in a function body AST.
+ */
+export function collectReferencedIdentifierNames(node: DynamicToolAstNode): Set<string> {
+  const names = new Set<string>();
+
+  const visit = (
+    current: DynamicToolAstNode,
+    parent: DynamicToolAstNode | undefined,
+    parentKey: string | undefined,
+    ancestors: readonly DynamicToolAstNode[],
+  ): void => {
+    if (
+      current.type === "Identifier" &&
+      current.name &&
+      isIdentifierReference(parent, parentKey, ancestors)
+    ) {
+      names.add(current.name);
+    }
+
+    const nextAncestors = [...ancestors, current];
+    for (const [key, value] of Object.entries(current)) {
+      if (Array.isArray(value)) {
+        for (const child of value) {
+          if (isAstNode(child)) {
+            visit(child, current, key, nextAncestors);
+          }
+        }
+      } else if (isAstNode(value)) {
+        visit(value, current, key, nextAncestors);
+      }
+    }
+  };
+
+  visit(node, undefined, undefined, []);
+  return names;
+}
+
+function isIdentifierReference(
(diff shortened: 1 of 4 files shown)


ctgowrie (approved):

packages/eve/src/internal/workflow-bundle/dynamic-tool-ast-references.ts
```diff
@@ -0,0 +1,140 @@
...
+  ancestors: readonly DynamicToolAstNode[],
+): boolean {
+  if (!parent || !parentKey) return false;
+
+  if (
+    parentKey === "typeAnnotation" ||
+    parentKey === "returnType" ||
+    parentKey === "typeParameters" ||
+    parentKey === "typeArguments" ||
+    ancestors.some((ancestor) => ancestor.type?.startsWith("TS"))
```
AndrewBarba: [P1] Preserve runtime expressions nested under TypeScript wrappers. This blanket ancestor check also rejects the expression side of `TSAsExpression`, `TSNonNullExpression`, and `TSSatisfiesExpression`. For example, `const fallback = "x"; execute() { return (fallback as string).length; }` now emits `__closureVars: {}` while the hoisted function still references `fallback`, so replay throws `ReferenceError`. Please ignore only the type-bearing fields while continuing to visit each wrapper's runtime expression.

packages/eve/src/internal/workflow-bundle/dynamic-tool-ast-references.ts
```diff
@@ -0,0 +1,140 @@
...
+      parentKey === "label")
+  ) {
+    return false;
+  }
+
+  if (
+    parent.type === "ObjectPattern" ||
+    parent.type === "ArrayPattern" ||
+    ancestors.some(
+      (ancestor) => ancestor.type === "ObjectPattern" || ancestor.type === "ArrayPattern",
```
AndrewBarba: [P1] Keep executed expressions inside binding patterns. Computed keys and default initializers are runtime references even though the surrounding pattern is a binding position. With `const fallback = "x"; execute(input) { const { value = fallback } = input; return value; }`, `fallback` is dropped here, the transform emits `__closureVars: {}`, and the hoisted execute throws when the default is needed. Please skip the bound identifiers but still traverse computed keys and `AssignmentPattern.right`.

AndrewBarba (commented):
I found two runtime-reference cases that regress from the previous text scan. Both leave a free identifier in the hoisted function while emitting an empty closure snapshot.

PR #1142, opened 2026-07-23
fix(eve): make Ctrl+C exit the dev TUI quickly and safely

chadhietala:
### Description

**Before:** Ctrl+C in the `eve dev` TUI could hang the command, leave the local server running, or leave the shell in raw mode with a hidden cursor. Shutdown behavior differed between the TUI, headless mode, and `eve start`.

**After:** the local dev server runs in a forked child that is `detached` (its own process group) and is stopped through its IPC channel: the parent sends a `shutdown` message on close, and the child also stops itself when the channel disconnects, so it never outlives the parent. The design rationale is documented in `local-server-process.ts`.

- One CLI shutdown lifecycle (`#cli/shutdown.js`), also used by `eve start`: the first SIGINT/SIGTERM requests a stop, a second one force-exits with 130/143, and an optional backstop bounds cleanup.
- Ctrl+C/Ctrl+D on an empty prompt starts the server close at the keypress, concurrent with terminal teardown; Ctrl+C on a non-empty prompt clears the draft instead.
- `close()` escalates IPC shutdown → process-group SIGTERM → SIGKILL; a test enforces the total budget stays under the forced-exit backstop, so `eve dev` exits 0 within a second.
- A forced exit (second Ctrl+C or the backstop) synchronously restores the terminal: raw mode, cursor, bracketed paste.
- Each dev run records a cleanup-intent file; the next start reconciles intents from dead owners and stops leaked sandbox resources.

- `pnpm fmt`, `pnpm lint`, `pnpm typecheck`, `pnpm guard:invariants`
- `pnpm test:unit` and `pnpm test:integration` (full suites)
- `vitest run --config vitest.scenario.config.ts test/scenarios/dev-server-shutdown.scenario.test.ts` — asserts the whole `eve dev --no-ui` tree exits 0 within one second of SIGINT/SIGTERM with no forced kill, including a signal delivered mid-startup
- New unit coverage: close-escalation staging and budget-vs-backstop invariant (`local-server-process.test.ts`), lifecycle listener hygiene (`shutdown.test.ts`), forced-exit terminal restore and Ctrl+C draft-clearing (`terminal-renderer.test.ts`), close-await ordering (`run.test.ts`)

### PR Checklist

### DIFF
--- a/packages/eve/src/cli/dev/local-server-process.ts
+++ b/packages/eve/src/cli/dev/local-server-process.ts
@@ -1,0 +1,196 @@
+/**
+ * The `eve dev` server child: forked `detached` (its own process group) and
+ * bound to the CLI's lifetime through the IPC channel, never through the OS.
+ *
+ * Why the child is not bound to the parent at the OS level (shared process
+ * group, inherited terminal signals):
+ *
+ * - The TUI puts stdin in raw mode, which disables ISIG: Ctrl+C never becomes
+ *   a terminal-generated SIGINT, so same-group signal delivery does nothing in
+ *   the mode where Ctrl+C matters most.
+ * - In headless mode, same-group delivery would signal parent and child
+ *   simultaneously, racing the parent's graceful IPC shutdown against the
+ *   child's own signal handler.
+ * - macOS has no parent-death signal. IPC `disconnect` is the portable
+ *   equivalent: the child self-stops whenever this process exits, for any
+ *   reason including SIGKILL (see local-server-child.ts).
+ *
+ * `detached` makes the child a process-group leader, so close() can escalate
+ * to a group-wide SIGTERM/SIGKILL that reaps the entire server tree. Sandbox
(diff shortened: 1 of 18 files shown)


packages/eve/src/cli/dev/local-server-process.ts
```diff
@@ -0,0 +1,196 @@
...
+    const active = child;
+    if (active === undefined) return;
+    child = undefined;
+    active.stdout?.destroy();
+    active.stderr?.destroy();
+    if (active.connected) active.disconnect();
+    active.unref();
+  };
+
+  const start = (): Promise<DevelopmentServerHandle> => {
```
ruiconti, line 85: Is the spawning and IPC overhead noticeable in `eve dev`'s boot time?

chadhietala: Not that I can tell.

ruiconti (approved):

PR #1148, opened 2026-07-23
feat(eve): add slack-specific response instructions around mention syntax

benpankow:
### Description

We already have some platform-specific response instructions for Teams, Telegram etc. Adds instructions for Slack which includes tips on @-mentioning a user using the `<@USER_ID>` syntax Slack requires, since in the wild eve has done this incorrectly (sending bare `@USER_ID`).

deployed manually, tested that eve @-ed me properly, unit test

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/.changeset/fuzzy-slack-mentions.md
+++ b/.changeset/fuzzy-slack-mentions.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Slack turns now tell the agent to format user mentions as `<@USER_ID>` so generated replies notify the intended user instead of displaying a bare ID.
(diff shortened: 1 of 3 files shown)


benpankow:
Closing as a stale micro-fix against an older Slack prompt implementation. If incorrect Slack mention syntax is still reproducible on current `main`, we should add focused guidance and coverage against the current channel model context.

PR #1154, opened 2026-07-24
feat(eve): add registry commands

cmpadden:
## Summary
- add `eve add <item>` for installing official, configured, or URL-addressed registry items
- add `eve registry add|list|search|view` for registry configuration and discovery
- store third-party namespace mappings in `package.json#registries`
- resolve bare item paths through eve’s official registry and aggregate official plus configured registries by default
- use shadcn 4.16’s programmatic registry APIs and explicit registry configuration instead of invoking its CLI
- keep the eve surface intentionally small, with only `--overwrite` on `eve add` and `--registry` for scoped list/search

## Package size
Vendoring the shadcn registry installer adds approximately **13 MB raw / 3.3 MB gzip** to the published eve package. The installer currently pulls in shadcn’s project-transformation dependencies. The shadcn maintainers plan to extract registry functionality into a separate package in the near future; eve can switch to that smaller package when it is available.

## Validation
- `pnpm --filter eve build`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/cli/commands/registry.test.ts src/cli/run.test.ts`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`
- local HTTP smoke test installing `@local/sdk-smoke` through `package.json#registries`

The full unit suite also ran; its only failures were two existing package-manager expectation assertions that expected `pnpm install` while the test environment selected `npm install`.


docs/reference/cli.md
```diff
@@ -77,6 +79,19 @@ eve extension build
...
 
+## `eve extensions`
+
+Commands for installing extension mounts through [shadcn registries](https://ui.shadcn.com/docs/registry). Bare item slugs use eve's official registry; URLs and configured registry addresses pass through to shadcn unchanged.
+
+```bash
+eve extensions add agent-browser --yes
+eve extensions add https://example.com/r/my-extension.json --overwrite
+eve extensions registry add @acme=https://example.com/r/{name}.json
+eve extensions add @acme/my-extension --yes
```
ctgowrie: Wonder if this should just be `eve add`. The fact that it's an extension is not important. We could also move our default channels to this too. What do you think?

docs/reference/cli.md
```diff
@@ -77,6 +79,21 @@ eve extension build
...
+```bash
+eve add extension/agent-browser --yes
+eve add https://example.com/r/my-extension.json --overwrite
+eve registry add @acme=https://example.com/r/{name}.json
+eve registry search browser
+eve registry view @acme/my-extension
+eve add @acme/my-extension --yes
+```
+
+`eve registry add` adds configured sources. `eve registry list` and `search` aggregate the official catalog and all configured sources by default, or browse one supplied URL or namespace. The wrapped shadcn CLI still requires its project configuration, including `components.json`.
```
ctgowrie: Worth elaborating on these commands a bit more in the docs imo

cmpadden: Added in child PR. Good call.

packages/eve/src/cli/commands/registry.ts
```diff
@@ -0,0 +1,164 @@
...
+  validateRegistrySource(source);
+  const config = await readRegistryConfig(appRoot);
+  const sources = source ? [source] : [OFFICIAL_CATALOG, ...configuredRegistrySources(config)];
+  const result = await searchRegistries(sources, {
+    config,
+    continueOnError: sources.length > 1,
+    query,
+  });
+  printSearchResults(logger, result.items);
+  for (const error of result.errors ?? []) {
```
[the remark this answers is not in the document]
cmpadden: Fixed. Registry searches now suppress the empty-result message when every source fails, report each source error, and set a failing exit code whenever any source fails. Added regression coverage for the total-failure case.

ctgowrie (approved):

ctgowrie (approved):

PR #1158, opened 2026-07-24
fix(docs): remove duplicate GitHub link in mobile nav menu

### LINKED ISSUE #1157: Duplicate "GitHub" link in docs site mobile nav menu
### Description

The docs site's mobile nav menu shows two identical "GitHub" links stacked on top of each other.

### Cause

`apps/docs/geistdocs.tsx` adds an explicit `GitHub` entry to the `nav` array, but the geistdocs `MobileMenu` already appends its own "GitHub" link derived from the `github` config (`internal/navbar/mobile-menu.js` renders `nav.map(...)` followed by an unconditional GitHub `NavLink` whenever `githubUrl` is set). With both in place, the mobile menu renders the link twice. The desktop navbar is similarly doubled: the nav text link plus the built-in GitHub icon button.

### Reproduction

1. Open https://eve.dev on a mobile viewport.
2. Open the hamburger nav menu.
3. Two "GitHub" links appear between "Templates" and "Vercel OSS".

Image

### Proposed fix

Remove the redundant `GitHub` entry from the `nav` array in `apps/docs/geistdocs.tsx` and rely on the layout's built-in GitHub affordances from the `github` config.

lucasdoell:
### Problem

Fixes #1157. The docs site's mobile nav menu shows two identical "GitHub" links: the geistdocs `MobileMenu` appends its own GitHub link derived from the `github` config, on top of the explicit `GitHub` entry in the site's `nav` array.

### Solution

Disable the layout's built-in GitHub affordance with `navbarGithub: { enabled: false }`. The `nav` entry remains the single source of the link:

- **Mobile menu**: one "GitHub" link (from `nav`) instead of two.
- **Desktop navbar**: keeps the "GitHub ↗" text link unchanged; the redundant GitHub icon button next to "Ask AI" is removed.

Verified locally at desktop (1280px) and mobile (390px) viewports.

Docs-app-only change, so no changeset.


lucasdoell:
One tradeoff to be aware of: `navbarGithub` also controls the small GitHub icon button that sat next to "Ask AI" on the desktop navbar (≥768px), so this change removes that too. It was redundant with the "GitHub ↗" text link in the nav, but if we want to keep both, the alternative is an upstream fix in [vercel/geistdocs](https://github.com/vercel/geistdocs) — e.g. having `MobileMenu` skip its auto-appended GitHub link when `nav` already contains one pointing at the repo.

PR #1192, opened 2026-07-24
feat(eve): add agent-semantic OTel provider

chadhietala:
## Summary

Phase 3 of local observability, stacked on #1191.

This PR adds the first lifecycle provider and defines eve’s agent-first OpenTelemetry semantics:

- Injected `Tracer`; no global OTel registration
- Provider-owned serializable session/turn trace-state store
- `SpanContext` restoration through `wrapSpanContext()`
- Zero-duration session and turn start markers
- Retrospective turn terminal markers
- `agent.step` and `agent.action` spans scoped to one atomic attempt
- Stable `agent.*` identity, framework, lineage, model, usage, and action attributes
- Tool/action failures remain localized to the action span
- Content is not captured

Only attempt/model/tool state that starts and terminates within one atomic invocation remains process-local. No live `Span` object is expected to survive a Workflow boundary.

The same provider can later run with a private local tracer or the global tracer configured by `@vercel/otel`.

## Scope

This PR does **not** register the provider, wire harness-native lifecycle events, persist the state store, call `registerOTel()`, or change production behavior. The provider is exercised only through real in-memory OTel tests.

## Testing

- Session root ignores ambient parent context
- Provider recreation restores session trace identity
- Multiple turns share one trace ID
- Turn, attempt, and action logical hierarchy
- Session event placement
- Structural attributes and no content capture
- Action failure does not fail the turn
- Provider-neutral bridge tests
- Tool-loop regression tests
- Package typecheck, build, lint, formatting, and invariants

### DIFF
--- a/packages/eve/src/harness/agent-otel-provider.ts
+++ b/packages/eve/src/harness/agent-otel-provider.ts
@@ -10,6 +10,7 @@ import {
 } from "@opentelemetry/api";
 
 import type {
+  InstrumentationAttemptScope,
   InstrumentationAttemptStartedEvent,
   InstrumentationAttemptTerminalEvent,
   InstrumentationModelCallTerminalEvent,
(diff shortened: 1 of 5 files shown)


packages/eve/src/harness/agent-otel-provider.ts
```diff
@@ -0,0 +1,407 @@
...
+    readonly tools: Map<string, Context>;
+  } {
+    let state = executionContexts.get(scope);
+    if (state === undefined) {
+      state = { models: new Map(), tools: new Map() };
+      executionContexts.set(scope, state);
+    }
+    return state;
+  }
+}
```
ctgowrie, line 375: Does this work in serverless workflow runtime?

chadhietala: Yea it should work. I added a test for it in [this commit](https://github.com/vercel/eve/pull/1192/commits/2f63650b1b6595b24d09b6891ec290abce30464a)

ctgowrie (approved):

Issue #1170, opened 2026-07-24
Stream reactors: separate event producers from channel/callback consumers


ruiconti:
Today an event's delivery is welded to its producer's call stack: the emit path inside a workflow step synchronously invokes the channel adapter (`callAdapterEventHandler`) and the remote-callback forwarder (`forwardSessionCallbackNotification`) before writing the durable stream. Consumers only run when the producer's compute is awake.

This coupling causes concrete failures for notification-class events (a remote callee's `authorization.required`/`authorization.completed` surfaced on the caller's stream, #1167):

- A notification callback can only land while the caller's turn is parked at its inbox `iterator.next()`; parked-for-input or between turns, `resumeHook` throws and the event is silently dropped.
- Resuming a durable workflow (context deserialization, session hydration, a journaled step) to write one rendering-only event.
- Local and remote children take different producer paths, so channels implement the same rendering twice.
- Adapter state persists through the producer's cursor, so nothing outside the workflow can safely invoke a channel.

Proposal: producers just deliver; consumption moves to a per-session **event consumer** — a sidecar run — an auxiliary workflow run deployed 1:1 with the session, with its own journal — parked on `<sessionId>:events` that invokes the channel's existing adapter event handler and appends the wrapped `subagent.event` to the session stream, with no session-state write-back. "Notification" is one event class on the lane; the consumer is the general entity, normalizing event consumption and paving the way for stream multiplexing and an eventual `defineEventStream({ onEvent })`. v1 scopes the consumer to the caller side (where the queue wake already exists today); design in `research/event-consumers.md`.

PR #1194, opened 2026-07-24
feat(eve): prototype the per-session notification consumer

ruiconti:
**Stacked on #1167** (base: `rui/remote-agent-callback-notifications`) — only the top three commits are this PR. Prototype for the plan in #1172 / issue #1170: validate the consumer design end to end and surface what the real implementation has to deal with. Draft on purpose.

## What

- **`workflow//eve//notificationConsumer`** — one companion run per session, started by the entry workflow with the driver's stream writable and serialized channel context, parked on `<sessionId>:notify`. Each delivery is one journaled step: invoke the channel's existing adapter event handler against the curated child event, append the wrapped `subagent.event` (now carrying `childSessionId`) to the session stream. The session workflow is never woken for notifications.
- **Producers ring it fire-and-forget.** Callback metadata bakes a `notifyUrl` targeting the consumer hook; the callee posts notifications there (legacy `url` fallback for pre-consumer callers); the route rings the token and 202s once the payload is well-formed — a disposed consumer logs and drops.
- **`fix(eve)`: OpenAPI `baseUrl` resolves from the live authored module.** Found while making the fixture work without the `eve/tools` exposure: the compiled manifest bakes the build-time value, where the deployment's own origin doesn't exist yet, while `auth`/`spec`/`operations` already resolve live. Standalone fix, extractable.
- **Fixture drops the `eve/tools` exposure** (reverts that commit from the base branch): auth triggers are now interactive OpenAPI connections fronting a bearer-gated `/probe/credential` route on the fixture's own deployment, with a custom `AuthFn` minting the user principal interactive auth requires. Existing public API only.

All three evals pass through the consumer — 34/34 gates, ~17s locally — including the new coverage that motivated everything: `hitl-collision` receives and completes an authorization notification **while the turn is parked for input**, the state where today's architecture silently drops it (`resumeHook` → 404).

## What the prototype surfaced

- **Session-opening race**: the consumer creates its hook on its first queue invocation, so a fast callee can ring before it exists. Bounded `HookNotFoundError` retry in the route covers it; hooks are creator-owned so entry-side pre-creation isn't available.
- **Stable-id trap**: `start(reference)` silently creates a run that later dies `WorkflowNotRegisteredError` unless the function name is in `STABLE_WORKFLOW_NAMES` — that set is the whole contract between the bundler and the reference template.
- **Eval-side win**: with delivery decoupled from turn state, the hitl eval reads the wrapped notification off the durable stream instead of blind-polling a derived hook URL, which also killed its ordering flake.

## Known-red, deferred to the real change

- `childSessionId` on `SubagentChildEventStreamEvent` trips the rule-36 capability-epoch ceremony on four contracts (`dynamicTool`, `hook`, `dynamicSkill`, `dynamicInstructions`) — additive/compatible, ceremony not run here, so `guard:invariants` fails on this branch.
- Local subagent proxying still uses the legacy re-emit (the one-path migration is the plan's scope, not the prototype's).
- Dispose is a poison-pill ring from the entry's `finally`; a crash before it leaks a parked run.
- No changesets; not for merge as-is.

Unit suite 5368 passed, typecheck 33/33, lint clean.


ruiconti:
Restacking: this lands before #1167, not after — replacement PR based on the research branch follows.

PR #1196, opened 2026-07-24
feat(eve): ingest session notifications inline in the callback route

ruiconti:
**Stacked on #1172** (base: `rui/notification-reactors-research`) and **lands before #1167**. Prototype for issue #1170, scoped to the caller-side ingestion of notification-class events — no remote-agent propagation here. #1167 restacks on top and carries the remote lane (callback envelope, callee forwarding, the `agent-remote-callbacks` fixture and its evals).

## What

Notification-class child events (`authorization.required`/`authorization.completed` surfaced on a caller's stream) reach the channel and the stream **with no run, no hook, and no queue wake** — the event arrives as an HTTP callback, so the compute is already paid. `POST /eve/v1/callback/<sessionId>:events`:

1. validates the closed notification vocabulary and wraps the event as `subagent.event` (carrying `childSessionId`);
2. loads the session entry run's serialized channel context (`world.runs.get`, read-only — no dedicated persistence) and runs the channel's **existing** adapter event handler against it, so push channels render with zero authoring changes;
3. appends the wrapped event to the session's durable stream (`world.streams.write` by the run-derived name) for followers (TUI, web, evals).

Both effects are best-effort per the lane's fire-once contract: a finished session logs and drops; nothing retries.

Three commits: two standalone harness fixes (same-step approval + runtime-action drop; live-module OpenAPI `baseUrl`) and the ingestion itself.

## Why inline, not a run

Earlier iterations delivered through a per-session **sidecar run** (a workflow parked on `<sessionId>:events`, catching up over the stream by cursor). It works, and the replay-cost argument for it is real — but it over-serves *this* lane. The notification class is fire-once and drop-ok: it needs neither the sidecar's journaled exactly-once, its ordering, nor its consumer-held state. Inline ingestion in the callback's own compute is the minimal shape that satisfies the contract. The sidecar survives in the [research doc](https://github.com/vercel/eve/pull/1172) as the **graduated tier** for event classes that genuinely need cursor semantics, ordering, or state (progress, proxied HITL), gated behind the frequency-admission rule.

The load-bearing feasibility questions were verified on disk: `world.streams.write` and `world.runs.get` both work from plain route code (no workflow context), and the wrapped `subagent.event` was observed on the parent's durable stream, written by the HTTP handler.

## Known-red / deferred

- `childSessionId` on `SubagentChildEventStreamEvent` trips the rule-36 capability-epoch ceremony on four contracts (additive/compatible; ceremony not run), so `guard:invariants` fails on this branch.
- The default stream name (`strm_<id>_user`) is a vendored-engine naming contract; the engine should eventually expose a resolver or accept name-less default writes.
- End-to-end validation (the remote fixture + evals) lands with #1167 once it restacks — the ingestion mechanism is proven here, but the full child→caller→reply loop needs the coherent remote lane, not a hand-spliced mix.

Unit suite green, typecheck clean.

### DIFF
--- a/packages/eve/src/execution/notification-ingestion.ts
+++ b/packages/eve/src/execution/notification-ingestion.ts
@@ -1,0 +1,69 @@
+import { buildAdapterContext } from "#channel/adapter-context.js";
+import { callAdapterEventHandler } from "#channel/adapter.js";
+import { deserializeContext } from "#context/serialize.js";
+import { appendSessionStreamEvent } from "#execution/session-stream-append.js";
+import { getWorld } from "#internal/workflow/runtime.js";
+import { createLogger, logError } from "#internal/logging.js";
+import type { SubagentChildEventStreamEvent } from "#protocol/message.js";
+import { ChannelKey } from "#runtime/sessions/runtime-context-keys.js";
+
+const log = createLogger("execution.notification-ingestion");
+
+/**
+ * PROTOTYPE (issue #1170): ingests one notification-class event for a
+ * session, inline in the caller's already-running compute — no run, no
+ * hook, no queue wake.
+ *
+ * Two effects, both best-effort:
+ *
+ * 1. Run the channel's existing adapter event handler against the curated
+ *    child event — the same handler the local proxy step invokes today, so
+ *    channels render with zero changes. The channel context is loaded from
+ *    the session entry run's input (`world.runs.get`), read-only.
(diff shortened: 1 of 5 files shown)


ruiconti:
Superseded — reparenting the stack (now `1202 → 1196 → 1172`) tripped GitHub's stacked-branch auto-merge detection and flipped this to MERGED without anything reaching main. Reopened as a fresh PR based on #1202.

PR #1204, opened 2026-07-25
feat(eve): ingest session notifications inline in the callback route

ruiconti:
**Relay lane** for #1170. Stacked on #1202 (terminal/control lane); the research doc PR stacks on top of this one. Recreated after a stack reparent flipped the prior PR to a false MERGED state and auto-deleted its branch — nothing reached `main`.

## What

Notification-class child events (`authorization.*`) reach the channel and the stream with **no run, no hook, no queue wake** — the event arrives as an HTTP callback, so the compute is already paid. `POST /eve/v1/callback/<sessionId>:events`:

1. validates the reserved notification vocabulary and wraps the event as `subagent.event` (carrying `childSessionId`);
2. loads the session entry run's channel context (`world.runs.get`, read-only) and runs the channel's existing adapter event handler against it — push channels render with zero authoring changes;
3. appends the wrapped event to the session's durable stream (`world.streams.write`) for followers.

Best-effort per the lane's fire-once contract: a finished session logs and drops; nothing retries. Activates the `"notification"` status that #1202 reserved on the wire.

## Two-lane split

Relay lane (fire-once, drop-ok, rendering-only) — counterpart to #1202's control lane (exactly-once, never-drop, durable resume).

## Known-incomplete (prototype)

- **Caller-side ingestion only.** The callee-side forwarder that *sends* `authorization.*` to `:events` isn't in this PR yet (lives in #1167's commit, needs adapting), so nothing produces on the lane end-to-end until it lands; the fixture + evals come with it.
- `childSessionId` on `SubagentChildEventStreamEvent` trips the rule-36 capability-epoch ceremony (additive; not run), so `guard:invariants` fails on this branch.
- Default stream name (`strm_<id>_user`) is a vendored-engine naming contract.

Typecheck clean; route + unit tests green.

### DIFF
--- a/packages/eve/src/execution/notification-ingestion.ts
+++ b/packages/eve/src/execution/notification-ingestion.ts
@@ -1,0 +1,69 @@
+import { buildAdapterContext } from "#channel/adapter-context.js";
+import { callAdapterEventHandler } from "#channel/adapter.js";
+import { deserializeContext } from "#context/serialize.js";
+import { appendSessionStreamEvent } from "#execution/session-stream-append.js";
+import { getWorld } from "#internal/workflow/runtime.js";
+import { createLogger, logError } from "#internal/logging.js";
+import type { SubagentChildEventStreamEvent } from "#protocol/message.js";
+import { ChannelKey } from "#runtime/sessions/runtime-context-keys.js";
+
+const log = createLogger("execution.notification-ingestion");
+
+/**
+ * PROTOTYPE (issue #1170): ingests one notification-class event for a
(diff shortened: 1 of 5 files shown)


ruiconti:
Closing per design correction: authorization.* is low-frequency control, not high-frequency progress, so it belongs on the main run's durable resumeHook path (handled in workflowEntry like a local subagent), not this best-effort route-direct relay. Rebuilding on #1202 as remote authorization via the durable path; the progress relay this branch prototyped is reserved until a concrete progress event type exists. See #1205 for the corrected framing.

PR #1205, opened 2026-07-25
docs(eve): research - event consumers plan

ruiconti:
**Research plan** for #1170. Top of the stack: `main → #1202 (terminal) → #1204 (relay) → this`. Design only, no runtime changes. Recreated after a stack reparent closed the prior PR (#1172) — the branch and commits are intact.

## What

Normalizes event consumption around eve's durable session stream. Two lanes, distinguished by delivery guarantee:

- **control** — `session.completed/failed`, `input.required`: exactly-once, never-drop, durable resume of the parked parent turn. (Implemented in #1202.)
- **relay / telemetry** — `authorization.*`, future `working`/progress: fire-once, drop-ok, rendering-only. Route-direct ingestion in v1. (Implemented in #1204.)

The doc frames the relay lane as the on-ramp for every event class that should propagate without waking a session, and records the **sidecar run** as the graduated tier (its own run, its own journal — the replay-cost firewall) for event classes that later need cursor semantics, ordering, or consumer-held state.

## Note

The doc's iteration history explored a per-session sidecar consumer before the prototype (#1204) showed route-direct ingestion satisfies the v1 relay contract with far less machinery. The doc retains the sidecar as the graduated tier rather than the v1 default. Fifteen commits of design iteration are preserved deliberately as the archaeology of that decision.

### DIFF
--- a/.changeset/remote-authorization-durable.md
+++ b/.changeset/remote-authorization-durable.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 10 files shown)


ruiconti:
Reparenting: #1204 was closed and replaced by #1207 (remote authorization via the durable path). Recreating this doc PR on top of #1207.

PR #1214, opened 2026-07-25
feat(eve): add disableSubagent helper

muratsu:
## Summary

- add `disableSubagent()` as a public `eve` helper
- omit disabled local or remote subagents from the compiled graph and model-visible capability surface
- document conditional subagent configuration and add a patch changeset

## Testing

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/definitions/subagent.test.ts src/compiler/normalize-manifest.test.ts`
- `pnpm docs:check`
- `pnpm guard:invariants`
- focused `oxlint` checks

### DIFF
--- a/.changeset/tidy-subagents-rest.md
+++ b/.changeset/tidy-subagents-rest.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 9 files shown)


ctgowrie (commented):
Let’s make defineDynamic work with subagents instead. Then simply return no subagent when env is not set.

PR #1218, opened 2026-07-25
fix(eve): prefer known Telegram attachment media type over HTTP content-type

### LINKED ISSUE #1217: Telegram channel serves wrong media type for photos (breaks vision models)
## Link to a minimal reproduction

A deterministic unit-test reproduction is included in the linked PR branch (no live Telegram account required, since the bug is in media-type resolution, which is mockable):

- https://github.com/harlley/eve/blob/fix/telegram-attachment-media-type/packages/eve/src/public/channels/telegram/attachments.test.ts

The test `createTelegramFetchFile > prefers the known attachment media type over the HTTP content-type header` **fails on `main`** and **passes with the fix**.

## Steps to reproduce

The reproducible path is a unit test that mirrors a real Telegram photo upload:

1. An inbound Telegram photo has `kind: "photo"`, so the channel records the attachment media type as `image/jpeg` (see `toTelegramFilePart`).
2. On `fetchFile`, the channel calls Telegram's `getFile` + file-download endpoint. Telegram's file endpoint frequently responds with `Content-Type: application/octet-stream` (or no content-type) for photos.
3. `createTelegramFetchFile` currently resolves the media type as:

   ```ts
   const mediaType =
     response.headers.get("content-type") ?? ref.mediaType ?? "application/octet-stream";
   ```

   So the HTTP header (`application/octet-stream`) wins over the known `image/jpeg`.
4. The resulting `mediaType: "application/octet-stream"` is handed to the model. Vision models key off the declared media type and either reject the part or cannot interpret it as an image.

## Current vs. expected behavior

**Current:** A photo sent to a Telegram-channelled eve agent is delivered to the model with `mediaType: "application/octet-stream"`. Vision-capable models (e.g. Gemini) cannot recognize the image and respond as if no image arrived. With an `uploadPolicy: { allowedMediaTypes: ["image/*"] }`, the photo is also **rejected outright** before reaching the model:

```
Telegram file rejected — photo.jpg has media type "application/octet-stream"
which is not allowed by this route. Allowed: image/*.
```

(That error string is exactly what the regression test asserts against on unpatched `main`.)

**Expected:** The channel already knows the attachment's media type from the inbound message (`image/jpeg` for photos). That known type should win, so the model receives a recognized image media type and the upload policy matches.

## Root cause

`packages/eve/src/public/channels/telegram/attachments.ts` — `createTelegramFetchFile` prioritizes the HTTP `content-type` header over `ref.mediaType`. Telegram's file endpoint is an unreliable source for the content type, so the known attachment media type should take precedence.

## Proposed fix

Invert the priority:

```ts
const mediaType =
  ref.mediaType ?? response.headers.get("content-type") ?? "application/octet-stream";
```

One-line change plus the regression test linked above. PR to follow.

## eve version

`eve@0.27.6`

## Environment

```sh
OS: macOS 26.5.2 arm64
Node: v24.8.0
Package manager: pnpm 11.15.0
```

## Where does the bug occur?

- Local dev (`eve dev`)
- Production (`eve start` / deployed)

harlley:
Closes #1217.

### Description

When a photo is sent to a Telegram-channelled eve agent, `createTelegramFetchFile` resolves the file's media type by prioritizing the HTTP `content-type` header from Telegram's file-download endpoint over the media type the channel already knows from the inbound message. Telegram's file endpoint frequently responds with `application/octet-stream` (or no content-type) for photos, so vision models receive `mediaType: "application/octet-stream"` and cannot recognize the image. With an `uploadPolicy: { allowedMediaTypes: ["image/*"] }`, the photo is rejected before it ever reaches the model:

```
Telegram file rejected — photo.jpg has media type "application/octet-stream"
which is not allowed by this route. Allowed: image/*.
```

The fix inverts the priority so `ref.mediaType` (the known attachment type, e.g. `image/jpeg` for photos) wins, falling back to the HTTP `content-type` and then `application/octet-stream`.

Ran the Telegram attachments unit tests against the `eve` package:

```bash
cd packages/eve
pnpm run build:compiled
pnpm exec vitest run --config vitest.unit.config.ts \
  src/public/channels/telegram/attachments.test.ts
```

- **With the fix:** all 4 tests pass.
- **Without the fix (reverted to `main`):** the new regression test fails with the exact production error string:

  ```
  FAIL  ... > prefers the known attachment media type over the HTTP content-type header
  Error: Telegram file rejected — photo.jpg has media type "application/octet-stream"
  which is not allowed by this route. Allowed: image/*.
  ```

  This confirms the test guards the regression: it fails on the old behavior and passes on the fixed one.

The regression test mocks Telegram's `getFile` (returning a `file_path`) and the file download (returning a body with `Content-Type: application/octet-stream`), then asserts the resolved `mediaType` is `image/jpeg` — the type the channel already knew from the inbound photo.

### PR Checklist

- [ ] I linked an issue with prior discussion confirming this change is wanted. #1217 is linked, but maintainer confirmation is still pending.
- [x] I ran the relevant checks from `CONTRIBUTING.md`. Formatting, lint, typecheck, invariants, the focused test, and integration tests pass. The full unit suite has two unrelated macOS telemetry path failures in `preference.test.ts`.
- [x] I added tests — regression test in `attachments.test.ts`
- [x] I added a changeset — `.changeset/telegram-attachment-media-type.md` (`eve`: patch)
- [x] DCO sign-off passes for every commit (`git commit --signoff`) — commit is signed and GitHub-verified (SSH signing key)

### DIFF
--- a/.changeset/telegram-attachment-media-type.md
+++ b/.changeset/telegram-attachment-media-type.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Telegram channel: prefer the known attachment media type over the HTTP `content-type` header when fetching files. Telegram's file endpoint frequently returns `application/octet-stream` (or no content-type) for photos, which broke image recognition in vision models that key off the declared media type.
--- a/packages/eve/src/public/channels/telegram/attachments.test.ts
+++ b/packages/eve/src/public/channels/telegram/attachments.test.ts
@@ -110,4 +110,39 @@ describe("createTelegramFetchFile", () => {
     await expect(result).rejects.not.toThrow("PRIVATE_FILE_ID");
     await expect(result).rejects.not.toThrow("123456:PRIVATE");
   });
+
+  it("prefers the known attachment media type over the HTTP content-type header", async () => {
+    // Telegram's file endpoint frequently returns `application/octet-stream`
+    // (or no content-type) for photos. The channel already knows the photo is
+    // `image/jpeg` from the inbound message; that known type must win so vision
+    // models receive a recognized image media type.
+    const fetchMock = vi
+      .fn()
+      .mockResolvedValueOnce(
+        new Response(JSON.stringify({ ok: true, result: { file_path: "photos/file.jpg" } }), {
+          headers: { "content-type": "application/json" },
+        }),
+      )
+      .mockResolvedValueOnce(
+        new Response("JFIF", { headers: { "content-type": "application/octet-stream" } }),
+      );
+
+    const fetchFile = createTelegramFetchFile({
+      api: { apiBaseUrl: "https://telegram.example", fetch: fetchMock },
+      credentials: { botToken: "123456:ABCDEF" },
+      policy: { allowedMediaTypes: ["image/*"], maxBytes: 1024 },
+    });
+
+    const result = await fetchFile(
+      String(
+        createTelegramFileUrl({
+          fileId: "photo-id",
+          filename: "photo.jpg",
+          mediaType: "image/jpeg",
+        }),
+      ),
+    );
+
+    expect(result?.mediaType).toBe("image/jpeg");
+  });
 });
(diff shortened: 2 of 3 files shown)


harlley:
Hi, friendly ping on this PR. I rebased it onto the latest main, resolved the conflicts, and reran the relevant checks. Could a maintainer please confirm the approach and review it when possible? The Vercel preview may also need team authorization. Thanks!

PR #1239, opened 2026-07-27
feat(eve): stamp every stream event with a stable id

AndrewBarba:
Refs #1193.

## Summary

- Add a durable `meta.id` to every session stream event. Each id is an `evt_`-prefixed ULID stamped once before persistence and shared with authored hooks.
- Make `MessageStreamEvent` the canonical public stream type. `HandleMessageStreamEvent` remains as a deprecated alias.
- Deduplicate replayed events in `EveAgentStore` and the dev TUI while continuing to admit legacy id-less events.
- Update the stream protocol to version 20, including public types, extension contracts, documentation, tests, and evals.

## Breaking change

`MessageStreamEvent` now requires both `meta.id` and `meta.at`. Code that constructs unstamped events using `HandleMessageStreamEvent` must add the envelope or use the appropriate event-specific input type.

Events persisted before stream version 20 may have `meta.at` without `meta.id`. eve passes these events through, but they cannot be deduplicated. Retried steps produce new emissions with new ids; replaying an existing event preserves its id.

## Verification

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm guard:invariants`
- `pnpm docs:check`
- Targeted unit coverage for ULIDs, event ids and deduplication, client store/reducer behavior, and TUI replay handling

## PR Checklist


packages/eve/src/client/ndjson.ts
```diff
@@ -73,7 +73,7 @@ export async function* readNdjsonStream(
     // Yield any trailing content without a final newline.
     const trailing = buffer.trim();
     if (trailing.length > 0) {
-      yield JSON.parse(trailing) as HandleMessageStreamEvent;
```
ruiconti, line 76: Should we `MessageStreamEvent.safeParse(JSON.parse(trailing))`? 

Risk is that an old event is assumed to have a `meta.id` when it doesn't

ruiconti (approved):

PR #1241, opened 2026-07-27
fix(eve): prevent SSRF in web_fetch

ctgowrie:
### Description

Prevent the built-in `web_fetch` tool from accepting HTTP URLs or reaching localhost, private, link-local, metadata, and reserved network targets.

Each HTTPS request uses built-in `fetch` with an Undici dispatcher whose socket lookup resolves the hostname, rejects the host if any address is unsafe, and passes those exact validated results to the connection. This closes DNS-rebinding bypasses without a separate resolution-and-pinning phase. Redirects are not followed automatically: the tool returns the resolved HTTPS target in its error so the agent can call `web_fetch` again, and every call receives the same validation. Response-size enforcement happens while streaming instead of after the body is fully buffered.

This is a focused follow-up to #662, limited to `web_fetch`.

- `oxfmt --check` and `oxlint` on changed files
- `pnpm guard:invariants`
- `pnpm --filter eve test:unit` — 5,450 passed, 1 skipped
- `pnpm --filter eve build`
- Node 24 packaged-output smoke tests for dispatcher DNS validation and agent-mediated redirects
- Added an `agent-tools` E2E eval that fetches `https://example.com/` and asserts the structured tool result and stable `Example Domain` marker

### PR Checklist

- [ ] I linked an issue with prior discussion confirming this change is wanted (security fix; focused follow-up to #662)


packages/eve/package.json
```diff
@@ -299,7 +299,8 @@
     "test:vercel": "vitest run --config vitest.vercel.config.ts"
   },
   "dependencies": {
-    "nitro": "3.0.260610-beta"
+    "nitro": "3.0.260610-beta",
+    "undici": "7.28.0"
```
AndrewBarba, line 303: Should we use latest v8?

ctgowrie: Node24 doesn't support the v8 agent apis yet

AndrewBarba (approved):

PR #1244, opened 2026-07-27
feat(eve): persist zero-config local traces

chadhietala:
## Summary

- install an internal process-wide instrumentation runtime for `eve dev` when authored instrumentation is absent
- emit agent, AI model/tool, and nested user spans with explicit agent-first parentage
- preserve session/turn trace context through Workflow serialization, cancellation, continuation, retry, and worker replacement
- persist agent-owned spans as immutable OTLP/JSON segments under `.eve/traces/v1`
- vendor `@vercel/otel` and the OTLP transformer without adding runtime dependencies

## Architecture

The harness consumes one generic `{ hooks, runInContext, forceFlush }` runtime. Local development is its first installer, but the boundary is intended for later Vercel production and authored instrumentation installers.

`AgentTraceSpanProcessor` owns process-local classification of agent trace IDs and forwards matching spans to replaceable child processors. `LocalTraceSpanProcessor` is a stateless persistence sink. Future query, Vercel, or custom sinks can attach without reimplementing eve lifecycle semantics.

## Persistence

Immutable OTLP/JSON segments are the canonical local spool. They are safe for overlapping HMR workers and remain available after `eve dev` exits. A future query index is derived and rebuildable rather than a second source of trace identity.

## Scope

This draft does not change existing production or authored `instrumentation.ts` tracing. The local installer is omitted when authored instrumentation exists. Production/custom cutover and the inspection UI are later work built on the same runtime boundary.

## Testing

- `pnpm fmt`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`
- targeted unit tests: 176 passed
- targeted scenario tests: 20 passed


packages/eve/src/harness/local-trace-span-processor.ts
```diff
@@ -0,0 +1,71 @@
...
+    const { spanId, traceId } = span.spanContext();
+    if (!isHexId(traceId, 32) || !isHexId(spanId, 16)) return;
+    const payload = JsonTraceSerializer.serializeRequest([span]);
+    if (payload === undefined) return;
+
+    this.#queue = this.#queue
+      .then(async () => {
+        const directory = join(this.#appRoot, ".eve", "traces", "v1", traceId, "segments");
+        await mkdir(directory, { mode: 0o700, recursive: true });
+        await chmod(directory, 0o700);
```
ctgowrie: Why 700?

ctgowrie (commented):
We should probably add some pruning of local trace files. Perhaps we can follow same practice as revisions

packages/eve/src/harness/local-trace-span-processor.ts
chadhietala: Changed to 777. Was slop that I didn't catch. Probably confused about the immutability constraint.

ctgowrie (approved):

PR #1246, opened 2026-07-27
fix(eve): keep a single tool_result per tool-use id in persisted history

ruiconti:
Opus 5 turns fail intermittently with:

```
messages.2.content.2: each tool_use must have a single result. Found multiple `tool_result` blocks with id: toolu_...
```

A provider-executed tool call (gateway `web_search`) whose arguments fail to parse comes back with **two** results for one id:

- the provider's own outcome, inline in the assistant message: `{"type":"error-json","value":{"error":"invalid_input","message":"JSON parse error..."}}`
- a separately synthesized `AI_InvalidToolInputError` tool message from the AI SDK, in `result.response.messages`

eve persisted both, and the next model call replayed two `tool_result` blocks for one `tool_use` — which Anthropic rejects with `statusCode 400, isRetryable false`, killing the turn.

## Scope

This is independent of the raw-JSON-string tool-argument handling in #1240. The tool-call input in the failing case is `{}` (a valid object), so no coercion path runs and no eve-synthesized error is involved — the AI SDK emits the duplicate itself. The bug is structural: eve assumed one result per tool-use id but never enforced it. Fixing it off `main` so it lands independently.

## Fix

`dedupeToolResultsByCallId` runs after `normalizeProviderToolHistory` and keeps the first `tool_result`/`tool_error` per tool-use id — the provider's authoritative inline outcome, which precedes any synthesized follow-up — dropping later duplicates and pruning any message left empty. `tool-loop` logs the dropped ids rather than truncating silently. Untouched messages keep their identity (no needless copies).

## Verification

- Captured the exact duplicate history from a live Opus 5 `web_search` eval; the unit test encodes that structure.
- Reproduced **4/20** on the live eval before the fix, **0/30** after.
- Full unit tier green.

### DIFF
--- a/.changeset/dedupe-provider-tool-results.md
+++ b/.changeset/dedupe-provider-tool-results.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fixed a turn failure when a provider-executed tool (e.g. gateway `web_search` on Opus 5) reported its result both inline in the assistant message and as a separately synthesized tool-error for the same call. The harness now keeps a single `tool_result` per tool-use id, so the next model call no longer fails with "each tool_use must have a single result".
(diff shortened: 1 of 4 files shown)


ruiconti:
Fix is upstream https://github.com/vercel/ai/pull/17400.

PR #1252, opened 2026-07-27
feat(eve): add stateful Slack `threadReplies` hook

benpankow:
## Summary

This exploratory PR adds a durable, in-workflow policy for deciding whether eve should respond to a reply in an active Slack thread.

### 1. Respond only to explicit mentions

This remains the default:

```ts title="agent/channels/slack.ts"
export default slackChannel({
});
```

### 2. Respond to every human reply in active eve threads

```ts title="agent/channels/slack.ts"
export default slackChannel({
  threadReplies: true,
});
```

Messages outside threads with an active eve session remain ignored.

### 3. Track custom thread-reply state

`threadReplies` can hold durable state and define an `onReply` policy that runs inside the hydrated session workflow. For example, this policy stops responding after someone else is mentioned, then resumes when eve is mentioned directly:

```ts title="agent/channels/slack.ts"
type ThreadReplyState = "followingThread" | "ignoringThread";

export default slackChannel({
  threadReplies: {
    initialState: "followingThread" as ThreadReplyState,
    onReply(message, ctx) {
      if (ctx.isBotMentioned) {
        return {
          respond: true,
          state: "followingThread",
        };
      }

      if (ctx.state === "ignoringThread") {
        return {
          respond: false,
          state: ctx.state,
        };
      }

      const mentionsSomeone = /<@[A-Z0-9]+(?:\|[^>]+)?>/.test(message.text);

      return mentionsSomeone
        ? {
            respond: false,
            state: "ignoringThread",
          }
        : {
            respond: true,
            state: ctx.state,
          };
    },
  },
});
```

Returning `respond: false` skips only that reply. The eve session remains active and `onReply` runs again for a later reply.

### 4. Classify ambiguous replies

`onReply` can also make a small structured AI SDK call to decide whether a thread reply needs a response:

```ts title="agent/channels/slack.ts"
import { generateText, Output } from "ai";
import { z } from "zod";

export default slackChannel({
  threadReplies: {
    async onReply(message, ctx) {
      if (ctx.isBotMentioned) {
        return { respond: true };
      }

      await ctx.thread.refresh();
      const conversation = ctx.thread.recentMessages
        .slice(-12)
        .map((entry) =>
          `${entry.isMe ? "agent" : (entry.user ?? "unknown")}: ${entry.text}`,
        )
        .join("\n");

      const { output } = await generateText({
        model: "google/gemini-3.1-flash-lite",
        system:
          "Decide whether the final Slack message expects the agent to answer.",
        prompt: conversation,
        output: Output.object({
          schema: z.object({ respond: z.boolean() }),
        }),
        temperature: 0,
        maxOutputTokens: 100,
        maxRetries: 0,
        timeout: 1_500,
      });

      return { respond: output.respond };
    },
  },
});
```

## Test Plan

- Added unit tests for default mention-only and `threadReplies: true` behavior.
- Added unit tests for arbitrary durable thread-reply state and explicit-mention transitions.
- Added workflow coverage for running `onReply` policies inside hydrated eve context.
- Manually verified the handoff behavior with a deployed Slack agent.


benpankow:
Closing this early ambient Slack reply exploration because the channel API and product direction have moved on. Current work such as #3656 is validating ambient replies through the existing `onMessage` path; any supported stateful participation API should be designed from that current behavior rather than this branch.

PR #1253, opened 2026-07-27
docs(eve): show AI-classified Slack thread participation

benpankow:
## Summary

Builds on stateful Slack thread participation with a structured AI SDK classifier example. It refreshes bounded Slack thread history and uses an AI Gateway model with the same ambient credentials as the main eve agent.

Also runs durable channel delivery policies inside the hydrated eve context so they can use `SessionContext`, `defineState`, and Slack thread handles.

## Test Plan

- Focused Slack, workflow-step, and send unit tests
- `pnpm docs:check`
- `pnpm guard:invariants`

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a></sub>

### DIFF
--- a/.changeset/slack-thread-unsubscribe.md
+++ b/.changeset/slack-thread-unsubscribe.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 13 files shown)


benpankow:
Folded into #1252 so the thread participation API, hydrated workflow execution, and classifier example can be reviewed together.

PR #1268, opened 2026-07-28
feat(docs): add Sanity, Typefully, and GitHub agent templates

bensabic:
Adds three vercel-labs templates to the gallery at `/templates`.

| Template | Slug | Integrations | Source |
| --- | --- | --- | --- |
| Sanity | `sanity-copilot` | Slack, Sanity, Notion | [sanity-copilot-eve-template](https://github.com/vercel-labs/sanity-copilot-eve-template) @ `f076b80` |
| Typefully | `typefully` | Slack, Typefully, Notion | [typefully-eve-template](https://github.com/vercel-labs/typefully-eve-template) @ `3627b22` |
| GitHub | `kody` | GitHub, Linear, Resend | [kody-eve-template](https://github.com/vercel-labs/kody-eve-template) @ `8427e3c` |

Each entry mirrors the template's core authored surface — `agent.ts`, instructions, sandbox, channels, connections, tools, `SKILL.md` files, subagents, and (where present) schedules. The deep `skills/*/references/*.md` and `rules/*.md` docs are excluded: they run to 355 KB on the Sanity template alone, and leaving them out also avoids needing a `json` variant on `TemplateFile["language"]` for Typefully's `banned-words.json` files. Every embedded file was verified byte-for-byte against its upstream clone.

Setup prompts are each template's own "quick start with an AI coding agent" prompt from its README, which points at the hosted setup doc. That differs from the three existing entries, whose prompts describe the copy-and-configure steps inline.

### File browser ordering

`categoryOrder` in `file-viewer.tsx` renders any unlisted name *after* everything else, so the new templates surfaced `instructions.ts`, `sandbox.ts`, and `schedules` stranded below `lib` — and Kody's order differed from the other two because only `instructions.md` was listed. Standalone files now lead, folders follow, and all three render the same sequence:

```
agent.ts → instructions.* → sandbox.ts → channels → connections → schedules → skills → tools → subagents → lib
```

This also moves the Design template's `sandbox/` folder from last position into the same slot the other templates' `sandbox.ts` occupies — the one behavior change to an existing template.

### Integrations

Adds `Sanity`, `Typefully`, `GitHub`, and `Resend` to `TemplateIntegration`. `githubLogo` and `resendLogo` already existed. `sanityLogo` wraps `SiSanity`, and `typefullyLogo` is an inline mark (Typefully isn't in simple-icons). Both inherit `currentColor` rather than a brand hex, so they match the neighbouring icons and work in dark mode — simple-icons registers Sanity as `#0D0E12`, which rendered near-black and ignored the `grayscale` class.

### Notes

- The GitHub template runs on `anthropic/claude-fable-5`; the other six are `claude-sonnet-5`.
- It's also the first entry with no chat-surface tag — its surfaces are GitHub, Linear, and email via Resend.
- Docs-only, so no changeset.


cmpadden (commented):
Happy to land this for now - thanks!

@bensabic something I'd like to discuss is how templates and registry items can coexist. Something I could imagine is being able to "build your own" template composing extensions / connections from the registry, then we could scaffold those components into a project for the user.

This would aid in template maintenance, but would add increased complexity.

PR #1284, opened 2026-07-28
fix(skills): resolve skill file reads home-root first as documented

### LINKED ISSUE #1234: Skills file reading fallback is the opposite as described in docs

later

### Steps to reproduce

1. In a nextjs withEve() environment create a skill with defineSkill()
2. add files

https://eve.dev/docs/skills#how-loading-works this docs say that it will look for files in `$HOME/.agents/skills/<skill>/` and then fallback to `/workspace/skills/<skill>/`

But as you can see here in practice is the opposite, it first look for the files in `/workspace/skills/<skill>/` and then fallback to `.agents/skills/<skill>/`
Image

This makes execution 2 seconds slower

### eve version

eve@0.27.6

### Environment

```shell
vercel sandboxes
```

### Where does the bug occur?

### Deployment

D8mmHHGMevF6Piw1DPbtycuaHmcV

### Build and runtime logs

```shell
Error

File not found: /workspace/skills/amable-conti-internals/references/creativa-ventas.md. Verify the path exists and is accessible in the sandbox.

ACTION_RESULT_FAILED

Input

{
filePath: "/workspace/skills/amable-conti-internals/references/creativa-ventas.md",
limit: 300,
offset: 1
}
Output

File not found: /workspace/skills/amable-conti-internals/references/creativa-ventas.md. Verify the path exists and is accessible in the sandbox.
```

### Additional context

_No response_

Mohith26:
Fixes #1234

## What

The docs and system prompt say skill files resolve from `$HOME/.agents/skills/<skill>/` first with `/workspace/skills/<skill>/` as fallback, and skill packages are seeded into the home root. But the model-facing read_file tool rejected the advertised `$HOME/...` form outright (absolute-path validation) and did no cross-root fallback, so the only literal path that ever worked was the workspace form, and reads of seeded skills hard-failed with "File not found", costing a slow model-driven retry loop (~2s per miss).

This aligns the code with the documented order: skill-path reads resolve home-root first, then fall back to the workspace root, via a shared `skill-paths` helper used by the read tool.

## Tests

35 unit tests cover the resolution order, fallback on miss, and the previously rejected home-root form. 10 fail on main, all pass with the fix. Full unit tier: 5485 passed, 0 failures.

Credit to @pedro757 for the report in #1234.

### DIFF
--- a/.changeset/skill-file-resolution-order.md
+++ b/.changeset/skill-file-resolution-order.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Resolve skill file reads in the documented order: `$HOME/.agents/skills/<skill>/` first, with `/workspace/skills/<skill>/` as the fallback. `read_file` now accepts the symbolic `$HOME/.agents/skills/...` form and falls back across skill roots instead of failing when the model names the workspace path.
--- a/packages/eve/src/execution/sandbox/read-file-tool.ts
+++ b/packages/eve/src/execution/sandbox/read-file-tool.ts
@@ -6,6 +6,7 @@ import {
   setReadFileStamp,
 } from "#runtime/framework-tools/file-state.js";
 import { validateAbsoluteFilePath } from "#execution/sandbox/require-sandbox.js";
+import { resolveSkillFilePathCandidates } from "#shared/skill-paths.js";
 import type { SandboxSession } from "#shared/sandbox-session.js";
 import { capLineLength, MAX_OUTPUT_BYTES } from "#execution/sandbox/truncate-output.js";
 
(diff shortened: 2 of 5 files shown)


cmpadden:
Thanks for investigating this and adding focused regression coverage. The underlying issue was that model-facing file tools did not expand the advertised `$HOME/...` path. That has since been fixed in #1459 across `read_file`, `write_file`, `glob`, and `grep`, with the path resolution consolidated further in #1466.

This implementation also makes `read_file` resolve `/workspace/skills/...` differently from `write_file`, which could cause a read and subsequent write to target different files. Since the original issue is resolved and the current implementation preserves symmetric file-tool behavior, I’m closing this as superseded by #1459.

PR #1286, opened 2026-07-28
feat: guide agents to use registry commands in generated `AGENTS.md`

cmpadden:
## Summary

- direct coding agents to discover integrations with `eve registry search` and `eve registry list`
- direct coding agents to inspect and install integrations with `eve registry view` and `eve add`
- cover the generated guidance in the scaffold integration test

## Testing

- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/setup/scaffold/index.integration.test.ts`
- `./node_modules/.bin/oxfmt --check packages/eve/src/setup/scaffold/create/project.ts packages/eve/src/setup/scaffold/index.integration.test.ts`

### DIFF
--- a/packages/eve/src/setup/scaffold/create/project.ts
+++ b/packages/eve/src/setup/scaffold/create/project.ts
@@ -270,9 +270,9 @@ installed \`eve\` package location first and read its \`docs/\` directory. If
 package docs are unavailable, use https://eve.dev/docs as a fallback.
 
 Before implementing an integration yourself, use
-\`eve registry search <query> --json\` or \`eve registry list --json\` to discover
-available integrations. Inspect one with \`eve registry view <item>\`, then
-install it with \`eve add <item>\`.
+\`eve registry search <query>\` or \`eve registry list\` to discover available
+integrations. Inspect one with \`eve registry view <item>\`, then install it with
+\`eve add <item>\`.
 `,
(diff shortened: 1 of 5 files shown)


packages/eve/src/setup/scaffold/create/project.ts
```diff
@@ -269,6 +269,11 @@ from the installed eve package docs. In most installs, those docs are at
 \`node_modules/eve/docs/\`. In workspaces or local package installs, resolve the
 installed \`eve\` package location first and read its \`docs/\` directory. If
 package docs are unavailable, use https://eve.dev/docs as a fallback.
+
+Before implementing an integration yourself, use
```
ruiconti, line 272: There's also a hand-off prompt on `eve init` for other boot-strapping via harnesses that should be updated too

cmpadden: Added the same registry discovery, inspection, and installation guidance to the shared init handoff prompt. This covers both the post-scaffold coding-agent handoff and the seeded REPL prompt. Tested in `agent-instructions.test.ts`.

cmpadden: Good catch, thanks @ruiconti. Updated the PR.

ruiconti (approved):
👍🏻

PR #1310, opened 2026-07-29
feat(eve): add declarative framework tool allowlists

### LINKED ISSUE #1309: Feature: declarative allowlist for framework built-in tools
## Problem

Eve has a filesystem-first override model: a production agent removes an unwanted built-in by adding a same-slug module that exports disableTool().

That is default-allow. An Eve dependency upgrade can expand a production agent capability surface when the framework adds a built-in, unless every consuming application already knows to add the matching tombstone module. This is risky for filesystem, shell, network, delegation, and skill-loading tools.

Filesystem discovery is a good implementation and convention layer, but it should not be the sole production authorization boundary.

## Proposal

Add a declarative built-in tool policy to defineAgent with explicit allowlist mode.

Example API:

    export default defineAgent({
      model: "openai/gpt-5.6",
      builtInTools: {
        mode: "allowlist",
        allow: ["ask_question", "todo", "web_fetch", "load_skill"],
      },
    });

Desired behavior:

- In allowlist mode, a framework built-in is unavailable unless named in allow.
- Authored tools and explicit same-slug wrappers or overrides continue to work normally.
- The compiled manifest and Eve info endpoint clearly show effective built-in policy and final tool surface.
- A framework upgrade that introduces a new built-in cannot silently grant it to an allowlisted production agent.
- Current behavior remains the backward-compatible default with a clear migration path.

A denylist API could also work, but an allowlist is the safer production primitive.

## Current workaround

We snapshot-test the public eve/tools/defaults export surface in CI and review every change. That helps, but it is incomplete as a framework guarantee: it depends on consumers maintaining a separate inventory, and not every conditional framework built-in is necessarily represented by that module.

Happy to help test or contribute an implementation.

bob-obringer:
### Description

Closes #1309.

eve currently enables every framework-provided tool unless the author creates filesystem disable sentinels one by one. That is fragile for capability-sensitive agents because a newly added built-in becomes available after an upgrade without an explicit application decision.

This adds an optional `builtInTools` allowlist to `defineAgent()`. Omitting it preserves the existing behavior. Enabling it denies unlisted framework tools while leaving authored tools and same-name authored overrides alone. The policy survives compilation, is enforced for static tools, `connection_search`, and queued recursive `agent` actions, and is reported by `/eve/v1/info`. Unknown names fail with an authoring error instead of silently weakening the policy.

- `pnpm fmt`
- `pnpm lint`
- `pnpm check:deps`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm typecheck`
- `pnpm test:unit` — 5,563 passed, 1 skipped
- `pnpm test:integration` — 566 passed, 1 skipped
- `pnpm build`
- `pnpm test:scenario` — 370 passed, 15 skipped; two unrelated `framework-next-build` cases stop during the fixture's `pnpm install` because local pnpm rejects ignored `esbuild` and `sharp` build scripts before exercising this change

### PR Checklist

### DIFF
--- a/.changeset/six-insects-smile.md
+++ b/.changeset/six-insects-smile.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add an opt-in `defineAgent({ builtInTools })` framework-tool allowlist. Allowlisted agents deny new framework tools by default while preserving authored tools and overrides.
--- a/docs/agent-config.md
+++ b/docs/agent-config.md
@@ -167,6 +167,35 @@ parent's quota, so a delegation tree can never outspend the budget configured
 at its root. An authored child limit applies only when it is tighter than the
 parent's grant; an uncapped parent delegates uncapped children.
 
+## Built-in tool allowlist
+
+By default, eve enables its framework-provided tools. To make production
+capabilities deny-by-default, opt into `builtInTools.mode: "allowlist"` and
+name the framework tools this agent needs:
+
+```ts title="agent/agent.ts"
+import { defineAgent } from "eve";
+
+export default defineAgent({
+  model: "anthropic/claude-opus-4.8",
+  builtInTools: {
+    mode: "allowlist",
+    allow: ["ask_question", "todo", "web_fetch", "load_skill"],
+  },
+});
+```
+
+The policy applies only to eve-built tools. Your authored tools, dynamic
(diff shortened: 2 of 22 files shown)


bob-obringer:
Closing this draft so Bob can review the implementation locally before any upstream publication. No merge requested.

PR #1329, opened 2026-07-29
feat(eve): own delegated children through the agent-handle lifecycle

ruiconti:
Part 2 of 4 in the agent-messaging stack. Base: `main`, after [#1328](https://github.com/vercel/eve/pull/1328). [#1330](https://github.com/vercel/eve/pull/1330) activates resumable child turns; [#1331](https://github.com/vercel/eve/pull/1331) exposes them to the model.

This replaces the fragmented child IDs on `pendingActionBatch` with one handle lifecycle used by dispatch, callback binding, settlement, cancellation, and proxy-input cleanup.

Before this change, child session IDs and local continuation tokens were recorded only after a child started. That was too late to represent a start in progress, omitted the remote continuation token, and forced every consumer to reconstruct ownership from different fields. A callback also proved which tool call it answered, but not which child session produced it.

## Data flow

```text
dispatch step
  callId
    -> starting handle { operation, target }
    -> start child
         -> failure: delete handle + return dispatch error
         -> success: running handle { operation, address }
                              |
callback { callId, sessionId? } |
    -> match running handle ----+   (sessionId absent: match by callId,
    -> clear proxy input            as older deployments send it)
       via stored continuationToken, when one exists
    -> settle terminal result
    -> delete handle
```

The operation ID derives from the parent session, parent turn, and tool-call ID, so ownership exists before the child session does. Local, self, and remote starts add a `starting` handle to the step's working state before the start side effect. Success confirms the handle with the child's address; failure rejects it. Every normal dispatch outcome returns that updated store.

This does not make child creation atomic with the dispatch-step commit. If the child accepts a start and the step crashes before its result commits, the durable step can replay the whole batch. The change gives that window one explicit owner and state machine; it does not remove the window.

## Identity at the callback boundary

Every child in a batch shares the parent's callback route, and possession of that callback token is the authorization to settle: results bind to their pending call by `callId` alone, exactly main's semantics. The `sessionId` on callback payloads is informational (tracing and diagnostics), never verified. This deliberately accepts the at-least-once dispatch window: a replay-orphaned duplicate child holds the same token and callId and may settle the call in place of the owned child — its output is computed from the same input. Inbox results matching no running handle (late duplicates, calls whose dispatch already failed) are dropped rather than allowed to overwrite a settled result.

Parent-generated dispatch failures remain separate, discriminated as `origin: "dispatch"`: they enter through the trusted step-result path, never the shared inbox.

Remote creation must return a `sessionId` (recorded on the handle for cancellation and continuation addressing); `continuationToken` is optional. An older deployment that returns only a `sessionId` dispatches normally — its handle records no token, the child completes as a task-mode one-shot, and #1330's delivery leg fails any continuation against it permanently. The continue route is resume-only: a dead token returns `404 SESSION_NOT_RESUMABLE` rather than silently creating a replacement session.

Cancellation reads the same handle store. Local and self children cancel by stored session ID. Remote children use the URL captured at dispatch, because registry resolution may now point at a different deployment.

## Stack boundary

All children still run in task mode here. The reachable lifecycle is `starting -> running -> deleted`.

The `parked` phase, continuation route, turn callbacks, caller field, and usage delta are present because they belong to the shared handle and wire contracts, but their first producers and consumers land in #1330. The model cannot supply `agentId` until #1331.

The changeset therefore describes the complete stack, not only behavior reachable in this slice.

## Review focus

- Accept the remaining start-side-effect/step-commit replay window; no transition here is atomic across that boundary, and a replay-orphaned duplicate child may settle its call (token + callId is the full authorization — see "Identity at the callback boundary").
- Accept the compatibility posture: no coordinated-version requirement — older peers degrade to task-mode one-shots — while resume-only continuation stays strict.
- Check the schema-free workflow-driver reader's trust boundary. It keeps zod out of the driver bundle, while write correctness depends on the transition constructors and previously validated state.


De-ASI-INTERFACE (approved):

AndrewBarba:
### [P1] Parent cancellation permanently leaks `running` handles

`cancelDescendantTurnsStep` requests cancellation for each running child but never transitions or removes the corresponding handle. The cancellation epilogue then clears the pending runtime-action batch and persists those handles unchanged. Because the old turn inbox is disposed, a later child settlement cannot repair the store.

This leaves each cancelled child permanently `running`. Subsequent parent cancellations retry the stale child up to eight times, and the handle can never enter the resumable `parked` lifecycle. The existing cancellation integration test verifies that the child aborts and that a follow-up turn runs, but it does not assert the persisted handle state.

Please add an explicit handle transition when the parent abandons a cancelled batch, plus coverage for the post-cancellation handle store.

ruiconti:
@AndrewBarba resolved in https://github.com/vercel/eve/pull/1330#issuecomment-5169828380.

AndrewBarba (approved):

PR #1330, opened 2026-07-29
feat(eve): conversation-mode children with explicit turn outcomes

ruiconti:
Part 3 of 4 in the agent-messaging stack, based on [#1329](https://github.com/vercel/eve/pull/1329). This is the engine slice: it turns the handle contracts below it into resumable child conversations. [#1331](https://github.com/vercel/eve/pull/1331) adds the model-facing schema, instructions, docs, and evals.

The base can own a child, but every child still runs as a one-shot task and every result is terminal. No handle can park, no continuation can create another child turn, and a persistent child has no per-turn path back to its caller.

## Runtime flow

```mermaid
flowchart LR
  P[Parent call] --> H[handle: running]
  H -->|local deliver or remote continue| C[child conversation turn]
  C -->|outcome + usageDelta| N[notify this turn's caller]
  N --> S[parent settles result]
  S -->|parked| W[handle parked; child waits]
  S -->|terminal| D[delete handle]
  W -->|next agentId| H
```

With `experimental.subagentPersistentSessions`, local, self, and remote children start in conversation mode. Task mode remains the default.

At an idle park, the child reports a required `AgentTurnOutcome`, not just success or failure. `parked` means the answer is complete and the same child can accept another message. `terminal` deletes the handle. A failed turn may still park, so error status no longer decides lifecycle.

Caller identity travels with each delivery. The child notifies that caller, clears it, then waits. Two callers continuing the same child do not collapse into one turn: each delivery gets its own answer path.

## Continuation delivery

An `agentId` continuation resolves through the handle store, moves a parked handle to running, then delivers through the local workflow hook or the stored remote session address.

Delivery is attempted once. A missing local session, remote 404, `SESSION_NOT_RESUMABLE`, a missing registry entry, or a remote handle with no stored `continuationToken` (its deployment predates continuations, so a token can never appear) proves the address is dead, so the handle is deleted. A network or server failure is ambiguous: the child may have accepted the message before the response was lost. The handle returns to parked and the parent receives `AGENT_UNREACHABLE`; eve does not resend and risk duplicate work.

At this layer, unknown IDs still return `AGENT_UNKNOWN`. #1331 changes the model-facing policy so an unknown or stale ID starts a fresh child instead.

## Usage and cleanup

Persistent children report per-turn usage deltas. The child stores the session totals already reported, subtracts that marker at settlement, then advances it. If totals are 100 after turn one and 150 after turn two, the parent receives `100 + 50`, not `100 + 150`.

When the parent completes, expires, or reaches the outer crash handler, it best-effort terminates every running or parked local/self child. One failed termination does not stop the rest.

Two cleanup gaps remain explicit. A `starting` handle has no confirmed child session to terminate. Crash cleanup sees only the latest completed snapshot, so a child started by the crashing turn can escape. Remote children also survive parent termination.

## Wire change

`turn.completed` and `turn.failed` callbacks now require the outcome envelope and drop the redundant top-level `usage`. Task-style `session.*` callbacks are converted to explicit terminal outcomes at the receiving boundary — including sessionId-less callbacks from older deployments, which bind by callId (#1329's compatibility posture).

There is no mixed-version shim. The extension contract epochs advance, and fixtures preserve the previous epochs as compatibility snapshots.

## Review focus

- Session creation currently runs before the initial caller is resolved. If creation fails, the delegated caller is not notified; this is the remaining unresolved review thread.
- Ambiguous continuation failures require caller-directed retry because an automatic retry can duplicate accepted work.
- Failed task sessions retain the existing zero-usage policy; only conversation failures settling through the turn path carry a delta.
- The callback route validates the outcome shape but trusts it to agree with the callback kind and top-level output.
- Remote termination and the mid-turn crash orphan window remain outside this slice.


packages/eve/src/execution/agent-handle-dispatch.ts
```diff
@@ -30,3 +75,240 @@ export type DispatchOutcome =
...
+          message: `Agent "${agentId}" is still working on its previous request. Wait for its result before continuing it.`,
+        }),
+        session: input.currentSession,
+      };
+    case "ready":
+      break;
+  }
+
+  const { handle } = prepared;
+  const delivery = await deliverToAgentHandle({
```
ruiconti, line 168: **Replay semantics:** this delivery is not exactly-once.

`prepareAgentContinuation` only updates the snapshot committed when the outer dispatch step returns. If delivery succeeds and that step dies first, replay starts from `parked` and calls this again.

This is the same commit window already accepted for child starts, not a continuation-specific regression. Could we document it here? The [transition contract](https://github.com/vercel/eve/blob/840c4d43c6158f4212c91a0f2368d34d3a925570/packages/eve/src/harness/handles/transitions.ts#L58-L65) currently says `running` must commit before delivery, which does not happen at this boundary.

The contract should either state that delivery may repeat after step replay, or continuations need receiver-side deduplication.

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -128,20 +151,33 @@ export async function workflowEntry(input: WorkflowEntryInput): Promise<Workflow
...
+        serializedContext: input.serializedContext,
+        status: "failed",
+      });
+      await notifyDelegatedParentStep({
+        result: createDelegatedSubagentErrorResult(input.serializedContext, error),
+        serializedContext: input.serializedContext,
+      });
+    } else {
+      await notifyTurnCallerStep({
+        caller: lifecycleState.caller,
```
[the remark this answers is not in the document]
ruiconti: Fixed in c96c5f01. `CrashCleanupState` now records whether `resolveInitialTurnCallerStep` ever ran: a crash before resolution (e.g. `createSessionStep` throwing) re-resolves the caller from the serialized context — which needs nothing from the failed steps — and rejects it terminally instead of parking the parent forever. A cell cleared because its reply already settled still notifies no one, and a failing re-resolution degrades to no notification rather than masking the original error. Covered by two new tests in `workflow-entry.test.ts` (pre-resolution crash; settled-and-cleared crash).

packages/eve/src/execution/agent-handle-dispatch.ts
ruiconti: Addressed in `c230c0a8`. The transition and dispatch docs now state the actual boundary: prepare updates the step working snapshot, durability arrives with the outer step result, and a crash before that commit can redeliver. This is explicitly the same at-least-once window as child starts, not an exactly-once guarantee.

ruiconti:
**Re: [P1] parent cancellation permanently leaks running handles — fixed in 378f7bc3.**

The exact leak sequence:

1. Parent delegates to a child: the handle store records it as `running`, and the parent waits on its turn inbox — the only hook a child settlement can resume.
2. User cancels the parent turn.
3. [`cancelDescendantTurnsStep`](https://github.com/vercel/eve/blob/378f7bc3/packages/eve/src/execution/cancel-descendant-turns-step.ts#L26-L65) tells the child to stop; the child aborts and parks itself. The step never touches the handle store.
4. The cancelled turn's teardown [disposes the turn inbox](https://github.com/vercel/eve/blob/378f7bc3/packages/eve/src/execution/turn-workflow.ts#L230-L238).
5. The epilogue persists the session with the handle still `running`.

After step 5 the record can never self-heal: the child's settlement targets the dead inbox, and [`resumeSettledTurnHook` swallows `HookNotFoundError` and drops it](https://github.com/vercel/eve/blob/378f7bc3/packages/eve/src/execution/delegated-parent-notification.ts#L270-L279). No other transition path reaches that handle.

Why this is worth a P1:

- Every later cancellation of the same session reads the stale handle and retries cancel against a dead child. [`requestCancellationWithRetry`](https://github.com/vercel/eve/blob/378f7bc3/packages/eve/src/execution/cancel-descendant-turns-step.ts#L144-L154) only exits early on `accepted`, so it burns all 8 attempts (~7.75s of backoff) while `finishCancelledTurn` blocks on it — and it compounds per leaked handle.
- The child parked itself and is resumable, but only `parked` handles are model-visible, so the child silently vanishes forever.
- It permanently violates the store's documented lifecycle ("a terminal child has no handle"; `parked` = idle and resumable) in durable state.

**Fix.** Both cancellation paths — turn-owned `finishCancelledTurn` and the driver's between-turns `cancel-turn` arm — converge on `settleCancelledTurnStep` as the final persistence point. The epilogue now runs [`abandonRunningAgentTurns`](https://github.com/vercel/eve/blob/378f7bc3/packages/eve/src/harness/handles/transitions.ts#L219-L250): every `running` handle parks as `"(cancelled)"`, mirroring `settleAgentTurn`'s cancelled outcome. If the child died instead of parking, the next continuation attempt discovers the dead session and `rejectAgentEffect` deletes the handle, so a wrong park self-corrects.

**Coverage.** Unit tests on the transition, plus [`settle-cancelled-turn-step.integration.test.ts`](https://github.com/vercel/eve/blob/378f7bc3/packages/eve/src/execution/settle-cancelled-turn-step.integration.test.ts) asserting the persisted handle store in the step's returned snapshot: running → parked `"(cancelled)"`, existing parked handle untouched. The assertion lives at the step boundary because full-flow tests can't read the persisted store — the snapshot rides inside workflow step results.

AndrewBarba (approved):

PR #1342, opened 2026-07-29
test(eve): make model output image smoke deterministic

cmpadden:
## Summary

Attempt to remove test flaky by not actually invoking image model, and instead confirming appropriate inputs.

- move the exact `toModelOutput` file-projection contract into a deterministic unit test
- keep the hosted image E2E as a smoke test for tool completion, without requiring a hosted model to classify randomized pixels

The previous E2E failed intermittently when the model omitted or confused a stripe color, even though the tool completed and the image content part was delivered. The unit test now verifies that raw tool output is given to the mapper while the model-facing result contains only the projected text and PNG file part.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/tools.test.ts`
- `pnpm --filter eve exec tsc -p tsconfig.json --noEmit`
- `pnpm exec oxlint packages/eve/src/harness/tools.test.ts e2e/fixtures/agent-tools/evals/static-tools/to-model-output-content-parts.eval.ts`

### DIFF
--- a/e2e/fixtures/agent-tools/evals/static-tools/to-model-output-content-parts.eval.ts
+++ b/e2e/fixtures/agent-tools/evals/static-tools/to-model-output-content-parts.eval.ts
@@ -1,40 +1,18 @@
-import type { HandleMessageStreamEvent } from "eve/client";
 import { defineEval } from "eve/evals";
 
 const TOOL_NAME = "render-stripes";
 
-// The stripe colors are randomized per run, so a blind model cannot pass by
-// guessing; the eval stays self-contained by validating the reply against
-// the answer key the tool records on action.result. The pixels reach the
-// model exclusively through `toModelOutput` content parts.
+// The pixels reach the model exclusively through `toModelOutput` content
+// parts. Unit coverage asserts their exact model-facing shape; this hosted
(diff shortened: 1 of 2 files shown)


e2e/fixtures/agent-tools/evals/static-tools/to-model-output-content-parts.eval.ts
```diff
@@ -1,40 +1,18 @@
...
-    await t.send(
-      "Without calling any tool, repeat the stripe colors left to right, comma-separated.",
-    );
-
-    t.succeeded();
-    t.calledTool(TOOL_NAME, { count: 1 });
-    t.eventsSatisfy("the replayed image still answers the follow-up", (events) => {
-      const answer = assistantAnswers(events).at(-1);
-      return answer !== undefined && namesColorsInOrder(events, answer);
-    });
```
ruiconti, line 37: Removing this means that we do not have a stable signal that the model has seen it in context window. For instance, this test also ensures that AI SDK is doing the right thing.

But adding a model call may not be needed - I could see an equally expressible test being done strictly by looking at the context window _before_ calling the model.

PR #1353, opened 2026-07-29
fix(eve): preserve concurrent approval authorization

benpankow:
## Summary

Hardens the chosen multi-responder semantics:

- merges concurrent pending authorization challenges instead of overwriting them
- stores the exact authorization name on each candidate for callback resume
- lazily expires overdue candidates at the next touch boundary
- includes responder principal correlation on candidate lifecycle events
- keeps Slack candidate-to-user presentation mapping durable across other responders clicking
- ensures rejection/failure feedback targets the candidate owner, not merely the latest session caller

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/authorization.test.ts src/public/channels/slack/defaults.test.ts src/execution/subagent-auth-proxy.integration.test.ts src/harness/approval-candidates.test.ts src/harness/input-requests.test.ts`
- `pnpm --filter eve typecheck`

## Stack

Depends on #1352.

PR #1354, opened 2026-07-29
fix(eve): bind approval callbacks to responders

benpankow:
## Summary

Closes the cross-user callback race for concurrent approval candidates:

- restores the persisted candidate responder into the active authorization context before resumed policy/token resolution
- prevents a later Slack click from substituting its principal into an earlier candidate OAuth callback
- retains candidate identity as the canonical responder for audit and settlement
- adds a regression with an original responder and a different latest session caller

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/input-requests.test.ts`

## Stack

Depends on #1353.


benpankow:

PR #1358, opened 2026-07-29
chore(eve): classify authorized approval contracts

benpankow:
## Summary

Classifies the additive extension-facing API changes introduced by the authorized-approval stack:

- bumps tool, dynamicTool, connection, hook, dynamicSkill, and dynamicInstructions capability epochs
- retains every prior epoch
- adds representative retained-contract fixtures
- generates the six new immutable API reports
- fixes the research issue frontmatter invariant

## Validation

- `pnpm update:extension-contracts`
- `pnpm guard:invariants`

## Stack

Depends on #1355.

PR #1360, opened 2026-07-29
fix(eve): target approval candidate feedback

benpankow:
## Summary

Fixes private candidate feedback targeting for both immediate and resumed outcomes:

- resolves responder identity from active or completed candidate audit history
- sends immediate rejection to the current verified responder even before a Slack candidate mapping exists
- continues to use durable candidate-to-user mapping after OAuth or intervening clicks
- adds immediate-rejection and runtime event regressions

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack/defaults.test.ts src/harness/tool-loop.test.ts`

## Stack

Depends on #1358.


benpankow:

PR #1361, opened 2026-07-29
fix(eve): emit approval events after persistence

benpankow:
## Summary

Moves approval lifecycle emission onto durable workflow boundaries:

- candidate creation returns a continuation before policy execution
- `approval.candidate` pending is emitted from persisted candidate state
- terminal settlement is emitted from persisted settlement audit state
- tool execution/batch resolution proceeds only after settlement projection
- emitted markers prevent replayed duplicate events
- updates focused tests for the multi-step transition

This enforces the invariant that channel UI is always a projection of durable state rather than an optimistic in-step notification.

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/input-requests.test.ts src/harness/tool-loop.test.ts src/public/channels/slack/defaults.test.ts`

## Stack

Depends on #1360.

### DIFF
--- a/packages/eve/src/harness/approval-candidates.ts
+++ b/packages/eve/src/harness/approval-candidates.ts
@@ -38,6 +38,7 @@ export interface ApprovalSettlementAuditRecord {
   readonly requestId: string;
   readonly settledAt: number;
   readonly candidateId?: string;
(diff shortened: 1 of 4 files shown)


benpankow:

PR #1362, opened 2026-07-29
fix(eve): emit candidate outcomes after persistence

benpankow:
## Summary

Applies durable-first ordering to rejected, failed, timed-out, and stale candidate outcomes:

- persists terminal candidate audit records before lifecycle emission
- emits pending terminal candidate history on the next workflow step
- marks emitted history to prevent replay duplicates
- keeps shared approval UI as a projection of committed candidate state

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/approval-candidates.test.ts src/harness/input-requests.test.ts src/harness/tool-loop.test.ts src/public/channels/slack/defaults.test.ts`
- `pnpm --filter eve typecheck`

## Stack

Depends on #1361.

PR #1364, opened 2026-07-29
fix(eve): isolate concurrent approval candidates

benpankow:
## Summary

Prevents a new explicit click from being attached to another responder’s already-pending candidate:

- resumes pending candidates only on an internal continuation with no new response
- explicit responses always create/dedupe against the clicking responder
- preserves concurrent independent candidates for the same request
- adds a two-responder regression

## Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/input-requests.test.ts`

## Stack

Depends on #1362.


benpankow:

Issue #883, opened 2026-07-16
Add a native MCP channel for external agent harnesses


AndrewBarba:
## What problem are you trying to solve?

Users want to bring their own agent harness while consuming the capabilities authored in an eve agent. Today there is no native MCP server surface for reading compiled instructions and skills or invoking eligible eve tools. Reimplementing this in a custom channel would duplicate compiler projection, auth, tool execution, sandbox, and deployment behavior.

The first version should stay intentionally constrained: expose static instructions, complete static skill packages, and static executable tools from the root and recursively nested local subagents. Remote subagents must never be exposed.

## Proposed solution

Add an explicit `mcpChannel()` mounted at `/mcp` using stateless Streamable HTTP. Project compiled instructions and skill files as MCP resources and eligible executable tools as MCP tools, with local subagent paths mapped into deterministic namespaces.

Direct tool calls should use normal eve validation, approval, authorization, output projection, and owning-node sandbox/skill behavior without creating a model run or durable eve session. An optional `onRequest` hook should let applications project harness session metadata and effective auth into tool context while keeping internal execution and sandbox identity private and call-scoped. The channel should expose `defaultMcpAuth()` to preserve the admitted route caller, consistent with other eve channels.

The accompanying research plan defines the public API, MCP protocol mapping, execution boundary, sandbox lifecycle, packaging, security constraints, and verification strategy.

## Alternatives considered

- Build a standalone MCP server beside eve. This loses eve compiler/runtime guarantees and duplicates deployment plumbing.
- Route every MCP operation through a durable eve model session. This adds model turns, history, and lifecycle semantics that external harnesses do not need.
- Expose only instruction and skill resources forever. Static reads are a useful first boundary, but direct invocation of eligible authored tools is the important follow-on capability and should fit the initial design.

enricodg:
looking forward! 👀

peter-at-barista:
+1 to this!

PR #1366, opened 2026-07-29
docs(eve): plan MCP agent channel

allenzhou101:
## Summary

- define a v1 MCP channel that exposes the actual eve agent as a durable invocation, rather than publishing compiled tools, skills, and instructions
- keep the authoring API minimal with existing `AuthFn` strategies and an optional generic `oauthResource()` decorator that mounts Protected Resource Metadata
- map MCP compatibility tools and the Tasks extension onto one protocol-neutral durable invocation service
- align the lifecycle with `eve invoke`, bind invocation access to the initiating principal, and document Deployment Protection constraints
- preserve PR #884's direct capability-export design as a future, separately reviewed surface

## Scope

This is a documentation-only design PR. It contains no runtime or public API implementation.

Related: #883, #884, #1203

## Validation

- `pnpm exec oxfmt --check planning/mcp-channel.md` against the local source document
- `git diff --check origin/main 7b4e0a87162cd24a7709b8e6f6dbd5fe5063a806`
- verified the PR commit contains only `research/mcp-agent-channel.md`

### DIFF
--- a/research/mcp-agent-channel.md
+++ b/research/mcp-agent-channel.md
@@ -1,3 +1,9 @@
+---
+issue: TBD
+status: proposed
+last_updated: "2026-07-29"
+---
+
 # MCP channel plan
 
 ## Summary
--- a/research/mcp-agent-channel.md
+++ b/research/mcp-agent-channel.md
@@ -1,0 +1,924 @@
+# MCP channel plan
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


allenzhou101:
Superseded by the hardened implementation in #1377, which is now based directly on main.

PR #1367, opened 2026-07-29
chore(e2e): coverage - split model and workflow world suites

AndrewBarba:
## Summary

- split e2e discovery into a Local live-model suite and deterministic Local, Vercel, and Postgres world suites
- run 8 meaningful live-model legs and 15 fixtures in each world, reducing focused coverage from 120 to 53 jobs while retaining every fixture
- extend the internal authored-model adapter for deterministic subagents, repeated tools, Workflow fan-out, silent completion, and schema-derived inputs
- preserve Vercel redeploy and Postgres persistence coverage, with stable aggregate checks for all three worlds
- document suite ownership and add a patch changeset for the internal eve adapter update

## Validation

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm check:deps`
- `pnpm guard:invariants`
- `pnpm guard:fixtures`
- `pnpm docs:check`
- `pnpm exec oxfmt --check .`
- `pnpm exec oxlint .`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/runtime/agent/mock-model-adapter.test.ts`
- discovery validation: 8 model legs, 15 world fixtures, 53 focused jobs
- `git diff --check`

CI is the official e2e validation.

### DIFF
--- a/.changeset/focused-e2e-world-models.md
+++ b/.changeset/focused-e2e-world-models.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Expand the internal deterministic model adapter for world-focused e2e coverage, including subagent, repeated-tool, fan-out, silent-completion, and schema-derived tool-input programs. Authored `eve-mock/*` models remain intact when automatic model replacement is enabled.
--- a/.changeset/focused-e2e-world-models.md
+++ b/.changeset/focused-e2e-world-models.md
@@ -2,4 +2,4 @@
 "eve": patch
 ---
 
-Expand the internal deterministic model adapter for world-focused e2e coverage, including subagent, repeated-tool, fan-out, silent-completion, and schema-derived tool-input programs. Authored `eve-mock/*` models remain intact when automatic model replacement is enabled.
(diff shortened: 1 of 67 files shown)


AndrewBarba:
Closing in favor of #1387, which lands the suite split with a different architecture:

- Every fixture stays in **both** the model suite and every world suite; membership is expressed with eval-level `real-model` tags (`eve eval --exclude-tag`) rather than an exclusive fixture partition, so no fixture loses live-model or world coverage.
- World suites use fixture-authored `mockModel()` responders (via the shared `@eve-e2e/config` package this PR introduced — thanks, that design carried over directly) instead of expanding the internal runtime adapter, keeping eval-prompt-specific logic out of `packages/eve`.
- Aggregate check names stay `e2e-local` / `e2e-vercel`, plus the new `e2e-postgres` (this PR's workflow, adapted to the mock convention, including the Postgres run-count durability assertion).

Salvaged from here:
- The turn-cancellation race fix + test → extracted to #1389 so the runtime fix gets independent review.
- The image-attachment, cancel-subagent, and agent-subagents eval determinism tweaks → cherry-picked into #1387 (image-attachment now runs mocked in all world suites).
- The mock directive programs (subagent delegation, Workflow fan-out, repeated tools, schema-derived inputs) remain the reference recipes for untagging `agent-subagents` / `agent-schedules` / `agent-skills` as fixture-side responders in follow-ups.

Issue #1021, opened 2026-07-21
Authorize who may respond to HITL tool approvals


AndrewBarba:
### What problem are you trying to solve?

eve can require human approval before a tool executes, but it cannot authorize **which human** may answer the approval. Once a channel renders an approval request, any user who can submit a matching response can approve the tool call.

For example, an agent that creates GitHub pull requests may be used from a shared Slack channel. The tool should execute only when the person approving it is a CODEOWNER for the target repository. The existing `approval` policy runs before the durable pause, so it can inspect the tool input and the caller who requested the action, but it cannot validate the different person who may respond seconds or days later.

Slack already authenticates the interaction actor and resumes the session with that actor as `session.auth.current`. However, the runtime treats any matching `inputResponse` as a terminal answer before authored policy can accept or reject that responder. Implementing this today requires replacing eve's approval flow with application-owned state, Slack callbacks, identity mapping, and manual continuation, which gives up the framework's built-in durable pause-and-resume behavior.

Source feedback: https://github.com/vercel/feedback/issues/152463

### Proposed solution

Add a first-class, async responder-authorization policy for tool approvals. It should run when an approval response is delivered, before the pending request is resolved or the tool executes.

The policy needs access to:

- the approval request (`toolName`, `toolInput`, `callId`, and `requestId`);
- the original request/session context;
- the authenticated responder as a channel-agnostic `SessionAuthContext`;
- the submitted decision (`approve` or `deny`).

An illustrative API could extend `approval` with separate request and response policies (exact naming is open):

```ts
export default defineTool({
  description: "Create a GitHub pull request.",
  inputSchema: z.object({
    repository: z.string(),
    // ...
  }),
  approval: {
    policy: always(),
    async authorizeResponse({ toolInput, responder, decision }) {
      if (decision !== "approve") return "allowed";

      const githubLogin = await identities.githubLoginFor(responder);
      return (await codeowners.includes(toolInput.repository, githubLogin))
        ? "allowed"
        : { type: "rejected", reason: "Only a CODEOWNER may approve this pull request." };
    },
  },
  async execute(input) {
    // ...
  },
});
```

Identity mapping and the external authorization lookup remain application concerns. eve should provide the durable enforcement point and authenticated responder context. Because channels already project actors into `SessionAuthContext`, the same policy should work for Slack, Teams, custom channels, and `eve/client` rather than introducing a Slack-specific callback.

When a response is unauthorized:

- the tool must not execute;
- the approval request must remain pending;
[...]

PR #1370, opened 2026-07-29
feat(eve): state machine for HITL authorization

benpankow:
## Summary

This PR is the bulk of approval authorization logic.

Introduces state machine for approval responses using the `request` / `response` approval API introduced in #1559.

This PR keeps approval state separate from conversation history. It records approval attempts, OAuth challenges, approval attempt outcomes, and the final approving or cancelling actor. This is necessary to enable nice multiplayer behavior, especially with blocking e.g. OAuth in the loop.

## Durable lifecycle

Approval coordination runs in durable phases:

```text
new Approval/Cancel request
→ apply Cancel request or persist new Approval candidates
→ run authorizers for active Approval candidates
→ persist rejection, OAuth, or settlement for each candidate
→ resolve the original input once every request is processed
```

If a responder needs to sign in, eve parks the candidate through the existing `authorization.required` lifecycle. A callback resumes the same attempt.

## State machine

| Current state | Input | Transition |
| --- | --- | --- |
| Pending request | Approve | Create an Approval candidate |
| Pending request | Cancel | Settle cancelled and stale active Approval candidates |
| Pending Approval candidate | Coordinator pass | Run authorizer for Approval candidate |
| Pending Approval candidate | Allowed | Settle other approved, stale candidates |
| Pending Approval candidate | Rejected, error, or expiry | Append terminal candidate history; request remains open |
| Pending Approval candidate | Authorization required | Persist the authz challenge,  park |
| Authorization required | Matching callback | Re-run authorizer for Approval candidate |
| Settled request | Any later response | No state change |

This PR does not contain Slack presentation or public candidate/settlement events; those are layered in #1371.

Depends on the approval response API PR: #1559.

## Validation

Validated as part of the consolidated stack with:

- `pnpm guard:invariants`
- `pnpm test:unit`
- focused candidate, batch, expiry, OAuth, dynamic-tool precedence, Cancel ordering, input-request, and tool-loop tests

### DIFF
--- a/packages/eve/src/context/build-dynamic-tools.ts
+++ b/packages/eve/src/context/build-dynamic-tools.ts
@@ -1,4 +1,5 @@
 import type { HarnessToolDefinition } from "#harness/execute-tool.js";
+import type { HarnessToolMap } from "#harness/types.js";
 import type { ContextKey } from "#context/key.js";
 import {
   SessionDynamicToolMetadataKey,
@@ -118,6 +119,20 @@ function buildReplayedApproval(
  * `LiveStepToolsKey`). Session/turn tools are replayed from durable
  * metadata via the bundler's registered step functions.
  */
+export function buildResponseAuthorizationTools(input: {
+  readonly authoredTools: HarnessToolMap;
+  readonly context?: { get<T>(key: ContextKey<T>): T | undefined };
+}): HarnessToolMap {
+  const tools = new Map<string, HarnessToolDefinition>();
+  for (const tool of input.context === undefined ? [] : buildDynamicTools(input.context)) {
+    if (!tools.has(tool.name)) tools.set(tool.name, tool);
+  }
+  for (const [name, tool] of input.authoredTools) {
+    if (!tools.has(name)) tools.set(name, tool);
+  }
+  return tools;
+}
+
 export function buildDynamicTools(ctx: {
   get<T>(key: ContextKey<T>): T | undefined;
 }): readonly HarnessToolDefinition[] {
--- a/packages/eve/src/harness/approval-delivery-coordinator.ts
+++ b/packages/eve/src/harness/approval-delivery-coordinator.ts
@@ -1,0 +1,553 @@
+import type { SessionAuthContext } from "#channel/types.js";
+import { buildCallbackContext } from "#context/build-callback-context.js";
+import { contextStorage } from "#context/container.js";
+import { AuthKey, SessionKey } from "#context/keys.js";
+import {
+  buildApprovalResponseAuth,
+  handleApprovalResponsePolicyError,
+} from "#execution/tool-auth.js";
+import {
+  createApprovalCandidate,
+  expireApprovalCandidates,
+  finishApprovalCandidate,
+  getActiveApprovalCandidate,
+  getApprovalAuditState,
+  markApprovalCandidateAuthorizationRequired,
+  settleAllowedCandidate,
+  settleDirectApprovalResponse,
+  type ActiveApprovalCandidate,
+} from "#harness/approval-candidates.js";
+import {
+  clearPendingAuthorization,
+  getAuthorizationResult,
+  getPendingAuthorization,
+  isAuthorizationSignal,
+  type AuthorizationChallenge,
+} from "#harness/authorization.js";
+import { isApprovalRequest } from "#harness/input-request-class.js";
+import { getPendingInputBatches } from "#harness/pending-input-batches.js";
+import type { HarnessSession, HarnessToolMap, StepInput } from "#harness/types.js";
+import type { InputRequest } from "#runtime/input/types.js";
+
+const UNAUTHENTICATED_APPROVAL_FEEDBACK = "Authentication is required to respond to this approval.";
+const TEXT_APPROVAL_FEEDBACK =
+  "Please use the Approve or Cancel buttons to respond to this approval.";
+const APPROVAL_AUTHORIZER_TIMEOUT_MS = 10_000;
+const APPROVAL_CANDIDATE_TTL_MS = 10 * 60_000;
+
+export interface ApprovalDeliveryResult {
+  readonly challenges: readonly AuthorizationChallenge[];
+  readonly feedback: readonly string[];
+  readonly kind: "continue" | "continue-coordination" | "authorization-required" | "park";
+  readonly session: HarnessSession;
+  readonly stepInput?: StepInput;
+}
+
+/**
+ * Advances approval state by one durable phase.
+ *
+ * | State | Input | Transition |
+ * | --- | --- | --- |
+ * | pending request | Approve | create responder-bound candidate |
+ * | pending request | Cancel | settle cancelled and stale candidates |
+ * | pending candidate | coordinator pass | run current authorizer |
+ * | pending candidate | allowed | settle approved and stale competitors |
+ * | pending candidate | rejected/error/expiry | append terminal history |
+ * | pending candidate | authorization required | persist private challenge |
+ * | authorization required | matching callback | re-run current authorizer |
+ * | settled request | any later response | no state change |
+ *
+ * Delivery ingestion returns before authorizer work so Cancel and candidate
+ * creation commit before long-running policy execution. Candidate results also
+ * commit before lifecycle events are projected by the next stack layer.
+ */
+/** Returns whether this invocation should prepare tools for persisted policy work. */
+export function shouldPrepareApprovalPolicyTools(input: {
+  readonly now?: number;
+  readonly session: HarnessSession;
+  readonly stepInput?: StepInput;
+}): boolean {
+  const batches = getPendingInputBatches(input.session.state);
+  const responses = [
+    ...(input.stepInput?.attributedInputResponses ?? []).map(({ response }) => response),
+    ...(input.stepInput?.inputResponses ?? []),
+  ];
+  if (
+    batches.some((batch) =>
+      batch.requests.some(
+        (request) =>
+          isApprovalRequest(request) &&
+          responses.some((response) => response.requestId === request.requestId),
+      ),
+    )
+  ) {
+    return false;
+  }
+
+  const now = input.now ?? Date.now();
+  return getApprovalAuditState(input.session.state).activeCandidates.some(
+    (candidate) =>
+      candidate.expiresAt > now &&
+      (candidate.status === "pending" ||
+        (candidate.authorizationChallenges?.some(
+          (challenge) => getAuthorizationResult(challenge.name) !== undefined,
+        ) ??
+          false)),
+  );
+}
+
+export async function coordinateApprovalDelivery(input: {
+  readonly now?: number;
+  readonly session: HarnessSession;
+  readonly stepInput?: StepInput;
+  readonly tools: HarnessToolMap;
+}): Promise<ApprovalDeliveryResult> {
+  const now = input.now ?? Date.now();
+  const expiredChallengeNames = getApprovalAuditState(input.session.state)
+    .activeCandidates.filter((candidate) => candidate.expiresAt <= now)
+    .flatMap(
+      (candidate) => candidate.authorizationChallenges?.map((challenge) => challenge.name) ?? [],
+    );
+  const expiredState = expireApprovalCandidates({ now, state: input.session.state });
+  let session: HarnessSession = {
+    ...input.session,
+    state: clearPendingAuthorization(expiredState, expiredChallengeNames),
+  };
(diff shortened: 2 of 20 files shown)


packages/eve/src/harness/input-requests.ts
```diff
@@ -149,7 +141,11 @@ export function resolvePendingInput(input: {
 
   // Pending batch exists -- only resolve if we have actual responses.
   const resolvedStepInput = resolveTextMessageInput(pendingBatch, stepInput);
```
ruiconti: **Typing "approve" as a message skips the authorization check entirely.**

A tool guarded by `approval.response` is supposed to run only after the policy validates whoever approved it. That holds when the user clicks the Approve button — but not when they type the word "approve" as a plain chat message, which eve has long supported as an equivalent way to answer.

Sequence:

1. The agent parks on an approval for `create_pull_request`, which has a `response` policy.
2. Anyone in the channel replies with the text "approve" (no button click).
3. `coordinateApprovalDelivery` looks only at `stepInput.inputResponses` — there are none, the delivery is just a message — so it does nothing and returns `continue`.
4. `resolvePendingInput` then converts the text into `{ optionId: "approve" }` (`resolveTextMessageInput`, this line) and resolves the batch immediately.
5. The tool executes. The `response` policy was never called.

So the entire feature this PR builds — "only a CODEOWNER may approve" — is bypassed by typing instead of clicking. The text path needs to feed the same coordinator flow as the button path, or refuse to auto-approve requests that require response authorization.

(The existing text-approval test at `input-requests.test.ts:395` doesn't set `responseAuthRequiredRequestIds`, which is why this passes today.)

packages/eve/src/harness/approval-delivery-coordinator.ts
```diff
@@ -0,0 +1,389 @@
...
+        settledAt: now,
+        state: session.state,
+      });
+      session = { ...session, state: settled.state };
+      didCommit ||= settled.result.kind === "settled";
+      continue;
+    }
+
+    if (response.optionId !== "approve") continue;
+    const responder = buildCallbackContext().session.auth.current;
```
ruiconti: **When two people message at nearly the same time, an approval can be checked against the wrong person.**

Sequence:

1. The agent asks for approval. Bob (not authorized) clicks Approve.
2. Before the turn wakes up, Alice (authorized) happens to send an unrelated message like "any update?" to the same session.
3. Both deliveries are batched into one turn. [`coalesceDeliveries`](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/harness/messages.ts#L231-L248) keeps *all* the payloads but only *one* sender identity — the most recent, Alice's.
4. `turnStep` installs Alice as the identity for the whole batch ([workflow-steps.ts:218](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/execution/workflow-steps.ts#L218-L222)).
5. This line reads that identity while processing Bob's Approve click, and asks the policy: "may **Alice** approve this?"
6. Policy says yes. The tool runs — approved by Bob, authorized as Alice.

This is the exact thing the feature exists to prevent: an unauthorized person's click going through. It only needs a timing overlap in a shared channel, which is the normal environment for this feature (the motivating use case in #1021 is a shared Slack channel).

The batching itself predates this PR and was harmless when sender identity carried no authorization weight. Now it does, so each approval response needs to keep the identity of the person who actually sent it.

packages/eve/src/harness/tool-loop.ts
```diff
@@ -673,11 +678,60 @@ export function createToolLoopHarness(config: ToolLoopHarnessConfig): StepFn {
...
+    });
+    session = coordinated.session;
+    if (coordinated.kind === "continue-coordination") {
+      const continuedSession =
+        coordinated.stepInput === undefined
+          ? session
+          : queueDeferredStepInput(session, coordinated.stepInput);
+      return { next: runStep, session: continuedSession };
+    }
+    if (coordinated.challenges.length > 0) {
```
ruiconti, line 795: **While one person is mid-sign-in, nobody else can cancel — or do anything.**

The PR's design says Cancel always works: "Active validation never blocks Cancel." The park created here breaks that once it reaches the driver, because the driver's wait loop ([workflow-entry.ts:307-328](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/execution/workflow-entry.ts#L307-L328)) reads **only** the OAuth callback iterator while an authorization is pending.

Sequence:

1. Bob clicks Approve. His policy check needs GitHub sign-in, so the turn parks here and sends him an OAuth link.
2. While parked with a pending authorization, the driver waits only on `authIterator`. Regular deliveries queue up unread.
3. Alice clicks Cancel. Nothing happens — her click sits in the queue.
4. Bob finishes signing in an hour later. His callback is processed *first*, his approval settles, and the tool runs.
5. Alice's Cancel is finally read — too late, it's discarded as stale.

Worse: if Bob abandons the sign-in, nothing ever wakes that loop. The 10-minute candidate expiry only runs when a step runs, and no step runs until a callback arrives — so the whole session hangs on one person's forgotten OAuth tab.

Two user-visible outcomes: a Cancel that was clicked first loses to an Approve that finished later, and one abandoned sign-in freezes the session for everyone. The driver wait needs to race the OAuth callback against ordinary deliveries whenever the park is approval-related.

(The Cancel-ordering unit test calls the coordinator directly and never reaches that loop, so it can't catch this.)

packages/eve/src/context/build-dynamic-tools.ts
```diff
@@ -118,6 +119,20 @@ function buildReplayedApproval(
  * `LiveStepToolsKey`). Session/turn tools are replayed from durable
  * metadata via the bundler's registered step functions.
  */
+export function buildResponseAuthorizationTools(input: {
+  readonly authoredTools: HarnessToolMap;
+  readonly context?: { get<T>(key: ContextKey<T>): T | undefined };
+}): HarnessToolMap {
+  const tools = new Map<string, HarnessToolDefinition>();
+  for (const tool of input.context === undefined ? [] : buildDynamicTools(input.context)) {
```
ruiconti, line 127: **A step-scoped dynamic tool with a response policy creates an approval that can never be approved.**

Sequence:

1. A `step.started` resolver produces a tool with `approval: { request, response }`. The model calls it; the agent parks and records that this request needs response authorization. So far correct — the tool exists at park time.
2. The user clicks Approve. A new step begins.
3. Step-scoped tools live only in virtual context, which is wiped at every step boundary ([run-step.ts:40](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/context/run-step.ts#L40)) and rebuilt by `step.started` dispatch.
4. But approval coordination runs *before* `step.started` fires ([tool-loop.ts:685](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/harness/tool-loop.ts#L681-L689) vs [tool-loop.ts:1250](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/harness/tool-loop.ts#L1250-L1254)). At that moment the tool doesn't exist, so `buildResponseAuthorizationTools` (this function) can't find its policy.
5. The coordinator fails closed: "Approval authorization is temporarily unavailable. Please try again."
6. The user tries again. Same sequence, same failure. Forever.

"Try again" is honest for a deployment gap, but here retrying can never succeed — the only exit is Cancel. Either run the step-tool resolvers before coordination on resume, persist the response policy the way turn/session scopes already do (`approvalResponseStepFnName`), or reject `approval.response` on step-scoped tools at definition time so nobody can author this trap.

ruiconti (commented):
Review of the durable approval coordination. Five findings, each anchored inline: (1) text approval bypasses the response policy, (2) batched deliveries can authorize the wrong person, (3) a pending OAuth blocks Cancel and can hang the session, (4) step-scoped dynamic tools create unapprovable requests, (5) malformed policy returns fail open. 1 and 2 look like blockers for the security goal of the stack; 3 points at pre-existing driver code that this PR's semantics make incorrect.

packages/eve/src/harness/approval-delivery-coordinator.ts
```diff
@@ -0,0 +1,389 @@
...
+        responder: input.responder,
+        session: {
+          id: context.session.id,
+          initiator: context.session.auth.initiator,
+          parent: context.session.parent,
+          turn: context.session.turn,
+        },
+      }),
+    );
+    if (outcome.status === "rejected") {
```
ruiconti, line 373: **A buggy policy approves instead of rejecting.**

This check treats exactly `{ status: "rejected" }` as rejection — and *everything else* as approval. A policy that returns `{ status: "reject" }` (typo), `{}`, or `undefined` (forgot the return value on one branch) settles the request as allowed and the tool runs.

```ts
async response({ responder }) {
  if (!isCodeowner(responder)) {
    return { status: "reject", safeReason: "..." };  // typo → tool runs anyway
  }
  return { status: "allowed" };
}
```

TypeScript catches this for authored tools, but dynamic tools cast the return value unchecked ([build-dynamic-tools.ts:110](https://github.com/vercel/eve/blob/cf5dc1662272810f82293173d48609257855a690/packages/eve/src/context/build-dynamic-tools.ts#L107-L112)), and durable replay across deployments means the running policy may not be the one that was typechecked.

An authorization gate should demand explicit proof of "yes", not absence of "no":

```ts
if (outcome?.status !== "allowed") { /* fail the candidate */ }
```

The PR already routes thrown errors to a non-terminal failure; a malformed return deserves the same treatment. Current tests only cover the exact `"allowed"`/`"rejected"` literals.

packages/eve/src/harness/input-requests.ts
benpankow: Ended up disabling text follow ups until the interaction is completed. Eventually we might want to support some sort of basic text matching (`I approve`).

packages/eve/src/harness/approval-delivery-coordinator.ts
benpankow: Good catch. Auth now piggybacks alongside every payload, so we no longer rely on the most recent identity.

packages/eve/src/harness/tool-loop.ts
benpankow: We now wait on OAuth callbacks and other interactions simultaneously, so parallel interactions work properly now.

packages/eve/src/context/build-dynamic-tools.ts
benpankow: We now resolve step scoped tools lazily as part of coordinating approvals, and avoid emitting step.started early - tested this fix manually in my deployed Slack/GH agent by setting up step scoped tools.

packages/eve/src/harness/approval-delivery-coordinator.ts
benpankow: Good catch, we now require exact `status: "allowed"` result.

ruiconti:
The E2E coverage here does not exercise the main path introduced by this PR. None of the HITL fixture tools define an `approval.response` policy, so the current approval evals still cover direct/legacy approval handling rather than responder-bound candidates and response authorization.

Before merging, I think we need E2E coverage for at least:

1. **Authorized approve:** `approval.response` allows the responder, the original tool executes exactly once, and the session returns to waiting.
2. **Durable settlement resume:** interrupt after the allowed settlement is persisted but before the pending batch consumes the synthetic approval; on resume, the settled request must still execute exactly once. This is an important crash boundary because settlement removes the active candidate.
3. **Rejected candidate:** the response authorizer rejects one attempt, the tool does not execute, and the shared request remains open for another responder/attempt.
4. **OAuth in `approval.response`:** Approve emits `authorization.required`; the callback resumes the same candidate, authorization succeeds, and the tool executes exactly once.
5. **Nonblocking OAuth:** while that approval-response OAuth challenge is open, an unrelated message completes without cancelling or executing the pending action; completing OAuth afterward still resolves the original approval.
6. **Cancel vs active candidate/OAuth:** one responder starts approval and parks on OAuth, another cancels; the request settles cancelled, the candidate/challenge becomes stale, and a late callback cannot execute the tool.

Follow-up coverage should include two responders racing to approve, rejection followed by another responder succeeding, candidate expiry/late callbacks, duplicate or late responses after settlement, mixed pending batches, and dynamic-tool response-policy replay.

The existing `button-cancel` and `text-approve` evals currently cover nearly the same immediate-cancel path. I would trade that duplication for the response-authorizer happy path and the durable/OAuth transitions above.

ruiconti (approved):

PR #1372, opened 2026-07-29
fix(eve): harden concurrent approval settlement

benpankow:
## Summary

Hardens multi-responder and durability semantics:

- concurrent challenge merging and candidate selection isolation
- exact callback-to-candidate correlation
- persisted responder restoration before resumed policy/token resolution
- lazy deterministic expiry
- durable-first pending, candidate-outcome, and settlement event ordering
- silent duplicate suppression and correct Slack feedback targeting

Depends on #1371.

## Validation

Covered by the consolidated top-stack checks.


benpankow:
Superseded: concurrency and durable-ordering hardening was folded into consolidated lifecycle PR #1371 so every stack layer builds independently.

PR #1387, opened 2026-07-30
test(e2e): split model and world suites

AndrewBarba:
## Why

The e2e matrix crossed every fixture with every model in both workflows (84 jobs/PR), so a full run rarely passed and every flake blocked merges. Per [the Slack discussion], most changes cannot plausibly break one flow on one model deterministically, and only a handful of fixtures vary per provider — so the cross product mostly bought flake, not coverage.

## What

Splits e2e into two suites:

| Suite | Workflow | World | Models | Jobs |
|---|---|---|---|---|
| Model suite | `e2e-local.yml` | local | real matrix models; only model-sensitive fixtures (`"e2e": { "modelMatrix": "full" }`) run on both, the rest run once on the default | 24 |
| World suite | `e2e-vercel.yml` | deployed Vercel | deterministic `mockModel()` via `EVE_E2E_MODEL=mock`; `real-model`-tagged evals excluded | 20 |
| World suite | `e2e-postgres.yml` | local production server backed by `@workflow/world-postgres` (service container) | same mock convention; asserts workflow runs persisted in Postgres | 20 |

**84 flaky jobs/PR → 64 jobs of which 40 are deterministic.** Aggregate check names `e2e-local` and `e2e-vercel` are unchanged; the new `e2e-postgres` aggregate satisfies the check the main ruleset already required (previously nothing reported it, leaving every PR blocked on a permanently "expected" status). Any additional world follows the same shape: consume `world_matrix`, set `EVE_E2E_MODEL=mock` + `EVE_E2E_WORKFLOW_WORLD`, run with `--exclude-tag real-model`.

The Postgres suite was validated locally against `postgres:18-alpine`: bootstrap → build → `eve start` → mock evals green, with hundreds of workflow runs persisted (`workflow.workflow_runs`).

### Pieces

- **`@eve-e2e/config`** (`e2e/fixtures/e2e-config`, extracted from #1367's design so that PR rebases down to its postgres-specific parts): `e2eAgentConfig` / `e2eSubagentConfig` / `e2eModel` / `e2eJudgeModel` resolve the matrix model, the mock sentinel, and the `EVE_E2E_WORKFLOW_WORLD` override in one place. Adopted across all 20 fixtures.
- **`eve eval --exclude-tag <tag...>`**: repeatable exclusion applied after `--tag`; a run where exclusion removes every match exits 0 ("this suite does not apply here"), while a no-match `--tag` stays exit 2. Unit-tested, documented, changeset included.
- **Discovery** now emits `model_matrix` (fixture × model legs) and `world_matrix` (one leg per fixture); the model registry lives in the script.
- **`real-model` tag**: evals whose assertions need a live model (judges, cache metrics, free-form tool planning). 90 evals start tagged; 10 already run mock-clean (verified locally with `EVE_E2E_MODEL=mock`): agent-workflow-stress, agent-session-timeout, agent-channels, agent-model (2/3), agent-basic-runtime (3/6), and the sandbox redeploy eval via a scripted directive-driven responder (it has no model-suite fallback). Untagging per fixture is the migration path for world-suite coverage.
- Docs: `e2e/README.md` rewritten around the two suites; stale AGENTS.md fixture reference fixed; leftover `agent-compaction-regressions-*` shell dirs removed.

## Testing

- `pnpm fmt` / `pnpm lint` / `pnpm guard:invariants` / `pnpm docs:check`
- eve typecheck + every fixture typecheck
- 5660 unit + 568 integration tests
- Local mock-mode eval runs (`EVE_E2E_MODEL=mock pnpm exec eve eval --strict --exclude-tag real-model`) green for the untagged fixtures

⚠️ First CI run to watch: the `agent-tools-sandbox` redeploy responder is the only untagged eval that needs Vercel credentials to validate — if it misbehaves, tag it `real-model` and iterate.


cmpadden (approved):
lgtm pending 🟢.

OwenKephart (approved):
naming's a bit odd for `--tag` / `--exclude-tag`, my guess is that in the future we'll want to have more generic `--include` / `--exclude` that can work based off of tags / paths / etc. but seems fine for now 🫡

PR #1389, opened 2026-07-30
fix(eve): make turn cancellation reliable for sessions parked on subagents

AndrewBarba:
## Why

Cancelling a session that was parked on an in-flight subagent was unreliable: the child could ride out its full in-tool fallback and complete uncancelled, and the parent could hold an accepted cancel indefinitely without ever emitting `turn.cancelled`. The `cancellation/cancel-subagent` e2e eval reproduced this intermittently on the Vercel world across both model providers.

Forensics across six CI incidents (timestamped event streams in the eval artifacts) separated the failure into two eve-side races in the turn loop plus one upstream root cause: **vercel/workflow#3183** — the workflow runtime could observe "idle" while a journaled `hook_received` delivery was still parked behind its deferral, raise a premature `WorkflowSuspension` carrying no work, and ack the wake. A run suspended racing `Promise.race([inbox.next(), cancelHook.next()])` — exactly eve's parked-on-subagents shape — went dormant with the accepted cancel journaled but unconsumed, until unrelated traffic woke it.

## What

Three eve-side fixes, each independent of the upstream bug:

1. **Cancel wins the settle race.** A cancellation observed while a durable turn step is returning now settles the turn as cancelled instead of losing to an ordinary completion. In `runTurnOwnedWorkflow`, the claimed cancel is honored whenever the signal has fired and the step produced no pending runtime-action batch (pending batches observe the signal in their own wait).

2. **The abort flips in the payload's own microtask.** The cancel signal aborts inside the hook read's continuation rather than a chained `.then`, so a cancel replayed alongside a completed step in one wake drain cannot lose the settle check by one task-queue hop.

3. **Descendant cascade hardening.** `cancelDescendantTurnsStep` retries undeliverable child cancels with backoff (covering the child adoption window), classifies `no_active_turn` reasons on `CancelTurnResult.reason`, and logs loudly instead of dropping a cancel silently.

And the root-cause fix:

4. **Upgrade the vendored Workflow DevKit to `@workflow/core@5.0.0-beta.38`** (plus latest sibling betas), which contains the #3198 fix for #3183. `@workflow/utils@beta.8` dropped `getWorldImport` in the same release; eve now owns the world-name→package mapping in `internal/workflow/world-target.ts`.

Two interim workarounds written against the unreleased upstream bug (request-time descendant fan-out at the cancel API boundary, and an explicit `reenqueueRun` nudge after accepted resumes) were removed once beta.38 shipped — the durable settle-time cascade is the principled owner of descendant cancellation now that the parent's cancel wake is reliable.

## Testing

- New unit tests for the settle race (`turn-workflow.test.ts`), the microtask abort ordering (`turn-cancellation-control.test.ts` — fails against the chained-`.then` implementation), the cascade retry/classification (`cancel-descendant-turns-step.test.ts`), and the world-target mapping
- Full unit (5,659) + integration (568) suites green against the re-vendored beta.38; fmt / lint / guard:invariants / typecheck / docs:check clean
- `e2e-vercel (agent-cancellation)` exercises the full parent+child cancellation path in CI; the failing runs and their artifacts are catalogued in the PR comments

## Provenance

Started as the settle-race fix extracted from #1367; the e2e failures it exposed led to the upstream root-cause investigation documented in the comments (six incidents, run ids, and timelines) and the internal issue filed against the workflow runtime.


AndrewBarba:
Added a second commit after digging into this PR's own e2e failures: the two `e2e-vercel (agent-cancellation, *)` legs exposed a **second cancellation bug**, distinct from the settle race this PR fixes.

**Failure forensics** (`cancel-subagent`, both model legs): the parent settled `turn.cancelled` correctly, but the child subagent never received its cascade cancel — its `wait-for-cancellation` tool rode out the full 90s fallback timeout, then the child ran another model step and completed `done` with zero `turn.cancelled`. Since the child's *next* step was serialized with an unaborted signal, the cancel-hook resume never reached the child run at all — a dropped delivery, not latency.

Every drop path in `cancelDescendantTurnsStep` was silent: missing pending batch, missing `childSessionIds`, and retry exhaustion (12×250ms, then return-as-success) all looked identical to a delivered cancel.

**The hardening commit**:
- exponential backoff on descendant cancel requests (~8s budget instead of a flat 3s)
- `no_active_turn` results now carry the classifying error (`HookNotFoundError` = target likely finished, `EntityConflictError` = transient contention, …)
- warn-level logs on every give-up path, so an uncancelled child is observable instead of invisible

This makes the drop far less likely and diagnosable when it still happens. A fully durable delivery (child consults durable cancel state at settle time, instead of best-effort hook resume) is a design change worth its own research doc — this keeps the fix reviewable.

The remaining e2e failures on the first run (`agent-tools` web-search legs) are the pre-existing live-model flake that #1387 quarantines.

AndrewBarba:
The single `e2e-vercel (agent-cancellation, anthropic-opus)` failure on the previous run (same child-never-cancelled signature, now intermittent — sol passed) pinned down a third defect, fixed in `4e222be1`.

**The microtask race**: when the child run wakes and replays a journaled cancel alongside a completed turn step, both continuations resolve in the same drain, payload first. But the abort was chained on `consumeMatchingCancel(...).then(() => controller.abort(...))` — one microtask hop behind the turn loop's own continuation. The loop checked `signal.aborted` before the abort callback ran, saw `false`, took the `done` arm, and `dispose()` swallowed the pending cancel permanently. The commit-1 settle check could lose by exactly one hop.

Fix: fire the abort inside the matching read's own continuation, so the signal is already flipped when any continuation from the same drain checks it. The new unit test reproduces the drain ordering (payload resolves before the step result) and fails against the chained-`.then` implementation.

**Residual risk, for the record**: the run→step abort bridge (the durable-stream side channel workflow-core uses to propagate an abort into an in-flight step invocation) swallows write failures; if that hop is lost, the child's tool still rides out its fallback timeout before the (now-reliable) run-side settle path catches the cancel at the next step boundary — slow, but no longer droppable by eve. Making descendant cancellation fully durable (child consults durable cancel state at settle) is follow-up design work.

AndrewBarba:
Root cause found and fixed structurally in `66ca1a28` — this is not another patch on the turn loop.

**The smoking gun** (from the run-3 artifact timestamps): the cancel was accepted at `02:27:52`, but the parent run — the only place the descendant cascade could execute — never woke until the child's completion delivery at `02:29:24`. The child's `wait-for-cancellation` tool reported *"fallback timeout reached without cancellation"*: no cancel ever reached it, because the cascade lived exclusively inside the suspended parent's workflow. The 9.9s gap between child completion (`02:29:24.301`) and the parent's `turn.cancelled` (`02:29:34.220`) is exactly the settle-time cascade's 8-attempt retry backoff against the already-completed child's missing hook. Every millisecond accounted for; commits 1–3 were working as designed and are why the parent settled cancelled at all.

**The fix**: `runtime.cancelTurn` now fans cancellation out to descendants immediately at request time, with no run wake required. Descendants are derived from the session's durable event stream (`subagent.called` minus `action.result`), read up to the tail observed on entry so a live stream can't block the request, and cancelled recursively — local children directly, remote children via the standard remote cancel route (which fans out on its own deployment). The settle-time cascade remains as the durable backstop, but terminal `no_active_turn` reasons now return after a single attempt instead of burning the 10s retry budget on children already cancelled at request time.

Deliberately avoided: extending the public `subagent.called` event (it would have bumped all four extension-capability contract epochs); remote nodes are re-resolved from the registry by dispatch URL instead.

Unit (5,670) + integration (568) green, guard/docs checks pass. The dropped/deferred hook-resume wake in the workflow world is still worth an upstream look — with this change eve no longer depends on it for cancellation correctness, but it cost this parent 92 seconds of suspension with an accepted cancel pending.

AndrewBarba:
Status after `a58d5982`:

**Request-time fan-out is confirmed working in production.** In the last two e2e runs the sleeper child was cancelled ~1.4s after the cancel request (previously it rode out its full 90s fallback). The remaining failure is the parent's own `turn.cancelled` emission, which still requires the workflow world to schedule the run that owns the cancel hook — the wake that the world has now dropped in 6 out of 6 failing incidents.

**Why the `a23dd001` nudge didn't take:** it re-enqueued `input.sessionId` — the session's *driver* run. The cancel hook is created inside the *turn workflow* run (`createTurnCancellationControl` executes in the turn workflow body), so the run suspended racing the payload is the hook's owning run, which `resumeHook` returns. The nudge woke a driver with nothing to do while the turn run stayed asleep — exactly the run-30542663066 signature (child cancelled in 1.4s, parent silent through the 240s timeout).

`a58d5982` re-enqueues `hook.runId` instead, with the session id as fallback. If this run of `cancel-subagent` goes green, the workaround holds and the upstream wake bug is fully contained; the internal issue write-up (6 incidents, run ids, hook tokens, timelines) is ready to file either way.

AndrewBarba:
**Root cause definitively identified — blocked on an upstream release, not on this PR's code.**

Dug into the workflow internals: the remaining `cancel-subagent` failure is **vercel/workflow#3183**, fixed by **#3198** (`b92c23ccb`, *"don't observe idle while a committed delivery is parked behind its deferral"*, merged 2026-07-29 23:12Z). That fix landed **~23 hours after `@workflow/core@5.0.0-beta.37` was published** (2026-07-29 00:21Z) — no published version contains it, and eve vendors the published beta.37 (verified: the vendored bundles contain zero trace of the guard).

Mechanism, matching all six incidents: the wake from `resumeHook` **is** delivered. During the replay, the turn workflow's `Promise.race([inbox.next(), cancelHook.next()])` has its journaled `hook_received` delivery parked between its serial queue slot and its detached `resolve()` (behind a macrotask yield); the idle check observes "idle" inside that gap, raises a premature `WorkflowSuspension` carrying no follow-up work, the runtime schedules nothing, and acks — dormant until unrelated traffic. This also explains why the `reenqueueRun` nudges (`a23dd001`, `a58d5982`) didn't cure it: every nudge invocation replays into the same deterministic-for-this-journal-shape race. And it explains the local/vercel asymmetry — e2e-local's in-process delivery timing rarely opens the gap.

**Plan:** wait for `@workflow/core@5.0.0-beta.38` (the `idle-check-parked-deliveries` changeset is already pending on their `main`), then bump eve's `@workflow/*` devDependencies and re-vendor. Everything in this PR remains correct and wanted: the settle-race fixes, the request-time descendant fan-out (children now cancel in ~1.4s instead of 90s — confirmed in production), and the reenqueue nudge (guards the wake-loss classes the upstream fix doesn't, e.g. producer-side namespace mismatch).

Remaining red is: `cancel-subagent` (this upstream bug) and `agent-tools` / `recursive-root-only` (provider streaming flakes, `MODEL_CALL_FAILED: Stream exceeded maximum duration`, occur across commits and providers — unrelated).

AndrewBarba:
`db803075` — upgraded to `@workflow/core@5.0.0-beta.38` (contains the #3198 fix for vercel/workflow#3183) and audited every change on this PR against the upstream fix. Net: **−887 lines**.

**Removed (existed only to compensate for the world bug):**
- Request-time descendant fan-out (`cancelTurnWithDescendantFanout`, ~780 lines with tests): with the parent's cancel wake reliable, the durable settle-time cascade is the principled owner of descendant cancellation and reaches children promptly. Proven by the pre-fanout opus passes where the wake happened to work.
- `reenqueueRun` nudge: `resumeHook` already enqueues the wake; the defect was in the *replay*, which beta.38 fixes. The nudge never helped and only added a redundant invocation per cancel.
- The retry reconciliation inside the settle cascade (existed to coexist with the fan-out); the cascade keeps its full adoption-window retry from `5c735a71`.

**Kept (eve-internal correctness, independent of the world bug):**
- Turn settle race: a cancel replayed alongside a completed step settles the turn as cancelled (`8d384a2b`).
- Microtask abort ordering: the cancel signal flips in the payload's own consumption drain (`4e222be1`).
- Settle-time cascade hardening + `CancelTurnResult.reason` classification (`5c735a71`).

**Upgrade notes:** `@workflow/utils@beta.8` dropped `getWorldImport` (static world-target injection revert); eve now owns the trivial mapping in `internal/workflow/world-target.ts`. All other `@workflow/*` pins bumped to latest betas.

Unit (5,659) + integration (568) green locally; scenario failures on my machine are internal-registry 401s in temp-app installs (environmental — CI validates). The remaining question is `e2e-vercel (agent-cancellation)` — this is the first run where the eval exercises the fixed core.

ctgowrie (approved):

PR #1403, opened 2026-07-30
feat(eve): add the /traces conversation viewer to the dev TUI

chadhietala:
Stacked on #1444 — **base is `trace-local-reader`, review that first.** This is the core layer of the `/traces` stack: capture (#1443) and reading (#1444) sit below; provider-executed tool cards (#1445), drag-to-select copy (#1446), and chrome polish (#1447) sit above.

`eve dev` writes a local trace spool while you chat, but inspecting it meant leaving the TUI for `eve traces` and reading span trees. This adds `/traces`: a full-screen viewer that re-tells a trace as the conversation it recorded.

**TUI with top-level tool calls**
image

**TUI with subagent dispatch**
image

## What changed

`/traces` opens on the terminal's alternate screen (the transcript's scrollback stays untouched underneath) and renders the trace as cards, discovered from the span tree and ordered by when their spans started:

- the **system prompt** collapsed at the top with a token estimate; **user** and **assistant** message cards; **tool** cards with `Input:`/`Output:` sections. Assistant cards carry model, duration, token, and gateway-cost metadata plus a dim `Thought:` reasoning section; error spans paint their whole card red
- work a **subagent** performed interleaves in time order — between the parent's dispatch and its resumed reply — with a dim `subagent:<name>` badge on every card of the delegated turn, and the delegated prompt rendered as a `task` card rather than a `user` message (uses the lineage attributes from #1433)
- cards expand/collapse with `←`/`→` or a mouse click (SGR reporting); the viewport scrolls by lines with keys or the wheel, so an expanded card taller than the screen is reachable to its end
- `Enter` opens a right-hand attributes drawer that stays metadata-only — status, timing, ids, non-payload OTLP attributes — since the cards already carry the payloads
- the spool is polled about once a second, so spans from the live session stream in (`● live` badge); `[`/`]` cycles traces, and the viewer opens on the current chat session's trace by matching session ids against the spool

The state is a pure reducer (`trace-viewer-state.ts`) with rendering split from it (`trace-view.ts`, `trace-conversation.ts`), a store that caches parsed segments so the 1 s poll only touches new files, and a session controller owning the poll/repaint lifecycle — all testable without a TTY. Keys survive in-flight polls: navigation during a slow disk read is re-applied against the fresh data rather than clobbered.

## Verification

Unit coverage for the store, reducer, frame renderer, and conversation builder (819 tests under `cli/dev/tui` at this layer). `test/tui-client/tui-traces.ts` drives the real runner through a pty: open → arrow-key fold → mouse-click expand → drawer metadata-only → subagent badge → trace cycling → escape → empty spool. Repeated end-to-end use against real spool data (fixtures through the gateway). Sweep green; full suite at the stack top.


chadhietala:
Restacked on #1433 (`research/subagent-trace-lineage`, which sits on #1410) and squashed to one commit. Beyond conflict resolution, three adaptations for the new capture shape:

- **Turns are discovered by name, not root position** — post-#1410 they parent to the `agent.session` window span (older spools with root turns still work)
- **The viewer opens on the current session's trace by matching session ids** — the derived-trace-id digest is gone, so the preference reads candidates newest-first and matches `sessionIds` (covers windowed sessions and subagent children recorded into a parent's trace); this also surfaced and fixed a latent bug where a preferred trace at index ≠ 0 was reverted by the post-read re-apply
- **Subagent turns are badged** — every card of a delegated turn carries a dim `subagent`/`subagent:<name>` label read from #1433's `agent.parent.turn.id` / `agent.subagent.name` attributes

The shared reader (`local-trace-reader.ts`) now carries `sessionIds` and `window` so `eve traces` and the TUI resolve identically, and the content-capture helpers moved to `agent-otel-content.ts` (provider was over the 700-line cap after the merge). Smoke suite gained a stage asserting the badge renders on a child turn under a session window root.

ctgowrie (approved):

PR #1419, opened 2026-07-30
feat(eve): trace inbound HTTP channel requests

karthikscale3:
**What changed:** Wrap every inbound HTTP channel request in one OTel `SERVER` span, named for the route, respecting an incoming `traceparent` and parenting nested channel/Workflow spans. Records only low-cardinality, non-sensitive attributes.

**Why:** Close the tracing gap so channel requests appear as a single root span in Datadog/Braintrust, correlated with downstream `hook.resume` and outgoing HTTP spans.


De-ASI-INTERFACE (approved):

packages/eve/src/internal/nitro/routes/channel-dispatch.ts
```diff
@@ -53,54 +54,69 @@ export async function dispatchChannelRequest(
...
+      );
+    }
 
-  let response: Response;
+    // Channel identity is known only after resolution; a 404 span carries
+    // just the route. `span` is undefined when instrumentation opted out of
+    // channel-request tracing.
+    span?.setAttribute("eve.channel.name", matchedChannel.name);
+    if (matchedChannel.adapter?.kind !== undefined) {
+      span?.setAttribute("eve.channel.kind", matchedChannel.adapter.kind);
```
AndrewBarba: Could we use `getChannelInstrumentationKind(matchedChannel.definition) ?? matchedChannel.adapter.kind` here? Behaviorless authored channels retain adapter kind `"http"`, so this currently reports `"http"` instead of the stamped `channel:<name>`.

karthikscale3: Good catch — fixed in 10054f1c. Now derives the kind via `getChannelInstrumentationKind(matchedChannel.definition) ?? matchedChannel.adapter?.kind`, so behaviorless authored channels report the stamped `channel:<name>` instead of the `"http"` adapter kind, matching the rest of the trace. Added a regression test covering a channel with adapter kind `"http"` and a stamped `channel:support`.

packages/eve/src/internal/nitro/routes/channel-request-instrumentation.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * This is observability-only: it never changes the response and performs no
+ * synchronous span export in the request path, adding only minimal in-process
+ * tracing overhead.
+ */
+export async function traceChannelRequest<T extends Response>(
+  input: TraceChannelRequestInput,
+  handler: (span: Span | undefined) => Promise<T>,
+): Promise<T> {
+  if (getInstrumentationConfig()?.traceChannelRequests === false) {
```
AndrewBarba: as discussed lets change default to no channel tracing
```suggestion
  if (getInstrumentationConfig()?.traceChannelRequests !== true) {
```

AndrewBarba (approved):
LGTM with one nit on default

packages/eve/src/internal/nitro/routes/channel-request-instrumentation.ts
karthikscale3: Done in 76c0648a — `traceChannelRequests` now defaults to `false` (`!== true` gate), so request spans are emitted only when explicitly enabled. Updated the prop doc, guide, and changeset to frame it as opt-in, and reworked the tests so the tracing suites enable the flag and a new suite asserts no span when it's unset or explicitly false.

PR #1421, opened 2026-07-30
feat(eve): add HITL tool approval authorization API

benpankow:
## Summary

Adds request- and response-time approval policies for authorizing the person who responds to a tool approval. The existing `approval: always()` function shorthand remains supported.

```ts
import { always } from "eve/tools/approval";

export default defineTool({
  description: "Refund a charge.",
  inputSchema: refundSchema,
  approval: {
    request: always(),
    async response({ request, response, responder, session, auth }) {
      const identity = await auth.getToken(billingApproverAuth);

      if (await canApproveRefund({ identity, request, response, responder, session })) {
        return { status: "allowed" };
      }

      return {
        status: "rejected",
        safeReason: "You are not permitted to approve this refund.",
      };
    },
  },
  async execute(input) {
    return refund(input);
  },
});
```

The callback receives:

- `request`: the stable approval request, tool call ID, tool name, and tool input;
- `response`: the submitted `{ decision: "approve" }`;
- `responder`: the identity produced by authenticated channel ingress;
- `session`: read-only session identity, initiator, turn, and parent lineage;
- `auth`: responder-bound `getToken` and `requireAuth` capabilities.

`auth` uses the explicit responder for token lookup, cache identity, challenge creation, callback completion, and eviction. It never substitutes mutable ambient session auth.

`TokenResult` now exposes the account represented by the credential:

```ts
interface TokenResult {
  token: string;
  expiresAt?: number;
  providerSubject?: string;
}
```

For example, a GitHub authorization provider can return the responder's stable GitHub user ID.

Dynamic tools retain both the request-time policy and response authorizer across durable replay. This PR also publishes the extension capability epochs required by the new public types.

This PR does not create candidates, settle approvals, or change channel/client presentation. Durable coordination is layered in #1370.

Depends on the Approve/Cancel protocol PR: #1385.

## Validation

- `pnpm guard:invariants`
- focused approval API, dynamic-tool, tool-auth unit and integration tests
- validated as part of the consolidated top-stack unit suite

### DIFF
--- a/packages/eve/src/public/definitions/approval.ts
+++ b/packages/eve/src/public/definitions/approval.ts
@@ -1,19 +1,16 @@
+import type { SessionAuthContext } from "#channel/types.js";
+import type { SessionParent, SessionTurn } from "#context/keys.js";
+import type { ToolAuthOptions, ToolAuthProvider } from "#public/definitions/tool.js";
 import type { SessionContext } from "#public/definitions/callback-context.js";
+import type { TokenResult } from "#runtime/connections/types.js";
 
 type ApprovalToolInput<TInput> = TInput extends object ? Readonly<TInput> : TInput;
 
 /**
- * Context passed to an {@link Approval} function.
+ * Context passed to an {@link ApprovalPolicy} function.
  *
  * Extends {@link SessionContext} so approval policies can make decisions from
  * the active session, current caller, and turn.
- *
- * `approvedTools` is the set of tool names (or compound approval keys)
- * already approved at least once in the current session. `toolName` is the
- * runtime name of the tool being evaluated. `toolInput` is the raw input the
- * model passed, available for input-aware decisions. `callId` is the id of
- * the call being evaluated — the same `callId` carried by the call's stream
- * events and its `execute` context.
  */
 export interface ApprovalContext<TInput = Record<string, unknown>> extends SessionContext {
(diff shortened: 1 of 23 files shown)


packages/eve/src/public/definitions/approval.ts
```diff
@@ -40,7 +32,63 @@ export type ApprovalStatus =
...
 
-/** Shared approval policy used by authored tools and connections. */
-export type Approval<TInput = Record<string, unknown>> = (
+/** Request-time approval policy shared by authored tools and connections. */
+export type ApprovalPolicy<TInput = Record<string, unknown>> = (
   ctx: ApprovalContext<TInput>,
 ) => ApprovalStatus | Promise<ApprovalStatus>;
+
+/** Stable tool request passed to a response authorizer. */
+export interface ApprovalResponseRequest<TInput = Record<string, unknown>> {
```
ruiconti, line 41: Is this just 

```suggestion
export interface ApprovalRequest<TInput = Record<string, unknown>> {
```

?

benpankow:
Superseded by #1559 with the identical signed commit to clear incorrect GitHub mergeability state.

packages/eve/src/public/definitions/approval.ts
benpankow: Good call, addressed in #1559.

Issue #100, opened 2026-06-19
eve dev as acp server


morlay:
### What problem are you trying to solve?

acp over tui

### Proposed solution

adapter to acp

https://agentclientprotocol.com/get-started/clients

PR #1423, opened 2026-07-30
feat: add Buzz ACP compatibility adapter

cmpadden:
### Summary

Related to #100. Builds on the generic ACP server and verified remote authentication from #1347.

Adds the experimental `@eve/buzz-acp-adapter` package so Buzz Desktop can run local or deployed eve agents and publish their final ACP response as a signed Buzz reply. The globally installed adapter discovers the authored eve model, configures the Buzz harness, authenticates protected Vercel targets, and uses the local eve application root for ACP sessions.

Buzz remains responsible for verifying senders. The adapter defaults to Buzz's owner-only gate, never treats sender text as an eve identity, and requires an explicit shared-principal opt-in for broader access. Reply publication is idempotent and fails closed when Buzz context or delivery confirmation is ambiguous; Buzz credentials remain outside the eve runtime.

The PR also adds Buzz to the integrations gallery as a gallery-only Channel with ACP setup and troubleshooting guidance. Review focus is the sender authorization boundary and reply behavior when delivery is uncertain.

### Relevant upstream Buzz issues

- [#2698 — buzz-acp reply delivery is implicit](https://github.com/block/buzz/issues/2698)
- [#3385 — Custom/BYOH ACP harnesses lack an authenticated reply path](https://github.com/block/buzz/issues/3385)
- [#3587 — buzz-acp can publish replies after a turn is terminal](https://github.com/block/buzz/issues/3587)

### Validation

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test:unit`
- `pnpm guard:invariants`
- `pnpm --filter @eve/buzz-acp-adapter test:unit` — 10 files, 39 tests passed
- `pnpm --filter @eve/buzz-acp-adapter typecheck`
- `pnpm --filter @eve/buzz-acp-adapter build`
- `pnpm --dir apps/docs exec vitest run lib/integrations/discovery.test.ts` — 12 tests passed
- packed the public npm tarball and verified its CLI, LICENSE, and NOTICE
- built the docs site and smoke-tested `/integrations/buzz`
- completed a live Buzz Desktop turn against the local weather agent, including an intentionally incorrect Buzz working directory

GitHub CI, scenario tests, and e2e suites pass. npm Trusted Publishing for `@eve/buzz-acp-adapter` must be configured separately before release.

### Checklist


packages/eve-buzz-acp-adapter/src/proxy.ts
```diff
@@ -0,0 +1,182 @@
...
+      if (message.method === "session/prompt" && message.id !== undefined) {
+        const text = promptText(message.params?.prompt);
+        turns.set(keyForId(message.id), {
+          route: parseBuzzRoute(text),
+          sessionId:
+            typeof message.params?.sessionId === "string" ? message.params.sessionId : undefined,
+          text: "",
+        });
+        message = addReplySinkInstruction(message);
+      }
```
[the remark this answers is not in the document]
cmpadden: Fixed in 724d6d38. The connector now listens for child stdin errors and fails subsequent writes cleanly when `eve acp` has stopped accepting protocol input.

packages/eve-buzz-acp-adapter/src/publication-state.ts
```diff
@@ -0,0 +1,97 @@
...
+import type { BuzzRoute } from "./types.js";
+
+export interface PublicationReservation {
+  commit(): Promise<void>;
+  release(): Promise<void>;
+}
+
+export function publicationKey(route: BuzzRoute): string {
+  if (!route.replyTo) throw new Error("Buzz reply routes require an event anchor");
+  return createHash("sha256").update(`${route.channelId}\0${route.replyTo}`).digest("hex");
```
ctgowrie: Buzz uses the thread root as `replyTo` for every turn in a thread. This gives every turn the same publication key. After the first reply all later responses are skipped as already published. Can we track the triggering event ID separately and use it for idempotency? `replyTo` should only determine where the response is posted.

packages/eve-buzz-acp-adapter/src/buzz-context.ts
```diff
@@ -0,0 +1,48 @@
...
+  const nextSection = afterContext.search(/^\[[^\n]+\]\s*$/m);
+  const context = nextSection < 0 ? afterContext : afterContext.slice(0, nextSection);
+
+  const channel = context.match(
+    new RegExp(`^Channel:\\s*(?:.*\\(#(${CHANNEL_ID})\\)|(${CHANNEL_ID}))\\s*$`, "im"),
+  );
+  if (channel === null) return undefined;
+
+  const replyTo = context.match(new RegExp(`--reply-to\\s+(${EVENT_ID})`, "i"))?.[1];
+  if (!replyTo) return undefined;
```
ctgowrie: Buzz omits `--reply-to` for a new top-level DM. This parser rejects the route even though the DM channel is valid. The eve turn completes but its response cannot be posted. Can we accept a channel-only route and publish a top-level DM when `replyTo` is absent?

packages/eve-buzz-acp-adapter/src/publisher.ts
```diff
@@ -0,0 +1,64 @@
...
+  route: BuzzRoute;
+  text: string;
+  timeoutMs: number;
+  signal?: AbortSignal;
+}): Promise<void> {
+  if (options.signal?.aborted) throw new Error("Buzz reply publication cancelled");
+  await new Promise<void>((resolve, reject) => {
+    const child = spawn(options.buzzCli, publicationArguments(options.route), {
+      env: options.environment,
+      stdio: ["pipe", "ignore", "pipe"],
```
ctgowrie: `buzz messages send` reports delivery as JSON with `event_id` and `accepted`. This discards that response and only checks the exit status. The CLI can exit successfully with `accepted: false` so the adapter may mark a rejected message as published. Can we capture stdout and commit only when `accepted` is true?

packages/eve-buzz-acp-adapter/src/proxy.ts
```diff
@@ -0,0 +1,233 @@
...
+            buzzCli: options.buzzCli,
+            environment: options.environment,
+            route: turn.route,
+            signal: turn.publication.signal,
+            text,
+            timeoutMs: options.publishTimeoutMs,
+          });
+          await reservation.commit();
+        } catch (error) {
+          if (turn.publication.signal.aborted) await reservation.commit();
```
ctgowrie: Cancellation does not tell us whether Buzz received the message. This marks every cancelled send as published even when the message may not have been delivered. It also allows a retry after every other error even when the message may have been delivered. That can either lose the reply or post it twice. Can the publisher return delivered not delivered or unknown and use that result to update the reservation?

packages/eve-buzz-acp-adapter/src/remote-target-auth.ts
```diff
@@ -0,0 +1,80 @@
...
+  prompter?: Prompter;
+  target: InstallTarget;
+  dependencies?: Partial<RemoteTargetAuthDependencies>;
+}): Promise<{ info: EveTargetInfo; vercelScope?: string }> {
+  const dependencies = { ...defaultDependencies, ...options.dependencies };
+  const targetOptions = eveTargetOptions(options.target, options.eveBin);
+  try {
+    return { info: await dependencies.readEveTargetInfo(targetOptions) };
+  } catch (error) {
+    if (!(error instanceof VercelDeploymentProtectionError) || options.target.kind !== "remote") {
```
ctgowrie: The runtime uses the eve verified remote client but this installer probe does not. It fetches `/eve/v1/info` directly and only starts authentication after a Vercel Protection redirect. A normal eve `401` stops installation before login. Can this probe use the same verified remote client as `eve acp`?

packages/eve-buzz-acp-adapter/src/proxy.ts
```diff
@@ -0,0 +1,233 @@
...
+        }
+        requests.set(keyForId(message.id), message.method);
+      }
+
+      if (message.method === "session/prompt" && message.id !== undefined) {
+        const text = promptText(message.params?.prompt);
+        turns.set(keyForId(message.id), {
+          cancelled: false,
+          publication: new AbortController(),
+          route: parseBuzzRoute(text),
```
ctgowrie: Buzz uses one ACP session per channel and can allow multiple people to trigger it. This route does not retain the verified sender identity. Every person who can trigger the agent therefore uses the same eve authentication and connections.

packages/eve-buzz-acp-adapter/src/publisher.ts
cmpadden: Addressed in a975dc99. The publisher now captures bounded stdout, parses the Buzz JSON response, and treats only `accepted: true` with a valid `event_id` as confirmed delivery.

packages/eve-buzz-acp-adapter/src/remote-target-auth.ts
cmpadden: Addressed in a975dc99. Remote installer inspection now uses the new eve-owned `inspectVerifiedRemoteAgent` setup API, which shares the exact-origin verified client and interactive Vercel login/Trusted Sources flow used by remote development clients.

packages/eve-buzz-acp-adapter/src/proxy.ts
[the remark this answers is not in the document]
cmpadden: Addressed in a975dc99. The proxy now aborts and settles its input reader when child output or exit wins, with a regression test covering child exit while Buzz input remains open.

packages/eve-buzz-acp-adapter/src/proxy.ts
cmpadden: Addressed in a975dc99. Publication now returns delivered, not-delivered, or unknown; confirmed rejection releases the reservation, confirmed delivery commits it, and unknown delivery is persisted to block automatic retries. In-flight publication is no longer aborted by ACP cancellation.

packages/eve-buzz-acp-adapter/src/publication-state.ts
cmpadden: Addressed in dd8d6f64. `BuzzRoute` now retains the triggering event separately from the optional reply anchor, and publication idempotency is keyed by `channelId + triggeringEventId`. The parser uses Buzz’s generated ACP prompt-block framing and selects the final triggering event in a batch.

packages/eve-buzz-acp-adapter/src/buzz-context.ts
cmpadden: Addressed in dd8d6f64. `replyTo` is now optional: a valid Buzz context plus triggering event produces a channel-only route, and the publisher omits `--reply-to` for top-level DMs. Added focused top-level DM coverage.

packages/eve-buzz-acp-adapter/src/buzz-context.ts
```diff
@@ -0,0 +1,89 @@
...
+    new RegExp(
+      `^\\[New (?:message — arrived while you were working|request — supersedes previous)\\]\\n\\n--- Event 1 \\([^\\n]*\\) ---\\nEvent ID:\\s*(${EVENT_ID})\\nChannel:`,
+      "i",
+    ),
+  );
+  if (newSingle !== null) return [newSingle[1]!];
+
+  const count = eventCount(block);
+  if (count === undefined) return [];
+  const event = block.match(
```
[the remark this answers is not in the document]
cmpadden: Fixed in 14dd2212. Batch parsing now enumerates every event header that has the generated `Event ID`/`Channel` shape and requires the headers to exactly match the declared count and 1..N sequence. A forged boundary therefore makes the batch ambiguous and route parsing fails closed instead of selecting the forged ID. Added a regression test for an earlier event injecting a complete fake final-event block.

packages/eve-buzz-acp-adapter/src/proxy.ts
cmpadden: Agreed. Addressed in cb517fd8: the adapter now fails before spawning eve unless Buzz’s effective author gate is `owner-only` (or inert `nobody`). Broader or unavailable gates require an explicit `--allow-shared-principal` opt-in, which warns that every accepted sender shares the configured eve authentication, connections, tools, and channel session. We intentionally do not derive an eve principal from the prompt’s `From:` text; true per-sender auth requires structured verified sender metadata plus an explicit ACP principal handoff.

De-ASI-INTERFACE (approved):

ctgowrie (approved):

PR #1444, opened 2026-07-31
refactor(eve): extract the local trace reader into a shared harness module

chadhietala:
Stacked on #1443 — **base is `trace-span-content`, review that first.**

Reading the local trace spool — listing trace directories, parsing OTLP/JSON segments, assembling spans into ordered traces — lived inside the `eve traces` CLI command. The `/traces` viewer up-stack needs the same reads without going through CLI printing, and the scenario tests were already importing a CLI module to get at them.

## What changed

The reading half of `commands/trace.ts` moves to `#harness/local-trace-reader.js`: `listLocalTraces`, `readLocalTrace`, `assembleLocalTrace`, `parseLocalTraceSegment`, `describeLocalTraceSpan`, and the `LocalTrace`/`LocalTraceSpan` types. The command becomes a thin CLI over the reader and keeps its behavior; the scenario test imports the reader directly instead of the command.

Two small additions ride along for consumers up-stack: assembled traces expose `sessionIds` (every `agent.session.id` seen in the trace, for matching a chat session to its trace) and `window` (the session-window index from #1410). The reader keeps its defensive posture — malformed segments are skipped, segments over 8 MiB are ignored, span ids are validated — so a spool being written or swept mid-read never throws.

## Verification

The `eve traces` integration tests pass unchanged (7 tests — behavior-preservation is the point), plus workspace typecheck at this layer. Full sweep at the stack top.

### DIFF
--- a/.changeset/shared-trace-reader.md
+++ b/.changeset/shared-trace-reader.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+The local trace spool reader behind `eve traces` moved into a shared internal module; command behavior is unchanged.
--- a/packages/eve/src/cli/commands/trace.integration.test.ts
+++ b/packages/eve/src/cli/commands/trace.integration.test.ts
@@ -4,12 +4,9 @@ import { join } from "node:path";
 
 import { afterEach, describe, expect, it } from "vitest";
 
-import {
-  listLocalTraces,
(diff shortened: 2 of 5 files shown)


ctgowrie (approved):
Approving since other tracing related things are here but harness seems like the wrong package for this

PR #1465, opened 2026-07-31
fix(eve): report evals without agent output to Braintrust

### LINKED ISSUE #1405: Braintrust reporter crashes the whole eval run when an eval produces no agent turn (output: null → "output must be specified")
## Summary

An eval that never drives an agent turn — e.g. one that calls `t.target.dispatchSchedule(...)` and then asserts database state with `t.check` — completes with `result.output === null` (per the documented derivation: *"the final turn's structured data when present, otherwise its last assistant message, or null when none was produced"*).

When the Braintrust reporter is active, `BraintrustReporter.onEvalComplete` forwards that value verbatim:

```js
this.#n.log({ id: e.id, input: ..., output: e.result.output, ... })
```

`braintrust`'s `validateAndSanitizeExperimentLogFullArgs` rejects null and undefined alike (`isEmpty(event.output)` → `throw new Error("output must be specified")`), and the throw escapes `onEvalComplete`, killing the entire `eve eval` run — remaining evals never execute and no artifacts are written, even though the crashing eval itself passed all its gates.

## Repro

1. Any eval whose `test` performs no `t.target` message turn (schedule-dispatch assertions over DB state are the natural case).
2. Run `eve eval <that-eval>` with `BRAINTRUST_API_KEY` set.
3. The eval passes (`✓ gates n/n`), then the run exits 1 with `output must be specified`.

Observed on eve 0.27.12 and 0.27.13 (reporter dist byte-identical across the two); braintrust 3.9.0.

## Expected

A no-turn eval is a legitimate shape (the eval API's own `output` doc anticipates `null`). The reporter should sanitize before logging — e.g. `output: e.result.output ?? ""` or omit/stub the field — and a reporter throw in `onEvalComplete` arguably shouldn't abort the run and drop the remaining evals + artifacts either.

## Workaround we're using

Split invocation: keyed `eve eval` for turn-producing evals (posts the experiment), a second un-keyed run for the no-turn evals (console gates only). Retires when this is fixed.

BlackishGreen33:
Closes #1405.

## What

Braintrust reporting stopped an `eve eval` run when an eval finished without producing an agent turn. Eve correctly records that output as `null`, but Braintrust requires a non-null `output` value.

## How

- Send an empty string to Braintrust when the eval output is `null` or `undefined`.
- Keep every existing non-null output unchanged.
- Add a regression test for an eval with no agent output.

## Tests / changeset

- Added focused unit coverage for the Braintrust reporter.
- Added an `eve` patch changeset.

## Verification

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/evals/runner/reporters/braintrust.test.ts`
- `pnpm test`
- `pnpm typecheck`
- `pnpm fmt`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm check:deps`
- `pnpm build`

### PR Checklist

- [x] I added a changeset because this touches the published `eve` package
- [x] DCO sign-off passes for every commit


BlackishGreen33:
CI note: all checks that do not require secrets unavailable to fork PRs passed, including lint, typecheck, unit tests, Ubuntu and Windows integration tests, scenario tests, DCO, Socket, and Postgres E2E.

The remaining failures are external to this patch:
- Vercel deploy jobs have empty `VERCEL_TOKEN`, `VERCEL_PROJECT_ID`, and `VERCEL_ORG_ID`, then fail with `No existing credentials found`.
- Real-model E2E and TUI jobs have an empty `AI_GATEWAY_API_KEY`, then fail with `AI Gateway received no credentials`.
- The docs preview requires Vercel team authorization.

No code changes were made for these fork-only credential failures.

Issue #1084, opened 2026-07-22
Background tasks: MCP Tasks-aligned task contract for tools and subagents


ruiconti:
Today, work the model starts can only resolve while its context is still alive: subagent results can only resume an open turn (the parent turn parks until every child completes), and HITL requests either lapse (questions cleared as ignored by the next message) or block (approvals defer all subsequent input). A response arriving after its request left \`pendingInputs\` is downgraded to plain text and never executes the original action.

Proposal: a durable **task** record aligned with the MCP Tasks extension (\`io.modelcontextprotocol/tasks\`, SEP-2663) — lifecycle \`working | input_required | completed | failed | cancelled\`, \`tasks/get\`/\`tasks/update\`/\`tasks/cancel\` semantics — where sync vs background is a property of the invocation, not the task. Background elections terminalize the tool-call position with a \`CreateTaskResult\`-shaped placeholder; status transitions reach subscribers as deliveries to registered notification endpoints (the generalization of the existing remote-agent callback), waking the parked driver through its existing deliver-only vocabulary.

Planned in five additive phases:

1. Task contract and implementation (no users)
2. Notifications contract: registered endpoints stored on session state
3. Wire tasks to notifications
4. \`defineTool()\` election via \`execution.taskSupport\`
5. Migrate the built-in \`agent\` tool and subagents

Research plan: \`research/background-tasks.md\` (PR to follow).

PR #1482, opened 2026-07-31
feat(eve): plan subagents as background tasks

ruiconti:
This is the contract for running subagents as durable background tasks after the agent-messaging stack lands. It changes no runtime behavior.

Today a subagent call leaves the model step as a private `runtimeAction`. eve starts the child, then keeps the parent turn open until every child in the batch finishes. Local children can forward input and authorization through that active turn; remote children can only report completion or failure.

That boundary works for synchronous delegation. It does not work for background work: the parent cannot move on while the child runs, progress has no shared local/remote path, and non-terminal communication depends on the original turn still being alive.

## Proposed flow

With `experimental.tasks`, each subagent call creates a durable task. The original tool call receives one task receipt, closing that call. Progress, input requests, authorization, completion, failure, and cancellation then move over the task boundary without keeping the parent turn open.

```text
parent model
    |
    v
parent harness ---- creates/indexes ----> task run
    |                                      | owns lifecycle state
    +---- resolves ---------------------> agent handle
    |                                      | owns child address
    +---- dispatches --------------------> child executor
                                           |
                                           +---- updates task run

parent returns to task through task_peek / task_await / notifications
```

The ownership split is the main design decision. The task run is the single writer for lifecycle transitions. The parent session stores only the `taskId -> taskRunId` index. Agent handles remain the source of child identity and routing. Credentials stay in private bindings and never enter model context or history.

The dedicated task run comes from the failure exposed by [#1190](https://github.com/vercel/eve/pull/1190): callback routes and child executors need to update task state, but neither owns the current threaded session snapshot. A separate durable run gives those writers a stable boundary and serializes competing updates.

## Delivery shape

The rollout is additive. Flag plumbing, task storage, gated controls, and delegated execution land inert first. A later stage moves the six child flows onto the task wire and lets the flag select the new mode. Local/remote normalization and flag removal come last.

This plan assumes the agent-messaging stack [#1328](https://github.com/vercel/eve/pull/1328)-[#1331](https://github.com/vercel/eve/pull/1331) as its addressing layer. Tasks compose with agent handles; they do not introduce another child-session identity or credential store.

The first implementation is subagents only. Authored tools, built-ins, connections, skills, dynamic workflows, and the public `defineTool` API stay unchanged. MCP Tasks provides lifecycle vocabulary and control semantics, not an MCP endpoint.

## Review focus

- **Sequencing:** the design and delivery plan currently disagree on whether agent messaging is already the baseline. They should both describe tasks as phase 2.
- **Creation:** define the task shape before child acknowledgement, including local metadata and dispatch-start failure.
- **Wake policy:** decide when terminal and `input_required` updates wake a parked parent instead of waiting for an explicit task control.
- **Busy semantics:** decide whether `task_send` queues behind a running child or returns `AGENT_BUSY`.
- **Trust and retention:** narrow which fields an executor may update, then define how long terminal tasks remain addressable.

Part of [#1084](https://github.com/vercel/eve/issues/1084).

### DIFF
--- a/research/subagents-as-tasks-implementation.md
+++ b/research/subagents-as-tasks-implementation.md
@@ -1,0 +1,197 @@
+---
+issue: https://github.com/vercel/eve/issues/1084
+status: draft
+last_updated: "2026-07-31"
+---
+
+# Subagents as tasks: additive delivery plan
+
+## Summary
+
+This plan sequences the implementation of the [subagents-as-tasks design] into strictly additive
+stages behind `experimental.tasks`. Every stage must land with this invariant: **with the flag
+absent, no existing codepath's behavior, wire shapes, or event stream changes.** The design doc
+owns the contract; this doc owns the order of construction, what each stage reuses from the
+agent-messaging stack, and how each stage is verified before the next begins.
+
+## Baseline
+
+The baseline is `main` plus the agent-messaging stack (`am/01-contracts` through
+`am/04-surface`, the phase-one agent2agent work). That stack supplies the session-addressing
+dependency the design doc declares:
+
+- **Durable child address**: the agent handle store (`eve.agent.handles` on session state) with
+  replay-safe operation ids, commit-before-side-effect dispatch, and `starting/running/parked`
+  phases. Handles hold the child `continuationToken` (local) or `url` + `continuationToken`
+  (remote); credentials never render to the model.
+- **Follow-up delivery**: `dispatchToAgentHandle` / `continueRemoteAgentSession`, with
+  `AGENT_BUSY` / `AGENT_MISMATCH` / `AGENT_UNREACHABLE` classification and `intent: "resume"`
+  sends that 404 instead of silently starting a new session.
+- **Reply envelope**: `TurnCaller` (`replyTo: hook token | callback URL`) and the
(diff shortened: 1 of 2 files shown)


research/subagents-as-tasks-implementation.md
```diff
@@ -0,0 +1,197 @@
...
+Task identity reuses the operation-id derivation, `hash(parentSessionId, parentTurnId, callId)`,
+so replayed creation for the same originating call yields the same task without new machinery.
+
+### Flag composition
+
+`task_send` follow-ups to a finished child require conversation-mode children and parked
+handles, which `experimental.subagentPersistentSessions` gates today. `experimental.tasks`
+therefore implies persistent-session children for subagent dispatch. Whether it sets the other
+flag or simply selects the same behavior internally is a stage-4 decision; the two flags must
+not produce a third hybrid mode.
```
ruiconti, line 51: Let's defer tool async communication to after a2a communication stack lands i.e., phase 2.

ruiconti: a2a has landed.

Rchie121 (approved):

AndrewBarba (approved):

PR #1485, opened 2026-07-31
feat(eve): support dynamic subagents

### LINKED ISSUE #696: Add conditional enable/disable support for declared subagents
## Summary

It would be useful if Eve allowed declared subagents to be conditionally enabled or disabled at runtime.

Eve currently supports dynamic tools, skills, instructions, and models through `defineDynamic`. I could not find an equivalent mechanism for conditionally controlling which declared subagents are available to a parent agent.

This request is intentionally broader than access control. A subagent may need to be disabled based on configuration, feature flags, environment, tenant capabilities, application state, or any other runtime condition.

## Use cases

Conditional subagents would be useful for:

- Feature flags and gradual rollouts
- Environment-specific capabilities
- Optional product modules
- Subscription or plan-based features
- Tenant-specific integrations
- Maintenance mode
- Regional restrictions
- Temporarily unavailable dependencies
- Application-level permissions
- Experiments and A/B tests

For example, an application might declare several specialists:

```text
agent/subagents/
├── sales/
├── finance/
├── inventory/
└── purchasing/
```

The application may want to enable only a subset for a particular session or turn without maintaining separate agent manifests.

## Why not dynamic instructions?

Dynamic instructions can tell the parent not to use a particular subagent, but they do not remove that subagent from the model-visible tool surface.

Framework-level conditional availability would be more reliable and would avoid:

- Exposing irrelevant capability descriptions
- Predictable failed delegation calls
- Unnecessary tool and token usage
- Depending on prompt compliance for capability composition

## Proposed API

One option would be an `enabled` or `available` resolver in the declared subagent configuration:

```ts
export default defineAgent({
  description: "Analyze financial and accounting data.",
  model: "openai/gpt-5.5",

  enabled: (_event, ctx) => {
    return ctx.session.auth.current?.attributes.modules?.includes("finance")
      ?? false;
  },
});
```

It could also support static configuration:

```ts
export default defineAgent({
  description: "Analyze financial and accounting data.",
  model: "openai/gpt-5.5",
  enabled: process.env.ENABLE_FINANCE_AGENT === "true",
});
```

Another option would be to use Eve’s existing dynamic capability pattern:

```ts
// agent/subagents.ts

import { defineDynamic } from "eve";

  events: {
    "session.started": async (_event, ctx) => {
      const enabledModules = await getEnabledModules(ctx);

      return {
        sales: enabledModules.includes("sales"),
        finance: enabledModules.includes("finance"),
        inventory: enabledModules.includes("inventory"),
      };
    },
  },
});
```

## Expected behavior

When a subagent is disabled:

- It is not included in the parent agent’s model-visible tool definitions.
- Its description is not included in the model context.
- The parent does not attempt to delegate work to it.
- A stale or manually constructed call is rejected at execution time.
- The resolved state is visible in traces and inspection metadata.

Conditional availability should ideally support:

- Build-time/static conditions
- `session.started`
- `turn.started`

Session-level resolution would probably cover most use cases, while turn-level resolution would support capabilities that can change during a conversation.

## Security considerations

Conditional visibility should not be presented as the only authorization boundary.

Sensitive tools and external actions should still validate authorization during execution. Conditional subagent availability would primarily provide capability composition, reduce unnecessary model context, prevent predictable failed calls, and improve routing.

## Open Questions

1. Is conditional subagent availability already supported through an undocumented API?
2. Would an `enabled`/`available` resolver fit the existing `defineAgent` API?
3. Could this use `defineDynamic` to remain consistent with dynamic tools, skills, instructions, and models?
4. Should disabled subagents be omitted from `/eve/v1/info`, or included with a disabled status for observability?

### Alternatives considered

_No response_

ctgowrie:
### Description

Closes #696.

Allow a declared subagent to export `defineDynamic` from its own `agent.ts`. Session- and turn-scoped resolvers return a `defineAgent(...)` configuration to expose the child, or `null`/`undefined` to omit it from direct and Workflow delegation. Resolver errors and invalid definitions also omit the child.

eve always compiles the subagent filesystem manifest. When selected, it injects the returned runtime agent configuration into that compiled child, so resolvers can choose different descriptions, models, model options, compaction settings, reasoning, output schemas, and limits without changing the child’s instructions, tools, skills, connections, sandbox, or nested subagents. Stale calls are rejected before dispatch.

This also documents the authoring contract, preserves selections across durable workflow steps and deployment revisions, and adds deterministic coverage for runtime model switching and nil omission.

- `pnpm typecheck`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm test:unit`
- `pnpm test:integration`
- focused compiler, lifecycle, model-selection, workflow-runtime, and runtime-graph unit tests after rebasing onto latest `main`
- `EVE_E2E_MODEL=mock pnpm exec eve eval --strict dynamic-subagent` from `e2e/fixtures/agent-subagents`

### PR Checklist

### DIFF
--- a/docs/guides/dynamic-capabilities.md
+++ b/docs/guides/dynamic-capabilities.md
@@ -51,10 +51,10 @@ compiled manifest. Each resolution can return a different model or other
 runtime agent settings. Runtime-selected models must use string model IDs;
 build and Workflow-world configuration cannot be selected at runtime.
 
-Dynamic subagents support `session.started` and `turn.started`, not
-`step.started`. A turn selection shadows the session selection for that turn,
-including when the turn handler returns nil. If a resolver throws or returns an
-invalid definition, eve logs the failure and omits the subagent.
+Dynamic subagents support `session.started` and `turn.started`. A turn selection
+shadows the session selection for that turn, including when the turn handler
+returns nil. If a resolver throws or returns an invalid definition, eve logs the
+failure and omits the subagent.
 
 The resolved set applies to direct delegation and the `Workflow` tool. eve
 also checks availability again before starting the child, so a stale or
--- a/docs/guides/dynamic-capabilities.md
+++ b/docs/guides/dynamic-capabilities.md
@@ -1,9 +1,9 @@
 ---
 title: "Dynamic Capabilities"
-description: "Resolve models, tools, skills, and instructions at runtime: dynamic model selection, defineDynamic resolver events, execution order, and durable dynamic tools."
+description: "Resolve models, subagents, tools, skills, and instructions at runtime with defineDynamic resolver events."
 ---
 
-`defineDynamic` resolves the model, tools, skills, and instructions at runtime from a session event instead of declaring them up front. Reach for it when the right capability isn't known until the session starts, because it hinges on who the caller is, what tenant they belong to, feature flags, or external data. The [tools](../tools), [skills](../skills), and [instructions](../instructions) guides each point here for their dynamic form.
+`defineDynamic` resolves the model, subagents, tools, skills, and instructions at runtime from a session event instead of declaring them up front. Reach for it when the right capability isn't known until the session starts, because it hinges on who the caller is, what tenant they belong to, feature flags, or external data. The [subagents](../subagents), [tools](../tools), [skills](../skills), and [instructions](../instructions) guides each point here for their dynamic form.
 
 ## Dynamic models
 
@@ -16,10 +16,49 @@ uncached prices. See
 [agent configuration](../agent-config#choose-the-model-dynamically) for the
 full contract.
 
-`fallback` is model-only: the agent always needs exactly one model, and the
-compiled fallback anchors build-time metadata. Tools, skills, and instructions
+The agent always needs exactly one model, so the compiled fallback anchors
+build-time metadata. A dynamic subagent also uses `fallback` to compile its
+child configuration, as described below. Tools, skills, and instructions
 default by authoring a static entry (or returning `null`), so `fallback` on
 their `defineDynamic` export is a build error.
+
+## Dynamic subagents
+
+Wrap a declared subagent's own `agent.ts` in `defineDynamic` when its
+availability depends on the caller, tenant, environment, or a feature flag.
+Keep the child definition in one constant, pass it as `fallback`, and return
+that same value to expose the subagent. Return `null` or `undefined` to omit it
+from the parent's model-visible tools.
+
+```ts title="agent/subagents/finance/agent.ts"
+import { defineAgent, defineDynamic } from "eve";
+
+const finance = defineAgent({
+  description: "Analyze financial and accounting data.",
+  model: "openai/gpt-5.5",
+});
+
+export default defineDynamic({
+  fallback: finance,
+  events: {
+    "session.started": (_event, ctx) =>
+      ctx.session.auth.current?.attributes.plan === "enterprise" ? finance : null,
+  },
+});
+```
+
+`fallback` is the definition eve compiles and runs when the subagent is
+selected; it does not make the subagent available by itself. A handler must
+return the exact fallback value. Dynamic subagents support `session.started`
+and `turn.started`, not `step.started`. A turn selection shadows the session
+selection for that turn, including when the turn handler returns nil.
+
+The resolved set applies to direct delegation and the `Workflow` tool. eve
+also checks availability again before starting the child, so a stale or
+manually constructed call fails with `SUBAGENT_UNAVAILABLE`. Treat conditional
+availability as capability composition, not as the only authorization
+boundary: sensitive child tools still need their own authorization and
+approval checks.
 
 ## Dynamic tools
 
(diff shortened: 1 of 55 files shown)


docs/guides/dynamic-capabilities.md
```diff
@@ -16,10 +16,52 @@ uncached prices. See
...
+
+eve always compiles the subagent's filesystem manifest, including its
+instructions, tools, skills, connections, sandbox, and nested subagents. When
+the resolver selects it, eve injects the returned agent configuration into that
+compiled manifest. Each resolution can return a different model or other
+runtime agent settings. Runtime-selected models must use string model IDs;
+build and Workflow-world configuration cannot be selected at runtime.
+
+Dynamic subagents support `session.started` and `turn.started`, not
+`step.started`. A turn selection shadows the session selection for that turn,
```
AndrewBarba: array of allowed events doesn't permit step.started

packages/eve/src/context/dynamic-subagent-lifecycle.ts
```diff
@@ -0,0 +1,157 @@
...
+  const matching = input.resolvers.filter((resolver) =>
+    resolver.eventNames.includes("session.started"),
+  );
+  const selections = await resolveSelections({ ...input, resolvers: matching });
+  input.ctx.set(SessionDynamicSubagentSelectionsKey, selections);
+  input.ctx.set(SessionDynamicSubagentRuntimeRevisionKey, input.runtimeRevision);
+}
+
+export function buildDynamicSubagentTools(input: {
+  get<T>(key: import("#context/key.js").ContextKey<T>): T | undefined;
```
AndrewBarba: nit: inline type imports can be top of file

packages/eve/src/context/keys.ts
```diff
@@ -166,6 +167,27 @@ export const LiveStepToolsKey = new ContextKey<
   import("#harness/execute-tool.js").HarnessToolDefinition[]
 >("eve.liveStepTools");
 
+export type DurableDynamicSubagentSelection = {
+  readonly agentConfig: import("#runtime/subagents/dynamic-agent-config.js").DynamicSubagentAgentConfig;
```
AndrewBarba: nit: weird mix of inline type import and top of file import. I prefer top of file

packages/eve/src/runtime/agent/resolve-model.ts
```diff
@@ -49,6 +50,10 @@ export async function resolveRuntimeModelReference(
     return bootstrapModel;
   }
 
+  if (reference.source === undefined && reference.id.startsWith("eve-mock/")) {
```
AndrewBarba: Could we keep `eve-mock/*` behind the existing test-only flag? As written, a dynamic model can select this in production and silently use the mock model instead of a real provider.

packages/eve/src/execution/workflow-steps.ts
```diff
@@ -382,10 +407,10 @@ export async function turnStep(rawInput: TurnStepInput): Promise<DurableStepResu
       ): Promise<StepResult> => {
         const refreshedSession = refreshSessionFromTurnAgent({
           compactionOverrides: {
-            thresholdPercent: bundle.resolvedAgent.config.compaction?.thresholdPercent,
+            thresholdPercent: effectiveAgent.thresholdPercent,
           },
           session: lifecycleSession,
-          turnAgent: bundle.turnAgent,
+          turnAgent: effectiveAgent.turnAgent,
```
AndrewBarba, line 417: Could we use this same effective agent when building the execution step below? The model call uses the selected model, but `session.started` still reports the placeholder model from the compiled manifest.

packages/eve/src/compiler/manifest.ts
```diff
@@ -247,15 +248,12 @@ export interface CompiledHookDefinition extends ModuleSourceRef {
...
-/**
- * Flattened compiled subagent node emitted by the compiler. `name` and
- * `description` are copied from `agent.config` for fast registry lookup.
- */
 export type CompiledSubagentNode = Readonly<
   ModuleSourceRef &
     Node & {
       agent: CompiledAgentNodeManifest;
-      description: string;
+      description?: string;
```
AndrewBarba: Can we keep `description` required here? The parent model needs it to know when it should call the subagent. For dynamic subagents, could we require the description up front and only resolve availability and model settings at runtime?

docs/guides/dynamic-capabilities.md
ctgowrie: That is what these docs say, but will clarify

packages/eve/src/runtime/agent/resolve-model.ts
ctgowrie: Good call. Will do

packages/eve/src/compiler/manifest.ts
ctgowrie: Yes. Improving types to make this more clearn

packages/eve/src/execution/workflow-steps.ts
ctgowrie: On it

AndrewBarba (approved):

PR #1492, opened 2026-07-31
docs(eve): additive delivery plan for subagents as tasks

ruiconti:
## Summary

- adds an implementation-sequencing plan for the [subagents-as-tasks design](https://github.com/vercel/eve/pull/1482), structured as strictly additive stages behind `experimental.tasks`
- takes the agent-messaging stack (`am/01-contracts` … `am/04-surface`) as the addressing baseline: agent handles, continuation dispatch, and the `TurnCaller` envelope satisfy the design doc's A2A dependency, so tasks composes with handles instead of introducing a second session-addressing mechanism
- stages 0–3 land inert (flag plumbing, task foundation, gated tools, delegated execution mode); stage 4 carries the six flows over the task wire and lets the flag select the mode; stage 5 normalizes and retires the flag

Also records the gaps to resolve in flight: `task_send` vs `AGENT_BUSY`, remote children outliving the parent, blocked-vs-computing visibility before the flag ships, and the MCP failure-taxonomy decision.

Companion to #1482. Part of #1084.

### DIFF
--- a/research/subagents-as-tasks-implementation.md
+++ b/research/subagents-as-tasks-implementation.md
@@ -1,0 +1,197 @@
+---
+issue: https://github.com/vercel/eve/issues/1084
+status: draft
+last_updated: "2026-07-31"
+---
+
+# Subagents as tasks: additive delivery plan
+
+## Summary
+
+This plan sequences the implementation of the [subagents-as-tasks design] into strictly additive
+stages behind `experimental.tasks`. Every stage must land with this invariant: **with the flag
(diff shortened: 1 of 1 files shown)


ruiconti:
Folded into #1482 — the plan doc is meaningless without the design doc it references, and both were unmerged single-doc branches off `main`, so splitting them only forced a review ordering that git wasn't enforcing (this branch was diverged from #1482, not stacked on it). `research/subagents-as-tasks-implementation.md` now rides on `rui/subagents-as-tasks` with the design-doc link pointing at the in-tree file. Closing.

PR #1495, opened 2026-07-31
fix(eve): prevent Vercel env pull setup hangs

OwenKephart:
## Summary
- run `vercel env pull` with stdin detached, including interactive setup flows
- abort a Vercel CLI child that remains alive after the environment files are written
- cover the bounded env-pull invocation in unit tests

## Root cause
Vercel CLI runs an optional Claude Code Vercel-plugin installer after a successful `env` command. Its confirmation helper checked only whether stdin was a TTY, not Vercel CLI's `--non-interactive` state. Therefore `vercel env pull --yes --non-interactive` could successfully write `.env.local` and then wait at the plugin-install confirmation when stdin remained inherited from eve.

The Vercel CLI fix is in vercel/vercel#17345. This eve change is still required to detach stdin for automation and bound a misbehaving child. With stdin ignored, the problematic CLI prompt declines immediately; the 30-second timeout remains a safety net for any future stuck subprocess.

## Validation
- `pnpm --filter eve build:compiled`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/setup/run-vercel-link.test.ts src/setup/boxes/deploy-project.test.ts src/setup/primitives/run-vercel.test.ts`
- `pnpm typecheck --filter eve`
- `pnpm lint`
- `pnpm fmt`

_Note: the sandbox uses Node 22.22.2 while the repository requires Node >=24; commands completed with engine warnings._

### DIFF
--- a/.changeset/bright-walls-sip.md
+++ b/.changeset/bright-walls-sip.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Prevent interactive setup from hanging after Vercel successfully pulls project environment variables. Environment pulls now run without terminal input and time out safely if the Vercel CLI does not exit.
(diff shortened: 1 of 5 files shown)


OwenKephart:
Confirmed root cause in Vercel CLI: after a successful `env` command, the optional Claude Code plugin installer prompts based only on `stdin.isTTY`, ignoring `--non-interactive`. The upstream fix is vercel/vercel#17345. This PR detaches stdin so the existing CLI version also declines that prompt immediately, plus retains a deadline as a fallback.

PR #1501, opened 2026-08-01
feat(eve): support dynamic remote subagents

ctgowrie:
Stacked on #1485.

## Summary

- allow dynamic subagent resolvers to return `defineRemoteAgent(...)`
- expose either the selected local or remote delegation at session and turn scope
- keep remote auth and headers out of durable workflow state and rehydrate them for outbound requests
- document the authoring contract and cover dynamic remote dispatch with unit and e2e fixtures

## Testing

- `pnpm test:unit`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`

### DIFF
--- a/packages/eve/src/internal/workflow-bundle/dynamic-remote-agent-transform.ts
+++ b/packages/eve/src/internal/workflow-bundle/dynamic-remote-agent-transform.ts
@@ -2,11 +2,11 @@ import { parseWithNitroRolldownAst } from "#internal/bundler/nitro-rolldown.js";
 import type { DynamicToolAstNode as AstNode } from "#internal/workflow-bundle/dynamic-tool-ast-references.js";
 
 interface CredentialsFactoryInfo {
-  readonly authSource?: string;
+  readonly authPropertySource?: string;
   readonly callEnd: number;
   readonly callSource: string;
   readonly callStart: number;
-  readonly headersSource?: string;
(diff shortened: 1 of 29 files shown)


AndrewBarba (approved):

packages/eve/src/internal/workflow-bundle/dynamic-remote-agent-transform.ts
```diff
@@ -0,0 +1,152 @@
...
+  };
+}
+
+function sliceNode(source: string, node: AstNode | undefined): string | undefined {
+  return node?.start === undefined || node.end === undefined
+    ? undefined
+    : source.slice(node.start, node.end);
+}
+
+function findProperty(object: AstNode, name: string): AstNode | undefined {
```
ruiconti: I think a bunch of these utilities can be re-imported from defineDynamic impl?

ctgowrie: Addressed in 619efccd. I moved the shared property lookup and recursive AST walker into `dynamic-tool-ast-references.ts`; both the existing `defineDynamic` transform and the remote-agent transform now import them.

ctgowrie: Good call. Reusing

PR #1503, opened 2026-08-01
fix(build): co-copy rollup content-hashed sibling .d.ts chunks (#1500)

### LINKED ISSUE #1500: Published package omits two declaration chunks its own .d.ts files import (chat/messages-BSoJG691, @chat-adapter/twilio/types-WYjTBVDi)

N/A

### Steps to reproduce

````
```sh
npm pack eve@0.28.0
tar xzf eve-0.28.0.tgz

# The vendored Chat SDK core declares an import for a sibling chunk:
grep -o "from '\./messages-[^']*'" package/dist/src/compiled/chat/index.d.ts
# → from './messages-BSoJG691.js'

# The Twilio chat adapter does the same:
grep -o "from '\./types-[^']*'" package/dist/src/compiled/@chat-adapter/twilio/webhook.d.ts
# → from './types-WYjTBVDi.js'

# Neither file is in the tarball:
ls package/dist/src/compiled/chat/
# _mdast.d.ts  _workflow-serde.d.ts  index.d.ts  index.js  jsx-runtime-_JEEAotp.d.ts  …
# (no messages-BSoJG691.d.ts)

ls package/dist/src/compiled/@chat-adapter/twilio/
# (no types-WYjTBVDi.d.ts)
```

Note that sibling hashed chunks *are* published — `jsx-runtime-_JEEAotp.d.ts` sits in the same directory — so this is specific to these two chunks rather than to hashed chunks generally.
````

````
**Current:** `dist/src/compiled/chat/index.d.ts` and `dist/src/compiled/@chat-adapter/twilio/webhook.d.ts` import declaration files that the published package does not contain. Because `skipLibCheck` defaults to `true`, TypeScript resolves the unresolvable imports to `any` rather than erroring, so every type re-exported through those chunks silently degrades. With `skipLibCheck: false` it surfaces as `TS2307: Cannot find module './messages-BSoJG691.js'`.

Roughly 120 exported names are affected across the Chat SDK surface — including `Adapter`, `Chat`, `ChatInstance`, `Message`, `Thread`, `Attachment`, `FileUpload`, `PostableCard`, `PostableMessage`, `StateAdapter`, `UserInfo` — plus the Twilio adapter's `TwilioWebhookUrl` and `TwilioVerifiedRequest`.

Confirmed through the public entry points, not only the vendored internals. Placing `@ts-expect-error` on an obviously-wrong assignment reports the directive as *unused*, which is the signature of a type that has become `any`:

```ts
import type { Attachment } from 'eve/channels/slack'
// @ts-expect-error — reported UNUSED, i.e. `Attachment` is `any`
const a: Attachment = 42
```

**Expected:** the two chunks are included in the published package, so the shipped declarations resolve and the exported types are real.
````

### eve version

Reproduced on 0.26.1, 0.27.6, 0.27.12 and 0.28.0 — not a recent regression. Also present in 0.29.3 at time of writing.

### Environment

```shell
- OS: macOS 26.5.2 (arm64)
- Node.js: v26.5.0
- Package manager: bun 1.3.14

`eve info` is not included: the defect is in the published tarball's declaration files, and the reproduction above is a pure `npm pack` inspection that requires no app. Happy to add it if useful.
```

### Where does the bug occur?

Build (`eve build`)

### Deployment

_No response_

### Build and runtime logs

```shell
N/A — reproducible from the published package alone, no deployment involved.
```

### Additional context

The real declarations exist in the upstream packages eve vendors, under the exact filenames eve's own declarations import:

- `chat@4.34.0` → `dist/messages-BSoJG691.d.ts`
- `@chat-adapter/twilio@4.35.0` → `dist/types-WYjTBVDi.d.ts`

The filenames are rollup content hashes, so a matching name implies matching content.

This is easy to miss because `skipLibCheck: true` is the common default — nothing fails, the affected types just quietly become `any`. We found it only when an obviously-invalid value type-checked without complaint.

As a workaround we vendor those two files and restore them via a `postinstall`, which we would like to delete once this is fixed.

iroiro147:
## Summary

Fixes #1500 — published `eve` package was silently missing two rollup content-hashed sibling declaration chunks that its own vendored `.d.ts` files import:

- `chat`'s `messages-BSoJG691.d.ts` (referenced from `chat/index.d.ts`)
- `@chat-adapter/twilio`'s `types-WYjTBVDi.d.ts` (referenced from `@chat-adapter/twilio/webhook.d.ts`)

The type hole degraded ~120 exported names across the Chat SDK surface (`Adapter`, `Chat`, `ChatInstance`, `Message`, `Thread`, `Attachment`, `FileUpload`, `PostableCard`, `PostableMessage`, `StateAdapter`, `UserInfo`, …) plus `TwilioWebhookUrl` / `TwilioVerifiedRequest` to `any` under the default `skipLibCheck: true`, and surfaced as `TS2307: Cannot find module './messages-BSoJG691.js'` under `skipLibCheck: false`.

## Root cause

`createDeclarationCopier` (`packages/eve/scripts/vendor-compiled/_shared.mjs`) had two ways to bring upstream `.d.ts` files into the vendored output:

1. **Explicit `files:`** — hardcoded list (twilio listed only its 5 entry points).
2. **`discoverExtraFiles`** — a per-package filename matcher (chat listed only `jsx-runtime-<hash>.d.ts`).

Meanwhile `collectExternalDeclarationImports` deliberately skips *relative* specifiers (`./…`), so nothing in the pipeline noticed when a vendored declaration referenced a sibling content-hashed chunk. The result: both hashed chunks were left behind.

## Fix

Add a transitive relative-import co-pass to `createDeclarationCopier`: after the existing files/extra-files copies, scan each written declaration for relative import specifiers (`from './x.js'`, `from '../y.d.ts'`, `export * from './z.js'`, side-effect `import './w.js'`), resolve them against the upstream package's `dist/`, and co-copy any matches — recursively, with cycle protection.

This **subsumes** `discoverExtraFiles` for the sibling-chunk class of problem: chat.mjs's `jsx-runtime-` regex remains in place as belt-and-braces but the new helper catches its matches too. The helper is generic across all vendored packages, so future content-hashed chunks in *any* vendored dependency are picked up automatically — no more per-package regexes.

## Verification

Locally (`pnpm build:compiled`):

```text
ls .generated/compiled/chat/
# LICENSE  _mdast.d.ts  _workflow-serde.d.ts  index.d.ts  index.js
# jsx-runtime-_JEEAotp.d.ts  messages-BSoJG691.d.ts       ← NEW

ls .generated/compiled/@chat-adapter/twilio/
# LICENSE  api.d.ts  api.js  format.d.ts  format.js  index.d.ts  index.js
# types-WYjTBVDi.d.ts  voice.d.ts  voice.js  webhook.d.ts  webhook.js  ← NEW
```

Both chunk filenames and contents match the upstream packages the reporter linked:
- `chat@4.34.0` → `dist/messages-BSoJG691.d.ts` ✓
- `@chat-adapter/twilio@4.35.0` → `dist/types-WYjTBVDi.d.ts` ✓

Test suite: `pnpm exec vitest run test/scenarios/compiled-vendor-assets.scenario.test.ts` — **8/8 pass** (7 existing + 1 new regression test).

Type check on the changed script + test: **clean** (pre-existing baseline of 124 errors in 8 unrelated files unchanged).

## Regression test

Added `(eve#1500) co-copies content-hashed sibling declaration chunks referenced by vendored entries` to `compiled-vendor-assets.scenario.test.ts`:

- Asserts both chunk files exist on disk.
- Asserts `chat/index.d.ts` still references `./messages-BSoJG691` and `webhook.d.ts` still references `./types-WYjTBVDi` (i.e. we copied the target alongside, not stripped the import).

## Why this shape and not the cheap fix

Alternative considered: just extend the regexes to include `messages-*.d.ts` and `types-*.d.ts`. Rejected because (a) rollup hashes change per version bump and would silently re-break on the next vendored dependency upgrade, and (b) the reactive-regex pattern would have caught this bug only for the two currently-known chunks. The transitive closure approach is immune to hash churn and automatically picks up new chunks in any vendored package.

### DIFF
--- a/packages/eve/scripts/vendor-compiled/_shared.mjs
+++ b/packages/eve/scripts/vendor-compiled/_shared.mjs
@@ -188,7 +188,112 @@ export function createDeclarationCopier({
         }),
       );
     }
+
+    // (eve#1500) Some upstream packages emit rollup content-hashed sibling
+    // chunks that their public `.d.ts` files reference via relative import
+    // (`from './messages-BSoJG691.js'`, `from './types-WYjTBVDi.js'`).
+    // `collectExternalDeclarationImports` intentionally skips relative
+    // specifiers, so without further handling those chunks would be left
+    // behind — the published declarations then resolve to `any` (or fail
+    // `skipLibCheck: false`). Scan each written declaration for relative
+    // import specifiers, resolve them against the upstream `dist/`, and
+    // co-copy any matches, recursively, with cycle protection. This
+    // subsumes `discoverExtraFiles` for the sibling-chunk class of problem.
+    const outputsAlreadyWritten = new Set(declarations.map((d) => d.output));
+    const extrasAlreadyWritten = new Set(
+      typeof discoverExtraFiles === "function" ? discoverExtraFiles(distEntries) : [],
+    );
+    const copiedChunks = new Set([...outputsAlreadyWritten, ...extrasAlreadyWritten]);
+    await coCopyRelativeDeclarationChunks(
+      declarations.map((d) => d.output),
+      destinationRoot,
+      distDir,
+      copiedChunks,
+    );
   };
+}
+
+/**
+ * Regex mirroring the relative side of `collectExternalDeclarationImports`:
+ * finds `from './x'`, `from '.../x'`, and side-effect `import './x'` specifiers.
+ */
+const RELATIVE_DECLARATION_IMPORT_PATTERN =
+  /^(?:import|export)\s+(?:type\s+)?(?:\{[^}]*\}|\*\s+as\s+[A-Za-z_$][\w$]*|\*)?\s*(?:from\s+)?['"](\.[^'"]+)['"]/gm;
+
+/**
+ * Walks a set of already-written declaration outputs, finds relative import
+ * specifiers in their source (looking at the post-rewrite output), and
+ * co-copies any matching upstream `.d.ts` chunk the package embeds. Recurses
+ * into newly-copied chunks so transitively-referenced chunks survive.
+ *
+ * Resolves the specifier against the chunk's own directory (which matches
+ * the upstream layout because we copy files verbatim with their basenames).
+ * Normalizes the `.js` specifier to `.d.ts` — rollup emits `.d.ts` files
+ * whose import clauses reference the `.js` companion.
+ */
+async function coCopyRelativeDeclarationChunks(
+  outputPathsRelativeToDestination,
+  destinationRoot,
+  distDir,
+  copiedChunks,
+) {
+  const queue = [...outputPathsRelativeToDestination];
+  const processed = new Set();
+  while (queue.length > 0) {
+    const relativeOutput = queue.pop();
+    if (processed.has(relativeOutput)) continue;
+    processed.add(relativeOutput);
+
+    const outputPath = join(destinationRoot, relativeOutput);
+    let sourceText;
+    try {
+      sourceText = await readFile(outputPath, "utf8");
+    } catch {
+      // The just-written declaration may not exist yet if a caller raced us;
+      // or the file may be a stub written via a different pathway. Skip.
+      continue;
+    }
+    const fromDir = dirname(relativeOutput);
+
+    for (const match of sourceText.matchAll(RELATIVE_DECLARATION_IMPORT_PATTERN)) {
+      let specifier = match[1];
+      // Skip anchor-only relative specifiers (`./foo.d.ts#Anchor`).
+      const hashIndex = specifier.indexOf("#");
+      if (hashIndex !== -1) specifier = specifier.slice(0, hashIndex);
+      if (specifier.length === 0) continue;
+
+      // Rollup outputs reference the `.js` companion in their `.d.ts`.
+      // Try the `.d.ts` twin first, then the specifier verbatim (in case
+      // the upstream file genuinely ends `.js`, e.g. a paired JS file).
+      const candidateNames = specifier.endsWith(".js")
+        ? [specifier.slice(0, -".js".length) + ".d.ts", specifier]
+        : specifier.endsWith(".d.ts")
+          ? [specifier, specifier.slice(0, -".d.ts".length) + ".d.ts"]
+          : [specifier + ".d.ts"];
+
+      for (const candidateName of candidateNames) {
+        const resolvedOutputRelative = posix.normalize(posix.join(fromDir, candidateName));
+        if (copiedChunks.has(resolvedOutputRelative)) {
+          queue.push(resolvedOutputRelative);
+          break;
+        }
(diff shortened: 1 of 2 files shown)


iroiro147:
FYI on the current CI failures: the `typecheck` and downstream E2E failures on this PR come from **5 pre-existing TypeScript errors on `main`** — not from the vendoring change in this PR.

Failing files (all in unrelated subsystems, all on `main`):

```
src/public/channels/chat-sdk/chatSdkChannel.ts(273,9): error TS2322
  Type 'Thread<unknown, unknown>' is not assignable to type
  'string | SerializedThread | Thread<Record<string, unknown>, unknown>'.

src/public/channels/photon/inboundContent.test.ts(9,5): error TS2741
src/public/channels/photon/photonIMessageChannel.test.ts(45,7): error TS2741
src/public/channels/photon/photonIMessageChannel.test.ts(68,7): error TS2741
src/public/channels/photon/photonIMessageChannel.test.ts(90,7): error TS2741
  Property 'fullName' is missing in type '{…}' but required in type 'Author'.
```

I verified by running `npx tsc --noEmit -p packages/eve` on a clean `origin/main` checkout — same 5 errors, same files, same line numbers. The E2E failures cascade from the same `pnpm run build:js` step hitting the typecheck error.

This PR itself only touches `scripts/vendor-compiled/_shared.mjs` and `test/scenarios/compiled-vendor-assets.scenario.test.ts`; my smoke test for the new vendoring path passed locally before push.

I'll keep the branch freshly synced with `main` and re-run CI the moment the upstream typecheck is fixed.

iroiro147:
Superseded by #1534.

This PR's recursive sibling-chunk walker is more general than the regex-broadened matchers in #1534, but it has a correctness gap: the co-copied chunks would land **verbatim** without the declaration-rewrite pass, which means chat's `messages-<hash>.d.ts` would still keep `import { Root } from 'mdast'` and `import { WORKFLOW_SERIALIZE, … } from '@workflow/serde'` — bare specifiers the published `eve` tarball cannot resolve (`_mdast.d.ts` / `_workflow-serde.d.ts` stubs are required instead). In other words, co-copying without rewriting re-creates the exact `skipLibCheck: false` failure mode this PR is trying to close, just on a different import.

#1534 fixes that by folding the discover-selected extras into the same `declarations` rewrite pass as the named entry files, applying the `mdast`/`@workflow/serde`/vendored rule matrix uniformly. It also broadens the chat matcher (only one regex change, covers both `jsx-runtime-<hash>` and `messages-<hash>`) and adds the analogous `types-<hash>` matcher to twilio, and cleans up the type-surface leaks the stricter chunk surfaces (chatSdkChannel cast + photon test mocks).

If the maintainers prefer the recursive walker shape from this PR as the long-term architecture, I'd rather rebase the rewrite pass onto this walker than keep both — but as-is, #1534 is the stronger candidate to land first. Closed in favor of #1534.

PR #1517, opened 2026-08-01
feat(eve): add the experimental.tasks flag, plumbing only

ruiconti:
First of five implementation PRs for the [subagents-as-tasks plan](https://github.com/vercel/eve/pull/1482), which this stacks on:

1. this — flag plumbing
2. #1518 — task foundation (durable task run, session index)
3. #1519 — flag-gated task tools
4. #1521 — delegated execution and the task wire
5. #1522 — `task_send`

## What

Adds `experimental.tasks` to the authored agent config. Nothing reads it: a flag-on agent compiles and runs identically to a flag-off one.

The field travels the same path as `subagentPersistentSessions`, field for field: authored `agent.ts` → authored-definition normalization → compiler copy → strict compiled-manifest schema → manifest clone → `ResolvedAgent.config.experimental`.

## Root-only enforcement

Subagent configs reject the flag at compile, same as `experimental.workflow`. That check's option was named `allowWorkflowConfig`; it now gates two root-only flags, so it's renamed `allowRootOnlyConfig`. Only rename outside the new field.

Part of #1084.

### DIFF
--- a/packages/eve/src/compiler/manifest.ts
+++ b/packages/eve/src/compiler/manifest.ts
@@ -429,6 +429,7 @@ const compiledAgentConfigSchema: z.ZodType<CompiledAgentDefinition> = z
     experimental: z
       .object({
         subagentPersistentSessions: z.boolean().optional(),
+        tasks: z.boolean().optional(),
         workflow: compiledAgentWorkflowDefinitionSchema.optional(),
       })
       .strict()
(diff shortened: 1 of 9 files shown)


ruiconti:
Superseded by the merge-safe v4 stack: inert final-shape kernel #1771; complete background runtime #1773.

PR #1518, opened 2026-08-01
feat(eve): add the durable task foundation, inert

ruiconti:
Part 2/5 of the `experimental.tasks` stack, on top of #1517. New `packages/eve/src/tasks/` modules plus the durable task run; nothing references them yet.

## What

- `TaskStatus`/`TaskView`/`TaskOutput` and a pure transition function: terminal is final, `working <-> input_required`, repeated cancel is a noop, and `failed` always carries its error as `TaskOutput.error` (failure is the state; the error output is its consequence — diverges from MCP, which reserves `failed` for protocol errors).
- The task run: a small stable-named workflow that is the single writer for one task's lifecycle.
- The `eve.tasks` session index and replay-idempotent task ids.

## Data flow

Task state never lives in session state. Writers and readers meet at the run:

- write: command → the run's private hook → transition function → full `TaskView` snapshot appended to the run's `eve.task` stream. Rejected commands (late completion after cancel) append nothing.
- read: `getRun(taskRunId).getReadable({ namespace, startIndex: -1 })` tail-read. Snapshots are trusted without re-validation because the run is the only writer.
- parent session: only `{ taskId, taskRunId, commandToken }` under `eve.tasks`. `commandToken` is a routing credential and never renders to the model.

The dedicated run comes from the #1190 spike: session state threads through step results, but callback routes and child executors must update task state without holding the current snapshot. The run also serializes competing completion/cancel/input transitions.

Task ids reuse the agent-handle operation-id derivation — `hash(parentSessionId, parentTurnId, callId)` — so replayed creation for the same call yields the same task. A duplicate run for the same task loses the hook claim and exits.

Part of #1084.

### DIFF
--- a/packages/eve/src/execution/tasks/run-control.ts
+++ b/packages/eve/src/execution/tasks/run-control.ts
@@ -1,0 +1,164 @@
+import {
+  EntityConflictError,
+  HookNotFoundError,
+  RunExpiredError,
+  WorkflowRunNotFoundError,
+} from "#compiled/@workflow/errors/index.js";
+
+import type { TaskRunWorkflowInput } from "#execution/tasks/run-workflow.js";
+import {
+  startWorkflowPreferLatest,
+  taskRunWorkflowReference,
+} from "#execution/workflow-runtime.js";
+import { getRun, resumeHook } from "#internal/workflow/runtime.js";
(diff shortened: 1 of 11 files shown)


ruiconti:
Superseded by #1771, which carries the final task/index/transport shape before any writer exists.

PR #1519, opened 2026-08-01
feat(eve): add flag-gated task tools, undiscoverable by default

ruiconti:
Part 3/5 of the `experimental.tasks` stack, on top of #1518. The model-facing controls; nothing dispatches them until part 4.

## What

With the flag on, the root node's harness tool map gains three tools (`task_peek`, `task_cancel`, `task_sleep`; `task_send` is part 5). Without it, the tool map, prompt, and event stream are unchanged (existing tests pass unmodified).

## How each executes

- `task_peek` / `task_cancel` are execute-less runtime actions with a new `task-control` kind. Before: only subagent calls carried `runtimeAction` and parked into the pending batch. Now these calls park the same way, and the dispatch step — which holds durable session state and world access that tool `execute` doesn't — runs them (part 4).
- `task_sleep` executes in-loop: it records the existing durable turn sleep and returns, same as the public `sleep` tool.

## Gating

Injection mirrors the implicit `agent` tool: inline definitions in `createNodeHarnessTools`, authored tools with the same name shadow, `disableTool(name)` removes, and advertised-tools hides them from any child-shaped session — root-node self-delegated children share the root's config, so session shape is the only signal separating them. Dynamic workflow interrupts reject `task-control` tools explicitly.

`task_send` is part 5; TUI presentation copy for the new names rides along.

Part of #1084.


Rchie121 (approved):

Rchie121 (approved):

Rchie121 (approved):

ruiconti:
Superseded by #1773. The flag and final model tools now land with their complete runtime rather than as a provisional public layer.

PR #1521, opened 2026-08-01
feat(eve): run subagents as background tasks behind experimental.tasks

ruiconti:
Part 4/5 of the `experimental.tasks` stack, on top of #1519. The flag now means something: every subagent dispatch (fresh start, agentId continuation, local and remote) returns a task receipt and the parent turn keeps moving.

## Data flows, before → after (flag on)

| Flow | Before | After |
| --- | --- | --- |
| Terminal child result | child resumes the parent **turn inbox** token; the turn blocks in `waitForRuntimeActionResults` until every batch result arrives | child's reply token is the **task run's** private hook; the originating call resolves immediately with `{ taskId, status: "working" }` |
| Child HITL / authorization | adapter forwards to the parent turn inbox | same payloads land on the task run → `input_required` snapshot (authorization blocks, `authorization.completed` unblocks) |
| Ready notification | n/a — parent was always waiting | task run resumes the parent **session** hook with a deliver payload; a parked parent starts a turn, an active turn sees it at the next boundary. Terminal and `input_required` are the only wake triggers |
| Agent handle | settled when the batch resolves with the child's outcome | receipt carries a `parked` outcome, so the unchanged resolve path parks the handle at receipt; the address stays for follow-ups |

The child wire itself is untouched — local notification, HITL forwarding, and the remote callback route all deliver to whatever token they were given. Delegation is a token-routing change, not a payload change.

## Ordering

The task run is created before the child dispatch side effect, so a fast child always finds a live command hook. The command token is derived deterministically from `(taskId, parent continuation token)`: a durable replay re-derives it and the duplicate run loses the hook claim.

## Tool execution

- `task_peek`/`task_cancel` run inside the dispatch step. Cancel commits `cancelled` on the run first, then propagates a cooperative abort through the handle address — a late child result can't revive the task.
- Parent finalization cancels live tasks before terminating children.

## Scope

- Flag off: dispatch behavior, wire shapes, and events unchanged — the agent-messaging scenario suite passes unmodified.
- The flag implies conversation-mode children, so `tasks` and `subagentPersistentSessions` never produce a third mode.
- There is no waiting primitive: parents are woken by ready transitions (see the flow table), never by blocking calls.
- Deferred: child usage accounting for delegated tasks (`usageDelta` reaches the task run but isn't folded into parent totals), progress/`task_message`.
- `subagent-start.ts` is code motion out of the dispatch step (file-length cap).

Part of #1084.

### DIFF
--- a/packages/eve/src/execution/dispatch-runtime-actions-step.ts
+++ b/packages/eve/src/execution/dispatch-runtime-actions-step.ts
@@ -47,11 +47,19 @@ import {
 } from "#protocol/message.js";
 import type {
   RuntimeActionRequest,
+  RuntimeActionResult,
   RuntimeRemoteAgentCallActionRequest,
   RuntimeSubagentCallActionRequest,
   RuntimeSubagentDispatchFailure,
-  RuntimeSubagentResult,
+  RuntimeToolCallActionRequest,
 } from "#runtime/actions/types.js";
+import {
+  beginDelegatedTask,
+  executeTaskControlAction,
+  failDelegatedDispatch,
+  isTaskControlAction,
+  settleDelegatedDispatch,
+} from "#execution/tasks/dispatch.js";
 import {
   createDurableSessionState,
   type DurableSessionState,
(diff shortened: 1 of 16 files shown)


Rchie121 (approved):

ruiconti:
Superseded by #1773. This GitHub-dirty intermediate layer is no longer part of the stack.

PR #1522, opened 2026-08-01
feat(eve): add task_send for blocked and finished background tasks

ruiconti:
Part 5/5 of the `experimental.tasks` stack, on top of #1521. Without this, a background child that parks on input is unreachable and a finished conversation can't be continued.

## Routing by task status

- `working` → `AGENT_BUSY` tool error. No queueing: matches handle-continuation semantics, and queueing on the task run is the reversible follow-up if busy errors prove noisy.
- `input_required` + `inputResponses` → delivered to the parked child session with the task run's token as the caller reply target, so the child's next settled turn reports to the same task. The task returns to `working`.
- terminal + `message` → the full delegated continuation from part 4, through the agent handle: new task on the same child session, new receipt back. The old task stays terminal.

Ownership is checked against the `eve.tasks` index; the child address comes from the agent handle. Input responses are local-only — remote children can't forward input requests in the first place, so a remote `input_required` never exists in this stack.

## One seam worth review

`task_send`'s own result is a `tool-result`, and the resolve path only settles handles from `subagent-result` outcomes — so the follow-up continuation parks its handle in-line instead of riding the receipt.

The task-control executors split into `execution/tasks/{delegate,control-shared,send}` (file-length cap, and it breaks an import cycle between the dispatch router and the send path). Includes the release changeset for the feature.

Part of #1084.

### DIFF
--- a/.changeset/tasks-experimental-subagents.md
+++ b/.changeset/tasks-experimental-subagents.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add experimental background tasks for subagents. With `experimental.tasks` on the root agent, subagent calls return a task receipt immediately instead of blocking the turn, and the model manages the delegated work with the new `task_peek`, `task_cancel`, `task_send`, and `task_sleep` tools. Terminal results and input requests wake the parent through the normal session delivery path. Without the flag, nothing changes.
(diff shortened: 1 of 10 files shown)


ruiconti:
Superseded by #1773, which introduces only the final message-only task_send and direct HITL contract.

PR #1533, opened 2026-08-02
feat(sandbox): add edit_file (str_replace) tool for partial edits — scaffold

### LINKED ISSUE #1523: Sandbox file tools: add a built-in `edit_file` (str_replace) tool so agents stop regenerating whole files
## What problem are you trying to solve?

The sandbox file toolset currently exposes `read_file`, `write_file`, `glob`, and `grep` (see #1459 for the current tool surface). There is no partial-edit primitive, so any change to an existing file forces the model to regenerate the **entire** file through `write_file`.

On large files this is a major latency and cost driver. A production run of a coding agent on `eve@0.27.6` (Vercel Agent Runs trace; private task data, so no run link) shows the pattern:

- A ~13.8k-character instructions file was fully regenerated **twice** via `write_file`, plus a ~13.2k-character test file once.
- Those three writes alone account for **~215 s of pure LLM output generation** in a ~19-minute run.
- Each full rewrite also re-enters the context as a complete file body, inflating input tokens for every subsequent step (the run averaged ~100k input tokens per step across ~40 tool calls).

The edits themselves were small relative to the file size — a str_replace-style tool would have reduced each of them to a few hundred output tokens.

Checked `eve` up to `0.29.4`: no built-in edit tool exists. The documented escape hatch is an authored tool via `defineTool` + `ctx.getSandbox()`, which works, but every agent project has to reimplement the same well-understood primitive (and its guidance prompt) independently.

## Current behavior

- Editing an existing sandbox file requires `read_file` followed by `write_file` with the complete new content.
- The framework already tracks read-before-write evidence for `write_file`, so the harness has the right foundation — but the only write primitive is whole-file replacement.
- Output tokens, wall-clock time, and context growth all scale with file size rather than edit size.

## Proposed behavior

Add a built-in `edit_file` tool alongside the existing sandbox file tools, with the widely used str_replace semantics:

- Input: `path`, `old_string`, `new_string`.
- The tool reads the file via the sandbox, requires `old_string` to occur **exactly once**, and replaces it.
- Zero matches or multiple matches fail with a structured error **without writing**, instructing the model to expand the match context (the established contract from Anthropic's text-editor tool and most coding harnesses, so models are already heavily trained on it).
- Reuse the existing read-before-write tracking: editing a file the model has not read in its current state should fail the same way `write_file` does.
- Optional but valuable: `replace_all: boolean` for intentional multi-site renames.

Built-in tool guidance should steer edits of existing files to `edit_file`, keeping `write_file` for new files or full rewrites — mirroring how Claude Code, OpenAI Codex, and other harnesses split the two.

## Bounded regression tests

| Case | Setup | Required result |
| --- | --- | --- |
| Unique match | `old_string` occurs exactly once | File updated; only the replaced span changes |
| No match | `old_string` absent | Structured error, no write, hint to re-read the file |
| Ambiguous match | `old_string` occurs twice | Structured error, no write, hint to expand context |
| Read-before-edit | File never read (or changed since last read) | Same failure contract as `write_file` |
| `$HOME` paths | Path starts with `$HOME` | Resolved like the other file tools after #1459 |

## Expected behavior

Small edits to large files cost output tokens proportional to the edit, not the file. Whole-file `write_file` remains available for file creation and full replacement.

## Version

- Observed on: `eve@0.27.6`
- Checked through: `eve@0.29.4` (no built-in edit tool)

iroiro147:
## What

Scaffold PR for #1523: adds a framework `edit_file` tool that performs a single unique substring replacement, avoiding the token cost of full-file `write_file` regeneration.

## Changes (139 LOC, 3 files)

| File | Change |
|------|--------|
| `runtime/framework-tools/edit-file.ts` (NEW, 129 LOC) | Zod input/output schemas, `executeEditFileOnSandbox` executor with unique-substring guarantee, `EDIT_FILE_TOOL_DEFINITION` definition |
| `runtime/framework-tools/index.ts` (+2) | Import + register EDIT_FILE_TOOL_DEFINITION alongside existing tools |
| `public/tools/defaults.ts` (+12) | Export `editFile` public tool surface matching readFile/writeFile pattern |

## Behavior

- `edit_file({filePath, oldString, newString})`
- Reads file → enforces oldString appears **exactly once** → replaces → writes back
- Throws descriptive error when oldString is missing, empty, or >1 occurrence
- Returns `{path, replacements: 1}`

## Design tradeoffs (intentionally deferred to maintainer review)

- **Atomicity**: read-then-write is one sandbox call (not two round trips). Rationale: agent already has file contents in context; the race window is bounded by turn-time.
- **Failure signaling**: descriptive errors ("not found", "ambiguous match") let the agent retry with more context rather than silent corruption. Mirrored from Claude Code's Edit tool behavior.
- **API surface**: only uses existing `SandboxSession` primitives — no new deps.

## Verification

- **Local typecheck: not run** (no node_modules in this workspace). `npx tsc --noEmit` in a dev checkout is the verification gate.
- Code review only. Maintainers: if design shape looks right, this is a clean scaffold to extend with tests/telemetry.

Closes #1523 (design-shaped; whether this is THE implementation or a starting point is maintainer call)

### DIFF
--- a/packages/eve/src/public/tools/defaults.ts
+++ b/packages/eve/src/public/tools/defaults.ts
@@ -4,6 +4,7 @@
  * `agent/tools/*.ts` files.
  */
 import { BASH_TOOL_DEFINITION } from "#runtime/framework-tools/bash.js";
+import { EDIT_FILE_TOOL_DEFINITION } from "#runtime/framework-tools/edit-file.js";
 import { GLOB_TOOL_DEFINITION } from "#runtime/framework-tools/glob.js";
 import { GREP_TOOL_DEFINITION } from "#runtime/framework-tools/grep.js";
 import { READ_FILE_TOOL_DEFINITION } from "#runtime/framework-tools/read-file.js";
@@ -46,6 +47,13 @@ export const readFile: ToolDefinition = toPublicToolDefinition(READ_FILE_TOOL_DE
  * Enforces read-before-write for existing files and stale-read detection.
  */
 export const writeFile: ToolDefinition = toPublicToolDefinition(WRITE_FILE_TOOL_DEFINITION);
+
+/**
+ * Framework-provided partial-edit tool (`edit_file`). Replaces one unique
+ * substring with a new one; fails if the substring is missing or ambiguous.
+ * Prefer `edit_file` over `write_file` for targeted changes on existing files.
+ */
+export const editFile: ToolDefinition = toPublicToolDefinition(EDIT_FILE_TOOL_DEFINITION);
 
 /**
  * Framework-provided HTTP fetch tool. Spread or wrap to customize.
--- a/packages/eve/src/runtime/framework-tools/edit-file.ts
+++ b/packages/eve/src/runtime/framework-tools/edit-file.ts
@@ -1,0 +1,129 @@
+import { z } from "#compiled/zod/index.js";
+
+import { requireSandboxSession } from "#execution/sandbox/require-sandbox.js";
+import { resolveAbsoluteFilePath } from "#execution/sandbox/require-sandbox.js";
+import type { ResolvedToolDefinition } from "#runtime/types.js";
+import type { ToolExecuteOptions } from "#shared/tool-definition.js";
+
+/**
+ * Typed input accepted by {@link executeEditFile}.
+ *
+ * Replaces exactly one occurrence of `old_string` with `new_string` in the
+ * target file. If `old_string` is not found (or is found more than once), the
+ * call fails with a descriptive error and the file is left untouched.
+ */
+export const EDIT_FILE_INPUT_SCHEMA = z.strictObject({
+  filePath: z
+    .string()
+    .describe("The absolute path to the file to edit. A leading $HOME is supported."),
+  oldString: z
+    .string()
+    .describe(
+      "The exact substring to replace. Must appear exactly once in the file. Use enough surrounding context to make the match unique.",
+    ),
+  newString: z
+    .string()
+    .describe("The replacement text. The file is rewritten as: before + newString + after."),
+});
+
+/**
+ * Structured result returned from {@link executeEditFile}.
+ */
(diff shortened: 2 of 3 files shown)


iroiro147:
Both VADE findings fixed in a8919ad5:

1. **read-file stamp refresh** — executeEditFileOnSandbox now refreshes the read-file stamp after a successful write, so a subsequent write_file on the same file no longer fails with "modified since it was last read". Mirrors the stamp-refresh pattern in executeWriteFileOnSandbox.

2. **$-pattern replacer** — String.replace(oldString, newString) was interpreting $$, $&, $`, $', $n in newString as special replacement patterns. Now passes a replacer function () => args.newString so the replacement text is inserted literally.

Typecheck still gated on CI (no local node_modules in this worktree).

iroiro147:
CI status note: typecheck + lint + test-integration(ubuntu) + test-scenario all green on a8919ad5. test-unit and test-integration(windows) fail identically to my other open eve PRs (#1528–#1532) — consistent with fork-scoped CI credentials, not the change (which is contained to packages/eve/src/runtime/framework-tools/). DCO signoff present. Ready for design/impl review when a maintainer gets a chance.

iroiro147:
Standing-health check (re-verified against current `main` `021dbbf2`, 2 commits past this branch's base):

- Merge dry-run vs current \`main\`: **clean, zero conflicts**. The 2 new commits touch \`cli/dev/tui\` + a version-package bump — disjoint from this PR's surface (\`runtime/framework-tools\`).
- CI: \`unit/windows/e2e/Vercel-deploy\` remain failing — these are **fork-CI infrastructure constraints** (Vercel deploy \`Authorization required\`, e2e needs Pro sandbox credentials), not regressions from this change. \`typecheck/lint/test-integration(ubuntu)/test-scenario\` green on the PR head (per prior run).
- This is a draft scaffold; the \`blocked\` merge state is the draft/review gate, intended.

Ready for a maintainer shape-review whenever convenient; no action required from contributors.

PR #1555, opened 2026-08-03
fix(eve): resolve skill file reads home-root first with a workspace fallback

shubhxho:
Skills live at `$HOME/.agents/skills/<skill>/`, with `/workspace/skills/<skill>/` as the documented fallback — but `read_file` resolved one root and read it once, so the other form hard-failed and the model wasted a retry guessing.

`read_file` now tries the home root first, then `/workspace/skills`. `write_file` resolves the same list so an edit lands on the file `read_file` served instead of a shadow copy under the other root.

### DIFF
--- a/.changeset/skills-file-read-fallback.md
+++ b/.changeset/skills-file-read-fallback.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Skill file reads now resolve `$HOME/.agents/skills/<skill>/` first and fall back to `/workspace/skills/<skill>/`, as the docs and system prompt already describe. Previously `read_file` resolved a single root and hard-failed when the model named the other one, forcing a wasted retry.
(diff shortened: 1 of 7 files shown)


shubhxho:
fixed it

cmpadden:
Thanks for following up on this. The original issue was resolved by #1459, which added `$HOME` expansion consistently across the model-facing file tools; #1466 subsequently consolidated that behavior.

This PR also changes the meaning of the workspace fallback. The documented contract is that `/workspace/skills` is used when `$HOME` is unavailable—not as a per-file fallback when a file is absent under HOME. When HOME is available, `/workspace/skills/...` remains an ordinary workspace subtree. Trying HOME first for every such path could therefore read or write a different file than the absolute path requested, and would allow `load_skill` to treat ordinary workspace content as an authored skill. The existing `sandbox-access` coverage explicitly preserves that boundary.

I’m closing this as superseded by #1459. Thanks again for the contribution.

PR #1557, opened 2026-08-03
fix(eve): decay stale dev runtime snapshots

chadhietala:
## Summary

Reports came in of `eve dev` retaining tens of GB of old runtime snapshots. Investigation showed the 30-minute decay only trimmed overflow above a **permanent floor**: the 5 newest retired generations were kept indefinitely (intended by #798), and an **undocumented** floor also kept the 5 newest never-activated *staged* snapshots forever — up to ~11 multi-GB directories with no age or size bound. Several silent failure modes could disable pruning entirely.

This PR makes stale generations actually decay:

- **Remove the staged forever-floor** — snapshots staged but never activated now get only the 30-minute in-flight grace, matching the documented policy.
- **Age-bound the retired safety net** — the 5-newest-retired floor now only covers generations superseded within the last 24 hours (`retainWindowMs`), so stale dev directories decay to just the active snapshot.
- **No more forever-protection on metadata failure** — when `retired.json` can't be read/written (e.g. ENOSPC, self-reinforcing on a full disk), the generation decays on its staging time instead of being retained forever.
- **Per-entry fault isolation** — one uninspectable snapshot now skips with a warning instead of zeroing the whole prune pass; a failed `rm` warns without aborting the other deletions.
- **Boot-time sweep** — `eve dev` now requests a background prune at startup (serialized with the existing commit-triggered passes), so leftovers from killed sessions are cleaned on the next run.

Docs (`cli.md`) updated to state the real policy; patch changeset included.

## Validation

- Integration: 22/22 pass — updated the staged-floor test (fixture that only survived via the staged floor now prunes), renamed the retired-floor test, and added coverage for the 24h window (out-of-window generations pruned despite `retainCount`), the metadata-failure fallback (deterministic via `retired.json` as a directory), and the staged-decay path:
  `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/internal/nitro/dev-runtime-artifacts.integration.test.ts`
- Unit: 43/43 pass — added background-prune and boot-sweep assertions:
  `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/internal/nitro/development-generation.test.ts src/internal/nitro/host/start-development-server.test.ts src/internal/nitro/host/dev-authored-rebuild-coordinator.test.ts`
- `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm guard:invariants`, `pnpm docs:check` all clean.


chadhietala:
Closing in favor of fixing snapshot size rather than sweeping harder.

Investigated where the tens of GB under `.eve/dev-runtime/snapshots` actually come from. `node_modules` is *not* copied — every `node_modules` in a snapshot is a 0B symlink farm (installed deps link out to the pnpm store, workspace deps link inward). The bytes are the authored workspace closure, and 30% of each snapshot is `.workflow-vitest`, a vitest-only cache that nothing at dev runtime reads:

```
32M  packages/eve/dist
22M  packages/eve/.workflow-vitest   <- vitest cache, gitignored, copied anyway
15M  packages/eve/src
```

It rides along because `SNAPSHOT_SKIP_NAMES` (`dev-runtime-source-snapshot-copy.ts`) has drifted from `IGNORED_INPUT_DIRECTORIES` (`workflow-bundle/builder-support.ts`), which already excludes it.

Separately, `SNAPSHOT_COPY_MODE = COPYFILE_FICLONE` is inert: Node/libuv on macOS creates the destination before `fcopyfile`, so no APFS clone happens. Measured free-space delta for a 200MB file — `fs.copyFile` with FICLONE: 187MB; `cp -c`: 8KB. With cloning working, N generations would cost roughly one.

Retention was bounding a cost that shouldn't be that large to begin with. Follow-up PRs will drop the cache from the copy and fix the clone path.

PR #1561, opened 2026-08-03
fix(eve): render background task turns in the dev TUI

ruiconti:
Stacked on #1522. Background tasks exposed four related failures in the dev TUI:

- The task receipt looked like child completion, so late child rows escaped their subagent section.
- Server-initiated wake turns stayed invisible while the prompt was idle.
- `task_await` and the completion wake could report the same ready transition in two parent turns.
- Completed background sections remained in the mutable prompt region and redrew on every input repaint.

These are one rendering lifecycle: a background child outlives its originating parent turn, wakes the parent independently, and finishes at its own stream boundary.

## Data flow

| Flow | Before | After |
| --- | --- | --- |
| Background dispatch | `subagent.completed` settled the section when the parent received a task receipt | explicit background-task metadata keeps the call cohort live until the child stream reaches its boundary |
| Late child output | rows appended at the transcript edge after later parent and user blocks | rows insert beside the original call cohort; later parent and user blocks render before active background sections |
| Parent wake turn | no parent stream reader existed while `readPrompt` owned input | one idle session follower renders complete wake turns, then aborts and commits its cursor before `send()` starts |
| Awaited completion | `task_await` returned the task and the buffered completion wake started another turn | the await claims that task ready transition; parked delivery consumes the matching one-shot wake |
| Child completion | the section reached `Done` but remained mutable until another parent boundary | the authoritative child success or failure boundary commits the completed cohort to scrollback immediately |

The idle follower and `send()` never read the parent stream concurrently. The follower preserves the active prompt key handler and draft, then hands the advanced durable cursor to the next send.

Wake suppression uses framework-owned task metadata rather than matching message text. Claims are one-shot and cancellation releases them, so `input_required` followed by `task_send`, ordinary background completion, and cancelled awaits still wake the parent.

The parent `subagent.completed` event remains a non-authoritative fallback because the final child message can still be in flight on its independent stream. Only the child stream boundary or turn cancellation finalizes the section.

Regression coverage follows the complete lifecycle: background receipt, parent and user turns while the child runs, idle wake rendering without replay or draft loss, await wake coalescing, late child events staying in their section, and authoritative success or failure finalization.

Part of #1084.


ruiconti:
Tasks 06 through 09 are coalesced in this PR as one background-task TUI rendering changeset.

ruiconti:
Superseded by the restack: protocol half → #1756, TUI half → #1760. Content preserved; see the stack thread on #1757 for the reorganization rationale.

PR #1564, opened 2026-08-03

ruiconti:
Stacked on #1561. The server already wakes and runs the parent turn when a background task completes; the dev TUI had no parent-stream reader while `readPrompt` was open, so that turn stayed invisible until the next user send replayed it.

## Data flow

| Phase | Before | After |
| --- | --- | --- |
| Idle prompt | no parent session reader | one persistent `session.stream()` follow from the stored cursor |
| Wake turn | durable events append with zero readers | follower splits the boundary-blind stream into complete turns and renders each immediately |
| User submits | `send()` reads from its stored cursor, replaying unseen wake events | idle follow aborts and is awaited first; its cursor commit happens before `send()` opens |
| Prompt redraw | n/a | idle rendering applies events without replacing the prompt key consumer or clearing its draft |

## Ownership choice

Uses option (a): pause the idle follower during `send()`. It is the smaller change and keeps existing request-scoped send rendering intact. There is exactly one parent-stream owner at a time:

```text
readPrompt: session.stream() follow
prompt resolves: abort follow → await cursor commit
send: existing boundary-scoped stream
turn ends: restart idle follow from the advanced cursor
```

The follower reuses the existing reconnecting client stream and durable `startIndex` cursor. No server route or push path changes.

The terminal renderer gets a separate idle-stream path: it applies the same tool/assistant events and paint logic, but does not detach input, replace the key handler, or clear the current editor. Persistent input typed before a wake remains submitted afterward.

Manual run against the background-task workshop: with no input after dispatch, `Checked task_…` and the assistant brief rendered on completion. A subsequent user message produced only its own answer; the task check and brief each appeared once.

### DIFF
--- a/.changeset/dev-tui-idle-session-follow.md
+++ b/.changeset/dev-tui-idle-session-follow.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Render server-initiated turns in the dev TUI while it is idle at the prompt. The idle follower hands its advanced stream cursor to the next user send, preventing wake turns from replaying.
--- a/packages/eve/src/cli/dev/tui/runner.test.ts
+++ b/packages/eve/src/cli/dev/tui/runner.test.ts
@@ -534,6 +534,147 @@ function sessionYieldingTurns(turns: ReadonlyArray<readonly unknown[]>): ClientS
   return session;
 }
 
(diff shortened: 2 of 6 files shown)


ruiconti:
Coalesced into #1561, which contains tasks 06 through 09 as one background-task TUI rendering PR.

PR #1571, opened 2026-08-03
fix(eve): reconcile channel state on delivery

### LINKED ISSUE #1554: GitHub channel reviews a stale checkout on synchronize: send() drops state when resuming a session
**Version:** eve 0.25.2 (production, Vercel backend); root cause verified still present on `main`.

## What happens

For a PR conversation, the first turn checks out the correct head. Every subsequent webhook-triggered turn on the same session (e.g. `pull_request.synchronize`) leaves the sandbox checkout at the *original* head sha. The agent sees a tree that doesn't match the webhook's `head_sha`, and any manual `git fetch` inside the sandbox fails with `fatal: could not read Username for 'https://github.com'` (expected — the token never enters the sandbox, but nothing re-establishes the auth-injecting network policy either, since the framework checkout never reaches its fetch step).

This contradicts the GitHub channel docs: "Before the first model call, every triggered turn checks out the relevant ref into the sandbox. […] Within a session, checkout is incremental across turns."

## Root cause

Three pieces interact:

1. `dispatchPullRequest` builds fresh state from each webhook (including the new `headSha` from `stateFromPullRequestEvent`) and passes it to `send(message, { auth, continuationToken, state })`.
2. `createSendFn` (`packages/eve/src/channel/send.ts`) only applies `state` on the session-*creation* path (`runtime.run`). When the continuation token resolves to an active session, it calls `runtime.deliver(...)` instead — and `DeliverInput` has no `state` field at all — so the updated `headSha` is silently discarded.
3. The default `turn.started` handler (`checkoutRepositoryForTurn`) checks out `state.headSha` from the stored session state. `checkoutGitHubRepository` short-circuits when the sandbox HEAD already equals that (stale) sha, so the tree never advances and no error is surfaced.

## Impact

Any GitHub-channel agent that handles more than one webhook per conversation reviews or operates on stale code after the first turn. In our case (a PR-review agent), every push after PR open silently produced a review of the original head; our agent detects the sha mismatch and aborts, but agents without that check would act on stale code with no signal.

## Reproduction

1. `githubChannel` with `onPullRequest` dispatching `opened` and `synchronize` events for the same PR.
2. Open a PR — turn 1 checks out correctly.
3. Push a new commit — turn 2's `<github_context>` shows the new `head_sha`, but `git -C /workspace rev-parse HEAD` still returns the turn-1 sha, and `git fetch` fails with the credential error above.

## Suggested fix directions

- Carry channel state updates through the deliver path (add `state` to `DeliverInput` and merge into adapter state before the turn), or
- Have the GitHub channel not trust the stored `headSha` for checkout freshness — e.g. re-resolve the PR head via the API at `turn.started` (the machinery already exists in `resolveCheckoutDescriptor` when `headSha`/`headRef` are null).

## Workaround we're using

Event handlers null the checkout-derived state fields (`headSha`, `headRef`, `baseSha`, `baseRef`) at end of turn, which forces `resolveCheckoutDescriptor` to re-fetch the PR head from the API on the next turn's checkout.

benpankow:
Previously, state passed to `send()` was dropped when resuming an existing session.

Now, resumed sessions carry updated state. Channels can opt into reconciling it with durable state via `updateState(current, incoming)`; channels without this hook retain the previous ignore-on-resume behavior. The GitHub channel uses this API so later PR webhooks, including `pull_request.synchronize`, update the head SHA before checkout instead of reviewing the session’s original commit.

Fixes #1554.

## Testing

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/channel/send.test.ts src/public/definitions/channel.test.ts src/public/channels/github/githubChannel.test.ts src/execution/workflow-steps.test.ts src/harness/messages.test.ts src/execution/workflow-runtime.test.ts src/execution/workflow-entry.test.ts src/runtime/channels/registry.test.ts`
- `git diff --check`

### DIFF
--- a/.changeset/channel-state-reconciliation.md
+++ b/.changeset/channel-state-reconciliation.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Let stateful channels reconcile state supplied to `send()` when resuming an existing session. Channels retain durable state by default and can explicitly apply fresh delivery metadata; for example, GitHub webhook turns now update the checkout ref before each model call.
--- a/packages/eve/src/channel/adapter.ts
+++ b/packages/eve/src/channel/adapter.ts
@@ -141,6 +141,16 @@ export type ChannelAdapter<TCtx extends ChannelAdapterContext<any> = ChannelAdap
   deliver?(payload: DeliverPayload, ctx: TCtx): StepInput | void | Promise<StepInput | void>;
 
   /**
+   * Reconciles the latest state supplied by resumed `send()` calls with
+   * durable state before a coalesced turn. Channels opt in so inbound snapshots
+   * cannot overwrite channel-owned mutations unless explicitly permitted.
+   */
+  updateState?(
+    state: Record<string, unknown>,
+    incoming: Readonly<Record<string, unknown>>,
+  ): Record<string, unknown>;
+
+  /**
    * Optional factory that builds the adapter context for this adapter.
    *
    * @internal
--- a/packages/eve/src/channel/routes.ts
+++ b/packages/eve/src/channel/routes.ts
@@ -73,8 +73,10 @@ export interface SendPayload {
 
 /**
  * Starts or continues a session on this channel. Accepts a plain string,
- * `UserContent`, or a {@link SendPayload}, plus {@link SendOptions}. Resolves
- * to the resulting {@link Session}.
+ * `UserContent`, or a {@link SendPayload}, plus {@link SendOptions} (auth,
+ * continuation token, run mode, and `state` for stateful channels). The state
+ * seeds a new session; an existing session applies it only when the channel
+ * defines an `updateState` reconciler. Resolves to the resulting {@link Session}.
  */
 export type SendFn<TState = undefined> = (
   input: string | UserContent | SendPayload,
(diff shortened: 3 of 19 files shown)


wantzjt:
Confirmed still present on current `main` after the session-operation unify, and on the `eve@0.27.13` tag that our public examples pin.

On `eve@0.27.13`, `createSendFn` builds `DeliverInput` without `state`, matching the routes docs (“ignored on subsequent delivers”). On current `main`, `packages/eve/src/channel/send.ts` is gone; the resume path is `channel-address.ts` → `dispatchContinuation` with a `SessionCommand` `send` that still has no state/adapterState field, while `state` is applied only on `createSession`. `githubChannel` on `main` also still lacks an `updateState` reconciler.

This PR’s direction (carry inbound state on resume + opt-in `updateState`, with GitHub merging webhook snapshots before checkout) still matches the failure mode in #1554. It looks blocked on a rebase onto the post-#1597 / `channel-address` + `dispatchContinuation` shape rather than on the old `createSendFn`/`runtime.deliver` path.

Happy to re-verify against a rebased head if useful. We are not opening a competing PR.

benpankow:
Closing as superseded by the newer work in #1900, which addresses refreshing GitHub checkout state on resumed turns against the current session and GitHub channel architecture.

PR #1574, opened 2026-08-03
fix(eve): coalesce task-await results with task wakes

ruiconti:
Stacked on #1564. The newly visible duplicate transcript was two real parent turns:

1. `task_await` reported the ready task inside the active turn.
2. The task run's unconditional completion wake was buffered, then started another turn that peeked and reported the same task again.

## Data flow

| | Before | After |
| --- | --- | --- |
| `task_await` starts | aggregator waits for task snapshots | aggregator waits, and parent session state claims those task ids' next ready wakes |
| task becomes ready | task run delivers a plain wake message | wake payload also carries framework-owned `{ taskId, status }` metadata |
| parent parks | buffered wake always starts a turn | parked delivery filter consumes a matching await claim and drops that wake; unrelated payloads still run |
| await turn cancels | no explicit wake ownership | cancellation releases outstanding claims, so later completion still wakes the parent |

The filter sits at the parked delivery boundary because that is where both facts are durable: the task-await dispatch's updated parent session state and any task wake buffered while the turn was active. It runs before child-response routing and before a new parent turn is started.

Claims are one-shot. A matching wake consumes its claim; `input_required` followed by `task_send` can therefore wake again on completion. `task_peek`, plain background completion, and cancelled awaits keep their existing wake behavior.

Manual transcript checks against shape 08:

- synchronous workshop: one `Awaited task`, one `BUILD LOG v1`, no queued `Checked task` turn;
- natural-language background continuation: receipt in 6s with no await, then one `Checked task` and one `BUILD LOG v2` on completion.

Part of #1084.

### DIFF
--- a/.changeset/task-await-wake-coalescing.md
+++ b/.changeset/task-await-wake-coalescing.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Prevent task completion from starting a duplicate parent turn when `task_await` already reports that ready transition. Cancelled awaits release their wake claims so later task completion still wakes the parent.
--- a/packages/eve/src/channel/types.ts
+++ b/packages/eve/src/channel/types.ts
@@ -153,6 +153,11 @@ export interface DeliverPayload {
   readonly message?: string | UserContent;
   readonly context?: readonly string[];
   readonly outputSchema?: JsonObject;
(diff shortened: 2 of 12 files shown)


ruiconti:
Coalesced into #1561, which contains tasks 06 through 09 as one background-task TUI rendering PR.

PR #1575, opened 2026-08-03
fix(eve): gate localDev() on the dev environment, not the request host

dnukumamras:
`localDev()` and the `vercelOidc()` local bypass now decide "am I a local dev server?" from the process environment (`EVE_DEV=1` from `eve dev`, or `VERCEL=1` + `VERCEL_ENV=development` from `vercel dev`) rather than the request URL host. On a self-hosted Node server the request host comes from the `Host` header, so header content no longer influences the auth decision.

## Changes
- `localDev()` reads the deployment environment; the `Request` no longer affects the result.
- Removed the `isLoopbackRequest` export (public API break, hence `minor`).
- Framework default is now `[vercelOidc(), localDev(), placeholderAuth()]`, matching the scaffold template: local dev still works, and production/self-hosted deployments fail closed with a `401`.
- Docs and tests updated to match.

## Verification
`pnpm fmt`, `pnpm lint`, `pnpm typecheck`, `pnpm test:unit`, `pnpm guard:invariants`, and `pnpm docs:check` all pass.

### DIFF
--- a/.changeset/local-dev-env-gated.md
+++ b/.changeset/local-dev-env-gated.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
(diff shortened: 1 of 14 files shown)


ctgowrie (commented):
Can we align on checking `NODE_ENV = "development"`

ctgowrie (approved):

Issue #786, opened 2026-07-14
Steering: a message sent mid-turn is deferred to the next turn — and can be consumed as the answer to a prompt raised after it was sent


Lukavyi:
*(Rewritten after re-checking the actual behavior: the messages are not dropped — they are deferred and then replayed at the wrong moment, which is worse. Original text described them as silently dropped.)*

### What we want

Steering: while a turn is running, a user sends another message ("actually, use staging", "stop, wrong repo") and the agent takes it into account at the **next step boundary of the current turn** — the way Claude Code / Codex CLI behave.

### What actually happens

The message is *not* dropped. It is held until the running turn finishes, and then folded into a **new** turn. Two consequences, both bad:

1. **Late and out of context.** By the time the message is processed, the agent has already finished the work the user was trying to redirect. The steering intent ("don't do X") arrives after X was done.
2. **It can be consumed as an answer to a prompt the user never saw.** If, while the first turn was running, the agent raised an input request — an `ask_question` select, or a HITL tool approval (`Yes` / `No`) — the queued message drains into `resolvePendingInput` and is resolved **against that request**. So a message the user typed *before the prompt existed* ("also add tests") becomes the answer to "Approve tool call: bash?". The user's session state and the agent's next action are then both wrong, and nothing in the transcript shows why.

(2) gets sharper for anyone who fixes #785: once a freeform message during a pending approval means "cancel this call and take my new context", a stale queued message will *cancel a tool call the user never intended to cancel*. The two behaviors are only safe together if the runtime can distinguish "the user answered the prompt" from "this message predates the prompt".

There is enough information to make that distinction — the delivery has an arrival time, and the pending input batch has `event: { sequence, stepIndex, turnId }`. Nothing currently compares them.

### Why steering looks close to free

The harness already has the plumbing:

- `queueDeferredStepInput` / `consumeDeferredStepInput` (`dist/src/harness/input-requests.js`) park a `StepInput` and re-inject it at the **start of the next step** (`executeStepBody` calls `consumeDeferredStepInput` before `resolvePendingInput`);
- the step loop keeps the turn alive while deferred input exists (`hasDeferredStepInput(session)` in the continue condition), so a mid-turn message could extend the current turn instead of waiting for the next one;
- `coalesceTurnInputs` already merges two `StepInput`s, which is what a burst of messages needs.

What is missing is a supported path from a channel delivery into that slot on an *active* turn, and an ordering rule.

[...]

PR #1578, opened 2026-08-03
feat(eve): define the HITL request lifecycle state machine

### LINKED ISSUE #1201: Approval dropped when a step requests both an approval-gated tool and a subagent call
When a single model step requests an approval-gated tool call **and** a subagent (or remote-agent) call in the same response, the approval is silently dropped and the turn fails on resume.

## Repro

A turn where the model, in one response, calls an `approval`-gated tool (e.g. `always()`) and delegates to a subagent. The approval request never surfaces; once the subagent result arrives and the step resumes, the turn fails.

## Runtime artifact

```
AI_MissingToolResultsError: Tool result is missing for tool call call_<id>.
```

## Cause

In the tool-loop park classification, the pending-runtime-action branch returns before the input-request park runs. So `input.requested` is never emitted for the approval, and the action-result resume calls the model with the gated tool call's `tool_use` still unanswered — which the AI SDK rejects.

## Fix

Persist both batches on park (the input batch with empty `responseMessages`, since the action batch owns the step's assistant messages) and emit `input.requested` immediately; on the action-result resume, re-park on the still-unanswered approval with the resolved results committed to history.

Fixed in #1200.

### LINKED ISSUE #1224: Freeform reply to a pending tool approval still kills the turn on 0.25.1–0.26.0 — #785 was closed but not fixed by #888
### Summary

The core repro from #785 still reproduces on `eve@0.25.1` (and the relevant code is unchanged through `0.26.0`): when the harness is waiting on a HITL **tool approval** and the user replies with freeform text instead of tapping Approve/Deny, the turn closes silently. The pending approval is never resolved and the message is parked in `deferredStepInput` forever, so the agent goes mute — every subsequent message defers again.

#785 was closed as fixed by #888, but #888 does not cover this case.

### Why #888 doesn't fix it

#888 ("deliver stale HITL responses as follow-ups") handles:
- stale option clicks on already-closed approvals/questions, and
- freeform text while a plain `ask_question` is pending (resolved via `resolveTextToResponses`).

It does **not** change the pending tool-approval path. In `dist/src/harness/input-requests.js` (`resolvePendingInput`), this branch is still present and unchanged from 0.25.1 through 0.26.0:

```js
if (batch.requests.some(isApprovalRequest) && hasUnansweredApproval({ pendingBatch: batch, responses })) {
    session = queueDeferredStepInput(session, compactStepInput(input));
    return { deferredMessage: true, outcome: "unresolved", messages, session };
}
```

Freeform text can never produce an approval response (`resolveTextToResponses` cannot map free text onto approve/deny), so `hasUnansweredApproval` stays true, the input is deferred, and the turn ends with no output.

### Repro

1. Trigger any tool with `approval: always()` so the harness posts an Approve/Deny gate (Telegram inline keyboard or Linear Agent Session — the bug is channel-agnostic).
2. Instead of tapping a button, send a plain text message.
3. The turn ends silently; the approval stays pending; the message sits in `deferredStepInput`; the agent never responds again.

Verified on `eve@0.25.1`; the `resolvePendingInput` branch above is byte-identical in the published `0.25.2`, `0.25.3` and `0.26.0` dist.

### Expected

Same as originally requested in #785: treat a freeform reply to a pending approval as "don't run this call, here is new context" — deny the pending call (terminal execution-denied result, as if the user tapped No), append the message as user context, and let the model continue.

### Workaround

We previously patched `input-requests.js` to synthesize `{ optionId: "deny" }` for unanswered approvals when a freeform message arrives; we dropped the patch when upgrading to 0.25.1 expecting #888 to cover it, and the hang came back immediately.

Refs: #785 (closed), #888.

### LINKED ISSUE #1608: Stale proxied subagent input response can fail parent session with HookNotFoundError

A public minimal reproduction is not available yet. This report is based on:

- A production reproduction on `eve@0.24.6` with Vercel Workflow event timing below.
- Source-level confirmation that the same failure path remains reachable in `eve@0.30.1`.
- The current forwarding code: https://github.com/vercel/eve/blob/eve%400.30.1/packages/eve/src/execution/workflow-steps.ts#L590-L613

Additional private deployment context can be shared if useful.

### Steps to reproduce

The production failure had this ordering:

1. A parent session dispatches a declared local subagent.
2. The child raises a HITL input request. The parent persists a proxy mapping from the request ID to the child's continuation token.
3. A response resumes the child and the child completes.
4. The child continuation hook is disposed before the parent processes the child's terminal result and clears its proxy mapping.
5. A duplicate or delayed input response for the same request reaches the parent during that window.
6. `routeProxiedDeliverStep` reads the stale proxy mapping and calls `resumeHook(childContinuationToken, ...)`.
7. `resumeHook` throws `HookNotFoundError`. The step retries three times and the parent session fails.

A focused integration reproduction should hold the parent before it consumes the child's terminal result, let the child complete and dispose its hook, then deliver another response for the proxied request.

Current behavior:

A stale proxied child route is treated as a fatal workflow step failure:

```text
FatalError: Step "step//eve@0.24.6//routeProxiedDeliverStep" failed after 3 retries: Hook not found
Caused by: HookNotFoundError
```

This fails the parent session even though the child has already completed successfully.

Expected behavior:

A late or duplicate response targeting an already-terminal child must not fail the parent session. Eve should classify the child route as stale and resolve it through a non-fatal contract, such as preserving it as follow-up input or returning a structured stale-response outcome. The response must not turn into `session.failed`.

### eve version

- Production reproduction: `eve@0.24.6`
- Confirmed reachable by source inspection: `eve@0.30.1`

### Environment

```shell
Runtime: Vercel production
Workflow world: Vercel Workflow
Node: 24.x
Package manager: pnpm 11.1.1
Channel: Slack
Agent shape: parent with declared local subagents
```

### Where does the bug occur?

Production (`eve start` / deployed).

### Deployment

Private Vercel deployment. Relevant Agent Run and error identifiers are included below for Vercel-side correlation.

### Build and runtime logs

```text
Root Agent Run:
wrun_41KZ6JD89X0GWEEAQ8MS67PZC3

Internal run executing routeProxiedDeliverStep:
wrun_41KZ6KDRBA0GPE4N9XJKRM4ZVZ

Error ID:
b3e62ecd-2c73-4cf3-a250-ee530e45a3b6

2026-08-04T14:59:19.518Z  child hook disposal begins
2026-08-04T14:59:20.967Z  final observed child hook disposed
2026-08-04T14:59:21.016Z  child run completed
2026-08-04T14:59:27Z      routeProxiedDeliverStep exhausts retries
2026-08-04T15:12:16Z      parent emits session.failed

Step "step//eve@0.24.6//routeProxiedDeliverStep" failed after 3 retries: Hook not found
```

### Additional context

The failure remains reachable in the published `0.30.1` source:

```ts
for (const forChild of routed.forChildren) {
  await resumeHook(forChild.childContinuationToken, {
    auth: input.auth,
    kind: "deliver",
    payloads: [forChild.payload],
  });
}
```

There is no `HookNotFoundError` handling around the child resume. The proxy mapping is cleared later, when the parent processes the child result:

https://github.com/vercel/eve/blob/eve%400.30.1/packages/eve/src/harness/runtime-actions.ts#L241-L252

That leaves a race window between child hook disposal and parent-side proxy cleanup.

The `0.25.0` stale HITL response change does not close this window because it runs only after a response reaches a live session. Here the failure occurs in `resumeHook` before the child can process the response. The newer Workflow concurrency and cancellation fixes also do not handle this missing child hook in `routeProxiedDeliverStep`.

I searched existing Eve issues and pull requests before filing. The closest reports are adjacent but not duplicates:

- #891 covers child-to-parent notification not waking the parent.
- #624 covers concurrent initial deliveries producing `HookConflictError`.
- #982 covers anchored root delivery hook rotation.
- #541 covers stale hooks on wedged root sessions.

Suggested regression coverage:

- A child raises a proxied HITL request.
- The response resumes the child.
- The child completes and its hook is disposed.
- Before the parent consumes the terminal child result, another response for the same request is delivered.
- The parent remains healthy and the delivery receives a documented non-fatal disposition.

ruiconti:
## What

- One research doc, [`research/hitl-request-lifecycle.md`](https://github.com/vercel/eve/blob/rui/hitl-request-lifecycle/research/hitl-request-lifecycle.md): motivation, state model, a per-case transition catalog, protocol events, and the staged implementation plan. The engine architecture and the decision record that were separate docs are folded in — reviewers read one document.
- An anchor-keyed eval suite in the existing HITL fixtures, gated behind `EVE_HITL_LIFECYCLE_CONTRACT=1`, with [`coverage.md`](https://github.com/vercel/eve/blob/rui/hitl-request-lifecycle/e2e/fixtures/agent-tools-hitl/evals/lifecycle/coverage.md) mapping every transition to its evidence, intended tier, or named blocker — plus the executable gate each implementation stage must pass.
- No runtime changes.

## Why

HITL has been our least stable surface, and the failures form one pattern:

- #1224 — a freeform reply to a pending approval mutes the session forever.
- #1201 — an approval is silently dropped when the same step also requests a subagent call.
- #1608 — a duplicate input response resumes a disposed child hook and fails the parent.
- #786 — a message is consumed as the answer to a question the sender never saw.
- #1830 and #1868 — two mitigations, months apart, for the same root cause: an obligation encoded as a blocked continuation ("the code is waiting on X, therefore only X can arrive") instead of as data.

The instability is structural. Interpretation is smeared across ten modules (6,780 lines across 14 modules, 39 files consuming fragments of the state APIs), reached through five entry points: channel POST, OAuth callback route, driver inbox, turn step, tool loop. State lives in three unrelated keys (`pending-input-batches`, `pendingAuthorization`, `deferredStepInput`), so behavior is the cross product of approvals × challenges × limit prompts × projections × actor guards — every fix touches an unknown subset of that product, and no seam ever sees the whole state. That is why both wedges could ship.

The third gap: no per-case specification. Expected behavior lived in scattered tests, so nobody — human or agent — could change one case and prove the others unchanged. Regressions surfaced as user reports.

## The answer

Obligations become rows in one durable ledger; one pure interpreter decides every transition; nothing the session receives is silently consumed.

```
before  delivery → channel POST / dedicated auth hook → driver wait-source,
        resolvePendingInput, stale conversion, limit special cases → 3 state keys
after   delivery / callback / timer / control → one FIFO stream →
        interpret(state, input) → persist → ordered effects
```

Every behavior is one catalog entry with a stable anchor (`machine.obligation.input.outcome[-guard]`) stating the machine state, the input, what happens on three planes — ledger, model transcript, turn — the required events, and a concrete example. Evals declare the anchor they prove. A change that regresses a case fails a named gate instead of a user. The catalog and the coverage matrix carry most of the document's length on purpose: they are what makes the behavior safely changeable.

Four invariants govern the admission seam: admission never blocks; admission does not necessarily invoke the model; every invoked model turn receives a safe projection of open obligations; a newly raised approval is checked against open approval intent before append.

## Scope and review notes

- The implementation lands as six staged PRs (store → interpreter → auth → single stream/projection → target semantics → lifecycle events), each with its gate in `coverage.md`; after stage 4 every behavior change is a diff to `interpret.ts`. This PR ships the contract and the gates, not the engine.
- Six active pre-machine evals that asserted text matching, approval-message queueing, stale-question-as-message conversion, or limit-prompt preservation were removed rather than kept beside their replacements.
- Wire compatibility: no existing event changes shape or meaning; all new schema is additive behind one stream-version bump. Two named breaks, both scoped: runtime text matching is removed (channel adapters keep `resolveTextToResponses`), and pre-cutover challenge URLs embedding the old `:auth` hook token need a one-release alias or an accepted pre-1.0 break.

Fixes #1224 (behavioral core shipped in #1868). Fixes #1201. Fixes #1608.


ruiconti:
Mitigating this in #1868.

research/hitl-request-lifecycle.md
```diff
@@ -0,0 +1,1097 @@
...
+An **obligation** is one open item the session owes an answer to. Four kinds:
+
+- **Approval** — human consent for one tool call.
+- **Question** — an `ask_question` prompt.
+- **Limit** — a session-limit continuation prompt, carrying a monotonic
+  `generation`.
+- **Challenge** — a connection-authorization challenge (OAuth credentials,
+  not consent).
+
+Each obligation is a small state machine. Terms:
```
chadhietala, line 105: The approval + challenge case [invariant 7](https://github.com/vercel/eve/pull/1578/changes#diff-c63e7d5f87276b1f6a9cd8d2f62ed8f8fa40d279742d83d1b80582c3ea10eb1cR271) I'm not sure about. 

Might be better to model these as [workflow nets](https://mlwiki.org/index.php/Workflow_Nets) where you can't actually wedge anything.

PR #1579, opened 2026-08-03

ruiconti:
Stacked on #1574. Completed background subagent sections reached `Done` but stayed in the TUI's mutable live region, so every prompt repaint kept redrawing a section that could no longer change.

## Data flow

| Completion source | Before | After |
| --- | --- | --- |
| Child stream reaches `session.waiting` / `session.failed` | pump marks `Done`; renderer waits for a later parent turn finalization | authoritative completion marks the whole call cohort non-live and commits it immediately |
| Parent `subagent.completed` fallback | same `Done` path | remains non-authoritative and mutable, because the child stream's final message may still be in flight |
| Turn cancellation | settles section, then waits for another paint boundary | authoritative completion commits the cancelled cohort immediately |

The distinction lives at the child-stream pump, where the source is known. The renderer still preserves the collapsed `※ subagent(name) └ Done` transcript record; it simply leaves the live prompt region and no longer redraws when the user types.

Regression coverage includes:

- successful child boundary → authoritative completion;
- failed child boundary → authoritative completion;
- parent fallback → non-authoritative completion for late child events;
- prompt repaint after completion does not redraw the committed subagent section.

Part of #1084.

### DIFF
--- a/.changeset/background-subagent-finalization.md
+++ b/.changeset/background-subagent-finalization.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Finalize completed background subagent sections immediately when the child stream reaches a success or failure boundary. Completed sections remain in transcript history but no longer occupy or redraw the live prompt region.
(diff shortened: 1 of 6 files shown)


ruiconti:
Coalesced into #1561. Its branch now contains the complete tasks 06 through 09 changeset against tasks/05-send.

Issue #1510, opened 2026-08-01
First-class memory


AndrewBarba:
## Summary

Track first-class memory for eve: durable, provider-owned context that outlives one session while eve owns trusted scope resolution, lifecycle boundaries, recalled-message projection, and model visibility.

## Proposed contract

- Filesystem-authored memory slots with path-derived identity.
- Provider-owned `recall`, optional `capture`, and optional model tools.
- Deployment-aware namespaces plus trusted principal, channel, tenant, or application scope.
- Keyed recalled messages with deterministic supersession and unkeyed immutable appends.
- One canonical projected-history boundary and memory-aware compaction that never launders private recall records into a free-form summary.
- A bounded `fileMemory()` reference provider with portable conditional document backends.

## Current foundation

The generic prerequisites are merged: projected history (#2352), durable dynamic callbacks (#2354), name-and-phase callback identity (#2384), the canonical source graph and binding table (#2404, #2516), and registered anchor-derived programmatic templates (#2539). The runtime-tool contribution proposal in #2347 is superseded.

Each selected memory binding induces `tools/<slot>.ts` through #2539's registered template API. The derived tool inherits the memory candidate's node and layer, depends on that selected binding, and yields to a direct target in the same layer. Definition factories materialize once per module namespace in each module-map load, so the memory lifecycle and derived provider tools share one provider instance without a global registry.

## Implementation sequence

1. **M1 — first-class memory core:** #2534 contains current `main` through the proposal merge at `6a2529f70`. It adds the authoring/compiler surface, registered provider-tool template, generic provider-instance materialization, namespace and scope locks, recall records and projection, lifecycle, compaction, agent-info v4, docs, changeset, extension contracts, and deterministic E2E coverage. It does not restack #2142. Exact-head CI and fresh approval after the merge are the remaining gates.
2. **M2 — `fileMemory()`:** after #2534 merges, rebase the preserved draft #2144 directly onto current `main`, retaining only provider document, backend, concurrency, and file-provider E2E work.

#2142 and #2145 are superseded and closed. #2144 remains the only implementation salvage branch, and M2 must not start on its pre-M1 base.

## Research and implementation

- `research/first-class-memory.md`, merged by #1581, is the public contract and implementation boundary.
- #2534 is the M1 core implementation.
- `research/programmatic-agent-sources.md` records the canonical source-graph architecture memory uses.

AndrewBarba:
## Implementation restructure — tracking

[...]

PR #1581, opened 2026-08-03
docs(eve): memory - propose first-class lifecycle

AndrewBarba:
### Summary

Related to #1510. This proposal defines path-authored memory with provider-owned recall, capture, and tools; trusted namespace and scope locks; keyed recall supersession; one projected-history boundary; and memory-safe compaction.

The source-composition design reflects merged #2539. Each selected memory binding induces `tools/<slot>.ts` through a registry-issued template anchored to the memory candidate. The derived module inherits the anchor's node and layer, records the memory binding as a dependency, derives a stable source ID, and yields to a direct target in the same layer. Memory therefore needs no custom source layer, registration mode, executable registry, or post-resolution tool merge.

The proposal also records the generic provider-instance boundary implemented in #2534: zero-argument definition factories are materialized once per module namespace in one module-map load, so a direct memory slot and its derived wrapper share one provider instance without a global cache. Provider tools otherwise use ordinary `defineDynamic` compilation, binding, hydration, approval, authorization, and replay.

The M1 core in #2534 is rebased and simplified on current `main`, with agent-info v4, generic cold callback rebinding, published docs, extension contracts, and deterministic E2E coverage. #2142 and #2145 remain superseded; #2144 remains the file-provider salvage branch to rebase only after M1 merges.

Reviewer attention should focus on the public namespace/scope contract, lifecycle failure semantics, durable recall projection, and compaction behavior. The wrapper-construction invariants themselves are supplied by #2539.

### Validation

Cross-checked the proposal against merged #2539, the rebased #2534 implementation, current dynamic-tool callback identity and rebind behavior, and the projected-history seam. `pnpm docs:check` and `pnpm guard:invariants` pass, and the proposal is formatted with oxfmt.

No runtime tests apply to this research-only change. Runtime validation lives in #2534.

### Checklist

Changeset not applicable: this PR changes research documentation only.

### Diff size

**Docs** — 1 file · `+1,281 / -0`

The research proposal defines the public contract and records the final source-template, provider-materialization, replay, lifecycle, projection, compaction, and two-PR delivery boundaries.

Not applicable.

Not applicable for a research-only change.

### DIFF
--- a/research/first-class-memory.md
+++ b/research/first-class-memory.md
@@ -801,14 +801,16 @@ Mounted extensions cannot contribute memory slots.
       thread](https://github.com/vercel/eve/pull/1581#discussion_r3807561187)
       once that boundary is unambiguous.
 
-- [ ] **Give the model a safe memory-slot purpose.** Decide whether the
-      consuming definition or provider configuration supplies a model-facing
-      description that distinguishes destinations such as personal preferences
-      and channel conventions. Keep raw namespace and scope values out of model
-      input, and state that descriptions guide tool choice rather than enforce
-      authorization. Verify that multiple `fileMemory()` slots expose enough
-      information for the model to choose the intended qualified tool. Then
-      close the [model-facing attribution
+- [x] **Give the model a safe memory-slot purpose.** The consuming
+      `defineMemory(...)` may supply an optional static `description`. eve
+      validates it and prepends it to every provider tool description before
+      storing durable dynamic metadata. Omitting it preserves provider
+      descriptions unchanged. The implementation neither derives nor exposes
+      namespace or scope values, and the docs distinguish model routing
+      guidance from authorization. Unit coverage verifies two slots sharing one
+      provider receive different descriptions without mutating provider tools;
+      file-memory integration and e2e coverage exercise described
+      `fileMemory()` slots. This resolves the [model-facing attribution
       thread](https://github.com/vercel/eve/pull/1581#discussion_r3807269437).
 
 - [ ] **Explain projection placement and tool invocation semantics.** Keep or
--- a/research/first-class-memory.md
+++ b/research/first-class-memory.md
@@ -33,12 +33,12 @@ turn-start recall through the same durable dynamic-capability machinery as a
 `turn.started` `defineDynamic` tool resolver.
 
 Memory definitions and provider tools compile through eve's canonical source
-graph. Each selected memory slot retains its authored binding and projects one
-compiler-owned programmatic `tools/<slot>.ts` wrapper that exports an ordinary
-`defineDynamic` resolver. The wrapper qualifies provider tool keys and passes
-them through the existing source composition, binding, module-map, dynamic-tool,
-approval, authorization, and replay paths. Memory does not add a runtime tool
-contribution seam or a second executable registry.
+graph. Each selected memory slot retains its direct binding and induces one
+derived `tools/<slot>.ts` module from a registered programmatic template. The
+template exports an ordinary `defineDynamic` resolver. It qualifies provider
+tool keys and passes them through the existing source composition, binding,
+module-map, dynamic-tool, approval, authorization, and replay paths. Memory
+does not add a runtime tool contribution seam or a second executable registry.
 
 ```text
 turn.started          ---> recall(phase: "turn.started") ---> tools
@@ -139,16 +139,35 @@ surface, because provider tools are ordinary dynamic tools.
 
 Memory is a source-graph primitive, not a post-compile runtime attachment. For
 `agent/memory/user.ts`, the compiler selects and binds the memory definition,
-then projects its provider-tool adapter as a programmatic source at
-`tools/user.ts`:
+then instantiates its registered provider-tool template at `tools/user.ts`:
 
 ```text
 memory/user.ts
   └─ selected memory definition
-       └─ programmatic tools/user.ts wrapper
+       └─ registered tools/user.ts template instance
             └─ compiled defineDynamic resolver
                  └─ user__<provider tool key>
 ```
+
+The instantiation API takes the memory candidate as its anchor and dependency:
+
+```ts
+instantiateProgrammaticTemplate({
+  anchor: memoryCandidate,
+  dependencies: { memory: memoryCandidate },
+  logicalPath: `tools/${slot}.ts`,
+  owner: { kind: "framework", feature: "memory" },
+  parameters: { memoryExportName, memoryLogicalPath, slot },
+  template: memoryWrapperTemplate,
+});
+```
+
+The registry-issued template handle fixes the loader and module identity. The
+constructor inherits the anchor's node and layer, derives the source ID from
+the template, target path, and anchor, and rejects dependencies from another
+node. A direct target in the same layer wins over the derived module. Memory
+therefore needs no special registration mode, source layer, or source-ID
+convention.
 
 The memory definition and wrapper each use the canonical candidate,
 composition, binding-validation, artifact, module-map, hydration, and runtime
@@ -172,20 +191,24 @@ slot, `tools: false`, `null`, or an empty result contributes no tools. Provider
 factories use eve's durable callback helpers when their callbacks cannot be
 stamped by authored-source transformation.
 
+Every module-map load also memoizes zero-argument definition-factory results
+within each module namespace. The direct memory binding and its derived wrapper
+therefore receive the same definition and provider instance, including when a
+module exports `() => defineMemory(...)`. A separately loaded compile or runtime
+module map gets a fresh materialization; there is no process-global provider
+registry.
+
 This architecture is a completion gate for the core implementation:
 filesystem, in-memory, generated, bundled, and hydrated module maps must load
 the same selected wrapper binding, and a parked provider tool must resume after
 a fresh-process callback rebind. A memory-only registry, post-resolution tool
 merge, or logical-path loading fallback fails that gate.
 
-Current eve automatically rebinds missing callbacks by rerunning
-`session.started` resolvers; it does not rerun an arbitrary
-`turn.started` resolver. The memory core must therefore prove that loading the
-selected wrapper binding registers the callbacks needed by persisted provider
-tools, or extend cold rebinding at the generic compiled-resolver boundary. Any
-such extension belongs to the ordinary dynamic-tool lifecycle, contains no
-memory types, preserves the parked call's captured closure values, and never
-reruns provider mutation.
+Compiled dynamic resolvers may opt into missing-callback rebinding at the
+generic compiled-resolver boundary. The memory wrapper uses that path to
+register the callback implementation needed by a persisted provider tool. The
+rebind machinery contains no memory types, preserves the parked call's captured
+closure values, and never reruns provider mutation.
 
 ## Model-facing description
 
@@ -197,7 +220,7 @@ interface MemoryDefinition {
   readonly namespace?: MemoryNamespaceDefinition;
   readonly provider: MemoryProvider;
   readonly scope: MemoryScopeDefinition;
-  readonly tools?: boolean;
+  readonly tools?: false;
   readonly visibility?: MemoryVisibility;
 }
 ```
@@ -515,7 +538,7 @@ boundaries. `recall` is required. `capture` and `tools` are optional.
 ```ts
 import type { ModelMessage } from "ai";
 import type { SessionContext } from "eve/context";
-import type { DynamicResolveContext, ToolDefinition } from "eve/tools";
+import type { DynamicResolveContext } from "eve/tools";
 
 interface MemoryRecallMessage {
   /** Provider context recalled into durable model history as a user-role message. */
@@ -597,7 +620,7 @@ interface MemoryToolsContext extends DynamicResolveContext {
   };
 }
 
-type MemoryToolSet = Readonly<Record<string, ToolDefinition>>;
+type MemoryToolSet = Readonly<Record<string, MemoryToolDefinition>>;
 
 interface MemoryProvider {
   recall(context: MemoryRecallContext): MemoryRecallResult | Promise<MemoryRecallResult>;
@@ -794,14 +817,13 @@ omit, or do nothing.
 
 Compaction is the canonicalization boundary for memory records. Trusted
 internal code partitions every attributed memory record — across every slot,
-namespace, scope, and visibility mode — away from ordinary conversation, folds
-keyed records to one live version or deletion tombstone per identity, and
-retains every immutable unkeyed append with its attribution. The free-form
-summary prompt is built from projected ordinary conversation only; no
-attributed memory record of any kind enters it. Canonical private memory state
-is reattached to the rewritten history, so items hidden from the currently
-active scope survive another scope's compaction, and sparse retrieval stays
-accumulative across compaction.
+namespace, scope, and visibility mode — away from ordinary conversation, keeps
+one live keyed record per identity, and retains every immutable unkeyed append
+with its attribution. The free-form summary prompt is built from projected
+ordinary conversation only; no attributed memory record of any kind enters it.
+Canonical private memory state is reattached to the rewritten history, so items
+hidden from the currently active scope survive another scope's compaction, and
+sparse retrieval stays accumulative across compaction.
 
 Because superseded versions and hidden scopes can grow while the visible
 prompt stays constant, eve triggers canonicalization on raw attributed-record
@@ -1146,9 +1168,10 @@ The generic prerequisites are on `main`: the projected-history seam
 ([#2352](https://github.com/vercel/eve/pull/2352)), durable dynamic callbacks
 ([#2354](https://github.com/vercel/eve/pull/2354)), name-and-phase callback
 identity ([#2384](https://github.com/vercel/eve/pull/2384)), and the canonical
-source graph and binding table
+source graph, binding table, and invariant-safe derived-template API
 ([#2404](https://github.com/vercel/eve/pull/2404),
-[#2516](https://github.com/vercel/eve/pull/2516)). The runtime-tool contribution
+[#2516](https://github.com/vercel/eve/pull/2516),
+[#2539](https://github.com/vercel/eve/pull/2539)). The runtime-tool contribution
 proposal in [#2347](https://github.com/vercel/eve/issues/2347) is superseded;
 memory uses the compile-time wrapper described above. The kernel-effects
 follow-up to #2516 is not a prerequisite because provider tools have ordinary
@@ -1157,12 +1180,13 @@ executors rather than native kernel effects.
 Implementation proceeds in two pull requests:
 
 1. The first-class memory core is implemented in
-   [#2534](https://github.com/vercel/eve/pull/2534), based directly on current
-   `main`. It adds the authoring and compiler surface, source-graph wrapper,
-   namespace and scope locks, recall records and projection, lifecycle,
-   compaction, agent-info v4, published documentation, and deterministic
-   end-to-end coverage. It does not restack or reuse the custom runtime
-   lifecycle from [#2142](https://github.com/vercel/eve/pull/2142).
+   [#2534](https://github.com/vercel/eve/pull/2534), rebased directly onto
+   current `main`. It uses #2539's registered template API and generic
+   per-module-map factory materialization, then adds the memory authoring and
+   compiler surface, namespace and scope locks, recall records and projection,
+   lifecycle, compaction, agent-info v4, published documentation, and
+   deterministic end-to-end coverage. It does not restack or reuse the custom
+   runtime lifecycle from [#2142](https://github.com/vercel/eve/pull/2142).
 2. After #2534 merges, the bounded `fileMemory()` provider in
    [#2144](https://github.com/vercel/eve/pull/2144) will be rebased onto current
    `main`. It retains only provider storage, document, backend, and concurrency
@@ -1170,7 +1194,8 @@ Implementation proceeds in two pull requests:
    [#2145](https://github.com/vercel/eve/pull/2145) is superseded.
 
 #2534 implements the core boundary through one selected source and binding
-authority for every memory definition and provider-tool wrapper, one projected
+authority for every memory definition and provider-tool wrapper, one
+materialization per definition factory in a module-map load, one projected
 history for message-bearing consumers, private memory canonicalization outside
 ordinary summarization, and generic cold callback rebinding with captured
 closures. Those invariants, the committed E2E fixture, and the required CI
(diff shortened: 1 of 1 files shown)


research/first-class-memory.md
```diff
@@ -0,0 +1,270 @@
...
+});
+```
+
+`provider` supplies the memory behavior. `scope` resolves the provider partition for the current turn. Descriptions, prompts, tool policy, and storage options belong to provider configuration rather than `defineMemory`.
+
+The same provider instance may back several slots. For example, `user.ts` may resolve `[userId]` while `workspace.ts` resolves `[workspaceId]`. Their path identities, scope keys, context contributions, tools, and event invocations remain separate.
+
+Local directory-form subagents may declare their own memory slots. Graph-node identity keeps root and subagent slots isolated even when they use the same path, scope parts, and provider.
+
+## Scope
```
chadhietala, line 333: We likely need to think of where permissions sit. The tuple might not be enough to express access control. For instance you might want to have a read-only memory (organizational knowledge) + read-write memory (channel knowledge).

research/first-class-memory.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+Scope parts come from trusted auth, application, or channel context. They cannot come from model input or unattested message fields. Parts have no built-in entity types, and no position has special meaning. `[userId, channelId]` means only that both values participate in the provider partition. `byPrincipal()` is sugar for the common authenticated-principal tuple.
+
+For each resolved slot, eve creates this scope value:
+
+```ts
+interface MemoryScope {
+  /** Stable eve namespace derived from application, environment, graph node, slot, and parts. */
+  readonly key: string;
```
chadhietala: Will this guarantee uniqueness?

ruiconti: This is up to user-space IMO.

research/first-class-memory.md
```diff
@@ -0,0 +1,270 @@
...
+
+After the compacted checkpoint has been written to durable history, eve dispatches `compaction.completed`. Each handler receives the settled post-compaction durable history. The event occurs only after successful compaction; a failed summarization preserves the prior history and does not emit it. Provider tools are not resolved at either boundary because compaction does not invoke the agent model.
+
+Both snapshots exclude transient provider context. If compaction occurs later in a multi-step turn, eve keeps the prepared-turn context separate from the compaction input and reapplies it to the remaining model steps without writing it into the checkpoint.
+
+### Turn preparation
+
+After turn admission and any required initial compaction, eve dispatches `turn.prepared` to the `events` map of each slot whose scope resolved at turn start. This is a new provider-resolution event rather than the existing stream `turn.started`: its name reflects that the newest input has been admitted and the model history has already been compacted. The event context includes that input and every model-visible user, assistant, and tool message retained after compaction.
```
ruiconti: In a HITL mid-turn user approval, I'm assuming that `turn.prepared` runs _after_ tool calls that have been approved so it has access to the most up-to-date moment before a tool call?

Not actionable but wanted to share - reading this I realized that there is a fundamental distinction between events that are supposed to be observational (what we currently expose in hooks) vs the ones that are supposed to be interfering (what `turn.prepared` does). `defineDynamic` changed this a bit but I think there is something to be said in having a clear distinction between those from a user-space POV, so it's clear on the lifecycles you can context-engineer the agent around.

research/first-class-memory.md
```diff
@@ -0,0 +1,431 @@
...
+        return {
+          forget: defineTool({
+            description: "Forget one saved preference.",
+            inputSchema: z.object({ key: z.string() }),
+            execute: ({ key }) => service.forget(scope, key),
+          }),
+        };
+      },
+    },
+  }),
```
ruiconti: I don't follow why would we want to define tools on a per-step (or even per turn) basis.

Assuming this is merely illustrative but would be useful to go over the reasoning/motivation for having dynamic tools to implement memory.

AndrewBarba: Depending on current user you could choose to add tools or not. Or add read only tools and not wrote. Etc.

research/first-class-memory.md
```diff
@@ -0,0 +1,431 @@
...
+
+After the compacted checkpoint has been written to durable history, eve dispatches `compaction.completed`. Each handler receives the settled post-compaction durable history and may return memory messages to append after the checkpoint. The event occurs only after successful compaction; a failed summarization preserves the prior history and does not emit it. Provider tools are not resolved at either boundary because compaction does not invoke the agent model.
+
+Both compaction boundaries are ordinary public lifecycle events: they are emitted in the session stream and available to hooks as well as memory providers. Memory adds no private compaction callback or synthetic `turn.prepared` event.
```
ruiconti: Assuming here that we'd now be exposing the current turn's message (ie the latest message) in `turn.started`?

research/first-class-memory.md
```diff
@@ -0,0 +1,431 @@
...
+import { byPrincipal, defineMemory } from "eve/memory";
+import { fileMemory, inMemory } from "eve/memory/file";
+
+const backend = inMemory();
+
+export default defineMemory({
+  provider: fileMemory({ backend, memoryLimit: 200 }),
+  scope: byPrincipal(),
+});
+```
```
ruiconti, line 492: I hadn't thought of `fileMemory` being a built-in provider and in hindsight it makes a lot of sense. Really fits in the framework IMO.

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+| ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------- |
+| `"scope"` (default) | Include only the projection whose scope key matches the active turn. Exclude projections recalled for earlier participants.                 | Filtering an anchored projection changes the prompt prefix and invalidates the affected cache.     |
+| `"session"`         | Keep projections recalled for earlier scopes visible, then add the active scope's first projection immediately before its first turn input. | A scope change keeps earlier messages in place; replacing or clearing a projection may still bust. |
+
+The default favors projection isolation over cache reuse when the authenticated
+participant changes.
+
+For a Slack thread where one authenticated participant follows another,
+`"scope"` removes the first participant's recalled memory before the second
```
benpankow, line 509: Do you envision `scope` mostly being used for e.g. user preferences?

I'd imagine for slack an implementation would typically want to store actual private information in some sort of scope for just a user's private channel/DMs rather than bound to their user ID, so that it never leaks in other channels?

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+projections in anchor order. Projections created at the same boundary use stable
+slot-path order. The slot and scope attribution remain attached to each
+projection throughout model assembly.
+
+A projection remains separate from ordinary conversation history:
+
+- It is not emitted as `message.received`.
+- It is not summarized by compaction.
+- It is not included in `ctx.messages` passed back to a provider.
+- Its slot and scope attribution are not exposed to the model as tool input.
```
benpankow: Wouldn't slot/scope attribution be useful for the model in determining which provider to call `__save_memory` on?

e.g. a user asks to update how a bot responds in a thread, the model needs to determine which existing preference (channel, user etc) to update based on this attribution?

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+query.
+
+The result is ready before the first model call. If automatic compaction runs at
+the same opening boundary, the projection remains outside compaction input and
+the post-compaction recall may replace it before the model runs.
+
+### Compaction save and recall
+
+Before automatic or manual compaction rewrites history, eve calls an implemented
+`save` method with `phase: "compaction.requested"`. The provider receives the
```
benpankow: Is there any mechanism for providers to negotiate, or be aware of which context other providers save? Wondering about duplicate writes here

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+    readonly scope: MemoryScope;
+    /** Path-derived slot identity, such as "memory" or "user". */
+    readonly slot: string;
+  };
+}
+
+type MemoryRecallContext = MemoryOperationContext &
+  (
+    | {
+        readonly phase: "turn.started";
```
ruiconti, line 582: Why not `event`?

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+```ts
+import { defineMemoryProvider } from "eve/memory";
+import { defineTool } from "eve/tools";
+
+const customMemory = defineMemoryProvider({
+  async recall(ctx) {
+    return loadProjection(ctx.memory.scope.key);
+  },
+  tools: async (ctx) => {
+    const policy = await loadToolPolicy(ctx.session.auth.current);
```
chadhietala: 👍

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+Each slot stores at most one projection for every scope key it encounters in a
+session. The durable projection state includes the slot, scope key, content, and
+prompt anchor. `ctx.memory.current` exposes only the projection belonging to the
+active scope.
+
+The first valid projection for a scope anchors one synthetic user-role
+message immediately before that scope's current turn input. Later recall results
+update the projection at the same anchor according to the result table above.
+Replacing, clearing, or filtering an anchored projection changes the prompt
+prefix and may invalidate the provider prompt cache.
```
ruiconti: Curious about this behavior or the consideration between this versus appending recall updates to the session history? Feel like this could be a very easy and potentially expensive foot-gun, but speaking out of ignorance of memory implementations.

Is it that memory is not necessarily additive? e.g.,

```
0:   recall('user:UAB129'): "rui's favorite color is orange"
N:   ...
N-1: recall('user:UAB129'): "rui's favorite color is blue"
```

and thus, we'd need to always update entry `0` to prevent confusion?

Though this could be worked around by having more granular memory scopes? e.g.,

```
memory/
  user.ts # general memory
  user_pref_color.ts # specific memory
```

Would then allow for additive memories to not require a full cache bust.

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+  },
+});
+```
+
+The result of `recall` updates the active scope's projection:
+
+| Result                          | Effect                                                  |
+| ------------------------------- | ------------------------------------------------------- |
+| `{ content: non-empty string }` | Replace the current projection for this slot and scope. |
+| `null`                          | Clear the current projection for this slot and scope.   |
```
chadhietala: Should we just provide a sentinel value here instead of `null` even if we bind it to `null`?

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+
+```ts
+import { defineMemoryProvider } from "eve/memory";
+import { defineTool } from "eve/tools";
+
+const customMemory = defineMemoryProvider({
+  async recall(ctx) {
+    return loadProjection(ctx.memory.scope.key);
+  },
+  tools: async (ctx) => {
```
chadhietala: What is the interaction between `recall` and a tool? Is it just that one runs deterministically and the other one runs stochastically based on the model?

research/first-class-memory.md
AndrewBarba: Intention is for recall to be additive as user messages. will update this. could also make it configurable, potentially provider configurable?

research/first-class-memory.md
AndrewBarba: providers can always inspect the full message array. we can attach metadata to those messages to indicate it came from a memory provider and attach the scope.

research/first-class-memory.md
AndrewBarba: if the eve agent defines multiple memory providers then yes the model does need to reason about which one to use as there could be multiple tools that are similarly named. we probably need a description somewhere so additional context about that scope can be provided

research/first-class-memory.md
AndrewBarba: scope is one way to achieve user preferences - make scope user id. but scope goes a lot further you could have multiple fileMemory memory files where scope on one of them is channel specific and scope on another is user specific

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+
+## Memory projection
+
+Each slot stores at most one projection for every scope key it encounters in a
+session. The durable projection state includes the slot, scope key, content, and
+prompt anchor. `ctx.memory.current` exposes only the projection belonging to the
+active scope.
+
+The first valid projection for a scope anchors one synthetic user-role
+message immediately before that scope's current turn input. Later recall results
```
chadhietala: Shouldn't we insert the synthetic recall message always as the true suffix? Since the model never asked for them there is no reasoning issue, it just means that synthetic recall messages are not part of the prefix cache.

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+import type { ModelMessage } from "ai";
+import type { SessionContext } from "eve/context";
+import type { DynamicResolveContext, ToolDefinition } from "eve/tools";
+
+interface MemoryProjection {
+  /** Non-empty provider context projected as one synthetic user-role message. */
+  readonly content: string;
+}
+
+interface MemoryTurnContext {
```
chadhietala, line 557: We should likely add optional linkage to trace and any token usage / cost here as well. A memory/continuous learning system likely wants to incorporate cost, errors, time etc into any synthesis it does

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+returns the recorded result. Provider mutations belong in the returned tool's
+`execute` function. eve never substitutes the current turn's definition or
+scope for a parked call's captured definition.
+
+### Completed-turn save
+
+After a turn reaches `turn.completed`, eve calls an implemented `save` method
+with `phase: "turn.completed"`. The provider receives the completed turn input
+and the settled durable history, including the assistant response and tool
+results. The method does not run for failed, cancelled, input-deferred, or
```
chadhietala: It may not run but we should probably propagate these things when they do run

research/first-class-memory.md
ruiconti: Being configurable makes sense.

Generally, as a user, I would want crystal clear understanding of when and how the harness is making changes to the session history. To me, that's the most sensible boundary as that directly influences:
- What potentially sensitive data is shared with the model (and thus, to other users)
- The efficiency of the session's prompt cache

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+
+```ts title="agent/memory/user.ts"
+import { defineMemory } from "eve/memory";
+import { byPrincipal } from "eve/memory/scope";
+import { customMemory } from "../lib/custom-memory";
+
+export default defineMemory({
+  provider: customMemory,
+  scope: byPrincipal,
+  visibility: "session",
```
ctgowrie, line 490: Can't you enforce the same behavior via scope? I think arbitrary strings are hard to reason about and can easily introduce sprawl

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+});
+```
+
+The result of `recall` updates the active scope's projection:
+
+| Result                          | Effect                                                  |
+| ------------------------------- | ------------------------------------------------------- |
+| `{ content: non-empty string }` | Replace the current projection for this slot and scope. |
+| `null`                          | Clear the current projection for this slot and scope.   |
+| `undefined` / no return         | Keep the current projection without changing it.        |
```
ctgowrie: Not a fan of the magic meaning behind these falsy values. I'd say falsy clears the value, if you want to keep it the same then return it

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+
+```ts title="agent/memory/user.ts"
+import { defineMemory } from "eve/memory";
+import { byPrincipal } from "eve/memory/scope";
+import { customMemory } from "../lib/custom-memory";
+
+export default defineMemory({
+  provider: customMemory,
+  namespace: () => `acme:${process.env.DEPLOYMENT_REGION ?? "local"}`,
+  scope: byPrincipal,
```
ctgowrie: Combine scope and namespace perhaps?

research/first-class-memory.md
```diff
@@ -0,0 +1,657 @@
...
+
+This option never changes provider scope. `recall`, `save`, and `tools` still
+receive only the active turn's locked scope. The provider cannot select the
+visibility mode, and eve never passes another scope or its projection into a
+provider method. A `null` scope suppresses all of the slot's projections in both
+modes.
+
+Projection visibility does not remove ordinary user, assistant, or tool
+messages already in the conversation. It also cannot undo information an
+earlier assistant response derived from memory. Applications that require hard
```
cmpadden: > Applications that require hard isolation between participants must use separate sessions.

Highly appreciate this being called out. 👍

research/first-class-memory.md
AndrewBarba: they used to be and I purposefully separated because the namespace has a set of defaults that scope shouldn't. specifically namespace by default on vercel uses things like project id. I want users to be able to override this, maybe they really do want cross-project memory

research/first-class-memory.md
AndrewBarba: No it solves a different thing. ex. multi-player multiple people start turns:
- session: previous user memories remain in message history. no cache busting
- scope: previous messages that dont match new scope are removed from history. busts cache on scope change.

research/first-class-memory.md
AndrewBarba: correct. a provider may decide to add a recall tool, that's up to them, but we will deterministically run recall here as well. providers can choose to ignore it that's on them

research/first-class-memory.md
AndrewBarba: The final API retains visibility as the constrained scope or session policy. scope filters attributed recall for that slot to the active scope key; session leaves all recalled messages in durable order. This does not alter provider scope. Prompt accounting is invalidated only when the visible recalled-message set actually changes. The implementation includes a second-participant coverage case in #2142.

research/first-class-memory.md
AndrewBarba: The final contract is append-only. A non-empty recall result is appended as an exact role: user message; null, undefined, no return, or blank content appends nothing. Earlier recall is never replaced or cleared. Session visibility therefore preserves the existing prefix, while scope visibility may filter attributed recall after a scope change and intentionally invalidate the cache. This is implemented and covered in #2142.

research/first-class-memory.md
AndrewBarba: The replacement and clear states have been removed, so a sentinel is no longer needed. MemoryRecallResult is a user message, null, or undefined. null, undefined, no return, and blank content all mean append nothing; false is not accepted, and no result can clear prior history. Implemented in #2142.

research/first-class-memory.md
AndrewBarba: The final placement keeps the actual turn input as the true suffix: turn-start recall is appended immediately before the admitted turn input, not after it. That preserves additive cache behavior while keeping the current human request as the final user message. At compaction.completed, recall is appended after the rewritten checkpoint and tail because there is no new turn input at that boundary. No prior message is replaced.

research/first-class-memory.md
AndrewBarba: Resolved in the final lifecycle. turn.started recall receives prior durable history in ctx.messages and the newly admitted normalized input separately in ctx.turn.input; the input is not folded into history until recall settles. The async tools(ctx) resolver runs afterward and sees prior history, appended recall messages, and the admitted turn input. The implementation and integration coverage are in #2142.

research/first-class-memory.md
AndrewBarba: The projection semantics that made this ambiguous are gone. false is not accepted. null, undefined, no return, or blank content appends nothing and never clears anything; a non-empty exact role: user message is appended. There is no longer a keep-current versus clear branch. Implemented in #2142.

research/first-class-memory.md
AndrewBarba: The final distinction is deterministic lifecycle recall versus model-selected tools. recall runs at turn.started and compaction.completed. tools(ctx) resolves once after turn-start recall through the existing defineDynamic machinery, and the model decides whether to call an exposed tool. A provider can offer optional search or deeper retrieval as a tool, but tools are not required for the baseline recall path. Implemented and tested in #2142.

research/first-class-memory.md
AndrewBarba: Implemented without adding provider coordination. Every recalled item is an attributed durable user message. Provider contexts receive the durable ctx.messages history, and getMemoryMessageAttribution(message) exposes the slot and locked scope for earlier recalls. That lets a provider deduplicate its own writes or snapshots without depending on private metadata; fileMemory uses it to suppress an identical latest recall for the same slot and scope. Providers remain independent and there is no implicit cross-provider mutation.

research/first-class-memory.md
AndrewBarba: The final contract keeps scope generic rather than principal-specific. Custom scope resolvers receive authenticated session and channel metadata, so an agent can partition one slot by principal and another by team, channel, or conversation. byPrincipal is only the convenience resolver and intentionally follows the caller across channels; the proposal now shows a channel-aware resolver and calls out that hard isolation requires separate sessions. This is implemented in #2142.

research/first-class-memory.md
AndrewBarba: We kept phase deliberately. The callback name already identifies the operation as recall or save, and phase discriminates the lifecycle boundary at which that operation runs. event elsewhere in eve generally refers to an actual stream, channel, or webhook event object, so using it here would blur two different concepts.

research/first-class-memory.md
AndrewBarba: Implemented as an optional static description on defineMemory. eve prepends that model-facing purpose to every tool exposed by the slot, while slot qualification keeps tool names distinct. Namespace and scope values remain private authorization and partitioning inputs rather than prompt content. The proposal, public docs, unit coverage, fileMemory integration, and e2e fixture all use the final API.

research/first-class-memory.md
AndrewBarba: We are keeping turn.completed save limited to successfully completed turns. Failed, cancelled, deferred, and adapter-consumed outcomes remain instrumentation concerns and do not invoke memory save. If outcome-driven learning becomes a requirement later, it should be introduced as an explicit typed lifecycle boundary rather than optional terminal fields on the successful save context.

research/first-class-memory.md
AndrewBarba: We decided not to add telemetry to the memory provider contract right now. Completed-turn save receives the semantic material it owns: durable messages, turn input, session.id, and turnId. Token usage, cost, latency, errors, and traces are instrumentation concerns; cost is not a stable cross-model value, and adding those fields would turn memory into a second observability API. Providers that consume instrumentation can correlate it with session.id and turnId. The compaction input-token field remains because it directly describes the context being compacted.

research/first-class-memory.md
AndrewBarba: The final API keeps them separate. namespace identifies the application or deployment domain and has the deployment-aware defaultNamespace implementation; callers can replace it completely when they intentionally want a different domain such as cross-project memory. scope identifies the trusted audience or container inside that namespace. The provider key is derived from exactly those two resolved values, with no hidden suffixes. Combining them would conflate deployment isolation with authorization scope.

ruiconti (approved):

PR #1587, opened 2026-08-03
feat(eve): add aligned session operation handles

ctgowrie:
Adds the non-breaking author-facing layer on top of the shared command inbox.

- adds fixed-ID session handles through `attachSession(sessionId)`
- adds additive ID-only HTTP plumbing and `client.sessions.create/attach` for the later route cutover
- retains every existing continuation-token helper and route for the later single breaking migration PR
- documents and tests the additive session/client APIs, including rejection of continuation tokens by ID-only operations
- deliberately does not expose `ChannelAddress`; direct channel operations land in the breaking migration PR

The final singular HTTP contract is documented and activated atomically in #1597. This PR does not publish its temporary non-conflicting route paths as the public API.

Checks:
- focused handle and channel unit tests (199 passed)
- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm guard:invariants`
- `pnpm docs:check`


ctgowrie:
Collapsed into #1597 so the session handles and breaking public migration are reviewed and merged as one implementation PR.

Issue #742, opened 2026-07-13
Local self-modification: dev-only selfmod subagent via eve/extensions/selfmod


ruiconti:
Let an agent running under `eve dev` read and modify its own source tree, so a developer can iterate on the agent by talking to it.

Proposed shape (full design in `research/local-self-modification.md`, PR #491):

- Declared as a subagent module, mirroring `defineRemoteAgent`:

  ```ts
  // agent/subagents/selfmod.ts
  import { defineSelfModifyingAgent } from "eve/extensions/selfmod";

  export default defineSelfModifyingAgent({ development: true });
  ```

- The capability is a framework-owned subagent authored as files in the eve package (`extensions/selfmod/`), grafted into the compiled manifest + module map only when the compile target is development. Hosted artifacts never contain it; dev and prod runtimes load artifacts identically.
- The lowering step is the internal `AgentExtension` contract (packaged agent files in, ordinary `CompiledSubagentNode` out), reusable by future `eve/extensions/<name>` capabilities.
- Its sandbox is the just-bash backend mapped over the real project root: default harness tools on the live tree, `ReadWriteFs` containment, pure-JS shell with no host binaries. Self-edits take effect through the existing dev source watcher.

New compiler input required: `compileAgent` gains `target: "development" | "hosted"`, folded into compile metadata and cache keys.

Open points:

- Dev and hosted compiles of the same tree produce different manifests; target participation in cache keys needs scrutiny.
- No real binaries inside the subagent; allowlisted host commands are a possible follow-up.
- `.eve/` is visible to the subagent (prompt-discouraged, not blocked) in v1.

PR #1632, opened 2026-08-04
feat(eve): add local self-modification

prha:
### Description

Introduces the public `eve-selfmod` package, a development-only source-editing subagent for eve applications.

The package provides:

- a declared subagent enabled only under `eve dev`;
- a just-bash sandbox that mounts the authored `agent/` tree read-write at `/source`;
- instructions for inspecting and directly editing authored source with eve's default bash and filesystem tools;
- agent, instructions, and sandbox exports for registry scaffolding.
- a skill for authoring eve agents pointed at the bundled eve docs

It also teaches the authored-module bundler to honor the `eve-source` export condition so linked workspace capability packages can run from source before their distribution build exists.

### Rationale

Developers should be able to iterate on an agent by talking to it. Packaging source editing as a development-only subagent isolates the writable source mount from the root agent while allowing local edits without an approval round trip.

Addresses #742.

### Testing

- `pnpm --filter eve-selfmod typecheck`
- `pnpm --filter eve-selfmod build`
- `pnpm --filter eve build`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/internal/authored-module-loader.scenario.test.ts`


packages/eve-selfmod/README.md
```diff
@@ -0,0 +1,52 @@
...
+    └── replace_in_file.ts
+```
+
+```ts
+// agent/subagents/selfmod/config.ts
+import { defineSelfmod } from "eve-selfmod/config";
+
+export const selfmod = defineSelfmod({
+  model: "anthropic/claude-sonnet-5",
+});
```
ctgowrie: What's the purpose of this config file? I'd expect to just defineSelfmodAgent with a model in agent.ts, otherwise just export agent from selfmod

packages/eve-selfmod/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+// agent/subagents/selfmod/agent.ts
+import { selfmod } from "./config";
+export default selfmod.agent;
+
+// agent/subagents/selfmod/sandbox.ts
+import { selfmod } from "./config";
+export default selfmod.sandbox;
```
ctgowrie: same: why not `export sandbox from selfmod`

packages/eve-self-modification/extension/tools/replace_in_file.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (content.indexOf(oldText, firstMatch + oldText.length) !== -1) {
+    throw new Error(
+      "oldText occurs more than once. Include more surrounding text to make it unique.",
+    );
+  }
+
+  return content.slice(0, firstMatch) + newText + content.slice(firstMatch + oldText.length);
+}
+
+export default defineTool({
```
ctgowrie: Why add this tool instead of base tool set?

packages/eve-selfmod/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+// agent/subagents/selfmod/extensions/selfmod.ts
+export { default } from "eve-selfmod";
+
+// agent/subagents/selfmod/tools/bash.ts
+import { disableTool } from "eve/tools";
+export default disableTool();
```
ctgowrie: Why?

packages/eve-selfmod/extension/instructions.ts
```diff
@@ -0,0 +1,17 @@
...
+
+Use write_file only to create a path that does not exist. Never use write_file to modify an existing file, even if you created or read it earlier in the turn. Use replace_in_file for every edit to an existing file, including a whole-file replacement. Its oldText must match exactly once, so copy enough surrounding text from read_file to make the replacement unique.
+
+Treat a successful file-edit tool result as confirmation; do not reread a file solely to verify that the edit succeeded. Do not approximate unavailable build or test commands with broad source searches.
+
+Source changes do not affect the current turn; after a successful development rebuild, they take effect on the next turn. Return a concise summary of files changed and any validation limitations.`,
```
ctgowrie: Not sure this line is necessary

packages/eve-selfmod/extension/instructions.ts
```diff
@@ -0,0 +1,17 @@
+import { defineInstructions } from "eve/instructions";
+
+export default defineInstructions({
+  markdown: `You are a source editor for an eve application.
+
+The live authored agent directory is mounted read-write at /source. The installed eve version's canonical documentation is mounted read-only at /eve-docs. Load the Eve authoring skill and read /eve-docs/README.md plus the guides relevant to the framework slots you will change before writing Eve code. Request independent documentation and source reads together. Use /workspace only for temporary files. The shell is intentionally unavailable. Runtime skill paths under $HOME/.agents and /workspace/skills belong to the caller's sandbox, are not authored source, and may not exist here. Never search those runtime paths even when the caller mentions them; discover the corresponding authored definition under /source instead. /source is the mounted authored agent directory, not the application or package root. Do not look for package.json, src/, or other project-level files. Do not assume conventional filenames such as tools/index.ts or skills/<name>/SKILL.md. Eve definitions are filesystem-first and may be flat files.
```
ctgowrie: Do we need the eve authoring skill or just read the docs?

packages/eve-selfmod/README.md
prha: I wanted a place to put additional configuration, which would affect more than just the agent.

For example, if we wanted to support the production workflow where we're publishing PRs instead of modifying source directly, this provides a single place to configure the git provider information, the sandbox, etc.

packages/eve-selfmod/README.md
prha: I was seeing a lot of bad performance on real repos (e.g. e0) where it wasn't parallelizing bash calls, because they were context (working directory) dependent.  I was able to coax better performance by forcing through the explicit read_file / write_file tool calls.

packages/eve-selfmod/extension/instructions.ts
prha: I can try making this more terse... I observed some runaway loops where the agent itself would try to confirm changed behavior and would burn tokens because it could not see the effects of its changes.

packages/eve-selfmod/extension/instructions.ts
prha: will try stripping this down to just point at the docs and remove the skill.

packages/eve-self-modification/extension/tools/replace_in_file.ts
prha: Might be an unnecessary micro-optimizations, but I'd see occasional attempts to write files without first reading them, despite the instruction.

packages/eve-self-modification/src/agent.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/** Default model used by self-modification when `defineSelfModification` does not override it. */
+export const DEFAULT_SELF_MODIFICATION_MODEL = "anthropic/claude-sonnet-5";
+
+/** Creates the development-only self-modification dynamic subagent definition. */
+export function createSelfModificationAgent(
+  model: AgentStaticModelDefinition = DEFAULT_SELF_MODIFICATION_MODEL,
+) {
+  return defineDynamic({
+    build: { externalDependencies: ["@eve/self-modification"] },
```
ctgowrie: Why is this an external dependency?

ctgowrie (approved):

PR #1649, opened 2026-08-04
Fix Markdown canonical ownership

JohnPhamous:
- Remove canonicals from smart Markdown misses while preserving the existing `200` and `noindex` behavior.
- Emit one canonical HTML URL for real Markdown pages.


allenzhou101 (approved):

JohnPhamous:
Upstream follow-up: vercel/geistdocs#224 moves canonical ownership into Geistdocs, which owns both the Markdown proxy rewrite and page resolution. After that release is adopted, the eve-specific wrapper in this PR can be removed.

PR #1650, opened 2026-08-04
Stabilize docs RSS dates

JohnPhamous:
- Stop assigning request-time dates to docs without source modification metadata.
- Keep every feed item with a stable fallback date while honoring real modification dates.


allenzhou101 (approved):
Reviewed locally. The stable RSS fallback makes sense as an eve-specific policy fix; a configurable deterministic fallback could be considered upstream in geistdocs.

PR #1653, opened 2026-08-04
Add responsible use to docs navigation

JohnPhamous:
- Link the responsible-use page from the server-rendered docs navigation.
- Remove its checker exemption so future unlinked pages fail the docs check.

### DIFF
--- a/docs/meta.json
+++ b/docs/meta.json
@@ -22,6 +22,7 @@
     "guides",
     "concepts",
     "patterns",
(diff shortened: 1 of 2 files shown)


allenzhou101 (approved):
Reviewed locally. This is correctly scoped to eve: it makes Responsible Use part of the docs navigation and enforces the repository sidebar-coverage policy.

PR #1676, opened 2026-08-05
fix(eve): compose the bridge with registered telemetry integrations

chadhietala:
Third in the instrumentation-provider stack, on top of #1670.

## Why

The AI SDK treats a per-call `integrations` list as replacing the globally registered ones, not adding to them:

> When provided, these integrations will take precedence over the globally registered integrations for this call.

eve passes `integrations` whenever it attaches its instrumentation bridge, so anything an authored `agent/instrumentation.ts` installed with `registerTelemetry` stopped receiving events for exactly the calls it cared about. eve worked around this by hand-recomposing `createOtelIntegration()`, which covered eve's own OpenTelemetry integration and nothing else.

That is survivable today only because the bridge is dev-only. It has to be correct before the bridge runs everywhere, which the provider work depends on.

## Approach

Read the registered integrations and compose the bridge ahead of them.

```ts
integrations:
  bridgeIntegration === undefined
    ? undefined
    : [bridgeIntegration, ...getRegisteredTelemetryIntegrations()],
```

Both shapes eve produces today are unchanged. With no authored instrumentation eve never calls `ensureOtelIntegration`, so nothing is registered and the list is just the bridge. With authored instrumentation eve's OpenTelemetry integration is registered, so the list is the bridge and that integration — the same pair as before, now the registered instance rather than a fresh one per call. What changes is that an integration eve did not install also survives.

`otel-integration.ts` becomes `ai-sdk-telemetry.ts`, since it no longer only wraps the OTel integration, and `createOtelIntegration` stops being exported for per-call composition.

## Verification

- `pnpm fmt`, `pnpm lint`, `pnpm guard:invariants`
- `pnpm test:unit` — 5,893 passed
- `pnpm test:integration` — 575 passed
- Falsified the new coverage: reverting `enrichTelemetry` to `[bridgeIntegration]` fails `composes the bridge with every registered integration`.

### DIFF
--- a/.changeset/quiet-integrations-compose.md
+++ b/.changeset/quiet-integrations-compose.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Telemetry integrations registered with the AI SDK's `registerTelemetry` no longer stop receiving events once eve attaches its own instrumentation bridge to a model call. eve now composes its bridge with every registered integration instead of replacing them.
--- a/packages/eve/src/harness/ai-sdk-telemetry.test.ts
+++ b/packages/eve/src/harness/ai-sdk-telemetry.test.ts
@@ -1,0 +1,25 @@
+import type { Telemetry } from "ai";
+import { afterEach, describe, expect, it } from "vitest";
+
+import { getRegisteredTelemetryIntegrations } from "#harness/ai-sdk-telemetry.js";
+
+describe("getRegisteredTelemetryIntegrations", () => {
+  const original = globalThis.AI_SDK_TELEMETRY_INTEGRATIONS;
+
+  afterEach(() => {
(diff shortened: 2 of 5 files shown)


chadhietala:
Not needed yet. The bridge only exists when `instrumentationPluginPath === undefined`, so no authored instrumentation — and therefore nothing registered — can coexist with it. `AI_SDK_TELEMETRY_INTEGRATIONS` is always empty when eve passes `integrations`, and the `[bridge, createOtelIntegration()]` branch this replaced is unreachable.

This becomes live once #1661 runs structural agent instrumentation alongside authored instrumentation. Branch `instr/03-compose-integrations` (ac4520d4) is left in place to rebase onto that.

PR #1680, opened 2026-08-05
refactor(eve): name the step lifecycle events for the span they produce

chadhietala:
Fourth in the instrumentation-provider stack, on top of #1677.

## Why

The bus called one model attempt `attempt.*` while the span it produces is `agent.step`. A provider handler and the span it writes disagreed on the noun.

The proposal makes the bus a public provider contract, and its event list is `step.started` / `step.completed` / `step.failed` / `step.metadata`. Doing the rename first means that surface never ships `attempt.*` and renames a public event name immediately after.

## Approach

Mechanical rename of the four event names and their payload types:

| before | after |
| --- | --- |
| `attempt.started` | `step.started` |
| `attempt.completed` | `step.completed` |
| `attempt.failed` | `step.failed` |
| `attempt.metadata` | `step.metadata` |
| `InstrumentationAttemptStartedEvent` | `InstrumentationStepStartedLifecycleEvent` |
| `InstrumentationAttemptTerminalEvent` | `InstrumentationStepTerminalEvent` |
| `InstrumentationAttemptMetadataEvent` | `InstrumentationStepMetadataEvent` |

`InstrumentationAttemptScope` deliberately keeps its name — it identifies one *attempt* of a step, which is exactly what `agent.step.attempt` records.

The started payload takes the `…LifecycleEvent` suffix because `InstrumentationStepStartedEventInput` already exists on the public surface for the unrelated `step.started` runtime-context hook. That collision resolves when the proposal deletes that hook; until then the two names stay distinct.

## Scope

Publishes nothing — these events are internal. No changeset-visible behaviour change; included as `patch` because it touches published `eve` source.

## Verification

- `pnpm fmt`, `pnpm lint`, `pnpm guard:invariants`
- `pnpm test:unit` — 5,894 passed
- `pnpm test:integration` — 575 passed

### DIFF
--- a/.changeset/lucky-actions-kind.md
+++ b/.changeset/lucky-actions-kind.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Trace spans now record what eve dispatched each action as. `agent.action.kind` was always `"tool"`, so a subagent or remote-agent call was indistinguishable from an ordinary tool in a trace; it now reports `subagent-call`, `remote-agent-call`, or `tool-call`, matching the kind on the corresponding `actions.requested` event.
--- a/.changeset/shiny-step-events.md
+++ b/.changeset/shiny-step-events.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Internal instrumentation lifecycle events named `attempt.*` are now named `step.*`, matching the `agent.step` span they produce. No authored surface changes — these events are not public yet.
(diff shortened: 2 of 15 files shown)


chadhietala:
Folded into #1670 — that PR is already about settling the lifecycle event names, so the `attempt.*` → `step.*` rename belongs in it rather than in a separate PR on top.

PR #1700, opened 2026-08-05
feat(eve): memory - add first-class provider lifecycle

AndrewBarba:
### Summary

Closes #1510. Stacked on #1901.

Adds first-class, path-authored memory slots with trusted scope resolution, durable provider-owned user messages, and scoped model tools. Scope resolvers may await an external lookup with `ctx.abortSignal`; eve resolves them once and locks the result across the turn and approval resume.

Every supported provider event—`session.started`, `turn.started`, `message.received`, both compaction events, and `turn.completed`—may return one `MemoryMessage` (`{ content }`) or an ordered array. Each item becomes a separate durable user message; bare strings are rejected. Opening-event messages append after the initial compaction decision, compaction-event messages append after automatic or manual compaction, and completed-turn messages append to settled history for the next turn. Replay of one stamped event is suppressed, but content from different events is deliberately not deduplicated.

Provider tools use the ordinary `defineTool` contract through `defineDynamic` at `step.started`. The public surface exposes provider, message, and scope concepts while keeping dynamic-sentinel and tool-set implementation types private. Storage remains provider-owned; #1703 adds the first file-backed provider.

### Validation

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm test:unit`
- `pnpm test:integration`
- Focused memory authoring and lifecycle unit tests: 9 passed
- Focused first-class memory integration test: 1 passed

### Checklist

- [x] I linked an issue with prior discussion confirming this change is wanted
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

PR #1703, opened 2026-08-05
feat(eve): memory - add portable file memory

AndrewBarba:
### Summary

Related to #1510. Stacked on #1700.

Adds the built-in `fileMemory()` provider behind a portable conditional-document backend. It recalls an indexed file at `session.started`, refreshes it at `compaction.completed`, and exposes scoped `save_memory(text)` and `remove_memory(index)` tools with stable indexes and conflict retries. Recalled content uses the structured `MemoryMessage` result contract introduced by #1700.

The portable surface lives on `eve/memory/file`; the Vercel-specific `vercelBlob()` backend is isolated to `eve/memory/file/vercel`. Vercel deployments select private Blob storage only when `BLOB_STORE_ID` or the legacy `BLOB_READ_WRITE_TOKEN` signals an attached store. Vercel deployments without a store and non-Vercel production deployments must configure a backend explicitly instead of silently losing memory on restart; development still defaults to process-local storage.

### Validation

- `pnpm fmt`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm guard:invariants`
- `pnpm docs:check` (83 documents compile)
- `pnpm test:unit`
- `pnpm test:integration`
- `../../node_modules/.bin/vitest run --config vitest.unit.config.ts src/public/memory/file` from `packages/eve` (22 passed)
- `../../node_modules/.bin/vitest run --config vitest.integration.config.ts src/execution/file-memory.integration.test.ts` from `packages/eve` (1 passed)
- `./node_modules/.bin/tsc -p packages/eve/tsconfig.json --noEmit`
- `pnpm --filter eve build`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts test/scenarios/file-memory-provider.scenario.test.ts` (1 passed)

Vercel Blob behavior is covered through its mocked SDK contract; no live Blob credentials were used locally.

### Checklist

### DIFF
--- a/docs/memory.md
+++ b/docs/memory.md
@@ -61,13 +61,14 @@ eve allocates each new memory one index above the current highest index and neve
 
 The provider stores at most 100 memories by default. Configure `memoryLimit` to change that count. At the limit, a failed save tells the model to remove an outdated memory by index and retry. The provider tells the model to keep stable context and omit secrets, instructions, and current-task details. It does not run a hidden capture model or persist complete transcripts.
 
-On Vercel, the default backend stores private `MEMORY.md` objects in Vercel Blob. Attach a Blob store to the project so `BLOB_STORE_ID` and Vercel OIDC are available, or provide `BLOB_READ_WRITE_TOKEN`. Outside Vercel, the default is process-local memory for zero-configuration development and tests; it is not durable across restarts.
+On Vercel, the default backend stores private `MEMORY.md` objects in Vercel Blob. Attach a Blob store to the project so `BLOB_STORE_ID` and Vercel OIDC are available, or provide `BLOB_READ_WRITE_TOKEN`. Outside Vercel, development and test environments default to process-local memory. A production deployment outside Vercel must pass an explicit backend so it cannot silently run with memory that disappears on restart.
 
 Pin Vercel Blob explicitly when you want to use it locally or customize credentials and pathnames:
 
 ```ts title="agent/memory/user.ts"
 import { byPrincipal, defineMemory } from "eve/memory";
-import { fileMemory, vercelBlob } from "eve/memory/file";
+import { fileMemory } from "eve/memory/file";
+import { vercelBlob } from "eve/memory/file/vercel";
 
 export default defineMemory({
   provider: fileMemory({
--- a/docs/memory.md
+++ b/docs/memory.md
@@ -61,7 +61,7 @@ eve allocates each new memory one index above the current highest index and neve
 
 The provider stores at most 100 memories by default. Configure `memoryLimit` to change that count. At the limit, a failed save tells the model to remove an outdated memory by index and retry. The provider tells the model to keep stable context and omit secrets, instructions, and current-task details. It does not run a hidden capture model or persist complete transcripts.
 
-On Vercel, the default backend stores private `MEMORY.md` objects in Vercel Blob. Attach a Blob store to the project so `BLOB_STORE_ID` and Vercel OIDC are available, or provide `BLOB_READ_WRITE_TOKEN`. Outside Vercel, development and test environments default to process-local memory. A production deployment outside Vercel must pass an explicit backend so it cannot silently run with memory that disappears on restart.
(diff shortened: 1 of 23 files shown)


De-ASI-INTERFACE (approved):

docs/memory.md
```diff
@@ -41,6 +41,43 @@ scope(ctx) {
...
+
+```ts title="agent/memory/user.ts"
+import { byPrincipal, defineMemory } from "eve/memory";
+import { userMemory, vercelBlob } from "eve/memory/user";
+
+export default defineMemory({
+  provider: userMemory({
+    backend: vercelBlob({
+      prefix: "my-agent/memory/user",
+      token: process.env.BLOB_READ_WRITE_TOKEN,
```
ctgowrie, line 77: Why do we need this token? Let's recommend OIDC

packages/eve/src/public/memory/file/provider.ts
```diff
@@ -0,0 +1,253 @@
...
+            backend,
+            document,
+            index: input.index,
+            key,
+            signal: context.abortSignal,
+          });
+        };
+
+        return {
+          forget_memory: defineTool({
```
ctgowrie: remove_memory is probably a better name here. more clear as the the effect

packages/eve/src/public/memory/file/provider.ts
```diff
@@ -0,0 +1,253 @@
...
+}): Promise<UserMemorySaveResult> {
+  const text = normalizeMemoryText(input.text);
+  let document = input.document;
+  let conflicts = 0;
+
+  for (;;) {
+    const entries = parseMemoryDocument(document?.content ?? "");
+    const existing = entries.find((entry) => entry.text === text);
+    if (existing !== undefined) return { index: existing.index };
+    if (entries.length >= input.memoryLimit) {
```
ctgowrie, line 137: Why number of entries vs length? The latter enforces concise entries as well

packages/eve/src/public/memory/file/provider.ts
```diff
@@ -0,0 +1,253 @@
...
+ * Creates a bounded persistent user profile recalled before each turn and
+ * maintained through scope-bound `save_memory` and `forget_memory` tools.
+ */
+export function userMemory(options: UserMemoryOptions = {}): MemoryProvider {
+  const backend = options.backend ?? defaultUserMemoryBackend();
+  const memoryLimit = normalizeMemoryLimit(options.memoryLimit);
+
+  return defineMemoryProvider({
+    events: {
+      async "turn.prepared"(_event, context) {
```
ctgowrie: why not use hook event names? i don't think turn.prepared exists

ctgowrie:
e2e?

packages/eve/src/public/memory/file/provider.ts
AndrewBarba: number of entries felt easier to reason about as opposed to characters. I think in general all memories will be roughly same length but you dont necessarily know what that length is so to go configure a character limit seems nearly impossible to get right

packages/eve/src/public/memory/file/provider.ts
AndrewBarba: this was added in earlier pr in stack but agree its a bit of a weird name let me see if we can just slot this into turn.started I forget exactly how materially different started vs prepared is

packages/eve/src/public/memory/file/provider.ts
ctgowrie: character limit constrains verbosity, which is useful

packages/eve/src/public/memory/file/provider.ts
ctgowrie: I'd prefer we just have one.

Also are these `hooks`? Should we call it such?

packages/eve/src/public/memory/file/provider.ts
AndrewBarba: Agreed. Renamed the model-facing tool to `remove_memory(index)` throughout the provider, recalled prompt, docs, tests, changeset, and PR description.

AndrewBarba:
Superseded by the revised memory contract in #1581 and the replacement core implementation. The portable document backend, storage implementations, and optimistic concurrency work remain useful and will be transplanted into a clean file-memory provider PR using recall and tools directly.

PR #1706, opened 2026-08-06
fix(eve): announce parked agents as framework-injected user notes

ruiconti:
### Summary

Related to #1329 / #1331 (agent-handle lifecycle and agent-messaging model surface). Fixes a production parent-resume failure in agents with `experimental.subagentPersistentSessions` enabled.

**Problem.** When a persistent child settled, the harness appended the updated `<agents>` listing to model history as an `assistant` message. A settle resume carries no new user input, so the next request ended with that assistant message — Anthropic models reject this through AI Gateway (`This model does not support assistant message prefill. The conversation must end with a user message.`), failing the parent turn. Confirmed in production (21 failures across 8 deployments in one day).

**Solution.** The announcement is now a framework-injected **user-role**:

- On a no-input settle resume the note trails the tool results, so the request is user-final by construction; on input-carrying calls it precedes the turn's user message (previous positioning preserved).
- The append-only, diff-gated design is kept, so the provider prompt-cache prefix stays warm. A system-channel injection was considered and rejected: it re-caches the entire conversation on every child settle. Note this also corrects `docs/subagents.mdx`, which previously claimed the listing was a per-call system message — that was never the implementation.
- The static agent-messaging prompt block now declares the `[Agents]` label as framework-injected and not user-authored, addressing role misattribution.
- The diff gate now only matches user-role announcements, so model output echoing the `[Agents]` label cannot suppress a fresh listing.
- The deterministic mock adapter skips `[Agents]` notes when extracting the last user message, so world-suite directive parsing is unaffected.

**Why CI missed it.** `agent-subagents` ran only the default matrix model (`openai/gpt-5.6-sol`), which tolerates assistant-final requests. The fixture now opts into `"e2e": { "modelMatrix": "full" }`, so the agent-messaging evals (whose first turn is exactly the failing resume shape) also run against `anthropic/claude-opus-5`.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/handles/prompt.test.ts src/harness/tool-loop.test.ts src/runtime/prompt/compose.test.ts src/runtime/agent` — 219 passed, including a new regression test reproducing the production shape (parked handle + history ending in tool results + no step input) asserting the request ends with the user-role announcement and the system prompt stays clean.
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/execution/agent-messaging.scenario.test.ts` — 2 passed (scripted parent locates the `agentId` in conversation and continues the child, local + remote).
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/harness` — 36 passed.
- `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm guard:invariants`, `pnpm docs:check` — all green.
- Not validated locally: the real-gateway rejection path only reproduces in the `e2e-local` `anthropic-opus` leg (e2e cannot run locally). Reviewers should confirm that leg is green — it is the regression's authoritative check.

### Checklist


ruiconti:
Pushed e379387e fixing the world-suite failures (`e2e-vercel`/`e2e-postgres` × `agent-subagents`, `dynamic-subagent` eval).

Root cause from the eval artifacts: the parent session died with `session.failed` — `dispatchRuntimeActionsStep` failed after 3 retries with `Agent handle "ag_conditional-marker:..." already exists`. The deterministic mock adapter's reverse scan for the last tool result stopped at the first user message, so the new user-role `[Agents]` announcement masked the settled child's tool result; the adapter re-issued the same deterministic tool call (`call_conditional_marker`), which derives the same operation id and collides in `prepareAgentStart`. Mock-only — both real-model legs (openai-sol, anthropic-opus) passed.

Mock heuristics now treat `[Agents]` announcements as scaffolding (scan past them), with a unit regression test reproducing the masking sequence.

AndrewBarba (approved):

PR #1713, opened 2026-08-06
docs: mention React in homepage native interop heading

tomocchino:
## Summary
- Updates the eve.dev homepage heading from `Works natively with Next.js` to `Works natively with React and Next.js`.

## Validation
- Confirmed the updated string is present in `apps/docs/app/[lang]/(home)/components/nextjs-interop.tsx`.
- Local package checks not run; dependencies are not installed in this checkout and this is a copy-only change.


tomocchino (commented):
Requesting review from @rauchg

cmpadden (approved):

PR #1408, opened 2026-07-30
docs: propose durable sandbox authoring API

ctgowrie:
## Summary

- propose sandbox definitions that return durable `Sandbox` values directly
- preserve build-time prewarming through exported provider templates and opaque build references
- cover Vercel creation, cross-session reuse, parent/child sharing, custom providers, local filesystems, and a future Dockerfile path

## Validation

- `pnpm exec oxfmt --check research/sandbox-api.md`
- `pnpm guard:invariants`

Closes #1406

PR #1726, opened 2026-08-06
feat(eve): expose userspace sandbox drives

ruiconti:
## What

The full proposal and research trail are in [`research/sandbox-drives.md`](https://github.com/vercel/eve/blob/rui/userspace-sandbox-drives/research/sandbox-drives.md). The code in this PR is its tested prototype.

- define Drives as userspace-owned sequential durability rather than framework-owned shared live state
- re-export `Drive` from `eve/sandbox/vercel`
- accept `mounts` when creating a live Vercel Sandbox session
- keep drive mounts out of build-time template options
- vendor the drives-capable `@vercel/sandbox@2.10.0-beta.0`

## Data flow

Before: an author returned a durable Vercel sandbox, but eve’s vendored SDK and session options could not represent a drive mount.

After: author code calls `Drive.getOrCreate`, passes the resulting name to `template.create({ mounts })` or `template.getOrCreate({ mounts })`, and `createVercelSandboxResource` forwards that session option to `Sandbox.create`.

## Why this shape

Drives mount only during sandbox creation, so `mounts` belongs to live session-create options. It is excluded from template options so build prewarming never claims a userspace drive. Re-exporting `Drive` keeps CRUD on the same vendored SDK and credential path as sandbox creation rather than requiring applications to install a second SDK copy.

Drive naming, seeding, deletion, and attach-conflict policy remain application-owned. The platform currently permits only one attachment, so eve cannot make one live drive concurrently writable across sessions ([Drives guide](https://vercel.com/kb/guide/vercel-drives)).

## Stack and scope

Stacked on #1455; #1408 is the authoring proposal beneath that implementation. #509 is the earlier framework-owned per-session alternative.

This remains a draft until we verify stop/resume behavior: when eve restores an idle-stopped sandbox through `Sandbox.get`, the platform must either restore its persisted mount or return a detectable conflict if another sandbox claimed that drive. This PR covers and tests the fresh-create path; it does not add a resume policy without that runtime evidence.

### DIFF
--- a/.changeset/swift-drives-mount.md
+++ b/.changeset/swift-drives-mount.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
+Expose Vercel Sandbox Drives and allow authors to mount them when creating live session sandboxes.
--- a/packages/eve/src/execution/sandbox/bindings/vercel.test.ts
+++ b/packages/eve/src/execution/sandbox/bindings/vercel.test.ts
@@ -293,6 +293,34 @@ describe("prewarmVercelSandboxTemplate", () => {
 });
 
 describe("createVercelSandboxResource", () => {
+  it("forwards session drive mounts to Sandbox.create", async () => {
+    const sessionSandbox = createMockSandbox({ name: "session-key" });
+    const sandboxModule = {
+      Sandbox: {
+        create: vi.fn().mockResolvedValue(sessionSandbox),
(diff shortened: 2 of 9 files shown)


ruiconti:
Superseded by #1889, which ports the Drive API onto the current `main` sandbox lifecycle without the obsolete stacked durable-sandbox API.

PR #1745, opened 2026-08-06

ruiconti:
Server-initiated turns (task wakes, child notifications) previously rendered only while the TUI had an active client turn; an idle TUI dropped them until the next keypress-driven refresh.

Before → after: idle TUI ignores server-initiated `turn.started` → idle TUI subscribes to the session stream and renders those turns as they arrive.

Supersedes #1687 (closed during restack); same commit, rebased base.


ruiconti:
Superseded by #1761 (same content, rebased above the eval suite in the restacked chain).

PR #1746, opened 2026-08-06
fix(eve): complete background task lifecycle routing

ruiconti:
Closes the nuclear-review ordering and ownership gaps in `experimental.tasks`, and removes `task_await` entirely.

## What

- Human answers to task-owned HITL route through the owning task run, which delivers to the child before recording the batch as answered (`answered {requestIds}` replaces the unbound `resume`).
- Child input requests surface on the parent session directly; answering them does not run the parent model.
- One child session admits at most one nonterminal task; same-batch and later continuations reject with `AGENT_BUSY`.
- `POST /eve/v1/session` accepts `operationId` for create-once semantics scoped to the transport principal. Concurrent retries adopt one owner: `SendOptions.intent = "create-once"` resolves the existing owner without redelivering input, and a losing candidate run exits before its first turn instead of emitting `session.failed`.
- `task_await` is removed; `task_peek` is the non-blocking read and completion updates wake the parent.

## Flows, before → after

| Flow | Before | After |
| --- | --- | --- |
| Task HITL answer | parent sent responses straight to the child hook; unbound `resume` could clear a newer batch | task run verifies the outstanding batch, delivers, then records `answered` |
| Concurrent create with one `operationId` | both candidates started; loser crashed with `session.failed`/`HookNotFound` | alias owner adopted; loser exits quietly before turn 1 |
| Child busy | second continuation raced admission | `AGENT_BUSY` with the owning task id |

Validation: unit + focused integration suites green on this branch; end-to-end coverage lands in the stacked evals PR (#1724).

Supersedes #1688 (closed during restack).


ruiconti:
Superseded by the restack: lifecycle routing → #1757, usage retention (the second commit here) → #1763.

PR #1689, opened 2026-08-05
fix(eve): finalize background subagent sections at child boundary

ruiconti:
### Checklist

- [ ] I linked an issue with prior discussion confirming this change is wanted
- [ ] I ran the relevant checks from `CONTRIBUTING.md`
- [ ] I added tests and documentation where relevant
- [ ] I added a changeset if this touches the published `eve` package
- [ ] DCO sign-off passes for every commit (`git commit --signoff`)

PR #1747, opened 2026-08-06

ruiconti:
Background subagent sections in the dev TUI previously finalized on the parent’s turn boundary, so a child that kept streaming after the parent parked left its section truncated or interleaved.

Before → after: section finalization keyed on parent turn events → finalization keyed on the child stream’s own terminal boundary.

TUI-only change (`subagent-pump`, `terminal-renderer`). Supersedes #1689 (closed during restack); same commit, rebased onto the lifecycle PR.

### DIFF
--- a/.changeset/background-subagent-finalization.md
+++ b/.changeset/background-subagent-finalization.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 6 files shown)


ruiconti:
Superseded by #1762 (same content, rebased to the top of the rendering chain).

PR #1748, opened 2026-08-06
test(e2e): add background task golden paths

ruiconti:
Continuation of #1724 (closed to change base; content identical).

## What

- Add a deterministic `fixture-tasks` entry with `experimental.tasks` enabled.
- Cover the background-task golden paths: one-step fanout of ten tasks, continued parent interaction while those tasks run, and lifecycle updates from every task back to the parent session.
- Keep regression coverage for task HITL routing, concurrent session-create idempotency, input-batch ordering, and child-task exclusivity.
- Run the fixture with an authored `mockModel` in every suite so the same orchestration executes against each workflow world without provider planning variance.

## Golden paths

| Path | End-to-end assertion |
| --- | --- |
| Many-task fanout | One model step starts ten background child sessions and returns ten distinct task receipts. |
| Parent interactivity | All ten tasks first block on unreleased child approvals; the parent then accepts a normal follow-up and returns an ordinary tool-free reply. |
| Fanout parent updates | After release, every fanout task completes independently and its completed-task notification is observed on the parent session. |

## Regression properties

| Property | End-to-end assertion |
| --- | --- |
| Task HITL routing | A child approval surfaces on the parent session; its answer reaches the child without a parent model step. |
| Session-create idempotency | Concurrent and serial `/eve/v1/session` retries resolve one session identity for one principal, while the same operation id under another principal creates a different session. |
| Task input-batch ordering | Replaying Q1 after Q2 becomes current cannot clear Q2; the exact Q2 response still resumes the child and completes the task. |
| Child-task exclusivity | Two same-batch `task_send` calls to one persistent child produce exactly one `working` receipt and one `AGENT_BUSY` result. |

## Shape

The fanout workers block on approval rather than a timer. The eval records every queued parent turn, proves all ten tasks are nonterminal before interaction, releases them explicitly, then matches updates by task id against the parent stream. Coalesced turns cannot hide updates or the interactive message.

Task completion is polled with `task_peek`; this fixture has no dependency on the removed task-wait behavior.

Stacked on #1747 (tasks/09). The full fixture passes locally with 27 gates.


ruiconti:
Superseded by #1759: same seven evals plus a fix for the world-agnostic exclusivity-race flake seen in this PR's CI. The deterministic e2e-vercel fanout failure is a product bug, tracked in #1759's description.

PR #1754, opened 2026-08-07
fix(chat-sdk-lark): template uses thread.post instead of webhook-only send

### LINKED ISSUE #1753: chat-sdk-lark: generated channel can never reply — chatSdkChannel().send is webhook-only, Lark adapter is WebSocket-only
## Summary

`eve add channel/chat-sdk-lark` scaffolds a channel that cannot send a single reply. The generated template calls `chatSdkChannel().send()` inside `bot.onNewMention`, but `send` only runs inside an inbound HTTP webhook request — while the only available Lark adapter (`@larksuite/vercel-chat-adapter`) is WebSocket-only. Every inbound message throws:

```
[chat-sdk] Message processing error {
  error: Error: chatSdkChannel().send can only run during a Chat SDK webhook handler for this bridge.
    at bridgeSend (.../eve/.../chat-sdk/chatSdkChannel.js)
    at send (.../eve/.../chat-sdk/chatSdkChannel.js)
    at agent/channels/lark.ts   (bot.onNewMention)
}
```

## Repro

1. `bunx eve add channel/chat-sdk-lark` in an eve project.
2. Set `LARK_APP_ID` / `LARK_APP_SECRET`; in the Feishu console enable 长连接 (long-connection) event mode and subscribe to `im.message.receive_v1`.
3. Start the dev server and @mention / DM the bot.
4. The message **is** received (the WS connects and `bot.onNewMention` fires), but the reply throws the error above. Nothing is posted back to the user.

## Expected

Either the generated channel replies successfully, or `eve add channel/chat-sdk-lark` warns that this adapter is incompatible with the current `chatSdkChannel` send path.

## Root cause

- `chatSdkChannel().send` → `bridgeSend` requires an active webhook request context (the `chat-sdk.active-webhook` context key, set only by the channel's HTTP route handler). It throws when no webhook is running.
- `@larksuite/vercel-chat-adapter` delivers events over its **WebSocket long-connection** only. Its `handleWebhook()` returns `501 "webhook not supported"`. Inbound events fire `bot.on*` straight from the WS client, outside any HTTP request — so the webhook context is never established and `send` can never succeed.
- The adapter's WS-only nature is by design. The official doc (https://chat-sdk.dev/adapters/vendor-official/lark) states: *"Transport: WebSocket only. `handleWebhook()` returns HTTP 501. Webhook transport is on the roadmap."* So the adapter behaves as documented; the bug is in how the generated channel wires it to `send`.

## Workaround

Drive the agent through the eve Client API and post replies with the Chat SDK thread directly (the pattern the adapter is built for), instead of `chatSdkChannel().send`:

```ts
const eve = new Client({ host: process.env.EVE_CLIENT_HOST ?? "http://127.0.0.1:3000" });
const sessionsByThread = new Map<string, string>();

async function runTurn(thread: Thread, text: string) {
  let response;
  const sessionId = sessionsByThread.get(thread.id);
  if (sessionId) {
    response = await eve.sessions.attach(sessionId).send(text);
  } else {
    const created = await eve.sessions.create({ message: text });
    sessionsByThread.set(thread.id, created.session.state.sessionId);
    response = created.response;
  }
  const result = await response.result();
  if (result.message) await thread.post({ markdown: result.message });
}

bot.onNewMention(async (thread, message) => {
  await thread.subscribe();
  await runTurn(thread, message.text);
});
```

Verified working end-to-end. It forgoes the streaming/HITL niceties the channel would otherwise provide.

## Suggested fixes

- **Generator:** for WS-only adapters, emit the Client + `thread.post` bridge instead of `chatSdkChannel().send`, so `eve add channel/chat-sdk-lark` produces working code out of the box.
- **Framework:** let `chatSdkChannel` deliver for adapters that receive outside a webhook (synthesize a delivery context from the adapter when `bot.on*` fires), so `send` works for WS adapters. This also future-proofs any WS-only Chat SDK adapter.
- **Minimal:** if neither is short-term, have `eve add channel/chat-sdk-lark` print a warning that the adapter is WS-only and the current chatSdkChannel send path does not yet support it.

## Environment

- eve `0.31.0`
- `@larksuite/vercel-chat-adapter` `0.1.2`
- Node `v24.18.0`

## Reference

- Introduced by #1061 (Lark / Feishu catalog + docs entry pointing at the Chat SDK channel).

iroiro147:
## Problem
The `@larksuite/vercel-chat-adapter` is WebSocket-only — `handleWebhook()` returns HTTP 501 by design. `chatSdkChannel().send` requires the active webhook request context (set only by the chat-sdk HTTP route handler). When the Lark bot received a message over WebSocket and tried to reply via `send(...)`, it threw:

```
Error: chatSdkChannel().send can only run during a Chat SDK webhook handler for this bridge.
```

The scaffolded channel could receive messages but could never reply. Reported as #1753.

## Fix
Replace the reply path in the generated Lark channel template from `send(...)` to `thread.post({ markdown: ... })` — the adapter-native delivery mechanism that works regardless of transport. Also remove `send` from the destructured surface so new projects don't reach for the webhook-only path.

## Validation
- [x] `pnpm fmt` passes
- [x] `pnpm typecheck` passes (catalog types unchanged; only template code)
- [x] Manual: template now compiles and replies via `thread.post`

Closes #1753.

### DIFF
--- a/apps/docs/registry/channels/chat-sdk-lark.ts
+++ b/apps/docs/registry/channels/chat-sdk-lark.ts
@@ -3,7 +3,7 @@ import { createMemoryState } from "@chat-adapter/state-memory";
 import type { Message, Thread } from "chat";
 import { chatSdkChannel } from "eve/channels/chat-sdk";
 
-export const { bot, channel, send } = chatSdkChannel({
+export const { bot, channel } = chatSdkChannel({
   userName: "My Agent",
   adapters: {
     lark: createLarkAdapter(),
@@ -13,11 +13,11 @@ export const { bot, channel, send } = chatSdkChannel({
 
 bot.onNewMention(async (thread: Thread, message: Message) => {
   await thread.subscribe();
-  await send(message.text, { thread });
+  await thread.post({ markdown: message.text });
 });
 
 bot.onSubscribedMessage(async (thread: Thread, message: Message) => {
-  await send(message.text, { thread });
+  await thread.post({ markdown: message.text });
 });
 
 await bot.initialize();


iroiro147:
**CI note:** `Test TUI smoke tests` fails with `AI Gateway received no credentials` — the `test-tui` job runs against the real AI Gateway and reads `AI_GATEWAY_API_KEY` from org secrets, which are not available to workflows triggered from fork PRs (GitHub strips secrets on `pull_request` events from forks). This is the same infra gap documented on other fork contributions; it is not related to this template change (`apps/docs/registry/channels/chat-sdk-lark.ts`). Job passes on upstream-owned branches where the secret resolves.

De-ASI-INTERFACE (changes requested):
No change this pr we will never change the replay routes

PR #1756, opened 2026-08-07
feat(eve): carry background-task lifecycle receipts on the wire

ruiconti:
Protocol half of the background-task rendering work, split out so the wire contract lands before any consumer.

With tasks enabled, a delegated subagent call resolves twice from a client's perspective: once when the dispatch is acknowledged (receipt), once when the child actually finishes. Before this change both arrived as indistinguishable `subagent.completed` events; a client could not tell acknowledgement from completion.

- before: `subagent.completed` → one shape for both phases
- after: receipts carry a `backgroundTask` lifecycle marker; clients can key rendering (and any other consumer logic) off the phase

Second commit bumps the extension contract reports for the new field.

Scope: wire contract + emission path only. The TUI consumer lands later in the stack (`tasks/09-background-rendering`), after the semantics are proven by the eval suite.

Stacked on #1522.

### DIFF
--- a/.changeset/background-task-receipt-events.md
+++ b/.changeset/background-task-receipt-events.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Background task receipts now carry explicit lifecycle metadata so clients can distinguish dispatch acknowledgement from child completion.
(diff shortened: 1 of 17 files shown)


ruiconti:
Superseded by #1773; final receipt/stream protocol now lands with its producer and remote proxy.

PR #1757, opened 2026-08-07
fix(eve): complete background task lifecycle routing

ruiconti:
Settles the `experimental.tasks` semantics end to end. Everything below this PR defines the primitives; everything above consumes them unchanged.

## What

- **Run-owned HITL**: a blocked child's input requests surface on the parent session, and the client's responses route directly back to the child — no parent model turn in between.
  - before: child HITL parked on the parent turn's inbox → parent model saw the requests
  - after: the task run owns the request batch; the parent session is a pass-through
- **Batch-bound answers**: an `answered` batch binds responses to the exact requests they answer. Late or duplicate answers cannot cross into a newer batch.
- **Child exclusivity**: one child session owns at most one nonterminal task. Same-batch duplicate `task_send`s race admission; exactly one wins, the loser exits quietly (`HookConflictError` → empty output, no error surfaced to the model).
- **Create-once ownership**: session creation is keyed by an operation-scoped token (`eve:op:<sha256>`), so a concurrent or retried create adopts the existing owner instead of redelivering input.

## Why no waiting primitive

Parents are woken by task lifecycle transitions (terminal states and `input_required`), never by polling or blocking. An earlier draft had a `task_await` aggregation workflow; it is gone from this stack entirely — no layer introduces it, so no layer removes it.

Replaces the lifecycle half of the closed #1746 (its usage-retention commit moved to the top of the stack).

Stacked on #1756.

### DIFF
--- a/.changeset/remote-create-operation-id.md
+++ b/.changeset/remote-create-operation-id.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Make subagent dispatch replay-safe. Remote create-session requests now carry a replay-stable `operationId`, and the built-in `POST /eve/v1/session` route returns the child it already created for that operation instead of starting a second one. A replayed local start adopts the child holding its deterministic continuation token rather than reporting a start failure.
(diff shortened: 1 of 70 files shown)


Rchie121 (approved):

ruiconti:
Superseded by #1772 (generic create-once) and #1773 (task lifecycle/runtime).

PR #1758, opened 2026-08-07
feat(eve): give background task agents a stable identity

ruiconti:
Adds a stable `agentId` as the join key between agent handles and tasks.

The problem it solves: handles and tasks each held a partial view of "which agent is this and can I talk to it". Reusing an agent after its task finished resolved through a stale handle, and availability questions had no single authority.

- before: handle phase approximated availability; continuation admission keyed by session
- after: handles are the address book, tasks are the availability authority; `handle.identity.id = task.metadata.agentId` joins them, and continuation admission re-keys by `agentId` so a reused agent resolves to its live task

Adds the `addressed` handle phase for a handle that has an address but no live task.

Stacked on the lifecycle PR; the eval suite directly above proves the semantics of the whole chain.

### DIFF
--- a/.changeset/task-agent-ownership.md
+++ b/.changeset/task-agent-ownership.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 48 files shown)


ruiconti:
Superseded by #1771/#1773: final agentId shape is established before the first writer and activated with the runtime.

PR #1759, opened 2026-08-07
test(e2e): prove background task semantics end to end

ruiconti:
Seven deterministic mock-model evals over a new `fixture-tasks` fixture, sitting directly above the last semantic PR so every contract the stack ships is proven before any rendering code consumes it.

Each eval is named by the invariant it gates:

- `task-hitl-routing` — child input requests surface on the parent; responses route back without a parent turn
- `session-create-idempotency` — concurrent + serial create retries yield one `message.received` / one `step.started`
- `task-input-batch-ordering` — answers bind to the batch they answer
- `child-task-exclusivity` — same-batch duplicate sends admit exactly one
- `task-fanout-parent-updates` — 10-task fanout behind an approval-gated barrier; parent stays interactive; per-task completion updates
- `task-fan-in-join` — one wake per ready transition; model-side join predicate
- `task-cancellation` — cancellation is final; a late child result cannot revive the task

Replaces closed #1748. Two changes against that revision:

1. **Exclusivity race hardened.** The world-agnostic flake in #1748's CI (also seen on postgres at an earlier commit): a completion notification coalescing into the RACE turn's model request made the scripted model ACK instead of running the same-batch sends. The notification branch now delegates to the race script, same pattern the fan-in eval already used.
2. `task_await` comment references swept — the primitive no longer exists anywhere in the stack.

Known red: `task-fanout-parent-updates` fails deterministically on `e2e-vercel` only — the parked parent's continuation token doesn't resolve on the vercel session-driver resume path while 10 children hold HITL (`Session is not active and cannot be resumed`, session-continue route). Local and postgres pass. That is a product bug under investigation, not an eval assumption; it reproduced identically on both prior CI runs.

Local run at this revision: 7 evals, 43 gates, all green.

### DIFF
--- a/e2e/fixtures/fixture-tasks/.gitignore
+++ b/e2e/fixtures/fixture-tasks/.gitignore
@@ -1,0 +1,11 @@
+node_modules
+.env*
+.eve
+.vercel
+.next
+.output
+.nitro
+dist
+.DS_Store
+*.tsbuildinfo
+.env*.local
--- a/e2e/fixtures/fixture-tasks/.vercelignore
+++ b/e2e/fixtures/fixture-tasks/.vercelignore
@@ -1,0 +1,6 @@
+node_modules
+.eve
+.next
+.output
+.nitro
+dist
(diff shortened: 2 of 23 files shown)


ruiconti:
Superseded by #1774. The new acceptance suite fixes the false-positive oracles, adds hard timeouts, and is green on local/Postgres/Vercel.

PR #1760, opened 2026-08-07
fix(eve): keep background subagent output in its section

ruiconti:
First TUI consumer of the settled task semantics. Background subagent activity now stays folded under its original section after the parent turn continues, keyed off the receipt lifecycle metadata from #1756.

- before: once the parent turn moved on, a background child's continued output opened a new section at the bottom of the transcript
- after: the child's stream renders into the section where it started; dispatch acknowledgement and completion are distinguished by the `backgroundTask` receipt marker

TUI half of what was #1561 (closed); the protocol half is #1756. Sits above the eval suite so review here is rendering-only — every contract this consumes is already gated.

### DIFF
--- a/.changeset/background-subagent-rendering.md
+++ b/.changeset/background-subagent-rendering.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Keep background subagent activity folded under its original TUI section after the parent turn continues.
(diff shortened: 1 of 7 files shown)


ruiconti:
Superseded by unified TUI PR #1775.

PR #1761, opened 2026-08-07

ruiconti:
An idle dev TUI now renders turns the server initiates — task wake notifications land as visible turns instead of silently mutating session state until the next user input forced a redraw.

Pure TUI + client-follow change (replaces closed #1745, rebased; content unchanged). No protocol surface.

Stacked on #1760.


ruiconti:
Superseded by unified TUI PR #1775.

PR #1762, opened 2026-08-07

ruiconti:
Child streams now close their own TUI sections when the child reaches a boundary, instead of leaving sections open until the parent turn ended.

Replaces closed #1747 (rebased; content unchanged). Top of the rendering chain.


ruiconti:
Superseded by unified TUI PR #1775; authoritative completion is now present from the first TUI contract.

PR #1763, opened 2026-08-07
fix(eve): retain background child usage on terminal task snapshots

ruiconti:
A background child's reported token usage is retained on its terminal task snapshot (internal, not model-visible), so strict budget accounting can land later without data loss.

Background budgets stay best-effort for now: each local child is capped from the parent's remainder at dispatch time, but no reservation couples sequential dispatches. The research doc records that boundary.

Moved here from the closed #1746 (it was orthogonal to the lifecycle work it rode on).

### DIFF
--- a/.changeset/task-usage-retention.md
+++ b/.changeset/task-usage-retention.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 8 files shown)


ruiconti:
Superseded by #1773; usage retention is part of the final task snapshot/runtime contract.

PR #1771, opened 2026-08-07
feat(eve): add the inert task kernel in its final shape

ruiconti:
Adds the durable task state machine and workflow substrate in the final shape that the runtime will consume.

The kernel owns task identity, pure transitions, snapshots, task-run hooks, batch-bound HITL answers, private executor state, terminal usage retention, and an immutable terminal snapshot fallback for expired workflow runs.

This layer is inert. It adds no authored flag, framework tool, persistence writer, or production caller. Existing agents cannot start or command a task run. That lets the data model and transition invariants be reviewed without publishing an intermediate feature or requiring a later persisted-state migration.

Stack prerequisite: #1482. The next PR adds an independent create-once session API; background-task activation first appears in PR 3.


ruiconti:
Recreated after GitHub marked exact-ancestor stacked branches dirty following a force-push; replacement uses the same head/base.

ruiconti:
Replacement: #1804

PR #1772, opened 2026-08-07
feat(eve): add authenticated create-once sessions

ruiconti:
Adds `operationId` to authenticated session creation as an independently complete channel feature.

- before: retrying `POST /eve/v1/session` could start and dispatch more than one session
- after: the first active owner of `(principal, operationId)` wins; existing and concurrent retries return that session without dispatching duplicate input

Principal identity is a domain-separated tuple of authenticator, issuer, principal type, and principal ID. Anonymous callers cannot use the shared operation namespace. The included basic-runtime eval proves concurrent and serial retries produce one initial message/model step, and that the same operation under another issuer owns a different session.

This PR contains no task or subagent consumer. PR 3 uses the API for replay-safe remote child creation.

### DIFF
--- a/.changeset/session-create-once.md
+++ b/.changeset/session-create-once.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add authenticated create-once session requests through `operationId`. Concurrent or retried creates adopt the active session that first claimed the operation without dispatching duplicate input.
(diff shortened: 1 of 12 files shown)


ruiconti:
Recreated after stacked-base synchronization; replacement uses the same branch and diff.

ruiconti:
Replacement: #1805

PR #1773, opened 2026-08-07
feat(eve): run subagents as durable background tasks

ruiconti:
Introduces the complete `experimental.tasks` runtime contract in one merge unit. The flag first appears together with the final model tools, persistence writer, stable agent identity, local/remote lifecycle routing, and release metadata.

## Data flow

- before: a subagent call blocked its parent turn until the child settled
- after: dispatch returns one durable `{ taskId, agentId, status: \"working\" }` receipt; task transitions later wake the parent through replay-identified session deliveries
- before: child HITL targeted a parent turn inbox
- after: the task run validates the outstanding batch, routes the complete request/challenge through the parent channel, delivers the answer to the exact local or remote child, then records `answered`

## Correctness boundaries

Every accepted child turn rebinds local adapter or remote callback forwarding to its owning task. Duplicate action IDs are rejected before side effects. Task readiness is released only after the parent index commits, making a fast terminal child replay-recoverable. Cancellation waits for committed task state, aborts the executor, and reports one cancelled settlement with usage so the task hook can close.

One agent owns at most one nonterminal task. Failed and cancelled work retains usage without reviving terminal status. Terminal snapshots remain readable after task-run retention expires. Remote callbacks are schema-validated and task response capabilities must match the recorded child address.

Remote child streams use an authenticated parent-origin proxy. Clients receive a server-authored `childStreamPath`; parent credentials are never forwarded cross-origin and remote credentials never enter the event stream.

This replaces the semantic/protocol portions of the v3 stack (#1517-#1522 and #1756-#1758). PR 4 is the merge-gating E2E acceptance suite.


ruiconti:
Recreated after stacked-base synchronization; replacement contains the correctness pass and same branch.

ruiconti:
Replacement: #1806

PR #1774, opened 2026-08-07
test(e2e): gate the final background task contract

ruiconti:
Seven deterministic task evals gate the final runtime contract before any TUI consumer lands.

The suite covers task-owned HITL across repeated requests, stale-answer batch isolation, same-batch and later-turn agent exclusivity, `task_send` continuation followed by new HITL, fanout parent interactivity, exact two-task fan-in, and cancellation finality. Every wait is bounded to 120 seconds.

The prior fixture could pass with queued duplicate creates, partial fan-in output, stale Q1 unblocking Q2, status-only cancellation, exclusivity limited to one batch, or continued child events still targeting the previous task. These oracles now inspect exact task IDs, complete model-visible views, and the relevant later turn. Generic create-once coverage remains in #1772, where that API is introduced.

Local mock acceptance at this revision: 7 evals, 48 gates, all green. This PR and #1773 must not merge until local, Postgres, and Vercel worlds all pass.

### DIFF
--- a/e2e/fixtures/fixture-tasks/.gitignore
+++ b/e2e/fixtures/fixture-tasks/.gitignore
@@ -1,0 +1,11 @@
+node_modules
+.env*
+.eve
+.vercel
+.next
+.output
+.nitro
+dist
+.DS_Store
+*.tsbuildinfo
+.env*.local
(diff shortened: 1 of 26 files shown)


De-ASI-INTERFACE (approved):

ruiconti:
Recreated after stacked-base synchronization; replacement includes continuation-HITL acceptance.

ruiconti:
Replacement: #1807

PR #1775, opened 2026-08-07
fix(eve): render background tasks as one owned TUI section

ruiconti:
Combines background section rendering, idle wake following, and child-boundary finalization into one final TUI state machine.

A subagent section records its originating parent turn and whether it became a background task. Parent completion is provisional; delayed child output can reopen the section, while the child's own boundary is authoritative. Cancelling turn B closes only B's foreground descendants — unrelated background tasks from A or B keep streaming.

Idle server turns render without taking input ownership. Session failure recovery is tied to the exact source `ClientSession`, so a stale failure cannot discard a fresh replacement. Reconnect exhaustion reopens the child stream with its cursor and event dedupe intact.

Local and remote children both open the server-authored `childStreamPath` through the parent client. The TUI never uses a remote URL or credential directly.

This replaces #1760-#1762 as one review unit because their completion and ownership semantics cannot be reviewed independently.

### DIFF
--- a/.changeset/background-task-tui.md
+++ b/.changeset/background-task-tui.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Render background subagent activity in one persistent dev TUI section across parent turns. Idle task wakes render while the prompt remains active, remote child streams use the authenticated parent proxy, and child boundaries finalize sections without later-turn cancellation closing unrelated background work.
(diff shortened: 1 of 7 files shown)


Rchie121 (approved):

ruiconti:
Recreated after stacked-base synchronization; replacement uses the same unified TUI branch.

ruiconti:
Replacement: #1808

Issue #1786, opened 2026-08-07
Make integration setup flows deterministic for coding agents


OwenKephart:
## Problem

Built-in integration setup flows are optimized for interactive terminals. Coding agents cannot reliably answer branching prompts, and replaying a setup command is unsafe once an earlier choice has mutated external state.

## Proposed direction

Use the existing keyed `Asker` abstraction for every semantic setup decision, separate decision collection from mutation in complex integrations, and let headless callers provide answers by stable key. Shared one-time prerequisites such as Vercel project linking should be completed separately in headless mode rather than unexpectedly opening an interactive flow.

Interactive CLI and TUI behavior should remain unchanged, including specialized controls such as editable selects.

This issue tracks migrating the built-in integration setup flows and defining the eventual structured CLI/setup-command contract used by `eve add` and coding agents.

jseramn:
Working on a first slice of this: headless Asker composition for the integration setup runner, Discord gather→perform, and fail-closed Vercel project linking in headless mode. Opening a PR shortly.

PR #1787, opened 2026-08-07
refactor(eve): prepare integration setup before apply

OwenKephart:
### Summary

Related to #1786. Stacked on #1790.

Replace built-in integrations' monolithic setup hooks with a strict in-memory `prepare → apply` lifecycle. Prepare owns questions, read-only discovery, and caller-owned prerequisites; apply receives no `Asker` (so it cannot ask followup questions that lead to bad partial application states)

Pretty big PR because I migrate all existing setup flows at once to avoid having to do annoying compat things, but for the most part the migration is quite rote (just breaking up one big function into two smaller ones).

### Validation

- `pnpm --filter eve exec tsc -p tsconfig.json --noEmit --pretty false`
- Focused unit suite on the full stack: 162 tests passed
- Focused `/add` registry-session tests passed
- `git diff --check`

### Checklist

### DIFF
--- a/packages/eve/src/setup/integrations/shared/ui.ts
+++ b/packages/eve/src/setup/integrations/shared/ui.ts
@@ -1,32 +1,44 @@
-import { confirm, SkippedSignal, type Asker } from "../../ask.js";
+import type { Asker } from "../../ask.js";
 import type { Prompter } from "../../prompter.js";
+import type { SetupApplyContext, SetupPrepareContext, SetupPresentation } from "../types.js";
 
-/** UI capabilities available to a channel-owned setup hook. */
-export interface IntegrationSetupUi {
-  readonly asker: Asker;
-  readonly prompter: Prompter;
-  confirm(input: { key: string; message: string; recommended?: boolean }): Promise<boolean>;
-  nextSteps(lines: readonly string[]): void;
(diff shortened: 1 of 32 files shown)


packages/eve/src/setup/integrations/shared/ui.ts
```diff
@@ -1,32 +1,50 @@
...
+    externalAction(input) {
+      if (onExternalAction !== undefined) {
+        onExternalAction(input);
+        return;
+      }
+      prompter.log.message(input.message);
+      prompter.log.message(input.url);
+      if (input.userCode !== undefined) prompter.log.message(`Code: ${input.userCode}`);
+    },
+  };
```
ruiconti: It's not clear when this will run. Can we make this API time-bound e.g `onComplete`, `postInstall`, etc.

OwenKephart: ah yeah this is confusing -- the basic idea is that this marks the point where the setup flow starts waiting for external input (canonical example being pulling up the browser for auth verification). renamed / refactored this to be `beginExternalAction`

from there, the integration setup flow decides when to invoke this, and the presenter makes sure we surface that consistently

AndrewBarba (approved):

Issue #1658, opened 2026-08-05
Denying a tool approval permanently fails the durable session (OpenAI provider)


SammyTourani:
# Denying a tool approval permanently fails the durable session (OpenAI provider)

## Summary

When a human **denies** a gated tool call (`approval: always()` from `eve/tools/approval`),
the denied result is recorded with `output.type = "execution-denied"`. That type is not part
of the AI SDK's `ToolResultOutput` contract (`text | json | error-text | error-json | content`),
so when the next turn replays the transcript, `@ai-sdk/openai` cannot map it: its output
switch falls through and the request is sent with

```
{ type: 'function_call_output', call_id: 'call_…', output: undefined }
```

OpenAI rejects the request with `400 Missing required parameter: 'input[N].output'`, the turn
fails with `MODEL_CALL_FAILED`, and the session ends in `session.failed` — one human "No"
permanently kills a durable session. **Approving is unaffected**; only denial poisons the
transcript.

## Reproduction

Verified on **eve 0.30.2 and 0.30.6** (identical behavior), `ai@7.0.5x`, `@ai-sdk/openai@2.x`,
model `gpt-5-mini`, both with the Postgres workflow world and the default local world.

1. Agent with one gated tool:

```ts
import { defineTool } from "eve/tools";
import { always } from "eve/tools/approval";
import { z } from "zod";

export default defineTool({
  description: "Delete a memory by id.",
  approval: always(),
  inputSchema: z.object({ id: z.number(), reason: z.string() }),
  async execute() { return { ok: true }; },
});
```

2. `POST /eve/v1/session` with a message that makes the model call the tool → turn parks,
   `input.requested` carries the approval.
3. Answer with `inputResponses: [{ requestId, optionId: "deny" }]`.
4. The resumed turn fails immediately:

```
step.failed   Missing required parameter: 'input[7].output'.  (upstream 400, invalid_request_error)
turn.failed   …
session.failed …
```

The `action.result` event shows the recorded denial:

```json
{ "status": "rejected", "result": { "isError": true, "output": {
  "approval": { "status": "denied" }, "code": "TOOL_EXECUTION_DENIED",
  "tool": { "result": "not_run" } } } }
```

and the model-call dump shows the transcript part reaching the SDK with
`output: { "type": "execution-denied", "reason": "Tool execution was denied." }`.

## Expected

A denied approval should reach the next model call as a well-formed tool result — e.g.
[...]

PR #1829, opened 2026-08-08
test(eve): prove approval batches splice after intervening turns

ruiconti:
Acceptance gate for the first mitigation slice of the HITL non-blocking plan (research/hitl-request-lifecycle.md, #1224): before any behavior change lands, prove the AI SDK accepts an approval batch restored *after* later conversation turns.

## What

- New integration case in `tool-loop-generate-approval-resume.integration.test.ts`: session history already contains an intervening user/assistant exchange when the approval response arrives.
- Shared mock-model/tool setup factored into `createModel()`/`createConfig()`; the existing case is unchanged in behavior.

## The bet being gated

Today a message during a pending approval is hidden in `deferredStepInput`, so history never grows between park and resume. Once messages run as normal turns, resume splices the stored assistant tool-call message after that exchange:

- before → history at resume: `user` (approval batch restored immediately after)
- after → history at resume: `user, user, assistant` (batch restored behind the intervening turn)

The test pins: one model call, one tool execution, the intervening turn precedes the restored tool-call message in the provider prompt, committed history is exactly `user, user, assistant, assistant, tool, tool, assistant`.

It passes against the current harness with no runtime change — SDK core already accepts the spliced shape. Provider converters remain unproven (#1658 family); that risk stays with the behavior-stage e2e against real providers.

No runtime changes, no changeset (test-only).

### DIFF
--- a/packages/eve/src/harness/tool-loop-generate-approval-resume.integration.test.ts
+++ b/packages/eve/src/harness/tool-loop-generate-approval-resume.integration.test.ts
@@ -33,7 +33,7 @@ const approvalRequest = {
   type: "tool-approval-request" as const,
 };
 
-function createPendingApprovalSession(): HarnessSession {
+function createPendingApprovalSession(history?: readonly ModelMessage[]): HarnessSession {
   const session: HarnessSession = {
     agent: {
       modelReference: { id: "generate-approval-resume-model" },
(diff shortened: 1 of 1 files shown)


ruiconti:
Coalesced into #1868 — the splice-gate test rides as that PR's first commit, so the acceptance gate and the behavior it gates review together.

Issue #624, opened 2026-07-08
Serialize concurrent deliveries to one continuation token


ruiconti:
### What problem are you trying to solve?

eve does not serialize concurrent deliveries that target the same continuation token. The token is both the session resume address and an exclusive workflow hook. If two invocations race before either session is established, each tries to claim the same hook and one can fail with `HookConflictError`.

GitHub webhooks make this easy to hit. Creating a pull request with a label can deliver `opened` and `labeled` close together. If an application opts into both actions, the GitHub channel starts both dispatches independently with `waitUntil`. Both derive the same per-PR continuation token, then call `send`. The losing delivery is logged by [`sendGitHubTurn`](https://github.com/vercel/eve/blob/2afed3b7d3d1c1a130befba7210de55f648cd6c2/packages/eve/src/public/channels/github/dispatch.ts#L297-L334), while hook ownership rejects the conflicting run ([source](https://github.com/vercel/eve/blob/2afed3b7d3d1c1a130befba7210de55f648cd6c2/packages/eve/src/execution/hook-ownership.ts#L3-L13)).

This race happens before there is a shared session. `defineState` cannot coordinate it because its state is scoped to one existing session. The current documentation therefore tells every channel or application to build an external per-session queue ([docs](https://github.com/vercel/eve/blob/2afed3b7d3d1c1a130befba7210de55f648cd6c2/docs/concepts/execution-model-and-durability.md#L55-L61)).

### Proposed solution

Provide framework-owned admission control for deliveries sharing a continuation token. The default policy should preserve accepted work rather than throwing or silently dropping it. A durable queue is the clearest contract, though an explicit configurable coalescing policy could also work for channels that want it.

Required behavior:

- Two verified deliveries for the same continuation token can arrive concurrently before the session exists without producing `HookConflictError` for normal admission contention.
- The framework creates or resumes one session and handles accepted deliveries in a documented order.
- Coordination works across processes, regions, cold starts, and retries. An in-memory mutex is insufficient.
- A crash after admission does not lose the delivery, and retrying the same delivery does not execute it twice when the caller supplies a stable delivery id or idempotency key.
- The caller receives or can observe the disposition of every delivery: started, queued, coalesced, deduplicated, or rejected. Nothing disappears behind a swallowed log.
- Different continuation tokens remain independently concurrent.
- Authentication, channel state, message context, and attachment metadata stay attached to the correct queued delivery.
- Admission contention does not become `session.failed` and does not post an unrecoverable error to the user.
[...]

PR #1843, opened 2026-08-08
Add durable channel delivery idempotency

### LINKED ISSUE #1842: Support optional channel delivery idempotency keys

Channels commonly receive at-least-once deliveries from webhook providers. A provider can retry the same verified event after a timeout, pod restart, or network failure, and multiple application instances may receive those attempts.

Today a channel can suppress duplicates only outside Eve (for example with process-local state or an external queue/store). Process-local deduplication does not work across replicas, while external deduplication has to guess when Eve has durably admitted a delivery. This makes it difficult to safely acknowledge a webhook without either losing an event or starting the same agent turn twice.

This is related to #624, which already calls for durable admission and mentions a stable delivery ID. This issue proposes the focused, channel-agnostic public API for supplying that ID. It should work for Slack, web, GitHub, Telegram, and custom channels rather than adding Slack-specific state.

### Proposed solution

Add an optional idempotency key to the channel-agnostic `send` options:

```ts
await send(message, {
  auth,
  continuationToken,
  idempotencyKey: providerDeliveryId,
});
```

Suggested semantics:

- The option is additive; unkeyed sends retain their current behavior.
- The key is scoped to the resolved session / continuation token, not globally. Different sessions may reuse the same provider ID.
- The same key is honored for both initial session creation and follow-up delivery, including concurrent ownership races.
- Eve durably records admission atomically with the turn. Retrying after durable acceptance returns the accepted session/disposition without starting another turn.
- The key and deduplication state are reconstructed by workflow replay, so the behavior works across processes, regions, cold starts, and retries.
- Retained keys are bounded with deterministic eviction (for example, a documented recent-key window) so a long-lived web session cannot grow durable state without limit.
- Channel implementations select the provider-native stable ID; Eve does not interpret its format.

Acceptance tests should cover:

- concurrent retries of one keyed initial delivery;
- concurrent retries of one keyed follow-up delivery;
- replay/restart after admission;
- the same key used by two different sessions;
- unchanged behavior for unkeyed sends;
- deterministic eviction at the configured/documented bound.

This can be the idempotency primitive used by the broader serialization/admission work in #624; it does not by itself define ordering between distinct deliveries.

### Alternatives considered

- Keep a process-local set in each channel. This is lightweight but does not deduplicate across replicas or restarts.
- Put Redis, SQS, or a database in every application. Those systems can buffer retries, but without a durable Eve admission contract the application still has to coordinate queue deletion and session creation itself.
- Make idempotency mandatory. Existing callers and channels may not have a provider delivery ID, so an optional API is more compatible.

ewhauser:
### Summary

Closes #1842. At-least-once channel deliveries can reach different eve replicas, while Slack's process-local event cache cannot prevent duplicate turns across pods or restarts and drops some timeout retries before durable admission. This adds an optional `idempotencyKey` to channel sends, enforces it in the durable session driver with a bounded 1,024-key per-session window, and uses Slack `event_id` values while removing process-local retry suppression. Unkeyed sends remain unchanged; custom channel-handler side effects outside `send` are deliberately not deduplicated. Related to #624.

### Validation

- `pnpm --filter eve test:unit` (580 files passed; 6,046 tests passed; 1 skipped)
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/workflow-entry.integration.test.ts -t "parks in conversation mode and resumes via runtime delivery"` (1 passed; 11 skipped)
- `pnpm --filter eve build:js`
- `pnpm docs:check`
- `pnpm guard:invariants`
- `pnpm lint`

### Checklist


komly:
We also need this for the built-in HTTP channel and a database-backed application outbox. We are maintaining a downstream patch on eve 0.54.0; this PR looks like the right place to converge rather than adding a second idempotency implementation upstream.

The integration cases that have been useful for us are:

1. Submit the same caller-owned key twice to an existing session, including while a foreground workflow is waiting in `ctx.ask`. After answering the input, exactly one matching `message.received` event should exist.
2. Restart the actual eve server, replay the session, and submit the old key again. The original message must remain deduplicated.
3. Carry the key through the built-in HTTP route as well as custom-channel `send`. An outbox needs an acknowledgment that identifies the accepted submission before marking it delivered.
4. Define what happens when the same key is reused with different content. A caller should be able to distinguish that conflict from a duplicate of the original submission.

Our downstream checks use a real local PostgreSQL world, real HTTP requests, a real model provider, and a process restart. They passed the first two cases on the patched package: one message for duplicate foreground-workflow deliveries, stable event IDs after restart, and no additional message for the repeated key. These are results for our downstream patch, not a test result for this PR.

The documented 1,024-key retention window is also important to expose to HTTP/outbox consumers: a delivery older than that window no longer has the same deduplication guarantee. Related issue: #1842.

PR #1856, opened 2026-08-09
refactor(eve): define the flat instrumentation lifecycle protocol

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799.

Defines an internal, flat instrumentation lifecycle protocol with eve-owned payloads. AI SDK callbacks are mapped at the bridge boundary, and an invariant prevents SDK types from leaking into the contract. The protocol now includes generic `input.requested` and `input.resolved` pairs for user-input suspension, including tool approvals; later stack layers add replay identity, publication, and tracing.

The OpenTelemetry consumer moves to the flat events while preserving existing spans and attributes. Root `agent.session` spans carry `agent.trace.schema.version = 1`, including rollover windows. This PR adds no public authoring API.

### Validation

- Top-of-stack approval and provider integration suites: 2 tests passed
- `pnpm --filter eve build`
- `pnpm guard:invariants`

### Checklist

### DIFF
--- a/.changeset/flat-instrumentation-lifecycle.md
+++ b/.changeset/flat-instrumentation-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Instrumentation lifecycle events now use eve-owned payloads and flat event names, including `step.attempt.*` and durable `input.requested`/`input.resolved` boundaries, instead of AI SDK callback types and paired hooks. Existing spans and attributes are unchanged.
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "fix:deps": "syncpack fix-mismatches && syncpack format",
     "fmt": "oxfmt",
     "guard:fixtures": "node ./scripts/guard-test-fixtures.mjs",
-    "guard:invariants": "node ./scripts/guard-invariants.mjs",
+    "guard:invariants": "node --test ./scripts/guard-invariants-rule37.test.mjs && node ./scripts/guard-invariants.mjs",
```
ruiconti: Is the `--test` here intentional? Doesn't seem like something would expand nicely if every guard did this

chadhietala: Going to just fold it into the guard invariants file.

ruiconti (approved):

PR #1857, opened 2026-08-09
feat(eve): assemble OpenTelemetry from declarations

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799. Stacked on #1856.

Establishes an internal declarative OpenTelemetry runtime: singleton settings and ordered destinations are assembled into one eve-owned tracer provider, with the local trace spool treated as an ordinary processor. eve centrally verifies provider ownership and manages batching, flushing, and shutdown without changing recorded spans.

This intentionally exposes no provider API; provider authoring support lands later in the stack.

### Validation

- Focused unit suite: 223 tests passed
- `tsc -p packages/eve/tsconfig.json --noEmit`
- `node scripts/guard-invariants.mjs`
- Full branch validation: 6,164 unit and 618 integration tests passed

### Checklist

### DIFF
--- a/packages/eve/src/tracing/install-instrumentation-runtime.ts
+++ b/packages/eve/src/tracing/install-instrumentation-runtime.ts
@@ -1,0 +1,98 @@
+import { trace } from "#compiled/@opentelemetry/api/index.js";
+import type { SpanProcessor } from "#compiled/@vercel/otel/index.js";
+
+import {
+  createInstrumentationHooks,
+  type InstrumentationProviderDefinition,
+} from "#harness/instrumentation-lifecycle.js";
+import {
+  registerInstrumentationRuntime,
+  type InstrumentationRuntime,
+} from "#harness/instrumentation-runtime.js";
+import { createLogger, formatError } from "#internal/logging.js";
+import { ContextAgentTraceStateStore } from "#tracing/agent-trace-context-store.js";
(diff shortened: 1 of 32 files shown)


packages/eve/src/tracing/install-instrumentation-runtime.ts
```diff
@@ -0,0 +1,99 @@
...
+  const providers: InstrumentationProviderDefinition[] = [...input.providers];
+  let otelRuntime: RegisteredOtelPipeline | undefined;
+  let runInContext: InstrumentationRuntime["runInContext"] = (_operation, execute) => execute();
+
+  if (input.collected.declared) {
+    activateOtelSettings(input.collected.settings);
+    otelRuntime = registerOtelPipeline({
+      pipeline: input.collected.pipeline,
+      serviceName: input.serviceName,
+    });
```
ruiconti: @!e0 why is this registering an authored config via a globally defined value instead of lowered to the compiled agent artifact?

ruiconti: > put it on the existing process-wide InstrumentationRuntime and have the harness read it there

This makes sense to me

chadhietala: Fixed thise

ruiconti (approved):

PR #1859, opened 2026-08-09
feat(eve): add replay-stable instrumentation event identity

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799. Stacked on #1858.

Every lifecycle event carries a namespaced `idempotencyKey`, allowing a provider to upsert one logical operation across worker replays without conflating retries. Starts and terminals share a key derived from durable eve identity, while model keys deliberately exclude the AI SDK's non-replay-stable call ID.

Input requests use the originating session, turn, and stable request ID, so a response handled by another worker settles the same provider operation. Durable provider state and handler deadlines follow in #1860.

### Validation

- Input identity unit suite: 5 tests passed
- `pnpm guard:invariants`

### Checklist

### DIFF
--- a/packages/eve/src/harness/instrumentation-lifecycle.ts
+++ b/packages/eve/src/harness/instrumentation-lifecycle.ts
@@ -81,6 +81,39 @@ export type InstrumentationToolOutput =
   | { readonly type: "result"; readonly output: unknown }
   | { readonly type: "error"; readonly error: unknown };
 
+/** Replay-stable row identity for every lifecycle operation. */
+export function sessionIdempotencyKey(sessionId: string): string {
+  return `session:${sessionId}`;
+}
+
+export function turnIdempotencyKey(sessionId: string, turnId: string): string {
+  return `turn:${sessionId}:${turnId}`;
+}
+
+export function attemptIdempotencyKey(scope: InstrumentationAttemptScope): string {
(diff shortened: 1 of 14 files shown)


packages/eve/src/harness/instrumentation-lifecycle.ts
```diff
@@ -81,6 +81,39 @@ export type InstrumentationToolOutput =
...
+  return `step:${scope.attemptId}`;
+}
+
+/** One model call occurs per AI SDK step within an eve attempt. */
+export function modelCallIdempotencyKey(
+  scope: InstrumentationAttemptScope,
+  stepNumber: number,
+): string {
+  return `model:${scope.attemptId}:${String(stepNumber)}`;
+}
```
ruiconti, line 103: Shouldn't this be the other way around -- `model:${stepNumber}:${scope.attemptId}`? i.e., isn't the attempt of the step? @!e0

ruiconti (approved):

packages/eve/src/harness/instrumentation-lifecycle.ts
ruiconti: Oh yes, I would want to disambiguate between AI SDK step and Worfkflow step attempt.

- `scope.stepAttemptId`
- `aiSdkStepNumber`

PR #1860, opened 2026-08-09
feat(eve): persist provider state and bound instrumentation handlers

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799. Stacked on #1859.

Instrumentation handlers can carry JSON-serializable state across workflow workers. State is isolated by provider and operation, survives workflow serialization and cancelled-step rollback, and is released at operation terminals or when its owning scope ends.

Input requests are durable pairs rather than attempt-owned work: state written at `input.requested` survives the originating model attempt and is restored for `input.resolved`. A timed-out input start is durably abandoned for that provider, matching other correlated starts, while metadata and terminal timeouts do not abandon an operation.

Authored handlers run concurrently for each publication while framework instrumentation remains ordered around them. Every phase receives one immutable framework-owned event snapshot.

### Validation

- Provider lifecycle and state suite: 26 tests passed
- `pnpm guard:invariants`

### Checklist

### DIFF
--- a/packages/eve/src/harness/instrumentation-dispatch.ts
+++ b/packages/eve/src/harness/instrumentation-dispatch.ts
@@ -1,0 +1,207 @@
+import {
+  abandonInstrumentationState,
+  instrumentationStateSlot,
+  isInstrumentationStateAbandoned,
+  releaseAllInstrumentationAttemptState,
+  releaseAllInstrumentationState,
+} from "#harness/instrumentation-state.js";
+import { createLogger, formatError } from "#internal/logging.js";
+
+import type {
+  CreateInstrumentationHooksOptions,
+  InstrumentationDispatchGroups,
+  InstrumentationEvent,
+  InstrumentationEventHandler,
+  InstrumentationHooks,
+  InstrumentationHooksInput,
+  InstrumentationProviderDefinition,
+} from "#harness/instrumentation-lifecycle.js";
+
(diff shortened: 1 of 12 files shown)


packages/eve/src/harness/instrumentation-dispatch.ts
```diff
@@ -0,0 +1,207 @@
...
+
+export function createInstrumentationDispatcher(
+  input: InstrumentationHooksInput,
+  options: CreateInstrumentationHooksOptions,
+): InstrumentationHooks {
+  const handlerTimeoutMs = options.handlerTimeoutMs ?? DEFAULT_HANDLER_TIMEOUT_MS;
+  const groups = normalizeDispatchGroups(input);
+  const snapshots = new WeakMap<object, unknown>();
+
+  const publish = async (event: InstrumentationEvent): Promise<void> => {
```
ruiconti, line 31: What is the otel standard terminology for the action of emitting an event? `publish` here can make it unclear that it is instrumentation-bound.

@!e0

ruiconti (approved):

packages/eve/src/harness/instrumentation-dispatch.ts
ruiconti: `emitInstrumentation` makes sense to me as it makes pertty obvious what it does, in the absence of a namespace e.g., `instrumentation.emit`

PR #1862, opened 2026-08-09
feat(eve): trace durable runtime actions

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799. Stacked on #1861.

The OpenTelemetry provider gives durable runtime actions a complete trace lifecycle across event races, worker changes, retries, and child-agent dispatch. Durable turns and actions retain replay-stable identities, child turns inherit the invoking `agent.action`, and physical `agent.step` spans link to their Workflow delivery without making it the logical parent. Provider-neutral terminal metadata maps to exact span end time, `agent.action.outcome`, `agent.action.error.code`, standard `error.type`, and action-level token usage for Datadog and other exporters.

Remote session creation propagates the preallocated action context as standard W3C `traceparent`. Updated eve receivers adopt it so remote child turns join the caller action trace in shared backends; older receivers ignore the header and preserve existing control flow. Persistent-session continuations retain the trace chosen at creation.

Human tool approvals now appear as durable `agent.approval` spans under the originating action. The approval span measures request-to-decision wait time and records the normalized outcome, while the sibling `ai.toolCall` measures later execution. Request and response content follow destination capture policy. `agent.turn` records only turn lifecycle events; session transitions are no longer attached to it.

Model calls use the GenAI `chat` operation with standard self-contained input, system-instruction, and output attributes. Long message arrays remain valid JSON and drop their oldest entries only when they exceed the per-attribute size cap. Runtime actions retain their originating step index while the result-consuming chat starts in step N+1.

### Validation

- Focused remote wire, action timing, and tracing suites pass at the stack head
- Local instrumentation scenario suite: 2 tests passed
- `pnpm --filter eve build`

### Checklist

### DIFF
--- a/.changeset/durable-action-tracing.md
+++ b/.changeset/durable-action-tracing.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Reconstruct durable `agent.action` spans when runtime actions settle, including across worker replacement, and record each action's exact caller-accepted duration, kind, outcome, stable error code, and subagent usage. Remote eve sessions join the caller action trace through W3C `traceparent`; older receivers may ignore the header. Human approval waits appear as durable `agent.approval` child spans, while chat spans use standard self-contained model input and output attributes.
--- a/packages/eve/scripts/vendor-compiled/declarations/@opentelemetry/api.d.ts
+++ b/packages/eve/scripts/vendor-compiled/declarations/@opentelemetry/api.d.ts
@@ -1,14 +1,21 @@
+export interface TraceState {
+  get(key: string): string | undefined;
+  serialize(): string;
+  set(key: string, value: string): TraceState;
+  unset(key: string): TraceState;
+}
+
 export interface SpanContext {
   isRemote?: boolean;
   spanId: string;
   traceFlags: number;
   traceId: string;
-  traceState?: unknown;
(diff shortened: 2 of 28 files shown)


packages/eve/src/execution/workflow-entry.ts
```diff
@@ -305,49 +305,59 @@ async function runDriverLoop(input: {
...
-    for (const attemptId of collectedAuthPayloads.keys()) {
-      if (!expectedAttemptIds.has(attemptId)) collectedAuthPayloads.delete(attemptId);
+    const attemptKey = (attemptId: string): string => `attempt:${attemptId}`;
+    const legacyKey = (name: string): string => `legacy:${name}`;
+    const expectedKeys = new Set([
+      ...park.expectedAttemptIds.map(attemptKey),
+      ...park.expectedLegacyNames.map(legacyKey),
+    ]);
+    for (const key of collectedAuthPayloads.keys()) {
+      if (!expectedKeys.has(key)) collectedAuthPayloads.delete(key);
```
ruiconti: Why is this needed? @!e0

chadhietala: Yea this wasn't needed. Backed it all out.

ruiconti (approved):

PR #1863, opened 2026-08-09
feat(eve): add the instrumentation provider layout

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799. Stacked on #1862.

Behind `experimental.instrumentationProviders`, this exposes the final `agent/instrumentation/` layout and runtime: file-derived slots, branded provider and OpenTelemetry declarations, durable handler state, complete setup context, and coordinated finalization, flush, and shutdown.

The public provider event map includes typed `input.requested` and `input.resolved` handlers plus action settlement time, outcome, error code, and usage. Authors receive the same durable metadata the OTel mapper consumes without importing runtime protocol types. Local traces and production Agent Runs remain overrideable reserved destinations, and the legacy `instrumentation.ts` layout remains available while the flag is off.

### Validation

- Public provider unit suite: 12 tests passed
- Authored provider integration suite passed at the stack head
- `pnpm docs:check`

### Checklist

### DIFF
--- a/packages/eve/src/harness/instrumentation-config.test.ts
+++ b/packages/eve/src/harness/instrumentation-config.test.ts
@@ -1,5 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
+import type { InstrumentationSetupContext } from "#public/instrumentation/index.js";
+
 const RUNTIME_GLOBAL_KEY = Symbol.for("eve.instrumentation-runtime");
 
 beforeEach(() => {
@@ -31,7 +33,7 @@ describe("instrumentation-config chunk-isolation regression", () => {
     vi.resetModules();
     const moduleA = await import("#harness/instrumentation-config.js");
     const config = { functionId: "test.instrumentation.cross-module.alice" };
-    moduleA.registerInstrumentationConfig(config, { agentName: "test-agent" });
+    await moduleA.registerInstrumentationConfig(config, { agentName: "test-agent" });
 
     vi.resetModules();
     const moduleB = await import("#harness/instrumentation-config.js");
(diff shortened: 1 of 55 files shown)


packages/eve/src/execution/node-step.ts
```diff
@@ -262,7 +262,11 @@ function resolveHarnessToolDefinition(input: {
       scope: def.name,
     }),
     frameworkAction:
-      isFrameworkTool && def.name === LOAD_SKILL_TOOL_NAME ? "load-skill" : undefined,
+      input.node.agent.config?.experimental?.instrumentationProviders === true &&
+      isFrameworkTool &&
+      def.name === LOAD_SKILL_TOOL_NAME
+        ? "load-skill"
+        : undefined,
```
ruiconti: Why is this necessary? @!e0

packages/eve/src/internal/instrumentation-layout.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * `config` is the single `agent/instrumentation.ts` default export. `providers`
+ * is a directory of them, one provider per file, keyed by the slot name the
+ * file derives (`instrumentation/otel.ts` → `otel`). Which one an agent may use
+ * is decided by `experimental.instrumentationProviders`, never by what happens
+ * to be on disk.
+ */
+export type InstrumentationLayout =
+  | { readonly kind: "config"; readonly modulePath: string }
+  | { readonly kind: "providers"; readonly modulePathsBySlot: Readonly<Record<string, string>> };
```
ruiconti: Is "config" and "providers" the right distinction? I'd expect that the "config" one to also declare providers..? Isn't single-slot/multi-slot the right distinction?

@!e0

packages/eve/src/execution/node-step.ts
chadhietala: I was trying distinguish tools and skills in the trace but this isn't really needed so I remove it.

packages/eve/src/internal/instrumentation-layout.ts
chadhietala: I changed this to `file` and `directory` since all we supported before was a single file and this change makes it a directory.

packages/eve/src/harness/instrumentation-config.test.ts
```diff
@@ -1,5 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
```
ruiconti, line 2: Unrelated to this change but can we move all `instrumentation-` to a directory `instrumentation/`?

ruiconti (approved):

PR #1892, opened 2026-08-10
fix(eve): namespace dynamic-tool step ids by module path

HugoRCD:
### Summary

While running an agent with several dynamic-tool modules, resumed sessions started failing with `references step function "..." which is not registered`. Fresh sessions were fine; only resume broke.

The dynamic-tool bundler transform registers each `execute` into a process-global step map under ids like `eve:dynamic-tool//__eve_dynamic_exec_0`. That name is only a per-transform counter, so two independently transformed files (or an app module and one pulled from `node_modules`) can emit the same key. The later registration overwrites the earlier one, and replay looks up an id that no longer points at the original function.

I first worked around this with a local `pnpm` patch on the published `eve` package (`patches/eve@0.30.6.patch`), adding a stable per-module scope into the step id. This PR lands that fix upstream:

- step ids become `eve:dynamic-tool//{moduleScope}/{hoistedName}` (same idea for dynamic remote-agent credentials)
- `moduleScope` is a short hash of a normalized module path (cwd / pnpm `node_modules` stripped so it stays stable across machines)
- the hoist counter resets per file so adding or reordering other modules does not renumber existing ones

Regular `"use step"` functions already use path-qualified ids (`step//{module}//{name}`); dynamic tools were the under-scoped exception.

**Compatibility note:** open sessions that already persisted the old unscoped ids will not resolve those steps after upgrade; new turns and new sessions are fine.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/internal/workflow-bundle/module-step-scope.test.ts src/internal/workflow-bundle/dynamic-tool-transform.test.ts src/internal/workflow-bundle/dynamic-remote-agent-transform.test.ts` (69 passed)

### Checklist

### DIFF
--- a/.changeset/dynamic-tool-module-step-scope.md
+++ b/.changeset/dynamic-tool-module-step-scope.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Namespace dynamic-tool and dynamic-remote-agent step ids by module path so independently transformed files no longer overwrite each other in the global step registry.
--- a/packages/eve/src/internal/workflow-bundle/dynamic-remote-agent-transform.test.ts
+++ b/packages/eve/src/internal/workflow-bundle/dynamic-remote-agent-transform.test.ts
@@ -65,7 +65,9 @@ export default defineDynamic({
       stepId?: string;
     };
     expect(Object.keys(remote)).not.toContain("__eveResolveRemoteAgentCredentials");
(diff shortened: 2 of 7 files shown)


HugoRCD:
Tbh, it's a bug I encountered on https://github.com/HugoRCD/evlog/tree/main/apps/evi, but I don't know if the fix is worth anything 😬

PR #1896, opened 2026-08-10
feat(eve): add editable setup questions

OwenKephart:
## Summary

- add a compound setup question for selecting an option with an editable text value
- preserve separate stable answer keys for the selection and edited text
- support inline editing when available and select-then-text fallback otherwise
- expose editable questions through the wire-safe setup question format

## Validation

- `pnpm --filter eve exec tsc --noEmit --pretty false`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/setup/ask.test.ts src/setup/setup-question-wire.test.ts`

### DIFF
--- a/packages/eve/src/setup/ask-editable.ts
+++ b/packages/eve/src/setup/ask-editable.ts
@@ -1,0 +1,84 @@
+import type { Prompter } from "./prompter.js";
+import { optionById, requiredOptionId } from "./question-options.js";
+import type { SelectOption } from "./ask.js";
+
+export interface EditableSelectQuestion<T> {
+  key: string;
+  message: string;
+  options: readonly SelectOption<T>[];
+  recommended?: T;
+  required?: boolean;
+  editable: {
+    key: string;
+    value: T;
+    label: string;
+    recommended: string;
+    validate?: (raw: string) => string | null;
+  };
+}
+
+export interface EditableSelectAnswer<T> {
+  value: T;
+  text?: string;
+}
+
+function selectedValue<T>(question: EditableSelectQuestion<T>, id: unknown): T {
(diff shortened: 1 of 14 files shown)


packages/eve/src/setup/ask-editable.ts
```diff
@@ -0,0 +1,84 @@
...
+
+function selectedValue<T>(question: EditableSelectQuestion<T>, id: unknown): T {
+  const match = optionById(question, id);
+  if (match !== undefined) return match.value;
+  const ids = question.options.map((option) => option.id).join(", ");
+  throw new Error(`Invalid answer for "${question.key}": ${String(id)}. Expected one of: ${ids}.`);
+}
+
+/** Renders the specialized one-row editable select through a terminal prompter. */
+export async function renderEditableQuestion<T>(
```
ruiconti, line 33: Can we piggy back on the existing inline input that we use for API key validation UI?

ruiconti: cc @!e0 can you provide reference and perma?

OwenKephart: Confirmed: this already reuses the shared inline-edit row used by the API-key picker. `#readSetupEditableSelect` and `#readProviderPicker` both render through `SetupInlineEditRow`/`inlineEditOption`; only the editor-specific presentation and state differ (`rename` for visible synchronous text, `key` for masked async validation). Reusing more of the provider picker would couple generic editable questions to API-key phases and validation, so no further refactor is needed here.

ruiconti (approved):

AndrewBarba (approved):

PR #1918, opened 2026-08-10
feat(eve): measure live stream event read age

alangenfeld:
### Summary

Each stream event already carries its durable write timestamp in `meta.at`, so following a newly created run now emits a `workflow.stream.follow.read` span from that timestamp to the moment eve reads the event. Only the live-follow path is instrumented — replayed streams from `getEventStream` are excluded — and span emission failures never interrupt event delivery.

### Validation

Unit tests added; `pnpm typecheck`, `oxlint`, `pnpm docs:check`, and `pnpm guard:invariants` pass.

### Checklist

### DIFF
--- a/packages/eve/src/execution/workflow-runtime.ts
+++ b/packages/eve/src/execution/workflow-runtime.ts
@@ -27,6 +27,7 @@ import {
 import { resolveInstalledPackageInfo } from "#internal/application/package.js";
 import { isEveDevEnvironment } from "#internal/application/dev-environment.js";
 import { createLogger, logError } from "#internal/logging.js";
+import { SpanKind, trace } from "#compiled/@opentelemetry/api/index.js";
 import {
   getHookByToken,
   getRun,
@@ -83,6 +84,7 @@ export const STABLE_WORKFLOW_NAMES: ReadonlySet<string> = new Set([
 const STABLE_ID_BASE = EVE_PACKAGE_INFO.name;
 
 const log = createLogger("execution.workflow-runtime");
+const workflowTracer = trace.getTracer("workflow");
 
 interface WorkflowHookRecord {
   readonly runId: string;
@@ -201,7 +203,7 @@ export function createWorkflowRuntime(config: {
 
       let events: ReadableStream<MessageStreamEvent> | undefined;
       const getEvents = () => {
-        events ??= parseNdjsonStream<MessageStreamEvent>(() => getRun(run.runId).getReadable());
+        events ??= createLiveEventStream(run.runId);
         return events;
       };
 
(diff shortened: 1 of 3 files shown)


packages/eve/src/harness/instrumentation/lifecycle.ts
```diff
@@ -532,6 +549,7 @@ export interface InstrumentationProviderDefinition {
     readonly "channel.delivery.cancelled"?: InstrumentationEventHandler<InstrumentationChannelDeliveryTerminalEvent>;
     readonly "channel.delivery.completed"?: InstrumentationEventHandler<InstrumentationChannelDeliveryTerminalEvent>;
     readonly "channel.delivery.failed"?: InstrumentationEventHandler<InstrumentationChannelDeliveryTerminalEvent>;
+    readonly "workflow.stream.follow.read"?: InstrumentationEventHandler<InstrumentationWorkflowStreamFollowReadEvent>;
```
chadhietala: I would prefer that we do not expose this into the public API here. This is supposed to be an eve centric event bridge and shouldn't have any notion of workflows in it.

chadhietala (changes requested):

packages/eve/src/execution/workflow-runtime.ts
```diff
@@ -421,6 +423,44 @@ function shouldRouteToLatestDeployment(): boolean {
...
 }
 
+function createLiveEventStream(sessionId: string): ReadableStream<MessageStreamEvent> {
+  let sequence = 0;
+  return parseNdjsonStream<MessageStreamEvent>(() => getRun(sessionId).getReadable()).pipeThrough(
+    new TransformStream({
+      transform(event, controller) {
+        recordLiveStreamEventRead({ event, sequence, sessionId });
+        sequence += 1;
+        controller.enqueue(event);
```
chadhietala: I think it would be better to do something like:

```ts
sequence += 1;
const readAtMs = Date.now();
controller.enqueue(event);
recordLiveStreamEventRead({ readAtMs, event, sequence, sessionId });
```

This way we enqueue the event without the overhead of running the telemetry but logically still capture the span time.

chadhietala (approved):
One other question, do we actually want per-trace latency or do you want something more like a distribution? If workflows has any meter's setup internally you could consider using a histogram:

```ts
import { metrics, SpanKind } from "@opentelemetry/api";

// Create the meter + instrument ONCE at module scope
const meter = metrics.getMeter("workflow.stream");
const streamReadLatency = meter.createHistogram("workflow.stream.follow.read.latency", {
  unit: "ms",
  description: "write→read latency per streamed event",
});

// later in the file

function recordLiveStreamEventRead(input: {
  readonly event: MessageStreamEvent;
  readonly sequence: number;
  readonly sessionId: string;
  readonly readAtMs: number; // captured at arrival, passed in
}): void {
  const writtenAtMs = Date.parse(input.event.meta?.at ?? "");
  if (!Number.isFinite(writtenAtMs)) return;
  const latencyMs = input.readAtMs - writtenAtMs;

  try {
    // the distribution across all events. sessionId only as it bounds cardinality
    // sequence number can't be passed though
    streamReadLatency.record(latencyMs, {
      "workflow.run.id": input.sessionId,
    });
  catch {
     // swallow the error
  }
}```

alangenfeld:
>  If workflows has any meter's setup internally

we do not currently, will keep this idea in mind though

PR #1928, opened 2026-08-11
fix(registry): publish requirements with Eve releases

benpankow:
## Summary

Right now, the registry updates with merges to `main`, but `npm` releases happen as part of the Changesets PR process. This PR keeps the public registry compatible with the currently published Eve package, while staging next-release registry requirements alongside the feature work that needs them.

- `apps/docs/registry.json` powers the docs site and remains compatible with the current `packages/eve` version.
- `apps/docs/registry.staged-requirements.json` is a small, editable list of registry items that require the next Eve release.
- `pnpm version-packages` runs Changesets first, then materializes those items’ `meta.eve.requires` values using the selected `packages/eve/package.json` version.

CI validates that we don't erroneously bump registry versions outside of this process.

## Validation

- `node --test scripts/materialize-registry-release-requirements.test.mjs scripts/validate-public-registry-requirements.test.mjs`
- `node scripts/validate-public-registry-requirements.mjs`
- `pnpm --dir apps/docs registry:build`

### DIFF
--- a/apps/docs/registry.json
+++ b/apps/docs/registry.json
@@ -1200,7 +1200,7 @@
             "bin": "eve",
             "args": ["integration", "setup", "slack"]
           },
-          "requires": ">=0.31.4",
+          "requires": ">=0.27.8",
           "docs": "/docs/channels/slack",
           "implementation": "native"
         }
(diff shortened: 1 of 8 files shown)


cmpadden (approved):

apps/docs/registry.json
```diff
@@ -1200,7 +1200,7 @@
             "bin": "eve",
             "args": ["integration", "setup", "slack"]
           },
-          "requires": ">=0.31.4",
+          "requires": ">=0.27.8",
```
chadhietala, line 1203: Just to clarify this is the min-version correct?

PR #1935, opened 2026-08-11
feat(eve): trace durable channel deliveries

chadhietala:
### Summary

Related to the instrumentation providers proposal in #1799. Stacked on #1864.

Inbound channel work now has a durable, provider-neutral lifecycle from processing start through the resulting turn terminal. Each accepted send/respond receives an opaque delivery ID that survives session creation, inbox persistence, coalescing, active-turn buffering, and descendant routing; several deliveries may map to one turn without losing identity.

OpenTelemetry maps each lifecycle pair to an `agent.channel.delivery` `CONSUMER` span under the durable session window. When HTTP channel tracing is enabled, the short-lived `SERVER` span is linked with `eve.link.type=channel.request` rather than used as a parent. Adapter-only interactions close without inventing a turn, failures and cancellations retain explicit outcomes, content capture includes only known eve delivery fields, and metadata remains available without message content.

Delivery-first initialization preserves an incoming remote `traceparent`, so a remote agent's delivery and turn remain children of the dispatching `agent.action` rather than opening a disconnected trace.

Outbound platform sends are deliberately excluded. A turn may emit zero or multiple Slack/Telegram/Twilio/etc. messages with platform-specific acknowledgement semantics; those will be modeled as separate child operations in a follow-up rather than delaying the inbound delivery terminal.

### Validation

- 245 focused unit tests passed after rebasing off #1865
- 15 workflow integration tests passed
- 2 local instrumentation scenarios passed
- `pnpm --filter eve build`
- `pnpm docs:check`
- `pnpm guard:invariants`

### Checklist


packages/eve/src/harness/channel-delivery-instrumentation.ts
```diff
@@ -0,0 +1,121 @@
...
+      rootSessionId: input.rootSessionId,
+      sequence: input.sequence,
+      sessionId: input.sessionId,
+      turnId: input.turnId,
+    };
+    active.push(item);
+    await input.hooks.publish({
+      agentName: item.agentName,
+      delivery,
+      idempotencyKey: channelDeliveryIdempotencyKey(input.sessionId, delivery.deliveryId),
```
ruiconti, line 96: AFAIK there is already a delivery idempotency key, which is the channel's inbound requestId. It is already written to workflow attributes.

@!e0 add permalink receipts

packages/eve/src/execution/workflow-steps.ts
```diff
@@ -286,34 +325,39 @@ export async function turnStep(rawInput: TurnStepInput): Promise<DurableStepResu
...
-      }),
-      refreshDynamicSessionToolsForRuntimeRevision({
-        ctx,
-        resolvers: dynamicToolResolvers,
-        event: refreshEvent,
-        messages: initialSession.history,
-        runtimeRevision: dynamicRuntimeRevision,
-      }),
-    ]);
+  try {
```
ruiconti, line 329: In the year of our lord 2026 and GH doesn't have semantic diffs

packages/eve/src/execution/workflow-steps.ts
```diff
@@ -621,68 +668,3 @@ export function resolveEffectiveOutputSchema(input: {
...
-          parentSessionId: input.sessionState.sessionId,
-          requestId: input.delivery.kind === "deliver" ? input.delivery.requestId : undefined,
-          rootSessionId: readRootSessionId(input.serializedContext) ?? input.sessionState.sessionId,
-        }),
-      ),
-    },
-  );
-
-  return { runId: run.runId };
-}
```
ruiconti, line 692: Why did we move this out to its own module?

packages/eve/src/harness/channel-delivery-instrumentation.ts
```diff
@@ -0,0 +1,121 @@
...
+  type ActiveChannelDelivery,
+} from "#context/keys.js";
+import type {
+  InstrumentationChannelDeliveryOutcome,
+  InstrumentationChannelDeliveryTerminalEvent,
+  InstrumentationHooks,
+} from "#harness/instrumentation-lifecycle.js";
+import { channelDeliveryIdempotencyKey } from "#harness/instrumentation-lifecycle.js";
+
+export async function publishChannelDeliveryStarts(input: {
```
ruiconti: Can we instead do `instrumentChannelDelivery` with overloads for both start and end args (instead of two separate functions). `publish` is an overloaded term that obfuscates that this is running telemetry code.

packages/eve/src/execution/workflow-entry.ts
ruiconti, line 575: Curious if this won't break wire contract of turns from different deployments?

@!e0 to confirm.

De-ASI-INTERFACE (approved):

packages/eve/src/execution/route-proxied-deliver-step.ts
```diff
@@ -0,0 +1,129 @@
...
+    remainder:
+      parentPayloads.length === 0
+        ? undefined
+        : {
+            ...sourceDelivery,
+            deliveryMetadata: parentMetadata.length === 0 ? undefined : parentMetadata,
+            payloads: parentPayloads,
+          },
+  };
+}
```
ruiconti: Unclear why an instrumentation PR had to move these modules around.

packages/eve/src/harness/channel-delivery-instrumentation.ts
ruiconti: @!e0 then why is this binding named `channelDeliveryIdempotencyKey` if it's not a "delivery idempotency key"??

ruiconti: @!e0 so isn't this delivery the channel to workflow delivery??

ruiconti: @!e0 which legal scenario do we expect the same requestId to trigger multiple deliveries? isn't that an invariant we want to preserve?

packages/eve/src/execution/route-proxied-deliver-step.ts
chadhietala: moved it back. Stochastic things doing stochastic things

packages/eve/src/harness/channel-delivery-instrumentation.ts
chadhietala: Done

ruiconti (approved):

PR #1950, opened 2026-08-11
feat(eve): prepare public canary artifacts

OwenKephart:
### Summary

Add a Git-linked canary artifact project based on `vercel/workflow`'s tarball pattern, with durable storage in public Vercel Blob. Each checked-out commit becomes a SHA-versioned tarball; `/canary/latest` is the convenient moving entry point, while canary-created projects pin the exact immutable Blob URL embedded by their CLI.

The project uses deployment OIDC for Blob writes, keeping canaries out of npm history without registry login or long-lived publishing credentials. Stable `eve` releases continue scaffolding normal semver dependencies.

### Validation

- `pnpm --filter eve-canary-artifacts test`
- `pnpm check:deps`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/setup/scaffold/version-tokens.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts test/scenarios/post-build-scripts.scenario.test.ts`

### Checklist

- [ ] I linked an issue with prior discussion confirming this change is wanted *(intentionally omitted per maintainer direction)*


apps/package-artifacts/api/package.js
```diff
@@ -0,0 +1,49 @@
...
+  if (!SHA_PATTERN.test(sourceSha ?? "")) {
+    return packageNotFound(response);
+  }
+
+  const manifest = await resolveManifest(sourceSha);
+  if (manifest === undefined) return packageNotFound(response);
+
+  response.setHeader(
+    "Cache-Control",
+    ref === "main" ? "public, max-age=60" : "public, max-age=31536000, immutable",
```
AndrewBarba, line 22: any non-sha (ie: branch) should have a 60s max age

AndrewBarba (approved):

apps/package-artifacts/api/package.js
OwenKephart: yup only main handled currently, but will make sure to handle that when I expand this to handling non-main branches

Issue #1673, opened 2026-08-05
Design durable hierarchical progress projections


benpankow:
## Summary

eve currently lets channels react directly to low-level lifecycle events. Slack uses those events to replace a thread status, but the model does not represent progress durably or compose progress from parallel delegated agents.

Design a channel-neutral progress projection that:

- reduces agent lifecycle facts into durable desired progress state;
- allows an authored reducer to customize that projection;
- bubbles delegated-agent progress through each parent, keyed by stable call identity;
- lets the root channel reconcile the projection into its native presentation, such as one replaceable indicator, an updated message, or a hierarchical collection of blocks;
- remains deterministic and replay-safe across durable workflow steps.

## Questions

- Which lifecycle facts are reducer inputs, and should child sessions forward raw facts or reduced snapshots?
- Where does reducer state live, and how is its version migrated with pinned workflows?
- How do terminal, waiting, authorization, cancellation, and nested-child states compose?
- What is the boundary between a pure progress reducer and effectful channel reconciliation?
- Which guarantees are required to prevent duplicate posts while allowing in-place updates?
- How should high-frequency reasoning updates be coalesced or rate-limited without losing durable state?

## Initial scope

Research and authoring/API design first. Preserve existing channel behavior. Treat Slack thread statuses as one renderer, not as the progress data model.

luxons-techsaviness:
We're building on eve at Lembra (event-planning platform) and hit a gap that sits right against this design. Sharing a concrete use case, since it answers one of the open questions above — *"which lifecycle facts are reducer inputs?"*

**Use case.** One tool call runs a supplier research sweep that takes 60–180 seconds — several web searches, then extraction across the results. During that window the buyer sees nothing move. The work is real and the milestones are meaningful ("searching", "12 candidates found", "extracting details", "4 shortlisted"), but they all happen *inside a single tool invocation*.

**What we verified on `eve@0.27.0`** (read from installed `dist`, not docs):

- `ToolContext` is `{session, getSandbox, getSkill, abortSignal, callId, toolName, getToken, requireAuth}` — no emit/publish member.
- `HandleMessageStreamEvent` is a closed union in `dist/src/protocol/message.d.ts`, with no custom/progress member (`meta` reads as framework timing rather than an author slot).
- No `emitEvent` / `publishEvent` / `emitProgress` / `customEvent` symbol across `dist/src/public` or `dist/src/protocol`.
- Hooks are observe-only, so they can carry such a fact but can't originate one.

[...]

PR #1957, opened 2026-08-11
research: durable hierarchical progress

benpankow:
### Summary

Related to #1673. This research revises the durable hierarchical progress proposal with evidence from the implementation spike and merged background-task updates in #2113.

The proposal keeps a framework-owned work graph as the canonical projection of observed turn, action, blocker, and task lifecycle state. It drops child-stream polling from the proposed architecture and instead composes the existing child-only `task_update({ message })` transport into the graph. The tool remains a free-form semantic update; stable task, turn, step, and call identity is added internally so the runtime can reduce updates deterministically without letting a child claim lifecycle state.

Slack remains the proving ground. Its status, task-card rendering, message ownership, terminal cleanup, and possible streaming-task integration stay channel-specific and internal. The proposal defers `ctx.reportProgress()` and a generic channel API until task updates and tool partials show a concrete missing contract. This PR remains a design artifact and changes no runtime behavior.

### Review focus

- Observed task lifecycle versus child-authored `task_update` annotations
- Reusing task inbox/callback delivery instead of polling child streams
- Keeping `task_update` free-form while structuring its internal identity
- Separating parent model awareness from immediate channel presentation
- Whether the current task message should eventually persist in `TaskView`

### Validation

- `pnpm exec oxfmt research/durable-hierarchical-progress.md`
- `git diff --check`
- No runtime tests were run; this PR changes only a research proposal

### Checklist

### DIFF
--- a/research/durable-hierarchical-progress.md
+++ b/research/durable-hierarchical-progress.md
@@ -8,246 +8,352 @@ last_updated: "2026-08-05"
 
 ## Summary
 
-eve should represent progress independently from any channel presentation. Agent
-lifecycle facts reduce into a durable progress projection; delegated agents
-publish their projection to their parent, which incorporates it into its own
-projection. Only the root channel renders the resulting projection into an
-external presentation.
+eve should maintain a durable, framework-owned work graph for the active turn
+and its delegated work. The graph is derived from authoritative runtime
+lifecycle state, composes through child sessions, and remains independent of
+channel effects. A root channel may render a safe, selected view of that graph.
 
 ```text
-child lifecycle facts ──reduce──▶ child projection
-                                      │ durable child update
-                                      ▼
-parent facts + child projections ──reduce──▶ parent projection
-                                              │
-                                              ▼
-                                       channel renderer
-                                  indicator | message | blocks
+runtime lifecycle + direct child state
+                │
+                ▼
+       durable work graph
+                │
+                ▼
+    channel-specific presentation
 ```
 
-A Slack thread status is one possible rendering. A replaceable thread message,
-a Block Kit tree of parallel work, a terminal spinner, or no visible rendering
-are equally valid. Progress is desired state, not a command to call a specific
-channel API.
+The first implementation should prove this boundary using existing lifecycle
+facts and local subagents. It should not initially define a general progress
+platform around plans, arbitrary annotations, public client events, remote
+capability negotiation, or custom channel APIs.
 
-## Current Slack behavior
+A later PR should add action-local `reportProgress()` and automatic MCP progress
+adaptation after action ownership, child propagation, and rendering semantics
+are stable.
 
-Slack status text currently comes directly from channel event handlers in
-`public/channels/slack/defaults.ts`:
+## Problem
 
-| Source                                         | Current projection                                                                                                    |
-| ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
-| inbound mention or DM                          | `Thinking...` before runtime dispatch                                                                                 |
-| `turn.started`                                 | `Working...`                                                                                                          |
-| `reasoning.appended`                           | first non-empty reasoning line, locally throttled                                                                     |
-| `actions.requested`                            | preceding assistant narration when available, otherwise a label derived from the first requested action and `+N more` |
-| `message.completed`                            | preserve narration before tool calls; otherwise post the answer or clear an empty status                              |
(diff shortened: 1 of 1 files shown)


research/durable-hierarchical-progress.md
```diff
@@ -0,0 +1,485 @@
...
+```
+
+The parent:
+
+1. binds the update to a running child handle by `callId` and session id;
+2. rejects stale revisions;
+3. replaces the child snapshot on the owning action;
+4. increments its own revision only when desired state changes;
+5. republishes its snapshot upward when it is itself delegated;
+6. rejects every child update after the action becomes terminal.
```
ruiconti: I have tried implementing this some time ago and the tl;dr is that because the parent runs in a serverless function that is bound to timeout, there isn't really a clean way to consume the child's durable stream without introducing a lot of overhead and orchestration.

ruiconti: This also becomes somewhat untenable once you have
- N concurrent children
- They are running in the background i.e., not turn-bound

research/durable-hierarchical-progress.md
```diff
@@ -0,0 +1,485 @@
...
+  readonly unit?: string;
+}
+
+export default defineTool({
+  async execute(input, ctx) {
+    await ctx.reportProgress({
+      message: "Indexing documents",
+      progress: 250,
+      total: 1_000,
+      unit: "documents",
```
ruiconti: How is this different than https://eve.dev/docs/tools#stream-preliminary-tool-results?

@!e0 can you help answer

ruiconti: @!e0 this is not clear to me

> Channels receive one canonical snapshot instead of reconstructing progress from each event stream

why is it desirable to have two sources-of-truth for session state? i.e., what is the current limitation that prevents such hierarchy from being derived from session events?

ruiconti: > but the child’s turn, step, and action events live in a separate session stream

So this proposal is mostly around child progress propagation? If that's the case, what is the proposed path for handling
1. consuming a child's event stream considering the runtime limitations of the parent running in a serverless workflow function? re-create the actor model used for the session loop just to consume the stream?
2. childs whose lifetime exceeds the turn i.e., childs running in the background

@!e0

ruiconti: > The parent turn already suspends on a durable hook iterator. A child update wakes the workflow, the parent validates it against the running handle, applies the latest revision, and republishes upward if needed.

How is this different than the `task_send` updates that will land shortly? `task_send` was designed explicitly for a child subagent to propagate updates to its parent - both remote and local.

@!e0

research/durable-hierarchical-progress.md
```diff
@@ -0,0 +1,485 @@
...
+
+For parallel work:
+
+```text
+Researching three approaches…
+```
+
+The renderer owns:
+
+- human labels for known action kinds;
```
ruiconti: This is a good one. Tool declaration should be the place where labels are defined.

research/durable-hierarchical-progress.md
ruiconti: > The remaining work here is the framework-generated payload and reduction semantics: deriving progress from lifecycle state, merging updates by task/call identity and revision, rejecting stale updates after settlement, and exposing the resulting view to channels without putting progress into model history

So most of the work at the channel layer reducing the event stream? @!e0

ruiconti: @!e0 note that the intent for `task_send` is to send messages between _agents_ i.e., the child decides when to send and the parent decides what reasoning gets exposed out of it.

research/durable-hierarchical-progress.md
benpankow: Hm, I wonder if some sort of periodic parent wake makes sense, e.g. https://github.com/vercel/eve/pull/1572 requires that anyway to keep Slack status alive, since it disappears after 2m of inactivity. That keeps the amount of work bounded for many concurrent children/non-turn-bound child work.

benpankow: Though I guess rolling up the child progress state into session state for the parent to probe would probably qualify as "a lot of overhead/orchestration"

benpankow:
Closing as superseded by the activity protocol, collector, observer, and rendering work that followed in #2293, #2294, #2306, #2304, and #2310. The later architecture also moved away from the child-stream polling explored here.

PR #1985, opened 2026-08-12
feat(eve): reduce active turn work graph

benpankow:
### Summary

Related to #1673. Add an internal, durable work-graph reducer for the active turn so eve has one framework-owned view of running, blocked, and settled actions rather than requiring channels to reconstruct it from stream events. The graph is derived from existing lifecycle facts and persisted in serialized context; it is not public and does not change channel rendering yet.

The graph preserves native turn and `stepIndex` boundaries, stable action `callId`s, authorization blockers, and local child-session ownership. Completed and terminal state is monotonic, so late child-dispatch events cannot reopen a settled action.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/work-graph.test.ts`
- `pnpm --filter eve exec tsc --noEmit -p tsconfig.json`

### Checklist

- [x] I linked an issue with prior discussion confirming this change is wanted
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [ ] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

PR #1986, opened 2026-08-12
feat(eve): preserve dispatched local subagent work

benpankow:
### Summary

Related to #1673. Build on #1985 by preserving work-graph changes made when the runtime dispatches a local subagent outside the ordinary harness event emitter. A local subagent is an authored `subagent-call` that runs through this eve deployment; remote-agent progress remains out of scope for this layer.

The dispatch step adopts updated serialized context only when local-subagent work state changed, so the parent’s durable cursor retains direct local-subagent ownership during its wait. This intentionally does not publish live child snapshots: parent-owned live rendering needs a separate cross-session snapshot-query design rather than waking the parent once for every committed child step.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/turn-workflow.test.ts src/execution/workflow-steps.test.ts`
- `pnpm --filter eve exec tsc --noEmit -p tsconfig.json`

### Checklist

### DIFF
--- a/packages/eve/src/execution/dispatch-runtime-actions-shared.ts
+++ b/packages/eve/src/execution/dispatch-runtime-actions-shared.ts
@@ -8,6 +8,7 @@
  */
 
 import { buildAdapterContext } from "#channel/adapter-context.js";
+import type { ContextContainer } from "#context/container.js";
 import {
   callAdapterEventHandler,
   type ChannelAdapter,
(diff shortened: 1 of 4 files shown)


benpankow:
Closing as superseded by the later activity collector and observer architecture in #2294 and #2306. Local subagent activity is no longer being pursued through this serialized work-graph snapshot design.

PR #2005, opened 2026-08-12
docs: add native shared-sandbox example

ruiconti:
## What

- add a runnable app where a parent agent and declared worker share the parent's live sandbox through `parent.sandbox`
- add an eval that proves files written by either agent are visible to the other
- add the example to the workspace typecheck
- make `docs:check` fail when a code block titled with an `examples/...` path drifts from its source file

## Sandbox flow

PR #1889 adds the runtime path: a child sandbox definition returns `parent.sandbox`, and eve restores the parent's persisted sandbox identity for the child session.

This PR adds the runnable consumer of that API. The parent creates a file, dispatches the worker, and the worker reads that file and writes a proof file back into the same `/workspace`. Agent conversations and durable state remain separate; only the sandbox is shared.

## Docs/source flow

Before: example snippets in docs could drift from runnable source without CI noticing.

After: `scripts/check-doc-examples.mjs` resolves code blocks titled with an `examples/...` path and compares them byte-for-byte with the referenced file as part of `docs:check`.

## Stack

- base: #1889
- this PR contains only the example, eval, docs navigation, and docs/source consistency check
- the native `parent.sandbox` API and runtime implementation are reviewed in #1889

### DIFF
--- a/docs/patterns/meta.json
+++ b/docs/patterns/meta.json
@@ -5,6 +5,7 @@
     "multi-tenant-memory",
     "dynamic-scheduling",
     "multi-tenant-auth",
-    "multi-tenant-approvals"
+    "multi-tenant-approvals",
+    "shared-sandbox"
   ]
 }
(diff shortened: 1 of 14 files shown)


ruiconti:
Closing this example PR. The parent-sandbox API needs to account for the current skills/extensions and workspace-mount lifecycle first; that work will move to a new draft PR.

PR #2043, opened 2026-08-13
feat(eve): support agent workspace member packages

OwenKephart:
### Description

Allows individual agents in multi-agent projects to have their own package.json entries as long as the root package.json claims that subdirectory as a member of its workspace.

### PR Checklist

### DIFF
--- a/packages/eve/src/internal/vercel/build-agent-workspace.integration.test.ts
+++ b/packages/eve/src/internal/vercel/build-agent-workspace.integration.test.ts
@@ -1,4 +1,4 @@
-import { lstat, mkdir, mkdtemp, readFile, writeFile } from "node:fs/promises";
+import { access, mkdir, mkdtemp, readFile, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
@@ -43,12 +43,9 @@ describe("buildAgentWorkspace", () => {
         src: "^/eve/agents/support/eve/v1/(.*)$",
       },
     ]);
-    const serviceRoot = await lstat(join(root, ".eve", "vercel-services", "eve-support"));
-    const serviceOutput = await lstat(
-      join(root, ".eve", "vercel-services", "eve-support", ".vercel", "output"),
-    );
-    expect(serviceRoot.isDirectory()).toBe(true);
-    expect(serviceOutput.isSymbolicLink()).toBe(true);
+    await expect(
+      access(join(root, ".eve", "vercel-services", "eve-support")),
+    ).resolves.toBeUndefined();
 
     expect(config.services["eve-support"]).toEqual({
       buildCommand:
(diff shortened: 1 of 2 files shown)


packages/eve/src/setup/primitives/pm/pnpm.ts
```diff
@@ -205,6 +205,26 @@ export function findClaimingAncestorPnpmWorkspaceRoot(projectRoot: string): stri
...
 
+/** Returns the workspace root only when its manifest explicitly claims `projectRoot`. */
+export function findStrictlyClaimingAncestorPnpmWorkspaceRoot(
+  projectRoot: string,
+): string | undefined {
+  const workspaceRoot = findAncestorPnpmWorkspaceRoot(projectRoot);
+  if (workspaceRoot === undefined) return undefined;
+
+  try {
+    const patterns = parsePnpmWorkspacePackagePatterns(
```
ruiconti: This rejects a valid pnpm workspace.

Concrete repro:

```yaml
packages:
  - "agents/*" # deployable agents
```

`pnpm list --depth -1 --json` includes `agents/support`, so pnpm treats it as a workspace member. `resolveAgentWorkspace()` instead throws `agents/support/package.json defines a child package that is not a member...`.

The parser captures the inline comment as part of the glob. The new strict check therefore makes a valid YAML comment change membership semantics. I reproduced this at `5478a3b` with a focused integration test.

Expected: if pnpm claims the package, eve accepts it.

Actual: adding an inline YAML comment makes eve reject it.

Can we handle comments here and keep this exact case as a regression test?

packages/eve/src/internal/vercel/build-agent-workspace.integration.test.ts
```diff
@@ -79,6 +79,20 @@ describe("buildAgentWorkspace", () => {
...
+    await writeFile(join(root, "pnpm-workspace.yaml"), 'packages:\n  - "agents/*"\n');
+    await writeFile(
+      join(root, "agents", "support", "package.json"),
+      JSON.stringify({ private: true, scripts: { build: "generate && eve build" } }),
+    );
+
+    const workspace = await loadAgentWorkspace(root);
+    const output = await buildAgentWorkspace(workspace);
+    const config = JSON.parse(await readFile(join(output, "config.json"), "utf8"));
+    expect(config.services["eve-support"].buildCommand).toContain("pnpm run build");
```
ruiconti: Can we add a real build scenario for the behavior this test represents?

This assertion proves only that the generated command contains `pnpm run build`. It stays green if the command runs from the wrong directory, the custom build step does not run, or a sibling workspace dependency cannot be resolved.

The concrete scenario I used was:

1. `agents/support/package.json` depends on `@fixture/shared` through `workspace:*`.
2. Its build script writes `agent/generated.mjs`, then runs `eve build`.
3. The agent imports both the generated module and `@fixture/shared`.
4. A real `vercel build` must produce the support service function.

That scenario passes at `5478a3b`, which is good. Keeping it in `hostless-agent-workspace.scenario.test.ts` would make the advertised package boundary a tested contract instead of an inferred one.

OwenKephart:
Recreated as #2792 after this PR was accidentally auto-closed when its branch was reset during `gh stack` adoption. #2792 preserves this implementation and the fixes from the review threads here, and is registered in GitHub stack #2793.

PR #2097, opened 2026-08-14
feat(eve): extensions - support channels

### LINKED ISSUE #2096: Allow extensions to contribute channels

Extensions can package tools, connections, skills, hooks, and instructions, but channel files are currently dropped when an extension is mounted. This prevents reusable inbound adapters and their routes from being distributed as part of an extension.

### Proposed solution

Compile channels from mounted extensions alongside the existing contribution slots. Namespace the channel ID with the extension mount, preserve the authored HTTP or WebSocket route paths, scope the module source so extension configuration resolves correctly, and advertise a versioned channel compatibility contract.

### Alternatives considered

Consumers can copy channel files into each agent, but that defeats extension ownership and upgradeability. Prefixing route paths with the mount namespace would avoid collisions but would unexpectedly change externally configured webhook URLs.

AndrewBarba:
### Summary

Closes #2096. Extensions previously dropped authored channels during composition; mounted extensions now contribute channel routes with mount-prefixed channel IDs while preserving their authored HTTP and WebSocket paths. Extension channel modules retain their scoped configuration, and the new channel capability epoch lets consumers reject incompatible packages. Directory mounts follow the existing consumer-wins override policy by channel ID and preserve every route from the winning channel; cross-channel route collisions remain the application route registry's responsibility.

### Validation

The installed-extension scenario builds a dist-only package, verifies its channel capability metadata, confirms the namespaced ID and unchanged route, and executes the handler with bound extension config. A focused merge test covers a multi-route consumer override alongside an unrelated packaged channel, and both installed and directory-override scenarios pass.

### Checklist

### Diff size

**Docs** — 4 files · `+13 / -6`

Documents extension-authored channel placement, namespaced identity, unchanged route paths, and directory-mount override precedence, with a concise release note.

**Implementation** — 8 files · `+72 / -9`

Keeps channel composition and override precedence in the existing extension compiler path and adds the required capability contract. The generated epoch-1 API metadata accounts for 17 additions.

**Tests** — 2 files · `+41 / -0`

Extends one installed-extension scenario to cover build metadata, composition, module resolution, route identity, and scoped configuration, plus focused coverage that preserves all routes from a winning override channel.


packages/eve/src/compiler/normalize-manifest.ts
```diff
@@ -205,6 +205,7 @@ async function compileAgentResources(
       consumerAgentRoot: manifest.agentRoot,
       externalDependencies,
     });
+    compiledChannels.push(...contributions.channels);
```
ruiconti, line 208: @!e0 how is this handling potential channel identity conflicts? I'd expect a similar treatment as other entities per lines below.

ctgowrie (approved):

packages/eve/src/compiler/normalize-manifest.ts
AndrewBarba: Fixed in d9f844650. `mergeContributions` now treats channels as name-grouped contributions: a consumer override removes the packaged same-name channel while preserving every route from the override definition. The focused test covers a two-route override plus an unrelated packaged channel, and the extension override docs now state the precedence explicitly.

PR #2142, opened 2026-08-14
feat(eve): memory - add append-only scoped lifecycle

AndrewBarba:
### Summary

Related to #1510 and stacked on #1581. This adds path-authored memory slots with provider-owned recall, save, and scope-bound tools, plus explicit namespace, contextual scope, recall visibility, and an optional model-facing slot description. Every non-empty recall becomes an attributed durable `role: "user"` message immediately before turn input or after compaction; `null`, `undefined`, and blank results append nothing and never replace or clear history. eve filters attributed recall before model calls and compaction, strips private attribution at model boundaries, and exposes `getMemoryMessageAttribution()` so providers can identify prior slot-and-scope recalls without depending on the internal metadata shape. Async provider tools reuse the existing `turn.started` `defineDynamic` engine and its durable approval, authorization, and replay behavior.

### Validation

After rebasing onto current `main`, the full unit suite passes 7,007 assertions with one skipped, and nine focused memory integration tests pass. Coverage exercises durable append order, no-op recall results, scope and session visibility, hidden-recall compaction safety, dynamic tool continuations, cold-start callback hydration, and the rule that recalled context is not replayed as a human-authored compaction prompt.

### Checklist

### Diff size

**Docs** — 7 files · `+463 / -126`

Documents authoring, append-only recall, trusted scope context, namespace and visibility choices, model-facing descriptions, lifecycle timing, and the release impact.

**Implementation** — 40 files · `+3015 / -230`

The cross-cutting runtime work covers discovery, validation, scope locking, durable attribution, prompt filtering, compaction, the `defineDynamic` memory-tool adapter, and parked-call origin restoration. Most changes stay in the existing compiler, context, and harness seams because recall must survive workflow replay while remaining absent from model metadata.

**Tests** — 26 files · `+4413 / -16`

The larger test surface locks down the public API and high-risk durability, multiplayer visibility, compaction, addressing, dynamic closure, cold-start hydration, continuation, and packed-declaration invariants.


packages/eve/src/context/memory-tool-lifecycle.ts
```diff
@@ -0,0 +1,326 @@
...
+  if (active === null) return session;
+  return anchorUnanchoredVisibleMemoryProjections({
+    anchorIndex: input.projectionAnchorIndex,
+    session,
+    slots: active.slots,
+  });
+}
+
+/** Replays parked tools from their captured dynamic metadata. */
+export async function resolveMemoryApprovalTools(input: {
```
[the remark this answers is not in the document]
AndrewBarba: Fixed by reusing the existing defineDynamic machinery. The resolver runs once after turn-start recall, and durable dynamic-tool metadata records the callback step functions and tool origin. On approval or inline-auth resume, eve reconstructs the exact originating definition before execute, approval.response, or toModelOutput, preserves the original scope, and does not rerun the resolver. The integration coverage includes parked approval and inline-auth continuations with an intervening participant turn.

AndrewBarba:
Freezing this branch (now draft): architectural review of the memory stack found generic prerequisites that should land on `main` first — one prepared-history seam for every message-bearing consumer (#2344), durable dynamic-tool callback descriptors (#2345), parked-call origins (#2346), and a generic runtime-tool contribution seam (#2347). The public surface is also being revised on #1581 (event-based recall changes, no memory-specific harness machinery).

The core memory lifecycle will be rebuilt on those foundations rather than iterated here, so please hold review on this PR. Reusable pieces (compiler/discovery/manifest/portability tests, provider types) will be harvested from this branch's amended tip.

packages/eve/src/harness/tool-loop.ts
```diff
@@ -815,10 +905,52 @@ export function createToolLoopHarness(config: ToolLoopHarnessConfig): StepFn {
...
+              session,
+            });
+      approvalMemoryTools = approvalMemoryResolution?.tools ?? new Map();
+      approvalTools = mergeMemoryTools(config.tools, approvalMemoryTools);
+    } catch (error) {
+      return failMemoryOperation(error, {
+        ...emissionState,
+        turnId: activeTurnId(emissionState),
+      });
+    }
```
ruiconti, line 949: I don't follow why we need to pass `approvalOriginCallIds` to `memory.resolveApprovalTools`? Would not expect "memory tools" to have privileged or special handling here @!e0

ruiconti: @!e0 but how is this different from defineDynamic returning tools? This should be using the same tool-loop access.

AndrewBarba:
Closing as superseded. The canonical source graph and binding table in #2516 replace this branch's custom memory compiler and provider-tool lifecycle. The memory core will restart from current main and compile each slot's provider tools through an ordinary programmatic defineDynamic wrapper; this branch will not be rebased or reused wholesale. #1581 remains the contract source of truth.

PR #2154, opened 2026-08-14
fix(eve): surface model catalog request errors

### LINKED ISSUE #2153: Catalog request failures are reported as missing model metadata
### Reproduction

1. Configure an agent with a model that is not in eve's built-in model metadata.
2. Make the AI Gateway model catalog request fail, for example with a network error.
3. Compile the agent.

### Current behavior

eve reports that the model does not have known AI Gateway context-window metadata. The catalog loader converts the request failure into a missing-model result, so the compiler cannot distinguish an unavailable catalog from a successful catalog response that omits the model.

### Expected behavior

Surface the catalog request failure, including its original error. Keep the existing missing-metadata diagnostic only when a catalog was loaded successfully and the model is absent. Models with eve-owned built-in metadata should continue to compile when the catalog is unavailable.

### Version and environment

Confirmed on current `main` at `c358fa939`. The failure is in compiler model-catalog resolution and is independent of operating system.

ruiconti:
### Summary

Closes #2153. A failed `/v1/models/catalog` request was collapsed into `null`, so compilation blamed the configured model for missing context-window metadata.

- Before: catalog request failure → missing-model diagnostic.
- After: catalog request failure → catalog-load diagnostic with the upstream error preserved as `cause`.

Fresh caches and eve-owned built-in limits still work. A successfully loaded catalog that lacks the model still produces the existing missing-metadata diagnostic; an expired cache no longer hides a failed refresh.

### Validation

The focused compiler suite passes all 27 tests, including catalog request failures through both lookup methods, offline built-in fallback, and preservation of the compiler error cause.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the required patch changeset for the compiler error behavior.

**Implementation** — 2 files · `+21 / -22`

Keeps the distinction between catalog transport failure and missing model at the catalog/compiler boundary while preserving built-in limits.

**Tests** — 2 files · `+48 / -6`

Replaces the masking expectation and covers both catalog lookup methods, offline built-ins, and the compiler-level error cause.

### DIFF
--- a/.changeset/surface-model-catalog-errors.md
+++ b/.changeset/surface-model-catalog-errors.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Surface AI Gateway model catalog request failures during compilation instead of reporting affected models as missing context-window metadata.
--- a/packages/eve/src/compiler/model-catalog.test.ts
+++ b/packages/eve/src/compiler/model-catalog.test.ts
@@ -245,16 +245,31 @@ describe("createCompiledRuntimeModelCatalogLoader", () => {
 
   it("falls back to built-in limits when fetch fails", async () => {
     vi.spyOn(globalThis, "fetch").mockRejectedValue(new Error("offline"));
-    const loader = createCompiledRuntimeModelCatalogLoader("/tmp/test-app");
+    const loader = createCompiledRuntimeModelCatalogLoader(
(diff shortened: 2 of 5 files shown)


ctgowrie (approved):

ruiconti:
Closed in favor of #2155.

PR #2167, opened 2026-08-14
feat(eve): workflow-tool subagents with per-call background launch and task_join

ruiconti:
### Summary

Related to #1084. Two layers on the frozen base (`tasks-v4/03-runtime` at `37314a9f`):

1. **Subagents as workflow tools** — with `experimental.tasks`, local and remote subagents are framework-owned `defineTool(...)` definitions whose `execute` values are separately compiled Workflow functions. Unchanged from the prior revision of this PR.
2. **Per-call detach/join (new)** — the sync/background choice moves from the agent-global flag to each call; a new `task_join` lets a turn block on a background task; a plain workflow-backed tool can be task-backed. Design and evidence live in `SPIKE.md` at the repo root.

### Layer 2 — per-call launch mode

Before → after: `experimental.tasks` made every delegation settle with a task receipt, decided per batch → the model chooses per call. `background: true` on a subagent call takes the receipt path; omitted means the call dispatches on the plain foreground wire and the turn parks on its result. Mixed batches work in one turn: receipts seed the wait loop's initial results while foreground keys resolve from the inbox.

Two boundaries worth review:

- Continuations (`agentId` resumes) always take the task path regardless of the flag — `AGENT_BUSY` session exclusivity is enforced there, and a foreground resume of a task-owned session would bypass it.
- Interrupt-sourced batches (dynamic workflow) carry no model arguments, so they now default to foreground under tasks agents. This is a behavior change from the prior revision, where they all became tasks.

The flag lives on a new `tasks` schema variant, so persistent-session-only agents don't see dead surface.

### Layer 2 — task_join

Before → after: a turn could not block on a task; the model paced itself with `task_sleep` + re-peek → `task_join(taskId)` settles immediately when the task is ready, otherwise the call's key stays pending and the turn's wait races its inbox against a 2s durable sleep, polling the task view per tick.

"Ready" is terminal **or** `input_required`: input wakes are message-less deliveries the driver buffers mid-turn, so a terminal-only join deadlocks against a task waiting on a human.

Every answer to a pending call — child-reported or framework-written — now enters the wait through one function, `acceptResult` in `turn-workflow.ts`. That is the seam a future mid-flight detach reuses: same entry point, fed a synthesized `{taskId, status: "working"}` receipt for a still-pending foreground call. `PendingTaskJoin` carries its result key precomputed because framework-tool modules cannot enter the workflow body bundle (their import graph reaches Node builtins).

### Layer 2 — tool tasks

A plain `defineTool` whose execute is a compiled Workflow function dispatches as a `kind: "tool"` task: a framework executor run starts the tool's Workflow, awaits it, and reports the terminal outcome to the task run, which stays the sole lifecycle writer. Not wired into model-facing dispatch — the integration test drives `dispatchToolTask` directly; this layer proves the seam, nothing more. `TaskMetadata.kind` widened to `"subagent" | "tool"`; tool tasks never project into `<agents>`.

### Layer 1 — subagents as workflow tools (prior revision, unchanged)

Task-mode delegation ran inside one parent dispatch step → task dispatch selects the local or remote framework tool definition and starts its compiler-assigned Workflow before re-entering the existing durable task lifecycle. Definitions under `runtime/framework-tools/subagent`; executors under `execution`, keeping definition-time schema and branding code out of the compiled Workflow program (payload delta at the time: +13,891 bytes per function vs. ~2.09 MB for the combined shape). The durable task run remains the sole lifecycle writer; local HITL uses the child continuation hook, remote HITL the task callback capability.

### Known residuals

- The wrapper Workflow `start()` inside a retryable step is at-least-once; identities stay deterministic but no wrapper-run idempotency key is persisted. The tool-task executor shares this.
- A joined task's own `:ready:` notification still replays as a user-visible delivery next turn; dedup needs the driver's `seenTaskDeliveries` machinery.
- A join on a task whose run died pends until turn cancellation; timeout policy is future work.
- e2e eval anchors are out of scope here — `fixture-tasks` lives on `rui/child-task-send`.

### Validation

Unit (631 files, 6,616 passed, 1 skipped) and integration (90 files, 646 passed) suites green; typecheck, lint, format, and invariant guards pass. The three integration tests named in `SPIKE.md` are the demo: mixed-batch dispatch, join over a real child/task run (pending → poll-settled → immediate), and the tool task reaching `completed` and joining.

### Checklist


ruiconti:
Superseded by #2255, which rebases this work onto `main` instead of the frozen `tasks-v4/03-runtime` base. Per-call background launch, `task_join`, and tool-backed tasks land separately, stacked on #2255.

PR #2203, opened 2026-08-17
Plan bounded new-project writes

OwenKephart:
### Summary

When an in-place `eve init` scaffold was allowed to replace existing files, it wrote generated files one at a time. A later destination conflict could therefore leave earlier generated files changed. The scaffolder now renders and validates every generated-file destination before it starts writing, so a known conflicting or non-directory destination fails without altering earlier template files.

### Before / after

Before, a blocked later destination could leave an earlier in-place scaffold file overwritten:

```console
# user chooses to scaffold into the current directory
# agent/agent.ts is written
# agent/channels is a file, so agent/channels/eve.ts cannot be written
# agent/agent.ts remains changed
```

After, eve validates every generated-file destination before it writes any template files:

```console
# detects that agent/channels is not a directory
# writes no template files
# existing agent/agent.ts is unchanged
```

### Validation

Focused scaffold integration coverage passed for a blocked later destination and confirms earlier in-place files are untouched. No broader validation was run for this PR separately.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

The changeset records the preflight behavior and its bounded scope.

**Implementation** — 1 file · `+42 / -12`

Keeps destination preflight alongside the scaffold that owns the generated-file set, reusing the existing write behavior.

**Tests** — 1 file · `+24 / -0`

A focused filesystem integration test proves a later blocked destination does not modify an earlier in-place template file.

### DIFF
--- a/.changeset/bounded-init-writes.md
+++ b/.changeset/bounded-init-writes.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Plan and revalidate new-project scaffold files before writing, reject symlink traversal and changed destinations, and roll back matching writes after in-process failures.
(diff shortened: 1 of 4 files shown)


OwenKephart:
Closing this layer: the bounded-write transaction is disproportionate to the rare in-place failure it covered. The remaining stack reports completed existing-project changes instead.

PR #2206, opened 2026-08-17
test: expand authoring benchmark cases

cmpadden:
### Summary

Adds four self-contained authoring cases for core eve workflows that were not represented by the scaffold and weather-tool benchmarks: an allow-listed OpenAPI connection, a packaged skill with a supporting reference, a conditionally approved tool, and a queued custom HTTP channel. The suite now publishes all six cases across every supported OpenCode model and both treatments. The OpenAPI contract is seeded locally, and graders accept equivalent valid channel implementations while enforcing the behavioral contract.

### Validation

The benchmark suite and typecheck pass. Canonical publication against `origin/main` completed all 72 cells (six models, two treatments, six cases, three runs); the docs aggregate and focused result tests also pass.

### Checklist

- [ ] I added a changeset if this touches the published `eve` package — not applicable; this changes private benchmark cases

### Diff size

**Docs** — 1 file · `+4 / -3`

Updates the private benchmark runbook for the expanded canonical suite.

**Implementation** — 2 files · `+53 / -1`

Promotes the cases into canonical publication and seeds a deterministic local OpenAPI contract.

**Tests** — 11 files · `+202 / -2`

Defines the four prompts and deterministic graders, plus focused configuration and setup coverage.

### DIFF
--- a/apps/benchmarks/evals/author-003-openapi-connection/CASE.ts
+++ b/apps/benchmarks/evals/author-003-openapi-connection/CASE.ts
@@ -6,7 +6,7 @@ export default defineAuthoringCase({
   setup: inventoryOpenApiSetup,
   async interact({ send }) {
     await send(
-      "Connect this agent to the inventory service described by the provided `agent/lib/inventory-openapi.ts` contract. Add an OpenAPI connection named `inventory` with base URL `https://inventory.example.com`. It should authenticate as the app with a bearer token from `INVENTORY_API_TOKEN`, and the model should only be able to discover the read-only `getStock` operation, not `reserveStock`.",
+      "Connect this agent to the inventory service described by the provided `agent/lib/inventory-openapi.ts` contract. Add an OpenAPI connection named `inventory` with base URL `https://api.northstar-fulfillment.com`. It should authenticate as the app with a bearer token from `INVENTORY_API_TOKEN`, and the model should only be able to discover the read-only `getStock` operation, not `reserveStock`.",
     );
   },
 });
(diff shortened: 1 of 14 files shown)


apps/benchmarks/evals/author-003-openapi-connection/CASE.ts
```diff
@@ -0,0 +1,12 @@
+import { defineAuthoringCase, simpleProject } from "../../lib/authoring-case.js";
+import { inventoryOpenApiSetup } from "../../lib/setups/inventory-openapi.js";
+
+export default defineAuthoringCase({
+  startingPoint: simpleProject,
+  setup: inventoryOpenApiSetup,
+  async interact({ send }) {
+    await send(
+      "Connect this agent to the inventory service described by the provided `agent/lib/inventory-openapi.ts` contract. Add an OpenAPI connection named `inventory` with base URL `https://inventory.example.com`. It should authenticate as the app with a bearer token from `INVENTORY_API_TOKEN`, and the model should only be able to discover the read-only `getStock` operation, not `reserveStock`.",
```
OwenKephart: i like to avoid "inventory.example.com" type stuff because it makes the model realize it's in a benchmark which can influence its behavior a fair amount

OwenKephart (approved):

apps/benchmarks/evals/author-003-openapi-connection/CASE.ts
cmpadden: Updated the fixture and its deterministic assertion to use `api.northstar-fulfillment.com` rather than an `*.example.com` benchmark-shaped hostname.

PR #2239, opened 2026-08-18
feat(eve): summarize initialized projects after running `eve init`

cmpadden:
### Summary

A successful `eve init` did not provide one authoritative summary of what it generated, so coding agents inspected `agent.ts`, `package.json`, channel files, and project trees to recover the path, model, instructions location, and installation state. Successful init runs now print those facts in a compact `Project ready` block and distinguish the eve default model from an explicit `--model` selection reducing sequential tool calls for agents.

### Validation

Focused helper and structural unit coverage passed 9/9, including the 700-line source invariant; typecheck also passed. Earlier focused init integration coverage passed 3/3 for default and explicit model summaries, and a one-run guided Sol benchmark passed in 178.7s with 14 tool calls.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds release metadata for the new init output.

**Implementation** — 2 files · `+17 / -0`

Keeps agent-only summary formatting with the existing handoff concern while leaving `init.ts` focused on control flow and below the source-size cap.

**Tests** — 2 files · `+20 / -3`

Covers helper formatting plus the human, default-model agent, and explicit-model agent paths.

### DIFF
--- a/packages/eve/src/cli/commands/init.ts
+++ b/packages/eve/src/cli/commands/init.ts
@@ -605,18 +605,6 @@ export async function runInitCommand(
     `${pc.green("✓")} Installed dependencies ${pc.dim(`in ${formatElapsed(result.installElapsedMs)}`)}`,
   );
 
-  const selectedModel = options.model ?? DEFAULT_AGENT_MODEL_ID;
-  logger.log(
-    [
-      "",
-      "Project ready:",
-      `  Path: ${result.projectPath}`,
-      `  Model: ${selectedModel}${options.model === undefined ? " (eve default)" : ""}`,
-      `  Instructions: ${join(result.projectPath, "agent/instructions.md")}`,
-      `  Dependencies: installed with ${result.packageManager}`,
-    ].join("\n"),
-  );
-
   if (result.kind === "created" && result.gitResult.kind === "failed") {
     logger.error(
       pc.yellow(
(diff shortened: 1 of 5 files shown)


chadhietala (approved):

packages/eve/src/cli/commands/init.ts
```diff
@@ -556,6 +556,18 @@ export async function runInitCommand(
     `${pc.green("✓")} Installed dependencies ${pc.dim(`in ${formatElapsed(result.installElapsedMs)}`)}`,
   );
 
+  const selectedModel = options.model ?? DEFAULT_AGENT_MODEL_ID;
```
OwenKephart: the above code already has some styling here for this sort of content (including dependency installs), we should make sure we align w/ that and don't duplicate any content

packages/eve/src/cli/commands/init.ts
```diff
@@ -556,6 +556,18 @@ export async function runInitCommand(
...
+  logger.log(
+    [
+      "",
+      "Project ready:",
+      `  Path: ${result.projectPath}`,
+      `  Model: ${selectedModel}${options.model === undefined ? " (eve default)" : ""}`,
+      `  Instructions: ${join(result.projectPath, "agent/instructions.md")}`,
+      `  Dependencies: installed with ${result.packageManager}`,
+    ].join("\n"),
+  );
```
ruiconti: @!e0 is this being emitted on _every_ `eve init` or just on `eve init`s that are initiated by an agent? IMO it should be exclusively for the latter.

packages/eve/src/cli/commands/init.ts
cmpadden: Updated. The existing styled path and dependency success lines remain the source of truth; the new output now adds only styled model and instructions lines, so it no longer duplicates content.

packages/eve/src/cli/commands/init.ts
cmpadden: Updated. The model/instructions lines now live inside the existing `result.agentLaunched` branch. Human `eve init` keeps its original output, with regression coverage asserting the unchanged four-message flow.

packages/eve/src/cli/commands/init.ts
cmpadden: Updated to align with other output, and ensure no duplication.

ruiconti (approved):

PR #2240, opened 2026-08-18
fix: isolate authoring benchmark artifacts to reduce distraction for benchmark agents

cmpadden:
### Summary

Authoring benchmark grader files were accidentally created under the agent-visible workspace, so `eve init .` saw `__agent_eval__` instead of an empty project and models repeatedly moved, deleted, and reconstructed benchmark infrastructure. Benchmark state now stays under `/tmp/eve-grader`; the Vitest grader files are materialized under `.eve-grader/` only after the agent finishes, so validation cannot influence the task and does not rely on symlink support.

The clean GPT-5.6 Sol guided run reduced tool calls from **37 to 28** (**9 fewer, 24.3%**), including init attempts from roughly three to one. Its valid single-run duration was **301.4s versus the prior 288.4s mean (+13.0s)**, so this PR claims the behavioral/tool-call improvement but not a wall-clock win from that noisy sample.

### Validation

The benchmark package typechecked and its focused non-export suites passed 19/19. A one-run GPT-5.6 Terra named-project benchmark passed in 165.7s with 18 tool calls, confirming the post-run grader directory is rooted correctly for nested projects.

### Checklist

Tests and a changeset are not applicable beyond the exercised private harness path; no published package changes.

### Diff size

Not applicable.

**Implementation** — 1 file · `+15 / -9`

Keeps benchmark artifacts outside the subject workspace and invalidates cached sandbox bootstraps.

The change is validated through the benchmark harness itself rather than a separate fixture.

### DIFF
--- a/apps/benchmarks/lib/harness-agent.ts
+++ b/apps/benchmarks/lib/harness-agent.ts
@@ -225,6 +225,7 @@ export function createAuthoringAgent(subject: {
             ? workspace
             : `${workspace}/${authoringCase.projectDirectory}`;
         const context = setupContext(activeSandbox, workspace);
+        const graderContext = setupContext(activeSandbox, projectWorkspace);
         await prepareGraderDirectory(context);
         await context.write(
           `${AGENT_EVAL_DIRECTORY}/results.json`,
@@ -235,19 +236,19 @@ export function createAuthoringAgent(subject: {
           JSON.stringify(transcript),
         );
         await Promise.all([
-          context.write(
+          graderContext.write(
             `${POST_RUN_GRADER_DIRECTORY}/EVAL.test.ts`,
             readFileSync(`${fixturePath}/EVAL.ts`, "utf8"),
           ),
-          context.write(
+          graderContext.write(
             `${POST_RUN_GRADER_DIRECTORY}/grader.ts`,
             readFileSync(new URL("./grader.ts", import.meta.url), "utf8"),
           ),
-          context.write(
+          graderContext.write(
             `${POST_RUN_GRADER_DIRECTORY}/paths.ts`,
             readFileSync(new URL("./paths.ts", import.meta.url), "utf8"),
           ),
-          context.write(
+          graderContext.write(
             `${POST_RUN_GRADER_DIRECTORY}/protocol.ts`,
             readFileSync(new URL("./protocol.ts", import.meta.url), "utf8"),
           ),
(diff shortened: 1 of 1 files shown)


apps/benchmarks/lib/harness-agent.ts
```diff
@@ -176,7 +176,7 @@ export function createAuthoringAgent(subject: {
         log("[grade] running deterministic assertions");
         const test = await resultOf(
           activeSandbox,
-          `vitest run ${workspace}/${AGENT_EVAL_DIRECTORY}/EVAL.test.ts`,
+          `ln -s ${AGENT_EVAL_DIRECTORY} .eve-grader && trap 'rm -f .eve-grader' EXIT && vitest run .eve-grader/EVAL.test.ts`,
```
chadhietala: Isn't this going to be a problem on windows?

cmpadden: These benchmarks currently only run on Vercel Sandboxes running Linux, but will update to be platform agnostic in case that changes in the future.

cmpadden: Good point. The benchmark runs in a Linux sandbox today, but the symlink introduced an unnecessary platform assumption. I replaced it by materializing `.eve-grader/` only after the agent turn completes; benchmark state remains under `/tmp/eve-grader`, so the agent still starts with a clean workspace. A one-run GPT-5.6 Terra project-creation benchmark passed with the new path in 161.7s (15 tool calls).

apps/benchmarks/lib/harness-agent.ts
```diff
@@ -160,27 +161,27 @@ export function createAuthoringAgent(subject: {
...
             readFileSync(new URL("./grader.ts", import.meta.url), "utf8"),
           ),
           context.write(
-            `${AGENT_EVAL_DIRECTORY}/paths.ts`,
+            `${POST_RUN_GRADER_DIRECTORY}/paths.ts`,
             readFileSync(new URL("./paths.ts", import.meta.url), "utf8"),
           ),
           context.write(
-            `${AGENT_EVAL_DIRECTORY}/protocol.ts`,
+            `${POST_RUN_GRADER_DIRECTORY}/protocol.ts`,
```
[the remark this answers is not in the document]
cmpadden: Fixed. I now create a `graderContext` rooted at `projectWorkspace`, so post-run `.eve-grader` files and the Vitest working directory agree for both root and named-project cases. A one-run GPT-5.6 Terra named-project benchmark passed end to end in 165.7s (18 tool calls).

OwenKephart (approved):

PR #2241, opened 2026-08-18
test: allow `eve init <agent-name>` outside of workspace in project creation benchmark

cmpadden:
### Summary

The new-project authoring benchmark forced agents to initialize the harness workspace in place, even though the normal project-creation flow is `eve init <name>`. The case now requests `wayfinder` as a named project, and authoring cases can declare the relative project directory where grading and validation should run.

In the sequential GPT-5.6 Sol guided benchmark, project creation improved from **288.4s and 37 tool calls** to **183.7s and 26 tool calls**: **104.7s faster (36.3%)** with **11 fewer tool calls (29.7%)**.

### Validation

The benchmark package typechecked, and its focused non-export test suites passed 20/20. A one-run guided GPT-5.6 Sol benchmark passed from the nested `wayfinder` project in 183.7s.

### Checklist

A changeset is not applicable because this changes only the private benchmark harness and case.

### Diff size

Not applicable.

**Implementation** — 2 files · `+9 / -3`

Adds a case-owned project-directory boundary and runs grading and package scripts from it.

**Tests** — 3 files · `+8 / -2`

Updates the project-creation eval and verifies that its declared output directory loads correctly.

### DIFF
--- a/apps/benchmarks/evals/author-002-new-project/CASE.ts
+++ b/apps/benchmarks/evals/author-002-new-project/CASE.ts
@@ -2,9 +2,10 @@ import { defineAuthoringCase, emptyProject } from "../../lib/authoring-case.js";
 
 export default defineAuthoringCase({
   startingPoint: emptyProject,
+  projectDirectory: "wayfinder",
   async interact({ send }) {
     await send(
-      "Create a new eve project directly in the current working directory for a concise travel assistant called Wayfinder. Do not create a subdirectory: initialize the current directory (for example, with `eve init .`). It should use the default model and be ready to build.",
+      "Create a new eve project named `wayfinder` for a concise travel assistant called Wayfinder. It should use the default model and be ready to build.",
     );
   },
 });
--- a/apps/benchmarks/evals/author-002-new-project/EVAL.ts
+++ b/apps/benchmarks/evals/author-002-new-project/EVAL.ts
@@ -4,7 +4,7 @@ import { expect, test } from "vitest";
 
 import { subjectDefaultAgentModel } from "./grader.js";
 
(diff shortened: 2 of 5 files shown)


OwenKephart (approved):

apps/benchmarks/evals/author-002-new-project/CASE.ts
```diff
@@ -2,9 +2,10 @@ import { defineAuthoringCase, emptyProject } from "../../lib/authoring-case.js";
 
 export default defineAuthoringCase({
   startingPoint: emptyProject,
+  projectDirectory: "wayfinder",
   async interact({ send }) {
     await send(
-      "Create a new eve project directly in the current working directory for a concise travel assistant called Wayfinder. Do not create a subdirectory: initialize the current directory (for example, with `eve init .`). It should use the default model and be ready to build.",
+      "Create a new eve project named `wayfinder` for a concise travel assistant called Wayfinder. It should use the default model and be ready to build.",
```
ruiconti, line 8: Do we have an eval case for the scenario we're deleting?

> Create a new eve project directly in the current working directory

cmpadden: We still have an eval for this in `init.integration.test.ts` but we don't currently have a benchmark.

The reason this change was made was because with the current workspace based benchmark structure, the agent spent several operations to update package.json and package-lock.json which are unnecessary when providing the agent name specifically.

ruiconti: Should we add a benchmark for it?

ruiconti (approved):

PR #2255, opened 2026-08-18
feat(eve): own subagent workflow tools in the framework

ruiconti:
### Summary

Related to #1084. With `experimental.tasks`, local and remote subagent calls become framework-owned `defineTool(...)` definitions whose `execute` values are separately compiled Workflow functions.

Before → after: task-mode delegation ran inside one parent dispatch step → task dispatch now selects the local or remote framework tool definition and starts its compiler-assigned Workflow before re-entering the existing durable task lifecycle.

The definitions live under `runtime/framework-tools/subagent`; the Workflow executors stay under `execution`, the compiler-scanned domain. That split matters: combining `defineTool` and a workflow directive in one source module embeds Zod and framework registration code in *every* Workflow function payload (~2.09 MB per function when measured on the original spike).

The model-visible subagent contract is unchanged. Local and remote executors stay separate through transport selection, then share task admission. The durable task run remains the sole lifecycle writer — local HITL over the child continuation hook, remote HITL over the task callback capability.

### Rebased onto main, and what that changed

This replaces #2167, which targeted a frozen base off `tasks-v4/03-runtime` (`37314a9f`, not an ancestor of `main`). Replaying onto `main` picked up 76 commits of drift and settled four things worth a reviewer's eye:

- **`task-event-callback.ts` keeps main's version.** The spike's `forwardTaskCallbackEvent` emitted locally *and* forwarded; main's `forwardTaskEventToSessionCallback` suppresses local emission so two TUIs can't both present the same HITL request. Main's behavior wins; the spike's `fireTaskEventCallbackStep` test is kept as added coverage main was missing.
- **Shared sandbox provisioning moved.** Main provisions from the plan inside `prepareRuntimeActionDispatch`, but this change splits planning out of preflight. Provisioning is now a helper called once per batch — deliberately *not* on per-entry rehydration, which would create a second sandbox (caught by `opens a shared parent sandbox for a background-task child`).
- **`sandboxSessionId`** is threaded into the workflow tool's `startSubagent` call, and `executeTaskControlAction` keeps main's `adapter`/`serializedContext` parameters.
- **`tasksEnabled` stays `boolean`** on the park result rather than narrowing to `true`.

The 12 spike commits are consolidated into one, since several superseded each other (prototype → inline → framework-owned).

### Known residual

The wrapper Workflow `start()` is at-least-once if the outer dispatch step retries after it succeeds. Task and child identities stay deterministic, so admission is replay-safe, but no wrapper-run idempotency key is persisted.

### Validation

Unit (652 files, 6841 passed, 1 skipped), integration (91 files, 661 passed), typecheck, build, lint, format, and `guard:invariants` all pass. `SPIKE.md` records the runtime path and boundaries.

### Checklist

### DIFF
--- a/.changeset/workflow-backed-task-subagents.md
+++ b/.changeset/workflow-backed-task-subagents.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Run each subagent delegation behind `experimental.tasks` as its own Workflow-backed tool execution while preserving the existing durable task receipts and local or remote agent behavior.
--- a/SPIKE.md
+++ b/SPIKE.md
@@ -1,0 +1,105 @@
+# Subagents as Workflow tools spike
+
+**Conclusion: this prototype is on eve's production delegation path: when the
+root agent enables `experimental.tasks`, each local or remote subagent call is
+started through its own Workflow-backed `defineTool` definition before both
+transports converge on the durable task lifecycle.**
+
+## Runtime path
+
+1. `turnStep` reads `resolvedAgent.config.experimental.tasks`. The turn
+   workflow selects `dispatchTaskStep` only when that value is `true`; plain
+   mode continues to select `dispatchRuntimeActionsStep`.
+2. The local and remote framework-tool modules each own a branded `defineTool`
+   value. Graph resolution does not load or register either definition. The
+   existing model-visible subagent descriptors still emit `runtimeAction`,
+   which is the parent session's park/resume boundary.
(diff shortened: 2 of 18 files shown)


ruiconti:
Closing as a duplicate of #2260, which owns the Layer 1 framework subagent-tool work in its separate workstream. The per-call background and task_join layer remains in #2258 and will be stacked on #2260; this PR should not independently own or duplicate Layer 1.

PR #2283, opened 2026-08-19
feat(cli): add telemetry

OwenKephart:
### Summary

Adds opt-out CLI telemetry for canonical command usage and structured outcomes, with a dedicated policy page that documents the current allowlist, exclusions, controls, storage, and versioned notice. `eve dev` records only local/remote target and TUI/headless UI dimensions; telemetry does not send command arguments, prompts, URLs, paths, environment values, file contents, error messages, or arbitrary error properties.

The telemetry identity model follows Next.js rather than Vercel CLI’s account-associated model: a random installation ID, a per-invocation session ID, and a salted one-way project ID. The raw Git remote, `REPOSITORY_URL`, working directory, and salt remain local. CI and Docker use in-memory identity state, and identity or project-resolution failures skip telemetry without changing the user command.

### Validation

Focused telemetry coverage passed with 24 tests, including command controls, salted project derivation, persisted and ephemeral identity behavior, disabled telemetry, failure isolation, and notice refresh. `pnpm --filter eve typecheck` and `pnpm docs:check` passed.

### Checklist

### Diff size

**Docs** — 4 files · `+72 / -5`

Adds the canonical telemetry policy page, links the CLI reference to it, and includes release notes.

**Implementation** — 6 files · `+715 / -327`

Implements telemetry controls, detached delivery, identity correlation, and preference storage. Extracting the existing dev command accounts for most of the mechanical implementation movement.

**Tests** — 5 files · `+314 / -0`

Covers command controls, event allowlisting, persistence, identity derivation, ephemeral environments, and failure isolation.

### DIFF
--- a/packages/eve/src/cli/telemetry/identity.ts
+++ b/packages/eve/src/cli/telemetry/identity.ts
@@ -1,0 +1,52 @@
+import { execFile } from "node:child_process";
+import { createHash, randomUUID } from "node:crypto";
+import { existsSync } from "node:fs";
+import { promisify } from "node:util";
+
+const runFile = promisify(execFile);
+const GIT_TIMEOUT_MS = 1_000;
+
+export type EveCliTelemetryIdentity = {
+  readonly installationId: string;
+  readonly projectSalt: string;
+};
+
+export function createEveTelemetryIdentity(): EveCliTelemetryIdentity {
+  return { installationId: randomUUID(), projectSalt: randomUUID() };
+}
+
+export function isEphemeralEveTelemetryEnvironment(): boolean {
+  return Boolean(process.env.CI) || existsSync("/.dockerenv");
+}
+
+export function hashEveTelemetryProject(identity: EveCliTelemetryIdentity, value: string): string {
+  return createHash("sha256").update(identity.projectSalt).update(value).digest("hex");
+}
+
+export async function resolveEveTelemetryProjectId(input: {
+  readonly cwd?: string;
+  readonly repositoryUrl?: string;
+  readonly getGitRemote?: (cwd: string) => Promise<string | undefined>;
+  readonly identity: EveCliTelemetryIdentity;
+}): Promise<string> {
+  const cwd = input.cwd ?? process.cwd();
+  const gitRemote = await (input.getGitRemote ?? getGitRemote)(cwd);
+  return hashEveTelemetryProject(
+    input.identity,
+    gitRemote ?? input.repositoryUrl ?? process.env.REPOSITORY_URL ?? cwd,
+  );
+}
+
+async function getGitRemote(cwd: string): Promise<string | undefined> {
+  try {
+    const { stdout } = await runFile("git", ["config", "--local", "--get", "remote.origin.url"], {
+      cwd,
+      timeout: GIT_TIMEOUT_MS,
+      windowsHide: true,
+    });
+    const value = stdout.trim();
+    return value === "" ? undefined : value;
+  } catch {
+    return undefined;
+  }
+}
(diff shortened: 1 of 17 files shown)


packages/eve/src/cli/dev/command.ts
```diff
@@ -0,0 +1,345 @@
+import { Command, InvalidArgumentError } from "#compiled/commander/index.js";
```
OwenKephart: note: this is a pure extraction from `run.ts`, which was getting too large on its own

packages/eve/src/cli/telemetry/preference.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+  if (process.platform === "darwin") {
+    return join(home, "Library", "Preferences", "eve", "config.json");
+  }
+  return join(
+    configuredDirectory(process.env.XDG_CONFIG_HOME, join(home, ".config")),
+    "eve",
+    "config.json",
+  );
+}
```
ruiconti, line 40: This steers away from the existing path that has the configuration being scoped to the immediate project.

Assuming this is intentional as we want to capture user intent across projects. If that's the case, I wouldn't scope this to `telemetry/` as it is config that's meant to be shared with other domains.

packages/eve/src/cli/telemetry/preference.ts
```diff
@@ -0,0 +1,128 @@
...
+
+const EveConfigSchema = z.looseObject({
+  telemetry: z
+    .looseObject({
+      enabled: z.boolean().optional(),
+      installationId: z.string().optional(),
+      noticeVersion: z.number().int().positive().optional(),
+      notifiedAt: z.string().optional(),
+      projectSalt: z.string().optional(),
```
ruiconti: If this is a looseObject (ie already accepts defaults), I'd expect for `enabled` to be a required prop.

packages/eve/src/cli/telemetry/identity.ts
```diff
@@ -0,0 +1,52 @@
...
+const GIT_TIMEOUT_MS = 1_000;
+
+export type EveCliTelemetryIdentity = {
+  readonly installationId: string;
+  readonly projectSalt: string;
+};
+
+export function createEveTelemetryIdentity(): EveCliTelemetryIdentity {
+  return { installationId: randomUUID(), projectSalt: randomUUID() };
+}
```
ruiconti, line 16: IIUC:
- installationId and projectSalt are both generated at random and persisted at a user-level config
- is read whenever present

How do we disambiguate between different eve projects? What is the point of having these ids?

@!e0

packages/eve/src/cli/telemetry/preference.ts
OwenKephart: Yep on the same page here -- this is definitely intentional, and path is now `<config dir>/eve/config.json`, with telemetry config just one key in there.

packages/eve/src/cli/telemetry/preference.ts
OwenKephart: mm good call updated to `z.boolean().default(true)`

packages/eve/src/cli/telemetry/identity.ts
OwenKephart: e0 was mostly correct here from my quick skim, but for the human response:

individual project ids are generated in-memory from (in order of priority):
 1. the local Git remote.origin.url
 2. REPOSITORY_URL
 3. the current working directory

that is hashed together with the projectSalt (same projectSalt regardless of project). this creates a random id that is impossible to reverse, but will be stable for a given project. this hashed identifier is sent along with each event.

the installationId is sent separately, and allows us to correlate events from different projects for the same installation.

AndrewBarba (approved):

PR #2293, opened 2026-08-19
[1/5] feat(eve): define activity protocol and reducer

benpankow:
### Summary

Related to #1673.

Adds an internal activity event protocol and reducer. Work, action, and blocker lifecycle reduces into one normalized snapshot, including replay and settlement-before-start handling.

### Validation

Adds focused reducer and parser coverage for replay, out-of-order settlement, terminal cleanup, bounded state, malformed known events, and mixed-version batches.

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the patch release note.

**Implementation** — 5 files · `+738 / -0`

Defines the versioned activity contract, deterministic work identities, text normalization, and reducer.

**Tests** — 2 files · `+356 / -0`

Exercises protocol parsing and reducer convergence.


packages/eve/src/execution/session-progress-command.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/** Invokes a best-effort progress observer without affecting session control. */
+export async function handleSessionProgress(
+  handler: SessionProgressHandler | undefined,
+  command: SessionProgressCommandV1,
+): Promise<void> {
+  try {
+    await handler?.handleProgress(command);
+  } catch {
+    // Progress observation cannot fail model execution or session control.
```
AndrewBarba: should at least log this

packages/eve/src/channel/types.ts
```diff
@@ -204,6 +204,7 @@ export type SessionCommand =
       readonly turnPolicy?: TurnPolicy;
     }
   | { readonly kind: "cancel"; readonly taskId?: string; readonly turnId?: string }
+  | import("#execution/session-progress-command.js").SessionProgressCommandV1
```
AndrewBarba: nit: I hate inline type imports. type imports should just be top of file

packages/eve/src/execution/wire/session-inbox-wire.v2.ts
```diff
@@ -0,0 +1,55 @@
...
+  const wire =
+    command.kind === "progress"
+      ? { ...command, version: VERSION }
+      : { ...encodeSessionCommandV1(command), version: VERSION };
+  const parsed = sessionInboxWireV2Schema.safeParse(wire);
+  if (!parsed.success) {
+    throw new SessionInboxWireError(
+      `Produced a session inbox payload that does not match wire version ${VERSION}: ${formatValidationError(parsed.error)}`,
+    );
+  }
```
ruiconti: If progress are not supposed to trigger a turn, why are we making changes to the inbox wire schema?

@!e0

packages/eve/src/channel/types.ts
benpankow: Updated

packages/eve/src/execution/session-progress-command.ts
benpankow: thanks, updated

packages/eve/src/execution/wire/session-inbox-wire.v2.ts
ruiconti: @!e0 why don't we consolidate strictly in the HTTP boundary?

benpankow: Agreed. The refactored stack removes progress from the session inbox and wire schema entirely. Producers report over HTTP to the independent collector, so this concern is no longer present in the current diff.

ruiconti (approved):

packages/eve/src/execution/activity-text.ts
```diff
@@ -0,0 +1,13 @@
+export const MAX_ACTIVITY_TEXT_LENGTH = 500;
```
ruiconti, line 1: This could be under `execution/activity` for better fs scoping

Issue #626, opened 2026-07-08
Allow declared subagents to inherit selected parent capabilities


ruiconti:
### What problem are you trying to solve?

eve currently makes authors choose between two bundled subagent shapes:

- The built-in `agent` is a copy. It shares the parent's live sandbox, tools, connections, and instructions.
- A declared subagent can have different instructions, but it inherits none of the parent's authored slots. It receives its own sandbox and its own connection set.

That prevents a common orchestrator and reviewer design. A GitHub channel checks the repository out in the root sandbox, then delegates review to specialists with different prompts. Declared reviewers have the right prompts but cannot see that checkout or use the root's GitHub connection. Built-in copies can see the repository and credentials but cannot have reviewer-specific instructions.

The isolation is intentional and documented in the [subagent slot table](https://github.com/vercel/eve/blob/2afed3b7d3d1c1a130befba7210de55f648cd6c2/docs/subagents.mdx#L54-L74). Runtime input only shares parent sandbox state for the built-in `agent` copy ([source](https://github.com/vercel/eve/blob/2afed3b7d3d1c1a130befba7210de55f648cd6c2/packages/eve/src/execution/subagent-tool.ts#L107-L125)). There is no opt-in middle ground.

### Proposed solution

Allow a declared subagent to opt into selected capabilities from its immediate parent while retaining its own instructions and other authored slots. For example:

```ts
export default defineAgent({
  description: "Review this change for security risks.",
  inherit: {
    sandbox: true,
    connections: true,
  },
});
```

The exact syntax is open. Inheritance must remain explicit and per capability.

Required behavior:

- Declared subagents remain isolated by default.
- Authors can independently share at least the parent's live sandbox and connection definitions. Sharing either one does not implicitly share instructions, tools, skills, hooks, state, channels, or schedules.
- A shared sandbox is the same live filesystem session, not a copied seed or a second checkout. Parent and child writes become mutually visible, with concurrent-write guidance matching the built-in `agent` contract.
- Shared connections reuse eve's normal auth resolution, approval rules, and current-session principal. Credentials must not be copied into prompts, durable JSON, or sandbox files.
- The child keeps its own `instructions.md` and identity. It can still narrow its own tools and skills.
- Nested inheritance resolves from the immediate parent and has documented behavior when the parent already inherited a capability.
- Invalid or unsupported combinations fail at build time. `eve info` shows each effective capability and whether it is owned or inherited.
[...]

PR #2298, opened 2026-08-19
feat(eve): expose effective model to dynamic resolvers

### LINKED ISSUE #2300: Expose the effective agent model to dynamic resolvers

Dynamic resolvers sometimes need to compose capabilities with the model their agent is about to use. For example, a dynamic subagent may inherit its parent agent's model, a tool resolver may omit tools unsupported by a smaller model, or an instruction resolver may choose a model-specific prompt dialect.

The shared resolver context currently exposes session, channel, and message data, but not model selection. Authors must hard-code a second model ID or duplicate their model policy, which can drift from session-, turn-, or step-scoped dynamic selection.

The value also needs precise timing semantics. A cached “active model” can describe the last synchronized selection rather than the model eve would actually use next, especially while a model resolver is replacing one scope.

### Proposed solution

Expose the agent's effective model on every dynamic resolver context:

```ts
ctx.model; // { readonly id: string } | null
```

“Effective” means the model eve would use if the resolver made no change, using one framework-owned precedence rule:

```text
step selection > turn selection > session selection > static model > null
```

The harness and resolver context should derive this value through the same code path rather than materializing and synchronizing a second active-model value.

Resolver ordering and scope should be explicit:

- Model resolution completes before other resolver kinds for the same lifecycle event.
- Non-model resolvers see the model selected for that event.
- A model resolver sees the effective model with its own event scope excluded, so it can make a relative choice without feeding back on its previous result.
- `ctx.model` is `null` only for a dynamic-only agent observed before its first applicable selection.
- The descriptor remains read-only, minimal, and serializable; live model instances and provider options are not exposed.

A dynamic subagent can then snapshot its parent's effective model while retaining an explicit fallback:

```ts
  events: {
    "session.started": (_event, ctx) =>
      defineAgent({
        description: "Analyze financial and accounting data.",
        model: ctx.model ? ctx.model.id : "openai/gpt-5.5-mini",
      }),
  },
});
```

The snapshot follows the resolver's lifecycle event. A later parent model change does not silently retarget an already resolved child config.

### Alternatives considered

- **Expose `ctx.agent.model`.** This reads like configured agent reflection rather than the runtime-effective selection and introduces an `agent.*` namespace for one field.
- **Expose `ctx.modelId` as a string.** This is smaller but prevents adding stable, serializable model metadata later.
- **Cache an active-model context key.** This duplicates the precedence rule and creates synchronization obligations whenever selection changes, allowing resolver context to drift from the harness.

prha:
### Summary

Closes #2300. Related to #626. Dynamic capability resolvers need to compose with the model eve would use if they made no change; they now receive that serializable value as `ctx.model`.

The API deliberately calls this the **effective** model rather than the active model. “Active” suggests a mutable, cached record of the last selection, while “effective” names a value derived at read time from the same precedence the harness uses: step, turn, session, static model, then `null`. Model resolvers see that precedence with their own event scope excluded, while other resolvers see the model selected earlier in the same event, avoiding feedback loops and drift between resolver context and the model call.

### Checklist

### Diff size

**Docs** — 3 files · `+26 / -10`

Documents `ctx.model` through the dynamic subagent inheritance example and includes release metadata for the new resolver context field.

**Implementation** — 12 files · `+114 / -49`

Centralizes effective-model precedence in one derivation shared by resolver context and the harness. This category also includes the required producer-epoch bumps and generated extension contract metadata for the four affected capability surfaces.

**Tests** — 14 files · `+207 / -23`

Covers precedence, the dynamic-only `null` case, model-resolver scope exclusion, and visibility across resolver kinds. Four retained compatibility fixtures prove extensions authored against the previous resolver context continue to compile; the mounted-extension scenario tracks the new producer epoch.

### DIFF
--- a/docs/guides/dynamic-capabilities.md
+++ b/docs/guides/dynamic-capabilities.md
@@ -68,21 +68,33 @@ availability depends on the caller, tenant, environment, or a feature flag.
 Return the child definition to configure and expose it. Return `null` to omit
 it from the parent's model-visible tools.
 
+The example below exposes a finance subagent to enterprise callers and gives
+it the model that the parent would use at that point:
+
 ```ts title="agent/subagents/finance/agent.ts"
 import { defineAgent, defineDynamic } from "eve";
 
 export default defineDynamic({
   events: {
-    "session.started": (_event, ctx) =>
-      ctx.session.auth.current?.attributes.plan === "enterprise"
-        ? defineAgent({
-            description: "Analyze financial and accounting data.",
-            model: "openai/gpt-5.5",
-          })
-        : null,
+    "session.started": (_event, ctx) => {
+      if (ctx.session.auth.current?.attributes.plan !== "enterprise") {
+        return null;
+      }
+
+      return defineAgent({
+        description: "Analyze financial and accounting data.",
+        model: ctx.model ? ctx.model.id : "openai/gpt-5.5-mini",
+      });
+    },
   },
 });
 ```
+
+`ctx.model` is the parent's effective model when the resolver runs. In this
+example, this dynamic subagent uses the parent's effective model when it is
+available falls back to `openai/gpt-5.5-mini` if the parent has not selected
+one yet. The returned child config snapshots the model ID; a later parent
+model change does not retarget the child.
 
 eve always compiles the subagent's filesystem resources, including its
 instructions, tools, skills, connections, sandbox, and nested subagents. It
--- a/.changeset/dynamic-resolver-effective-model.md
+++ b/.changeset/dynamic-resolver-effective-model.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Expose the effective model ID to model, subagent, tool, skill, and instruction resolvers through `ctx.model?.id`.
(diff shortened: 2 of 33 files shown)


docs/guides/dynamic-capabilities.md
```diff
@@ -66,22 +66,33 @@ availability depends on the caller, tenant, environment, or a feature flag.
...
+      return defineAgent({
+        description: "Analyze financial and accounting data.",
+        model: ctx.model ? ctx.model.id : "openai/gpt-5.5-mini",
+      });
+    },
   },
 });
 ```
 
+`ctx.model` is the parent's effective model when the resolver runs. This child
```
ctgowrie: maybe ctx.parent.model is better?

prha: i like that a lot better for subagents, but this is getting exposed for all dynamic resolvers.  in the case that it's a tool getting resolved, it would be the "owning" agent.  is it correct to call that a parent?

ctgowrie: Good point. Weird that it has different meaning in different places though. In dynamic agent I would assume it is this agents model

AndrewBarba (approved):

PR #2304, opened 2026-08-19
[1/2] feat(slack): render activity status

benpankow:
### Summary

Related to #1673. 

Adds opt-in `slackActivityStatus()` "activity" renderer. This replaces the event-stream-driven Slack status with one driven by the Activity stream, which flows through activity from all subagents. 

Status prefers an active approval, sign-in, or input wait, then the newest action or delegated work, and finally `Working…` for root work. Duplicate snapshots avoid Slack writes, and collector expiry clears the transient status.

In practice this ends up a little bit less expressive than the current event-stream-driven status since it lacks agent reasoning, but we can expand what flows through to the activity renderers to get closer to parity.

```ts
import { experimental_slackActivityStatus, slackChannel } from "eve/channels/slack";

export default slackChannel({
  activity: { renderers: [experimental_slackActivityStatus()] },
});
```

### Validation

Adds mocked Slack coverage for configuration, status priority, duplicate suppression, clearing, disposal, retry, and adapter rehydration.

### Diff size

**Docs** — 2 files · `+20 / -1`

Documents the activity status API and adds its release note.

**Implementation** — 3 files · `+227 / -2`

Adds Slack activity configuration and the transient status renderer.

**Tests** — 1 file · `+214 / -0`

Exercises status selection and provider behavior.

### DIFF
--- a/packages/eve/src/public/channels/slack/activity.ts
+++ b/packages/eve/src/public/channels/slack/activity.ts
@@ -1,55 +1,55 @@
-import type { ChannelProgressRenderer } from "#channel/progress-renderer.js";
+import type { ChannelActivityRenderer } from "#channel/activity-renderer.js";
 import type {
-  ProgressActionV1,
-  ProgressBlockerV1,
-  ProgressSnapshotV1,
-  ProgressWorkV1,
-} from "#protocol/progress.js";
+  ActivityActionStateV1,
+  ActivityBlockerStateV1,
+  ActivitySnapshotV1,
+  ActivityWorkStateV1,
+} from "#protocol/activity.js";
 import { callSlackApi, type SlackBotToken } from "#public/channels/slack/api.js";
 import { truncateTypingStatus } from "#public/channels/slack/limits.js";
 
-const SLACK_STATUS_PROGRESS_RENDERER_ID = "slack.status.v1";
-const SLACK_PROGRESS_RENDERER = Symbol("eve.slack.progress-renderer");
+const SLACK_ACTIVITY_STATUS_RENDERER_ID = "slack.status.v1";
+const SLACK_ACTIVITY_RENDERER = Symbol("eve.slack.activity-renderer");
 
-export interface SlackProgressRenderer {
-  readonly id: typeof SLACK_STATUS_PROGRESS_RENDERER_ID;
-  readonly [SLACK_PROGRESS_RENDERER]: true;
+export interface SlackActivityRenderer {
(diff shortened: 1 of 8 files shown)


packages/eve/src/channel/channel-address.ts
```diff
@@ -135,10 +135,17 @@ export function createChannelAddress<TState = undefined>(input: {
               ...input.adapter,
               state: { ...input.adapter.state, ...(state as Record<string, unknown>) },
             };
+      const progress = (adapter.progressRenderers?.length ?? 0) > 0;
```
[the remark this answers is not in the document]
benpankow: Fixed in b95e63ac2: state-overlay adapter cloning now explicitly preserves progressRenderers and progressDestination, with regression coverage using the Slack-style state override path.

benpankow:
@dco check

ruiconti (approved):

packages/eve/src/public/channels/slack/activity.ts
```diff
@@ -0,0 +1,155 @@
...
+        botToken,
+        context: {
+          teamId: typeof installationTeamId === "string" ? installationTeamId : undefined,
+        },
+        operation: "assistant.threads.setStatus",
+      });
+      if (response.ok !== true)
+        throw new Error(`Slack status disposal failed: ${response.error ?? "unknown_error"}`);
+    },
+    async render({ destination, snapshot, state }) {
```
ruiconti, line 70: @!e0 how is this different than the existing `todo` tool handling and the potential usage of `todo`? Be thorough and comprehensive.

ruiconti: @!e0 thanks. I think we should proper guidance on when and how to use each. We have several implementations that render `plan` based off `todo tool, for example.

ruiconti: @!e0 based on this PR, when would we want to render the slack `plan` block?

ruiconti: @!e0 what if an eve agent is already mapping `todo` to `plan` blocks?

ruiconti: @!e0 yeah I am mostly concerned with activity summary competing with todo? Based on activity summary's reducer architecture, isn't this something to resolve early?

benpankow: @ruiconti I think it would make sense to eventually subsume `todo` rendering into this renderer framework, this way a renderer can for example choose to display either todo or activity or both based on the presence of or absence of a plan.

PR #2294, opened 2026-08-19
[2/5] feat(eve): collect activity independently

benpankow:
### Summary

Related to #1673. Built on #2293.

Adds one independent activity collector per root session. Activity observers submit normalized batches through an opaque sink; the collector reduces them and renders a debounced snapshot without waking or blocking the agent session.

```mermaid
flowchart LR
  subgraph Sessions[Session tree]
    Root[Root session]
    Local[Local child]
    Remote[Remote child]
  end

  subgraph Activity[Activity presentation]
    Sink[Activity sink]
    Collector[Activity collector]
    Renderers[Channel renderers]
  end

  Root -. session events .-> Sink
  Local -. session events .-> Sink
  Remote -. session events .-> Sink
  Sink --> Collector
  Collector -->|reduced, debounced snapshot| Renderers
```

### Validation

Adds focused coverage for collector launch and cleanup, independent bounded retention when session timeouts are disabled, revision-based render scheduling, sink admission, malformed batches, renderer isolation and disposal, and Trusted Sources auth for Vercel self-callbacks.

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the patch release note.

**Implementation** — 16 files · `+524 / -20`

Adds the activity collector, sink route, renderer contract, and root-session lifecycle wiring.

**Tests** — 10 files · `+570 / -2`

Exercises collector ownership, ingress, callback auth, renderer behavior, and runtime launch failures.

PR #2306, opened 2026-08-19
[3/5] feat(eve): relay session events into activity

benpankow:
### Summary

Related to #1673. Built on #2294.

Projects canonical session events into activity through a fire-and-forget observer on each session deployment. The observer derives work, action, and blocker lifecycle during event emission and submits it to the root activity sink.

```mermaid
flowchart LR
  subgraph RootDeployment[Root deployment]
    RootSession[Root session event emission]
    RootObserver[Activity observer]
    Sink[Activity sink]
    Collector[Activity collector]
    Renderers[Channel renderers]

    RootSession -->|canonical event| RootObserver
    RootObserver -->|fire-and-forget activity batch| Sink
    Sink --> Collector
    Collector --> Renderers
  end

  subgraph RemoteDeployment[Remote deployment]
    RemoteSession[Remote session event emission]
    RemoteObserver[Activity observer]

    RemoteSession -->|canonical event| RemoteObserver
  end

  RemoteObserver -->|fire-and-forget activity batch| Sink
```

Activity observer configuration is private launcher state: local descendants inherit it through durable dispatch, while remote creation carries the callback-bound sink and work identity to the remote launcher.

### Validation

Adds coverage for canonical root/child lineage, fire-and-forget delivery under an unresolved sink, task-owned HITL projection despite channel suppression, replayed root lifecycle, delegated HITL remaining active across a parked turn, parent-owned settlement from canonical subagent results, background receipts, local inheritance, and remote observer configuration bound to delegated callback metadata.

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the patch release note.

**Implementation** — 26 files · `+668 / -25`

Adds best-effort activity observation at canonical event emission plus launcher-owned propagation across local and remote sessions.

**Tests** — 13 files · `+673 / -13`

Covers projection, observer lifecycle, launcher propagation, and remote ingress.


packages/eve/src/execution/session-progress.ts
```diff
@@ -99,11 +112,16 @@ export function reduceProgressCommand(
 }
 
 function reduceEvent(snapshot: ProgressSnapshotV1, event: ProgressEventV1): ProgressSnapshotV1 {
-  if (event.kind === "turn") {
-    const current = snapshot.turns[event.turn.id];
+  if (event.kind === "turn" || event.kind === "report") {
+    const incoming = event.kind === "report" ? { ...event.turn, report: event.report } : event.turn;
```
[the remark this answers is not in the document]
benpankow: Fixed in b4d576c3d: report messages are normalized at the reducer boundary, so local and remote callback submissions share the same control-character and length limits. Added a direct untrusted-report reducer regression test.

packages/eve/src/execution/submit-progress.ts
```diff
@@ -0,0 +1,19 @@
...
+}): Promise<void> {
+  const payload = progressPayloadV1Schema.parse({ events: input.events });
+  const command: SessionProgressCommandV1 = {
+    commandId: input.commandId,
+    kind: "progress",
+    payload: JSON.parse(JSON.stringify(payload)),
+    protocolVersion: 1,
+  };
+  await resumeSessionInbox(sessionCommandHookToken(input.rootSessionId), command);
+}
```
ruiconti: Can we always report progress via the same pathway ie via HTTP callback route? Not sure why we need to report it through the session inbox here granted that this shouldn't trigger a new turn.

ruiconti: That way we can get remote and local updates "for free".

packages/eve/src/execution/report-progress.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export async function reportProgress(input: {
+  readonly callId: string;
+  readonly message: unknown;
+}): Promise<{ readonly status: "queued" }> {
+  const message = typeof input.message === "string" ? normalizeProgressText(input.message) : "";
+  if (message === "") throw new Error("Provide a non-empty `message`.");
+  const ctx = loadContext();
+  const session = ctx.require(SessionKey);
+  const work = ctx.get(ProgressWorkKey) ?? rootProgressWork(session);
```
ruiconti: Why is this storing progress in `ProgressWorkKey` instead of simply emitting it like other events?

@!e0

benpankow:
Did a spin to drop `report_progress` tool - this stack now focuses on core lifecycle events (work start/settled). Should be relatively easy to iterate and add more event types down the line (e.g. #2356)

packages/eve/src/execution/submit-progress.ts
benpankow: Main benefit is just avoiding the network loopback - both end up in the session inbox regardless

ruiconti: I still don't understand why it needs to resume at session inbox if all it should do is receive events and propagate it to channel?

@!e0

ruiconti: @!e0 I hate that we have to resume the workflow just to tap into a writable stream. Thoughts:
- Doesn't workflow support a way to share access to a run's writable stream?
- If above not possible. Could we have a workflow that's used to host this?

ruiconti: Good bot. This approach seems directionally better than coupling this comms with the session's.

ruiconti: @!e0 that would also entirely decouple the session writes from the session's state i.e., the sprawl in workflow-entry and innards?

packages/eve/src/runtime/framework-tools/subagent/local.ts
```diff
@@ -240,6 +240,7 @@ async function dispatchSubagent(input: SubagentDispatchInput): Promise<SubagentD
           parentContinuationToken: input.task.taskInboxToken,
           parentTraceContext: prepared.parentTraceContext,
           persistentSessions: true,
+          progressWork: prepared.progressWork,
```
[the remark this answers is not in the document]
benpankow: Good catch. `PreparedRuntimeActionDispatch` already carried the callback, but this task/background start path only forwarded the work identity. Added `progressCallback: prepared.progressCallback`, so the child and nested descendants inherit the root collector callback.

packages/eve/src/execution/submit-progress.ts
benpankow: The replacement stack now follows this direction: all root/local/remote producers use the same HTTP callback, and an independent collector owns reduction and rendering. Progress no longer enters the session inbox or wire schema.

packages/eve/src/execution/subagent-start-remote.ts
```diff
@@ -88,9 +99,24 @@ export async function startRemoteSubagent(input: {
...
+          {
+            eventId: `${progress.workIdentity.id}:started`,
+            kind: "work.started",
+            startedAt: new Date().toISOString(),
+            work: { ...progress.workIdentity, sessionId: child.sessionId },
+          },
+        ],
+      });
+    }
+
```
ruiconti: @!e0 am I being too idealistic of not expecting any of "progress" work to be ever emitted directly from a subagent _directly_? i.e., what i'd expect

- when an agent is spawned as a subagent, it registers an observer (I see we already do this here) over the event stream

that's it, the entire interface is within the observer that consumes the event stream and propagates its events upwards. no need for anything internal.

benpankow: @!e0 how does the revised implementation match your anticipated model?

ruiconti: @!e0 how does the revised implementation match your anticipated model?

packages/eve/src/execution/remote-agent-dispatch.ts
```diff
@@ -54,6 +54,7 @@ export async function startRemoteAgentSession(input: {
   readonly auth?: SessionAuthContext | null;
   readonly callbackBaseUrl: string | undefined;
   readonly callbackToken?: string;
+  readonly eventRelay?: import("#channel/types.js").SessionEventRelayConfig;
```
ruiconti: @!e0 let's change AGENTS.md to not have dynamic import for types.

packages/eve/src/execution/workflow-runtime.ts
```diff
@@ -279,6 +289,19 @@ export function createWorkflowRuntime(config: {
         throw error;
       }
 
+      if (eventRelay !== undefined) {
+        const relayerInput: SessionEventRelayerInput = {
+          sessionId: run.runId,
+          sink: eventRelay.sink,
+          workIdentity: eventRelay.workIdentity,
```
ruiconti: [P1] Re-key activity for each continuation

This freezes the relayer to the `workIdentity` from the session's initial start. A persistent agent handle later delivers a new `callId` and parent turn into the same session without starting another relayer. Child events therefore continue to project under the first work ID, while the parent's new `subagent-result` settles an ID derived from the continuation call. The new work is never started, and its activity attaches to already-settled work. Please derive the active work identity from canonical per-turn caller data, or make both projection and settlement consistently session-scoped. A start -> result -> continue -> result projection test would cover the failure.

packages/eve/src/execution/dispatch-runtime-actions-shared.ts
```diff
@@ -282,12 +286,32 @@ async function prepareActionDispatch(input: {
...
+function resolvePreparedActivity(
+  eventRelay: SessionEventRelayConfig | undefined,
+  session: RuntimeSession,
+  turnId: string,
+): (SessionEventRelayConfig & { readonly workIdentity: ActivityWorkIdentityV1 }) | undefined {
+  if (eventRelay === undefined) return undefined;
+  return {
+    sink: eventRelay.sink,
+    workIdentity: eventRelay.workIdentity ?? {
+      id: `root:${session.sessionId}:${turnId}`,
```
ruiconti: [P1] Use the canonical root work ID

The root relayer creates `root:<sha256([sessionId, turnId])>`, but this fallback creates `root:<sessionId>:<turnId>`. `deriveChildWorkIdentity` then copies this value into the child's `parentId`. The collector never has a work node with that ID, so every root-to-child hierarchy is disconnected. Please reuse `deriveRootTurnActivityWorkId` (or the root identity helper) here and assert that the delegated work's `parentId` equals the projected root work's `id`.

packages/eve/src/execution/tasks/child/steps.ts
```diff
@@ -37,6 +43,36 @@ export async function appendTaskViewStep(input: { readonly view: TaskView }): Pr
   } finally {
     writer.releaseLock();
   }
+
+  const events = projectTaskActivitySettlement({
+    activityObserver: input.activityObserver,
+    settledAt: new Date().toISOString(),
+    view: input.view,
+  });
+  void submitActivity({ events, sink: input.activityObserver?.sink });
```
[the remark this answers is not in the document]
benpankow: fire-and-forget intentionally, activity is best effort

packages/eve/src/execution/workflow-runtime.ts
benpankow: Fixed. Continuations now have a fresh observer config.

packages/eve/src/execution/dispatch-runtime-actions-shared.ts
benpankow: │ Fixed. Dispatch now uses `deriveRootTurnActivityWorkId`

packages/eve/src/execution/wire/session-inbox-wire.v1.ts
```diff
@@ -187,8 +187,28 @@ const authSchema = z
...
+  .strict();
 const callerSchema = z
   .object({
+    activityObserver: z
+      .object({
+        sink: z.object({ url: z.string(), version: z.literal(1) }).strict(),
+        workIdentity: activityWorkIdentitySchema.optional(),
+      })
+      .strict()
+      .optional(),
```
ruiconti: We should be versioning the schema? The guard should've been picking up these changes 🤔

benpankow: Thanks, fixed.

https://github.com/vercel/eve/pull/2658 should enforce that we can't modify wire files already committed to main

packages/eve/src/execution/wire/session-inbox-wire.v2.ts
```diff
@@ -0,0 +1,90 @@
...
+  v1Deliver.extend({
+    caller: v1Caller.extend({ activityObserver: activityObserverSchema.optional() }).optional(),
+    version,
+  }),
+  v1[1].extend({ version }),
+  v1[2].extend({ version }),
+  v1[3].extend({ version }),
+  v1[4].extend({ version }),
+  v1[5].extend({ version }),
+]);
```
ruiconti, line 50: Regretting some decisions I made on migration ergonomics 😬

ruiconti (approved):

PR #2307, opened 2026-08-19
feat(selfmod): inherit the parent agent model by default

prha:
### Summary

The self-modification subagent currently defaults to Sonnet even when its parent has already selected another model, which can introduce an unexpected provider requirement. It now inherits the parent's effective model unless explicitly configured, while preserving Sonnet as the fallback when no parent model is available. Related to #742. Stacked on #2298.

### Validation

Not run for this implementation-only update.

### Checklist


AndrewBarba (approved):

AndrewBarba:
I dont totally agree with this change. I think we have to use a coding model. if the parent agent is a reasonable coding model then yes lets inherit but if its not we should pick one ourselves

PR #2310, opened 2026-08-19
[1/5] feat(slack): render activity messages

benpankow:
### Summary

Related to #1673. Built on #2304.

Renders blockers, actions, nested agents, and background tasks as activity grouped by originating turn. Nested relationships use Unicode tree rails, and completed top-level work collapses its descendant rows so finished activity stays readable.

Screenshot 2026-08-28 at 3 11 27 PM
Screenshot 2026-08-28 at 3 11 16 PM

### Validation

Reproduced the duplicate comments and premature completion on a deployed Slack agent using concurrent authored background tasks and nested background subagents. Adds regression coverage for descendant lifecycle, final top-level compaction, nested and orphaned work, recovery, and background lifecycle boundaries.

### Diff size

**Docs** — 2 files · `+10 / -0`

Adds release metadata without documenting the experimental renderers as supported APIs.

**Implementation** — 9 files · `+759 / -22`

Carries nested and background lifecycle through activity collection, adds the Unicode tree and native Slack stream presentations, and provides renderer-owned state and destination identity for durable updates.

**Tests** — 6 files · `+690 / -11`

Covers activity reduction and projection, background task boundaries, message recovery, Unicode hierarchy, native stream deltas, and final plan compaction.

### DIFF
--- a/packages/eve/src/public/channels/slack/activity-plan.ts
+++ b/packages/eve/src/public/channels/slack/activity-plan.ts
@@ -1,0 +1,247 @@
+import type { ChannelActivityRenderer } from "#channel/activity-renderer.js";
+import type { ActivitySnapshotV1, ActivityWorkStateV1 } from "#protocol/activity.js";
+import { callSlackApi, type SlackBotToken } from "#public/channels/slack/api.js";
+
+export const SLACK_ACTIVITY_PLAN_RENDERER_ID = "slack.experimental.plan.v1";
+type Phase = "running" | "completed" | "failed" | "rejected" | "cancelled" | "blocked";
+interface PlanState {
+  readonly streams: Readonly<
+    Record<
+      string,
+      {
+        readonly ts: string;
+        readonly seen: Readonly<Record<string, Phase>>;
+        readonly stopped: boolean;
+      }
+    >
+  >;
+}
+
+export function createSlackPlanRenderer(
+  botToken: SlackBotToken | undefined,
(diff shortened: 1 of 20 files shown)


packages/eve/src/public/channels/slack/activity-plan.ts
```diff
@@ -0,0 +1,247 @@
...
+      {
+        readonly ts: string;
+        readonly seen: Readonly<Record<string, Phase>>;
+        readonly stopped: boolean;
+      }
+    >
+  >;
+}
+
+export function createSlackPlanRenderer(
```
ruiconti, line 20: @!e0 what consumes this function? I don't see any callers to it in this PR.

packages/eve/src/public/channels/slack/activity.ts
```diff
@@ -43,10 +115,88 @@ export function buildSlackActivityRenderers(input: {
...
+    if (renderer.id === SLACK_ACTIVITY_STATUS_RENDERER_ID) {
+      return createSlackStatusRenderer(input.botToken);
+    }
+    assertCustomSlackActivityRenderer(renderer);
+    return createCustomSlackActivityRenderer(renderer);
   });
 }
 
+function assertCustomSlackActivityRenderer(
+  renderer: unknown,
```
ruiconti, line 139: Why is renderer unknown?

ruiconti (approved):

PR #2311, opened 2026-08-19
chore(eve): workflow - update beta dependencies

VaguelySerious:
### Summary

Moves eve's bundled Workflow SDK to `@workflow/core@5.0.0-beta.43` and pins the worlds to exactly what that core depends on (`world@beta.28`, `world-local@beta.37`, `world-vercel@beta.39`, catalog `world-postgres@beta.35`), so the lockfile keeps one copy of each and the vendored bundle cannot mix versions. Follows the same maintenance as #2023.

The reason to take this release is [vercel/workflow#3554](https://github.com/vercel/workflow/pull/3554): branch wake order, and therefore step correlation ids, could depend on how much of the event log an invocation had loaded, which corrupts runs under concurrent replays. [vercel/workflow#3497](https://github.com/vercel/workflow/pull/3497) also drops ignored-duplicate-event logging from info/error to debug, so a straggler no longer prints on every replay.

Worth a reviewer's attention: core beta.43 adds optional batched suspension transitions ([vercel/workflow#3025](https://github.com/vercel/workflow/pull/3025)). It engages only when the World implements `events.createBatch`, so eve's development World stays on the single-event path, but agents deployed to Vercel use `world-vercel`, which does implement it. That path is newly active for them, and `WORKFLOW_BATCH_TRANSITIONS=0` is the escape hatch if it misbehaves.

### Validation

Typecheck (39 tasks), build, 6897 unit tests and 659 integration tests all pass. `pnpm test:scenario` is 407 passed / 15 skipped / 2 failed; both failures are an `npm install` peer resolution for `ai` inside the scenario harness, and both reproduce identically on unmodified `origin/main` built at beta.42 in a separate worktree, so they are pre-existing and local. Confirmed the vendor stamp records the new versions, meaning `.generated/compiled` was rebuilt rather than reused, and the vendored `@workflow/core` carries the beta.43 batching code that consumers of the published package will get.

### Checklist

No tracking issue exists for routine SDK bumps; #2023 is the prior one. No new tests: the existing `development-world-spec-version.drift.test.ts` already guards the one hardcoded value that a world bump can invalidate.

### Diff size

**Docs** — 1 file · `+5 / -0`

The changeset for the published `eve` package.

**Implementation** — 3 files · `+83 / -68`

Four version pins, one catalog pin, and the lockfile they imply. The lockfile is the bulk of it and is generated; the churn is larger than four pins suggest because moving core also moves the worlds it depends on, which rewrites their integrity hashes and peer entries.

Not applicable. Existing drift coverage is described above.


AndrewBarba (approved):

VaguelySerious:
(AI) Note on the one red check.

`e2e-local (agent-tools, zai-glm)` failed the `static-tools/throw-recover` fixture at gates 3/5. It is not related to this change. The identical failure, same fixture and same gate, is on `codex/e0-redundant-task-wake-e2e` ([run 32312085482](https://github.com/vercel/eve/actions/runs/32312085482)), which does not touch the SDK:

```
✗  static-tools/throw-recover  gates 3/5
  ✗ messageIncludes(/\byes\b/iu) (0% < 100%): assistant messages did not include
    /\byes\b/iu; got: The `always-throws` tool failed as expected.
```

`agent-tools, zai-glm` failed on 6 of the 9 most recent `e2e-local` runs across four unrelated branches, rotating which fixture it kills (`to-model-output-content-parts`, `dynamic-tools/conditional`, `throw-recover`) and landing at 26 passed / 1 failed each time. The gate is a literal `/\byes\b/` match on model prose, so it turns on `zai-glm`'s phrasing.

The behavior under test passed: the tool threw, the agent recovered, and it reported the throw accurately. Only the wording missed the assertion's literal. `agent-workflow-stress`, the lane most likely to surface a replay or batching regression from this bump, passed on every model. Re-ran the failed jobs.

Issue #2331, opened 2026-08-20
Add audience-aware trace content policies


chadhietala:
## Problem

Agent trace destinations currently choose input/output capture statically. Messaging agents need to retain content for positively public conversations while redacting direct, restricted, and unclassified conversations. Channel instrumentation metadata also lacks a shared audience vocabulary that tracing can rely on.

## Proposal

- Add an optional channel instrumentation audience: `public`, `restricted`, `direct`, or `unknown`.
- Classify built-in messaging channels from durable platform state, defaulting ambiguous cases to `unknown`.
- Stamp normalized audience on agent session traces.
- Make local traces and Vercel Agent Runs retain content only for positively public sessions; all other audiences fail closed to metadata-only spans.

## Scope

Deliver as a three-PR stack: metadata contract, built-in classifications, then audience-aware trace filtering.

PR #2333, opened 2026-08-20
feat(eve): define channel audience metadata

chadhietala:
### Summary

Related to #2331. This is the first PR in a three-PR implementation stack. It adds an optional `audience` field to existing channel instrumentation metadata with `public`, `private`, and `unknown` values. Missing or invalid values normalize to `unknown`, preserving existing authored channels while giving later layers a fail-closed contract. The retained channel extension epoch verifies the type widening remains backward compatible.

The research proposal specifies the complete tracing surface: trace and export policy interfaces, public redaction/composition APIs, default behavior by audience, runtime ordering, failure semantics, compatibility behavior, and representative configuration. This API remains behind `experimental.instrumentationProviders` while we iterate on it as integrations adopt it.

### Validation

Focused channel-definition and projection coverage confirms all three audience values, invalid-value normalization, and compatibility with metadata that omits the field. After rebasing onto main's channel epoch 6, this PR retains epoch 6 and advances the audience-aware contract to epoch 7. The contract checker, installed-extension scenario, type-check, and invariant guard pass.

### Checklist

- [x] DCO sign-off passes for the commit (`git commit --signoff`)

### DIFF
--- a/docs/channels/custom.mdx
+++ b/docs/channels/custom.mdx
@@ -306,6 +306,8 @@ export default defineChannel({
 
 The projection is re-evaluated whenever adapter state changes after channel event handlers run. Dynamic tool resolvers read it via `ctx.channel.metadata` and narrow it with `isChannel`. See [Dynamic capabilities](../guides/dynamic-capabilities) for the full consumption pattern.
 
+Metadata may include an `audience` classification: `public`, `private`, or `unknown`. Omit it when the channel cannot classify the conversation confidently; eve normalizes missing and invalid values to `unknown`. Consumers must treat `unknown` as non-public.
+
 When a parent agent dispatches a subagent, the framework forwards the parent's channel metadata projection to the child. The same `metadata(state)` projector also serves instrumentation metadata resolvers.
 
 ## Continuation tokens
--- a/.changeset/tidy-cats-observe.md
+++ b/.changeset/tidy-cats-observe.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add an optional audience classification to channel instrumentation metadata.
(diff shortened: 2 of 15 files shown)


docs/channels/custom.mdx
```diff
@@ -306,6 +306,8 @@ export default defineChannel({
 
 The projection is re-evaluated whenever adapter state changes after channel event handlers run. Dynamic tool resolvers read it via `ctx.channel.metadata` and narrow it with `isChannel`. See [Dynamic capabilities](../guides/dynamic-capabilities) for the full consumption pattern.
 
+Metadata may include an `audience` classification: `public`, `restricted`, `direct`, or `unknown`. Omit it when the channel cannot classify the conversation confidently; eve normalizes missing and invalid values to `unknown`. Consumers must treat `unknown` as non-public.
```
ctgowrie: It's not obvious to me what these all mean? Why not just public or not?

chadhietala: I was trying to allow for nuance here and what you may want to trace:

- `public`: Completely open to all. Trace created with all content
- `restricted`: Open to select group of people. Trace outline but no content
- `direct`: One + N but is often done with even more privilege. Don't even take the trace.
- `unknown`: Used as the fail closed and has the similar behavior as `direct` but we categorize as an unknown type.

We can narrow the type to `public`, `private`, and `unknown`. We can always widen it if need be.

ctgowrie: Sounds good

chadhietala: Narrowed the contract to `public`, `private`, and `unknown` across the full stack. Direct messages and private/group-restricted conversations now both project `private`; `unknown` remains the fail-closed value.

chadhietala: Thank you clanker.

ctgowrie (approved):

PR #2355, opened 2026-08-20
Prevent session streams from hitting runtime limits

allenzhou101:
## Summary

- close each long-held eve session stream response after four minutes
- cancel the upstream reader and return a clean EOF so the existing cursor-based client reconnects
- document periodic stream rotation and add regression coverage

## Why

A durable run can outlive the HTTP request serving its event stream. On a five-minute serverless runtime, leaving that response open indefinitely causes the platform to terminate the function at 300 seconds. Rotating the response with headroom lets the invocation finish normally while the run continues and the client resumes from its current event index.

## Testing

- pnpm --dir packages/eve exec vitest run --config vitest.unit.config.ts src/public/channels/eve.test.ts
- pnpm --dir packages/eve run typecheck
- pnpm --dir packages/eve run build
- pnpm exec oxlint packages/eve/src/public/channels/eve.ts packages/eve/src/public/channels/eve.test.ts
- pnpm run docs:check

### DIFF
--- a/.changeset/calm-streams-reconnect.md
+++ b/.changeset/calm-streams-reconnect.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Close long-held session stream responses before common serverless runtime limits so reconnecting clients can continue from their event cursor.
--- a/docs/guides/client/streaming.mdx
+++ b/docs/guides/client/streaming.mdx
@@ -128,7 +128,7 @@ The stream can emit an interim `session.waiting` while the authorization callbac
 
 ## Reconnection
 
-HTTP connections can end before a run does. The client reconnects from the number of events already consumed, so long turns continue without replaying events. By default, a turn response keeps reconnecting until it reaches a turn boundary or is aborted, including while the turn is paused for authorization. A manually opened `session.stream()` eventually stops after repeated empty streams when it can no longer make progress.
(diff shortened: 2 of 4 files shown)


huntsyea:
Reproduction on a production Vercel deployment, in case it helps prioritise this.

**Setup:** eve 0.52.5, Node.js runtime on Vercel (Fluid Compute, default 300 s max duration), an agent with an iMessage channel plus the default eve HTTP API.

**Steps:** from a laptop, `eve invoke --url https://<production-host> "Reply with only the word hi."`. The turn completed in about two seconds and the CLI printed the `ready` / `completed` result and exited.

**Observed:** about five minutes later, one runtime log line at error level:

```
Vercel Runtime Timeout Error: Task timed out after 300 seconds
```

on `GET /eve/v1/session/<sessionId>/stream`. Nothing else in the logs; the turn itself, its trace, and its usage were all normal. It happened once per `eve invoke --url` call, and never for turns arriving through the channel webhook (those run through Workflow and never open the stream route).

**Impact:** no user-visible effect, but each `eve invoke --url` against a deployment costs one five-minute function invocation and leaves a red line in the error log that looks like a regression after a deploy until you trace the session id back to the CLI call. Rotating the stream response with headroom, as this PR proposes, is exactly what we'd want; the cursor-based client already handles a reconnect.

Happy to test a prerelease against this deployment if useful.

PR #2356, opened 2026-08-20
feat(eve): buffer action progress lifecycle

benpankow:
### Summary

Related to #1673. Built on #2310.

Adds the producer-side progress buffer. Events accumulate to the turn-step boundary, split at the protocol limit, and submit through the existing callback. Reduction and provider debounce stay in the collector; this does not add another durable queue or transport.

### Validation

Adds a focused buffer test for protocol-sized batches and flushing the remainder.

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the release note.

**Implementation** — 2 files · `+49 / -12`

Adds the single producer buffer and routes the observer through it.

**Tests** — 1 file · `+28 / -0`

Covers bounded submission and explicit flush.

### DIFF
--- a/.changeset/calm-actions-progress.md
+++ b/.changeset/calm-actions-progress.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Run progress reporting in an independent best-effort collector, with deterministic root, delegated work, action, and human-blocker lifecycle across local and remote agents.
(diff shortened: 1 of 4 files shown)


benpankow:
Closing to simplify the event-relay refactor. Producer-side batching is no longer the intended boundary and can be revisited later if the observer path needs it.

PR #2372, opened 2026-08-21
fix(eve): prepare Web Chat TypeScript config before install

ctgowrie:
### Summary

`eve add channel/web` installed Web Chat files that import through `@/`, but shadcn could leave the existing eve `tsconfig.json` without that alias or the required Next.js settings. The command now prepares and validates the project’s JSONC TypeScript configuration before shadcn installs files, preserving existing options and comments while rejecting conflicting `@/*` mappings. The registry no longer ships `tsconfig.json`, so `--overwrite` cannot replace project-owned configuration.

### Validation

Focused registry unit tests pass (56 tests), the eve package typechecks, and the complete docs registry check passes. A production Next.js build of the reported `../eve-scratch/eve-web-1` reproduction also passes after applying the new pre-install transformation. `guard:invariants` remains blocked by symlinks in the unrelated untracked ARC-AGI virtual environment.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the patch release note for the corrected Web Chat installation behavior.

**Implementation** — 5 files · `+140 / -13`

Keeps TypeScript configuration ownership in eve’s pre-install step, removes it from the shadcn payload, and extends the vendored JSONC declaration only with APIs needed to preserve comments during edits.

**Tests** — 2 files · `+53 / -0`

Covers JSONC preservation, compatible and conflicting aliases, and verifies that Web Chat preparation runs before registry installation.

### DIFF
--- a/apps/docs/scripts/validate-channel-registry.ts
+++ b/apps/docs/scripts/validate-channel-registry.ts
@@ -130,13 +130,19 @@ for (const [index, item] of items.entries()) {
   if (entry === undefined) throw new Error(`Unexpected channel registry item "${item.name}".`);
   const registrySlug = expectedSlugs[index];
 
-  if (
-    entry.slug === "eve" &&
-    item.dependencies?.some((dependency) => dependency === "ai" || dependency.startsWith("ai@"))
-  ) {
-    throw new Error(
-      `Registry item "${item.name}" must preserve the agent's existing AI SDK dependency.`,
-    );
+  if (entry.slug === "eve") {
+    if (
+      item.dependencies?.some((dependency) => dependency === "ai" || dependency.startsWith("ai@"))
+    ) {
+      throw new Error(
+        `Registry item "${item.name}" must preserve the agent's existing AI SDK dependency.`,
+      );
+    }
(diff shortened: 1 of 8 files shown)


AndrewBarba (approved):

apps/docs/scripts/validate-channel-registry.ts
```diff
@@ -130,13 +130,19 @@ for (const [index, item] of items.entries()) {
...
+    ) {
+      throw new Error(
+        `Registry item "${item.name}" must preserve the agent's existing AI SDK dependency.`,
+      );
+    }
+    if (item.files?.some((file) => file.target === "tsconfig.json")) {
+      throw new Error(
+        `Registry item "${item.name}" must let eve prepare tsconfig.json before shadcn installs files.`,
+      );
+    }
```
ruiconti, line 145: Does this mean that `eve add` will crash if `tsconfig.json` is already present? @!e0

ctgowrie: No. This prevents the registry from having tsconfig. eve add merges attrs in

ruiconti (approved):

PR #2374, opened 2026-08-21
fix(docs): preserve sidebar position when expanding sections

ctgowrie:
### Summary

Expanding a docs sidebar section reran Geistdocs' active-page scroll effect, moving the sidebar away from the section the reader had just opened. This patches Geistdocs so active-page scrolling still runs for route-driven ancestor changes, but not for user-driven folder toggles. Collapsed-by-default sections and active-page positioning are otherwise unchanged.

### Validation

Reproduced the jump before the patch and confirmed at desktop and mobile widths that expanding the last visible section preserves both its expanded state and the sidebar scroll position, with no mobile horizontal overflow. The docs unit suite passes (161 tests); a production build compiled and typechecked successfully but could not finish prerendering because template sync data requires network access.

### Checklist

### Diff size

Not applicable.

**Implementation** — 3 files · `+40 / -2`

The runtime change is a focused pnpm patch to Geistdocs; the remaining changes register and lock that patch until the dependency includes the fix upstream.

The regression is interaction-specific and was validated in the running docs site at desktop and mobile widths.

### DIFF
--- a/patches/@vercel__geistdocs@1.20.2.patch
+++ b/patches/@vercel__geistdocs@1.20.2.patch
@@ -1,0 +1,31 @@
+diff --git a/dist/internal/sidebar.js b/dist/internal/sidebar.js
+index eedcfef1f4a0657c91a5b626194d964bd7c97029..6c0f3767de684af2c97376bb8abff89106ef087d 100644
+--- a/dist/internal/sidebar.js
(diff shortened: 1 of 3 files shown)


ctgowrie:
Upstreamed

PR #2379, opened 2026-08-21
feat(eve): resume Web Chat sessions from URLs

ctgowrie:
### Summary

The generated Web Chat template lost its session identity and transcript on reload. Web chats now put the durable session ID at `/s/{sessionId}`, replay that session when the route loads, and remain attached when the durable turn is still streaming.

This adds `resume()` to the shared frontend store and React, Vue, and Svelte bindings. Resume first performs a bounded replay through the current durable tail, then follows only when that tail does not represent a settled session boundary.

### Validation

Focused client and frontend unit coverage passes with 186 tests. Real-socket and scaffold integration coverage passes all 71 tests, including idle cursor reconnect. Browser automation verified progressive rendering, fresh session creation, automatic replay after refresh, exact mid-transcript restoration, bottom-pinned refresh during active streaming, full-height scrollbar geometry, and visible scroll-to-bottom and New chat actions. Typecheck, docs validation, lint, and the invariant guard pass.

### Checklist

### Diff size

**Docs** — 5 files · `+59 / -20`

Documents declarative and imperative resume behavior across frontend bindings, explains the generated Next.js session routes, and includes the release changeset.

**Implementation** — 16 files · `+545 / -121`

Adds URL-addressed and sessionless routes to both Web Chat variants, bounded replay with live continuation, declarative `resume: true` support, browser-local transcript restoration, stale-stream reconnection, and frontend binding support. The generated scaffold artifact accounts for one implementation file; authenticated routes share one server component rather than duplicating auth logic.

**Tests** — 6 files · `+257 / -4`

Covers settled and idempotent replay, declarative auto-resume in React, Vue, and Svelte, interrupted-turn continuation, stale real-socket reconnection, and the resulting scaffold contract.


De-ASI-INTERFACE (approved):

cmpadden (approved):
super nice improvement.

docs/guides/frontend/overview.mdx
```diff
@@ -297,7 +298,19 @@ then save a final snapshot in `onFinish`.
...
+Call `resume()` after mounting with `initialSession` to rebuild the projection from the durable stream. If the stream ends at an in-flight turn, `resume()` stays attached until that turn reaches its boundary. If the stream already ends at a settled session boundary, it returns after replaying the history.
+
+```tsx
+const agent = useEveAgent({
+  initialSession: { sessionId, streamIndex: 0 },
+});
+
+useEffect(() => {
+  void agent.resume();
+}, [agent.resume]);
```
ruiconti: Why not encapsulate this behavior?


```suggestion
const agent = useEveAgent({
  initialSession: { sessionId, streamIndex: 0 },
  resume: true
});
```

and

```ts
export function useEveAgent(props) {
  // [...]
  useEffect(() => {
    if (!props.resume) return;
    void agent.resume();
  }, [agent.resume]);
  // [...]
}
```

ctgowrie: Good call. Will do

ctgowrie: Implemented. `useEveAgent` now accepts `resume: true`, validates that a session is attached, reports `submitted` on the initial render to avoid layout flicker, and schedules one idempotent replay after mount. The Web Chat template no longer owns a resume effect. Vue and Svelte expose the same option, while the imperative `resume()` method remains available.

PR #2392, opened 2026-08-21
fix(eve): protect non-public observability content

chadhietala:
### Summary

Hosted private and unknown conversations were excluded from retained agent traces, but AI SDK telemetry and durable Workflow data could still expose content through older observability paths. This change applies the audience decision before content projection: hosted private and unknown calls disable AI SDK input/output capture and deliver metadata-only provider events and runtime context, while public calls retain content. Each Workflow run also receives an `allowed` or `denied` metadata marker; denied session runs omit the content-derived title so content access can be denied using run metadata alone. Local `eve dev` keeps content for unknown TUI and HTTP sessions, while private sessions remain denied.

### Validation

Validated with `pnpm run typecheck`, 324 focused privacy and Workflow tests, and persisted-attribute integration coverage. The full unit run completed 7,019 tests with three unrelated timeouts; all 45 tests in those files passed when rerun directly.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

The changeset records the privacy behavior for the next patch release.

**Implementation** — 7 files · `+106 / -22`

The implementation centralizes audience-aware content capture, applies it at AI SDK and provider boundaries, and writes the Workflow access marker when runs are created.

**Tests** — 8 files · `+207 / -13`

Regression coverage verifies hosted and local capture behavior, metadata-only projections, durable delivery handling, and persisted Workflow attributes for sessions, subagents, and turns.


packages/eve/src/execution/eve-workflow-attributes.ts
```diff
@@ -209,11 +233,13 @@ export function buildSessionAttributes(input: {
   readonly inputMessage: unknown;
   readonly serializedContext: Record<string, unknown>;
 }): Record<string, EveAttributeValue> {
+  const traceContent = resolveWorkflowTraceContent(input.serializedContext);
   return {
     "$eve.channel_request_id": readChannelRequestId(input.serializedContext),
+    "$eve.trace_content": traceContent,
```
ruiconti: Is there a better name here? I'd have never guessed that the expected value for this is an access boolean.

Maybe `$eve.is_trace_content_visible: boolean` or `$eve.trace_content_visibility: "allowed"|"denied"`.

chadhietala: I changed this to `$eve.is_trace_content_visible` but workflow normalizes the values as strings so we have to do "true" | "false"

packages/eve/src/harness/instrumentation/runtime-context.ts
```diff
@@ -80,16 +82,21 @@ function buildInstrumentationStepStartedInput(
...
-    modelInput: snapshotForInstrumentation(input.modelInput, "modelInput") ?? {
-      instructions: undefined,
-      messages: [],
-    },
+    modelInput: capturesContent
+      ? (snapshotForInstrumentation(input.modelInput, "modelInput") ?? {
+          instructions: undefined,
+          messages: [],
+        })
+      : { instructions: undefined, messages: [] },
```
ruiconti, line 99: More of a meta-comment but doing sanitization ad-hoc like this is error-prone. I would be curious to see what a potentially consolidated sanitization solution could look like. Maybe hard though due existing data flows so sharing as a thought.

packages/eve/src/execution/workflow-entry.integration.test.ts
```diff
@@ -50,6 +51,12 @@ function buildSerializedContext(overrides: {
     "eve.channel": { kind: overrides.channelKind, state: overrides.channelState ?? {} },
     "eve.mode": overrides.mode,
   };
+  if (overrides.audience !== undefined) {
+    context["eve.channelInstrumentation"] = {
+      kind: overrides.channelKind,
+      metadata: { audience: overrides.audience },
+    };
+  }
```
ruiconti: Why not extend the existing `eve.channel` entry?

chadhietala: just stuck audience into `eve.channel`

ruiconti (approved):

PR #2409, opened 2026-08-22
feat(eve): add canonical agent source composition

AndrewBarba:
## Summary

Adds the source-neutral composition primitives described by #2404: immutable programmatic module sources, explicit registries, canonical slot identity, a single layer precedence order, and a namespace loader that handles filesystem and programmatic backing without virtual files.

Generated module maps can now emit static imports of a registry and bind its exact module namespace. This PR is stacked on #2407, which makes those bindings required in compiled artifacts.

## Why

Framework features currently enter eve after compilation through runtime catalogs and merge branches. Moving them into the effective source graph requires a deterministic pre-normalization selection step and a way to load an in-memory namespace through the same export/factory semantics as an authored module.

This PR establishes that boundary without migrating behavior yet, keeping the review focused on construction, identity, precedence, loading, and bundler reachability.

## Behavior and boundaries

- Defines frozen programmatic sources and shallow-frozen namespace containers while preserving branded definition values and callback identity.
- Validates normalized POSIX paths against eve's module-backed slot grammar.
- Rejects graph- or host-expanding modules in `all-local-nodes` registrations.
- Derives deterministic source IDs and framework ownership from explicit registrations.
- Canonicalizes extension variants, connection file/folder forms, and sandbox file/folder forms before composition.
- Applies one precedence order: framework default, extension package, extension override, application.
- Rejects same-layer collisions before loading either candidate.
- Loads selected programmatic exports through the ordinary export selection and zero-argument factory materialization path.
- Emits static registry imports in generated module maps and fails when a programmatic binding has no statically reachable registry.

The registry remains compiler-internal. This does not add a public mutable registration API or a global side-effect registry.

## Validation

- `pnpm fmt`
- `pnpm lint` (passes with one pre-existing `no-unsafe-optional-chaining` warning)
- `pnpm typecheck`
- `pnpm guard:invariants`
- 24 focused compiler unit tests

## Diff size

- Docs: None
- Implementation: Large
- Tests: Medium


packages/eve/src/compiler/programmatic-agent-source.ts
```diff
@@ -0,0 +1,96 @@
...
+  const segments = logicalPath.split("/");
+  const moduleName = getSupportedModuleBaseName(segments.at(-1)!);
+  if (moduleName === null || !isModuleBackedSlot(segments, moduleName)) {
+    throw new Error(`Programmatic module path "${input}" does not select an eve module slot.`);
+  }
+
+  return logicalPath;
+}
+
+function isModuleBackedSlot(segments: readonly string[], moduleName: string): boolean {
```
ctgowrie, line 81: This code smells to me. Is there a more canonical way to determine this?

AndrewBarba:
Superseded by the consolidated implementation in #2407. The PR boundaries were removed to eliminate stack overhead; no planned scope was dropped.

PR #2438, opened 2026-08-23
refactor(eve): map channel audience before the harness

chadhietala:
## Summary

- Evaluate `tracePolicy` at channel delivery and persist only its drop/input/output decision.
- Construct a decision-bound instrumentation scope whose provider graph, hooks, AI SDK integrations, OTel settings, and execution context already enforce that decision.
- Run the complete turn and every durable side-effect step with serialized session state inside a reconstructed scope.
- Remove channel audience from harness lifecycle events, trace state, span export context, runtime-context callbacks, and dynamic resolver metadata.
- Ferry decisions to persistent local subagents while remote agents classify independently.

`InstrumentationRuntime` exposes no raw hooks, provider factory, policy/settings object, or context runner. Execution can only resolve a decision and construct a capability; the harness receives only the resulting capabilities and does not branch on audience or the decision.

`tracePolicy` now returns `{ action: "drop" }` or `{ action: "record", recordInputs, recordOutputs }`. The default still records public conversations and drops private/unknown conversations. The opt-in, metadata-only `traceChannelRequests` server span remains independent because it starts before channel classification.

This is a breaking channel extension contract change (epoch 8) with a minor `eve` changeset.

Follow-up to #2335.

## Validation

- `./node_modules/.bin/tsc -p packages/eve/tsconfig.json --noEmit`
- Full unit suite: 668 files, 7,031 passed, 1 skipped
- Focused integration suite: 2 files, 28 passed
- `node ./scripts/guard-invariants.mjs`
- `oxlint` on changed TypeScript files
- Docs validation, snippets, and MDX compilation
- `git diff --check`

## Stack

This draft is stacked on #2400 to keep its diff isolated from the branch that was active when the work began. Retarget to `main` after #2400 merges.


chadhietala:
Replaced by #2448, rebuilt from the stack base as the minimal construction-only audience boundary.

PR #2477, opened 2026-08-24
feat(eve): run tool execute bodies as durable workflows

ctgowrie:
### Summary

Related to #1084. A tool that has to wait on a person, a webhook, or a timer had no durable place to do it: `execute` runs inside one model step, so anything longer than the step is lost. A tool's `execute` may now be a Workflow SDK workflow. eve starts one durable run per tool call and, by default, parks the turn until it returns; the run's return value is the tool result, however long it took. With `execution: "background"` the model gets a receipt and the agent is woken with the result.

Inside the body it is the Workflow SDK (`"use workflow"`, `"use step"`, `createHook`, `createWebhook`, `sleep`; `workflow/api` from steps). eve adds two things, both from `eve/workflow`: `ask(ctx, request)` puts a question to the human on the session's channel and returns the SDK hook the answer resumes, so it can be awaited or raced against a `sleep` deadline; and a body may be an async generator whose `yield`s are durable progress. `ctx.abortSignal` aborts when the run is cancelled and the run waits up to 30 seconds for the body to unwind before it ends. A turn or task opens the hooks its runs report on only when it is about to start one, so ordinary turns pay nothing for this. The tool's `ctx` otherwise keeps its shape; `getSandbox`, `getSkill`, `getToken`, and `requireAuth` throw in a body.

The bundler discovers directives with the SDK's own line-based pre-scan and validates placement with an AST pass, hoists the body into the workflow driver, and registers steps. The compiler derives the body's workflow id from the tool's source and records it as the tool's compiled handling; the runtime prepares a `workflow-tool-call` dispatch target from it, the same path the self-agent and task-control tools take since #2778, so the harness parks the turn instead of calling `execute`. The run reports to its owner (the parked turn or the owning task) on three hooks derived from the owner's inbox token; that protocol is internal. eve also serves the Workflow webhook route so `createWebhook()` URLs resolve.

Scope: framework tools (`ask_question`, `sleep`, subagents) keep their existing paths; rebuilding them on this primitive is a follow-up. Pre-1.0, so `ToolContext` changes without a shim and the compiled manifest version moves to 47. On current `main` only the `tool` capability's surface differs, so a single epoch is bumped (23 to 24) with a retained fixture; the other epochs this branch bumped earlier are identical to `main`'s and were dropped.

Worth a close look: `ask` is deliberately synchronous, because a `Hook` is thenable and an async `ask` could only ever hand back the answer; a request is answered once, so the hook is not a standing question. `ask` finds its run through a `Symbol.for` property on `ctx` because the driver's framework and application layers are separate bundles that share only the objects they pass around. `workflow/api` stays the SDK runtime everywhere; a body that calls it fails the driver build. The compiler derives the workflow id with the same function the transform uses; nothing inspects `execute` for it, and the runtime never learns about workflow tools except through the prepared dispatch target. The epoch bump used `--retain` because the contract classifier could not build the historical worktree in this environment (it fails to vendor the since-renamed `experimental-ai-sdk-code-mode`); the change is additive.

### Validation

The `agent-workflow-tools` e2e fixture runs six deterministic mock-model evals through the CLI: a waiting tool, a mid-body `ask`, a deadline race, a steer cancelling a parked run, a background tool with progress and a wake, and a fan-out that starts child runs from a step. Integration coverage exercises the same flows in-process and adds the cases the fixture cannot see: a deadline winning against an unanswered `ask`, a body cleaning up in `finally` after `ctx.abortSignal`, a background failure waking the agent, a waiting generator streaming `action.partial`, a finished run withdrawing its unanswered request, and `toModelOutput` applied to a run's result. Two of these regression tests were confirmed to fail against the code they fix. Locally on this tree, rebased onto `main` at f2c96a136: workspace typecheck, lint, guard invariants, docs and contract checks, unit 7548/7548, integration 699/699, the workflow-bundle, compiler, dev-generation, dynamic-tool source-edit, agent-info, and mounted-extension scenarios (30/30), and the e2e fixture 6/6 from a fresh local world. Not covered locally: the Vercel and custom worlds and `createWebhook` end to end.

### Checklist

### Diff size

**Docs** — 5 files · `+544 / -1`

A new `docs/tools/workflows.mdx` page for the authoring surface, the research plan updated to the shipped API, and the changeset.

**Implementation** — 83 files · `+2994 / -390`

Two halves of roughly equal size: the build layer that discovers, validates, hoists, and registers authored directives and composes the driver in framework and app layers, and the runtime under `execution/tool-run/` that runs a call as a workflow and connects it to the turn and task owners. The remainder is small edits where dispatch, delivery, cancellation, and the harness learn the new dispatch target. `pnpm-lock.yaml` accounts for 366 of the additions (the `@workflow/builders` dev dependency and a new e2e fixture importer); one generated contract report is a few lines.

Key files: `packages/eve/src/execution/tool-run/workflow.ts` (the run), `messages.ts` (`ask` and the owner protocol), `packages/eve/src/internal/workflow-bundle/authored-workflow-directives.ts` (directive validation and hoisting).

**Tests** — 43 files · `+2636 / -51`

Unit tests for the directive pass, the channel racer, and the run records; an integration suite for the run against real turn and task owners; scenario coverage for the build and dev-host fingerprint; the e2e fixture with six evals; and the retained compatibility fixture for tool epoch 23.


packages/eve/src/execution/tasks/child/workflow.ts
```diff
@@ -98,15 +110,27 @@ export async function taskRunWorkflow(input: TaskRunWorkflowInput): Promise<void
     let pendingAuthorizationEvents: TaskInboundAuthorizationEvent[] = [];
     let pendingUpdates: TaskInboundUpdate[] = [];
     let dispatchAcknowledged = false;
+    let runUpdateIndex = 0;
+    // The parent answers with the request's token; only this run knows it was a hook.
+    const answerHooks = new Map<string, AnswerHookRoute>();
```
ruiconti, line 122: Do we need to persist this? Analogous to SessionInbox's `pendingInput`?

@!e0 to confirm

ruiconti: @!eo but how do we durably store all the `AnswerHookRoute` that a particular workflow tool run has emitted?

packages/eve/src/execution/tool-run/messages.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/** Identifies the sending run to an owner that shares its channels across many runs. */
+export interface RunRef {
+  readonly callId: string;
+  /** The tool's parsed input, shown as the request's action when the run asks. */
+  readonly input: JsonObject;
+  readonly runId: string;
+  readonly stepIndex: number;
+  readonly toolName: string;
+  readonly turnId: string;
```
ruiconti, line 25: @!e0 should this map to `sequence`?

packages/eve/src/execution/tool-run/messages.ts
```diff
@@ -0,0 +1,111 @@
...
+import type { JsonObject, JsonValue } from "#shared/json.js";
+
+export interface ToolRunOwner {
+  readonly outcome: string;
+  readonly report: string;
+  readonly request: string;
+}
+
+/** Identifies the sending run to an owner that shares its channels across many runs. */
+export interface RunRef {
```
ruiconti, line 16: @!e0 if a workflow run is always bound to a task (if it isn't, under which circumstances it wouldn't be), shouldn't we be folding this unto Task state so the source of truth is not split?

packages/eve/src/execution/tool-run/messages.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export interface RunReport {
+  readonly from: RunRef;
+  readonly update: JsonValue;
+}
+
+/** `replyTo` is the token of the hook the human's answer resumes. */
+export interface RunRequestMessage {
+  readonly from: RunRef;
+  readonly replyTo: string;
```
ruiconti: @!e0 is `replyTo` used consistently for other hook references? I know we've also used `*token` directly, which reads clearer

packages/eve/src/execution/tool-run/messages.ts
```diff
@@ -0,0 +1,111 @@
...
+  readonly toolName: string;
+  readonly turnId: string;
+}
+
+export type RunOutcome =
+  | { readonly status: "completed"; readonly output: JsonValue }
+  | { readonly status: "failed"; readonly error: unknown }
+  | { readonly status: "cancelled"; readonly reason?: string };
+
+export interface RunReport {
```
ruiconti, line 34: @!e0 we should add comments here explaining that this is a message for propagating updates from the run upwards

packages/eve/src/execution/tool-run/workflow.ts
```diff
@@ -0,0 +1,161 @@
...
+  const iterator = result[Symbol.asyncIterator]();
+  let last: JsonValue | undefined;
+  let next = await iterator.next();
+  while (next.done !== true) {
+    last = next.value;
+    const report: RunReport = { from, update: next.value };
+    await resumeHookStep(input.owner.report, report);
+    next = await iterator.next();
+  }
+  return (next.value as JsonValue | undefined) ?? last ?? null;
```
ruiconti: @!e0 is it possible to have proper typing on `iterator.next()`? Even if by casting it away?

packages/eve/src/execution/tool-run/owner-channels.ts
```diff
@@ -0,0 +1,64 @@
...
+  return { channel, iterator: iterable[Symbol.asyncIterator](), landed: [] };
+}
+
+export type ChannelRead<R extends readonly ChannelReader<string, unknown>[]> = {
+  [I in keyof R]: R[I] extends ChannelReader<infer C, infer T>
+    ? { readonly channel: C; readonly next: IteratorResult<T> }
+    : never;
+}[number];
+
+const EXTRA = Symbol("extra");
```
ruiconti: @!e0 what are we using tagging here for?

packages/eve/src/execution/workflow-webhook-route.ts
```diff
@@ -0,0 +1,29 @@
...
+import { walkCauseChain } from "#shared/errors.js";
+
+/** The path `createWebhook()` mints, so external systems call back without knowing eve. */
+export const WORKFLOW_WEBHOOK_ROUTE_PATTERN = "/.well-known/workflow/v1/webhook/:token";
+
+export async function handleWorkflowWebhookRequest(
+  request: Request,
+  ctx: RouteContext,
+): Promise<Response> {
+  const token = ctx.params.token;
```
ruiconti, line 14: @!e0 make a seq diagram illustrating a real scenario that this route would be used; notably, scenarios in which the inline `createWebhook()` would not suffice

packages/eve/src/internal/workflow-bundle/authored-workflow-directives.ts
```diff
@@ -0,0 +1,308 @@
...
+  return {
+    executeWorkflow: HOISTED_EXECUTE_NAME,
+    hasDirectives: true,
+    hasWorkflowDirective,
+    source: hoistExecuteMethod(input.source, executeProperty, hoist.fn),
+  };
+}
+
+/** `execute: deploy`, where `deploy` is a top-level `"use workflow"` declaration. */
+function readExecuteReference(
```
ruiconti, line 160: @!e0 doesn't `workflow` exposes this parsing? or is it internal-only?

packages/eve/src/execution/workflow-webhook-route.ts
ruiconti: @!e0 mermaid diagram failed to compile.

packages/eve/src/execution/tasks/child/workflow.ts
ruiconti: @!e0 but how do we durably store all the `AnswerHookRoute` that a particular workflow tool run has emitted?

packages/eve/src/execution/tool-run/messages.ts
ruiconti: @!e0 point me to the snippet refs that allow a workflow tool to block current turn and to defer current turn (ie it's a background tool)

packages/eve/src/execution/tool-run/owner-channels.ts
ruiconti: @!e0 why is it needed? why the distinction is warranted?

packages/eve/src/execution/tasks/child/workflow.ts
ctgowrie: No — it's rebuilt on replay from the recorded request reads, same as `view`. Only history compaction would break that, which we don't do.

packages/eve/src/execution/tool-run/messages.ts
ctgowrie: Yep, good catch — it was hardcoded to 0. `RunRef` carries the turn sequence now and both emitters use it.

packages/eve/src/execution/tool-run/messages.ts
ctgowrie: Runs aren't always task-owned. Default execution parks the turn and settles the call directly; only `background` creates a task. So the run stays its own thing.

packages/eve/src/execution/tool-run/messages.ts
ctgowrie: Renamed to `answerToken`.

packages/eve/src/execution/tool-run/messages.ts
ctgowrie: Added.

packages/eve/src/execution/tool-run/workflow.ts
ctgowrie: Typed the stream at the boundary as `AsyncIterable<JsonValue, JsonValue | undefined, void>`. Cast is gone.

packages/eve/src/execution/tool-run/owner-channels.ts
ctgowrie: Wasn't needed. The only extra is the cancel sentinel, so it just takes `Promise<"cancel">` now.

packages/eve/src/internal/workflow-bundle/authored-workflow-directives.ts
ctgowrie: Only the line-based pre-scan is public. The SWC plugin has no AST API and only handles top-level functions, so the inline-method hoist stays on our side. (Small correction to e0: the downstream transform is ours, not the SDK's.)

ruiconti (approved):

PR #2491, opened 2026-08-24
research: extract eve registry service

prha:
### Summary

This proposes making `eve registry` and `eve add` the shared public interface for humans and agents, with machine-readable modes covering discovery and noninteractive installation. Constrained agents without host command access use a capability adapter for the same operation, including development-mode HITL approval, rather than a separate registry API. The internal extraction remains a thin eve-specific wrapper over the vendored shadcn registry client, while a public TypeScript API and production-specific guarantees are deferred until a concrete executor requires them.

Review should focus on whether the command surface provides sufficient agent coverage and whether the proposed internal boundary adds only eve-specific behavior.

### Validation

No tests were run because this is a research-only change; the branch diff passes whitespace validation.

### Checklist

### Diff size

**Docs** — 1 file · `+233 / -0`

Adds the focused research proposal covering command parity, shared internal registry behavior, development approval, and deferred production concerns.

Not applicable.

Not applicable for a research-only proposal.

### DIFF
--- a/research/registry-service-extraction.md
+++ b/research/registry-service-extraction.md
@@ -1,0 +1,205 @@
+---
+issue: TBD
+status: proposed
+last_updated: "2026-08-21"
+---
+
+# Registry service extraction
+
+## Goal
+
+Give registry discovery one eve-owned implementation that can be reused by:
+
+- `eve registry list/search/view`;
+- `eve add`;
+- the dev TUI's `/add` flow; and
+- `@eve/self-modification` discovery.
+
+The immediate public need is read-only discovery. `@eve/self-modification` is a
+separately published package, so it needs a narrow `eve/registry` entrypoint
+rather than a private CLI import or its own parser for the official index.
+
+Installation remains an internal eve capability in this phase. The CLI, TUI,
+and a development-mode agent action can share it without publishing a general
+project-mutation API. A production self-modification executor may create a need
+for a public mutation API later, but that API should be designed with the
+production executor rather than anticipated here.
+
+This extraction is not required for the initial guided self-modification
+handoff. Until it lands, self-modification can search the official index and ask
+the developer to type `/add <address>` manually. See
+[`selfmod-development-setup-actions.md`](./selfmod-development-setup-actions.md).
+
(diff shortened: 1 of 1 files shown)


research/registry-service-extraction.md
```diff
@@ -0,0 +1,205 @@
...
+Installation remains an internal eve capability in this phase. The CLI, TUI,
+and a development-mode agent action can share it without publishing a general
+project-mutation API. A production self-modification executor may create a need
+for a public mutation API later, but that API should be designed with the
+production executor rather than anticipated here.
+
+This extraction is not required for the initial guided self-modification
+handoff. Until it lands, self-modification can search the official index and ask
+the developer to type `/add <address>` manually. See
+[`selfmod-development-setup-actions.md`](./selfmod-development-setup-actions.md).
```
cmpadden, line 37: Can you expand on why this is necessary?

Shouldn't it be possible for the agent to invoke the same commands as a human: `eve registry` and `eve add`? In that structure we would be ensuring the tools have full coverage and utility for both humans and agents.

research/registry-service-extraction.md
```diff
@@ -0,0 +1,205 @@
...
+
+## Proposed boundary
+
+### Public `eve/registry` API
+
+The initial public API is read-only and serves the concrete cross-package use
+case:
+
+```ts
+import { searchRegistryItems } from "eve/registry";
```
cmpadden: Worth calling out that we're just a thin wrapper around shadcn registries, so we'll want to be careful not to implement unnecessary layers of abstraction.

research/registry-service-extraction.md
prha: the selfmod package does not have access to the eve binary in its sandbox... this change is adding some subset of the capabilities inside of a tool call

research/registry-service-extraction.md
prha: added a section here.... lemme update

De-ASI-INTERFACE (approved):

prha:
okay, going to abandon this for another approach

PR #2495, opened 2026-08-24
[1/3] feat(eve): define Connect manifest artifact

benpankow:
### Summary

Defines the private Connect manifest contract used by #2496. Requirements keep their opaque `target` identity, expose service endpoints through `interfaces`, and group Connect provisioning metadata under `connect`, including subject types and an inline provider manifest.

```json
{
  "target": "connector:oauth/linear",
  "interfaces": [
    { "protocol": "mcp", "url": "https://mcp.linear.app/mcp" }
  ],
  "connect": {
    "subjectTypes": ["user"],
    "service": "linear",
    "type": "oauth",
    "manifest": {
      "$type": "https://datatracker.ietf.org/doc/html/rfc7591"
    }
  }
}
```

### Validation

Adds unit coverage for empty manifests, opaque target preservation, MCP and custom interfaces, and inline provider metadata.


allenzhou101:
wondering if this should be framed around the connections spec rather than vercel connect? eve should be able to emit this independently of the host, with connect as the first consumer.

maybe `connections-manifest.json` / `ConnectionsManifest`? not just a rename though. would be good to separate what the app needs from connect-specific things like how we resolve a connector locator. dont think we need to finish the whole spec here, just get the boundary right.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -0,0 +1,46 @@
+import type { JsonObject } from "#shared/json.js";
+
+export const VERCEL_CONNECT_MANIFEST_FILENAME = "vercel-connect-manifest.json";
+export const VERCEL_CONNECT_MANIFEST_KIND = "vercel-connect-manifest";
+export const VERCEL_CONNECT_MANIFEST_SCHEMA_VERSION = 1;
+
+export type VercelConnectTarget =
+  | { readonly mode: "direct"; readonly locator: string }
```
dvoytenko: IMHO this won't know whether it's a direct or a binding - we should treat them the same and just record a name or a local name or call it just reference and let Connect decide how to bind it.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -0,0 +1,46 @@
...
+  | { readonly mode: "binding"; readonly reference: string };
+
+export interface VercelConnectRequirement {
+  readonly target: VercelConnectTarget;
+  readonly connector: {
+    readonly type: string;
```
dvoytenko, line 17: Will we always know this? Good if we so. Sometimes a range is possible here. E.g. OAuth can be used for Datadog, but it can also be per-user APP Keys. Not sure if we need to allow this range here.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -0,0 +1,46 @@
...
+    readonly configuration?: JsonObject;
+  };
+  readonly interface?: { readonly protocol: "mcp" | "openapi"; readonly url: string };
+  readonly providerConfiguration?: { readonly format: string; readonly path: string };
```
dvoytenko: IMHO we can always extract the provider configuration later. The reason why this might be harder is because the provider config would have some features that need to be updated by Connect such as `redirect_uri` or `trigger_url`. The idea of doing `trigger?: {...}` below is interesting, but then we'd need to do something like:

```
override: [
  {
     path: "oauth_settings/redirect_uris",
     kind: "redirect_uri"
  },
  {
     path: "event_subscriptions/webhook_url",
     kind: "trigger_url"
  },
  {
     path: "interactivity/url",
     kind: "trigger_url"
  },
]
```

I.e. instructions on how Connect should update the manifest. But IMHO that's too much. We can do something like `redirect_uris: ["https://well-known-url"]` that we know how to override and that would be enough for us potentially.

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Agreed - now `target` is just a string. Imagine we can have a temporary contract that e.g. `connect` can emit `id:abcd` for explicit ID binding in the short term or something.

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Agreed, I think this ends up being implicit on the connect end, e.g. for Slck connect knows to replace redirect, callback, interactivity URLs.

benpankow:
> wondering if this should be framed around the connections spec rather than vercel connect? eve should be able to emit this independently of the host, with connect as the first consumer.
> 
> maybe `connections-manifest.json` / `ConnectionsManifest`? not just a rename though. would be good to separate what the app needs from connect-specific things like how we resolve a connector locator. dont think we need to finish the whole spec here, just get the boundary right.

agree for eventual state, I do think having a more specific artifact for Connect is useful at least as we determine the specific end shape

dvoytenko (approved):

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: The particular Connect use-case specifies one type, I think, even if the Connector itself might be set up for both. e.g. the Connect-side reconciliation checks the requested types are a subset.

benpankow:
Superseded by the replacement stack starting at #3468, which moves the boundary to a generic eve external-resource snapshot.

PR #2496, opened 2026-08-24
[2/3] feat(eve): emit Connect & Slack app manifests

benpankow:
### Summary

Emits deployment requirements for direct Connect-backed connections and Slack channels. A production build writes `vercel-connect-manifest.json`; Slack requirements include an inline app manifest with Connect-managed callback and trigger URLs.

Built on #2495. Further Slack customization is handled in #2977. The coordinated `@vercel/connect/eve` producer supplies the Connect metadata consumed here.

### Data flow

The authored definition carries runtime behavior plus build-time Connect metadata:

```ts
export default slackChannel({
  botName: "Support agent",
});
```

During compilation, `slackChannel()`'s manifest builder and credential metadata are resolved into serializable route data. Connections follow the same path through their normalized auth definition:

```ts
{
  adapterKind: "slack",
  logicalPath: "channels/slack.ts",
  urlPath: "/eve/v1/slack",
  manifest: {
    $type: "https://docs.slack.dev/reference/app-manifest/",
    display_information: { name: "Support agent" },
    // Slack features, scopes, and settings...
  },
  vercelConnect: {
    connector: "connector:slack/my-agent",
    connectorType: "slack",
    principalTypes: ["app"],
    service: "slack",
  },
}
```

At the end of the production build—after Nitro output is assembled and before the staged output is published—`emitVercelConnectManifest()` reads the compiled manifest and flattens eligible connections and channel routes into `requirements`:

- Connection requirements use the compiled protocol and endpoint, with an RFC 7591 provider manifest.
- Slack requirements clone the generic compiled provider manifest and add Connect's OAuth callback and webhook URLs. The Slack-owned builder supplies the provider `$type`.
- Entries without the complete Connect metadata contract are omitted rather than emitting partial requirements.

The resulting output is:

```json
{
  "kind": "vercel-connect-manifest",
  "schemaVersion": 1,
  "generator": { "name": "eve", "version": "..." },
  "requirements": [
    {
      "target": "connector:slack/my-agent",
      "interfaces": [
        {
          "protocol": "custom",
          "url": "https://docs.slack.dev/apis/web-api/",
          "npm": "@slack/web-api"
        }
      ],
      "connect": {
        "subjectTypes": ["app"],
        "service": "slack",
        "type": "slack",
        "manifest": {
          "$type": "https://docs.slack.dev/reference/app-manifest/",
          "display_information": { "name": "Support agent" },
          "oauth_config": {
            "redirect_urls": ["https://connect.vercel.com/callback"]
          },
          "settings": {
            "event_subscriptions": {
              "request_url": "https://connect.vercel.com/trigger?path=%2Feve%2Fv1%2Fslack"
            },
            "interactivity": {
              "request_url": "https://connect.vercel.com/trigger?path=%2Feve%2Fv1%2Fslack"
            }
          }
        }
      },
      "uses": [
        {
          "kind": "channel",
          "name": "slack",
          "logicalPath": "channels/slack.ts"
        }
      ]
    }
  ]
}
```

`uses` keeps each requirement attributable to the authored filesystem definition. This PR intentionally emits only the Connect manifest for now; there is no separate Slack manifest artifact.

### Validation

Adds unit coverage for connection requirement assembly, inline Slack manifests, Connect-managed URLs, incomplete metadata, and non-mutating Slack manifest transformation.

🤖


packages/eve/src/internal/authored-definition/connection.test.ts
```diff
@@ -119,14 +119,20 @@ describe("normalizeMcpClientConnectionDefinition", () => {
           auth: {
             getToken: async () => ({ token: "x" }),
             principalType: "app",
-            vercelConnect: { connector: "oauth/mcp-linear-app" },
+            vercelConnect: {
+              connector: "oauth/mcp-linear-app",
+              connectorType: "oauth",
+              principalTypes: ["app"],
```
dvoytenko: Nit: ofc it's just a test, but realistically MCPs are most commonly "user".

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -1,8 +1,30 @@
+import type { CompiledAgentManifest } from "#compiler/manifest.js";
 import type { JsonObject } from "#shared/json.js";
 
 export const VERCEL_CONNECT_MANIFEST_FILENAME = "vercel-connect-manifest.json";
 export const VERCEL_CONNECT_MANIFEST_KIND = "vercel-connect-manifest";
 export const VERCEL_CONNECT_MANIFEST_SCHEMA_VERSION = 1;
+export const SLACK_APP_MANIFEST_FORMAT = "slack-app-manifest";
+
+export interface SlackAppManifest {
```
dvoytenko: Can we just use Slack SDK in devDependencies to pull this type from there? Also, it'd be good to keep this somewhere next to the relevant source code, such as `slackChannel`, so that this doesn't need to know about the exact structure.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -32,6 +54,140 @@ export interface VercelConnectManifest {
...
+    const name = channel.name.slice(0, 35);
+    manifests.set(slackAppManifestPath(channel.logicalPath), {
+      display_information: { name },
+      features: {
+        app_home: {
+          home_tab_enabled: false,
+          messages_tab_enabled: true,
+          messages_tab_read_only_enabled: false,
+        },
+        bot_user: { display_name: name, always_online: false },
```
dvoytenko: Isn't it always online usually?

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -32,6 +54,140 @@ export interface VercelConnectManifest {
...
+      },
+      oauth_config: { scopes: { bot: ["app_mentions:read", "chat:write"] } },
+      settings: {
+        event_subscriptions: { bot_events: ["app_mention"] },
```
dvoytenko: There are a lot more of these, right?

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -32,6 +54,140 @@ export interface VercelConnectManifest {
...
+      display_information: { name },
+      features: {
+        app_home: {
+          home_tab_enabled: false,
+          messages_tab_enabled: true,
+          messages_tab_read_only_enabled: false,
+        },
+        bot_user: { display_name: name, always_online: false },
+      },
+      oauth_config: { scopes: { bot: ["app_mentions:read", "chat:write"] } },
```
dvoytenko: heads up, there's also `bot_optional` that can be used to specify optional scopes.

packages/eve/src/internal/vercel-connect-manifest.ts
dvoytenko: Unfortunately, there's no such thing as `bot_events_optional`. I wonder if that'd ever be needed. But ok for now.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -32,6 +54,140 @@ export interface VercelConnectManifest {
...
+      settings: {
+        event_subscriptions: { bot_events: ["app_mention"] },
+        interactivity: { is_enabled: true },
+        org_deploy_enabled: false,
+        socket_mode_enabled: false,
+        token_rotation_enabled: false,
```
dvoytenko: The system should decide which to use. Though it's ok to have it - Connect would just override.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -32,6 +54,140 @@ export interface VercelConnectManifest {
...
+    );
+  }
+  for (const [path, slackManifest] of slackManifests) {
+    const outputPath = join(input.outputDirectory, path);
+    await mkdir(dirname(outputPath), { recursive: true });
+    await writeFile(outputPath, `${JSON.stringify(slackManifest, null, 2)}\n`);
+  }
+}
+
+export function buildSlackAppManifests(manifest: {
```
dvoytenko: Can this be somehow hosted with/on slackChannel? I.e. we use some build definition to call on slackChannel to call and extract from the source code definition. This way we can extract slash commands and whatever else `slackChannel()` declares and not worry about keeping track of all channels/connections in one place.

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Addressed in the latest stack update: Slack manifest construction now lives under `public/channels/slack/app-manifest.ts` and is checked against the official `@slack/web-api` manifest request type. The dependency is direct and dev-only, so it adds no eve runtime dependency.

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Addressed in the latest stack update: `slackChannel()` now owns a build definition that produces the final Slack-native manifest during compilation. The Connect emitter no longer constructs or writes Slack manifests; it only references an independently emitted provider artifact.

packages/eve/src/internal/authored-definition/connection.test.ts
benpankow: Updated

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Changed so eve does not specify a default; user can explicit configure if they want

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Yes, moved out of connect emitter into Slack channel implementation, we check there against the exported `@slack/web-api` type.

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: > added support to `slackChannel` for `bot_Optional`

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -31,6 +38,127 @@ export interface VercelConnectManifest {
...
+      }
+      return [
+        {
+          target: vercelConnect.connector,
+          interfaces: [{ protocol: connection.protocol, url: connection.url }],
+          connect: {
+            subjectTypes: vercelConnect.principalTypes,
+            service: connectService(vercelConnect.connector, vercelConnect.connectorType),
+            type: vercelConnect.connectorType,
+            manifest: { $type: RFC_7591_MANIFEST_TYPE },
```
dvoytenko, line 70: This won't always be the case, but we can start here, I guess. Just subject to change very soon.

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -31,6 +38,127 @@ export interface VercelConnectManifest {
...
+  ];
+}
+
+function buildConnectSlackManifest(manifest: JsonObject, triggerPath: string): JsonObject {
+  const copy = parseJsonObject(manifest);
+  const oauthConfig = objectValue(copy.oauth_config);
+  const settings = objectValue(copy.settings);
+  const eventSubscriptions = objectValue(settings.event_subscriptions);
+  const interactivity = objectValue(settings.interactivity);
+  const requestUrl = `${VERCEL_CONNECT_TRIGGER_URL}${triggerPath}`;
```
dvoytenko: encodeURICOmpoennt on triggerPAth?

packages/eve/src/public/channels/slack/app-manifest.ts
```diff
@@ -0,0 +1,44 @@
...
+          },
+          bot_user: { display_name: name, always_online: false },
+        },
+        oauth_config: { scopes: { bot: ["app_mentions:read", "chat:write"] } },
+        settings: {
+          event_subscriptions: { bot_events: ["app_mention"] },
+          interactivity: { is_enabled: true },
+          org_deploy_enabled: false,
+          socket_mode_enabled: false,
+          token_rotation_enabled: false,
```
dvoytenko, line 33: I don't think this property is required, so let's drop it. Connect will always set it to true.

packages/eve/src/internal/vercel-connect-manifest.ts
dvoytenko: IMHO it should be `true` by default

packages/eve/src/compiler/manifest.ts
```diff
@@ -500,6 +515,8 @@ const compiledChannelDefinitionSchema = z
     exportName: z.string().optional(),
     adapterKind: z.string().optional(),
     cors: compiledChannelCorsSchema.optional(),
+    slackAppManifest: jsonObjectSchema.optional(),
```
dvoytenko: Should it be here if we moved it to Slack side? Or maybe let's make it follow 

```
manifest: { $type: "string", ....theRestOfArgs? }
```

This way Slack will set `manifest: { $type: "https://docs.slack.dev/reference/app-manifest/", ....}`

dvoytenko (changes requested):

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Dropped, Connect can own

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Updated to encode the route with `encodeURIComponent`

packages/eve/src/public/channels/slack/app-manifest.ts
benpankow: Drop, Connect transformation now adds

packages/eve/src/internal/vercel-connect-manifest.ts
benpankow: Updated contract to carry service explicitly instead of deriving from connector reference

packages/eve/src/compiler/manifest.ts
benpankow: Updated to generic manifest field with `$type`

packages/eve/src/internal/vercel-connect-manifest.ts
```diff
@@ -31,6 +37,123 @@ export interface VercelConnectManifest {
...
+        vercelConnect.service === undefined ||
+        channel.adapterKind !== "slack" ||
+        channel.manifest === undefined
+      ) {
+        return [];
+      }
+      return [
+        {
+          target: vercelConnect.connector,
+          interfaces: [{ protocol: "custom" as const, url: SLACK_API_URL, npm: "@slack/web-api" }],
```
dvoytenko, line 96: Let's move both this and the buildConnectSlackManifest fully to slack channel, so this module is type-agnostic. It's ok for Slack channel to reference Connect URLs IMHO - we can always find another compromise later. But I'd like to have a strong ownership model where each channel/connection-type fully own their manifests and interfaces.

packages/eve/src/public/channels/slack/app-manifest.ts
dvoytenko: Did you push the change? I'm still seeing it here.

But to confirm "Connect transformation" should just be backend transformation, not in eve, since backend has a better info.

dvoytenko (approved):

benpankow:
Superseded by #3469. Final Connect manifest ownership and provider-specific conversion now live in vercel-internal#989.

PR #2502, opened 2026-08-24
research: let self-modification drive registry installs

prha:
### Summary

Related to #742. This research plan proposes letting the development-only self-modification subagent install official registry items through eve's existing noninteractive CLI contract, while handing setup-bearing items to the dev TUI. It keeps credentials and setup questions out of the model transcript and avoids widening the source sandbox or introducing a parallel registry API.

This is the first PR in a five-PR stack; the implementation continues in #2638. Review should focus on the host-tool/TUI boundary and the decision to wrap `eve add` rather than expose a new public registry API.

### Validation

No behavioral tests are applicable to this research-only PR; the complete stack adds focused coverage for the proposed contracts and flows.

### Checklist

### Diff size

**Docs** — 1 file · `+360 / -0`

The research plan records the security boundary, authoring contract, observable behavior, and staged implementation needed to review the proposal before its implementation PRs.

Not applicable.

Not applicable for a research-only proposal.

### DIFF
--- a/research/self-modification-registry-install.md
+++ b/research/self-modification-registry-install.md
@@ -1,0 +1,360 @@
+---
+issue: https://github.com/vercel/eve/issues/742
+status: proposed
+last_updated: "2026-08-26"
+---
+
+# Self-modification: driving registry installs
+
+## Proposal
+
+The self-modification subagent can search the registry but cannot install from
+it. Asked to "add a Slack channel", it finds `channel/slack` and then tells the
+developer to type `/add channel/slack` themselves. The developer has to leave
+the conversation to complete a request the agent already understood.
+
+The subagent should install registry items itself, under approval, and hand off
+to the dev TUI only for the steps that genuinely require a terminal.
+
+It should do that by driving the install contract eve already publishes —
+`eve add <item> --non-interactive` — through one tool, rather than through a new
+programmatic registry API. Anything that would ask the developer a question
+goes to the wizard built to ask it.
+
+The constraint that looks like it blocks all of this — the subagent's narrow
+sandbox — is not the constraint. Extension tools do not run in the sandbox.
+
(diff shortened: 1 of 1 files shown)


research/registry-service-extraction.md
```diff
@@ -0,0 +1,217 @@
...
+That command enters the existing TUI-owned interactive setup flow, which can
+open the browser, show authentication instructions, collect answers, handle
+cancellation, and report deployment follow-up. Extend `/add` argument parsing to
+recognize `--skip-install` and route it to setup without reinstalling files or
+packages. The model must not invoke the continuation itself or claim setup is
+complete before the developer finishes it.
+
+Items without declared setup complete in the original tool call. Clients that
+cannot present tool approval retain the manual `/add <address>` handoff and do
+not expose `selfmod__add_registry`.
```
OwenKephart: I think this bit can be simplified a lot here actually -- the setup functions are designed to be possible for an agent to interact with iteratively

e.g. with the CLI, you can do:

```
eve add xyz --skip-install --json
eve add xyz --skip-install --json --answer 'q1=answer1'
eve add xyz --skip-install --json --answer 'q1=answer1' --answer 'q2=answer2'
```

...and so on

It should be fairly straightforward to extend the add tool call and add a few instructions to let it navigate these iterative prompts inline.

For something like Claude Code, the agent is able to translate this CLI interaction into native TUI components (e.g. an option menu), and I imagine much the same would be possible here 

Basically a simpler version of what's being discussed in the next section

prha: Is that true of flows that require some external action (like OAuth) or manage some external state?  I'm thinking of some Vercel Connect flow that requires the creation of some connector.

prha:
Superseded by the implementation stack beginning with #2638.

PR #2509, opened 2026-08-24
experimental: add `harness_agent` tool

felixarntz:
### Summary

This PR implements a `harness_agent` tool, running variable coding agent harnesses via AI SDK's `HarnessAgent` layer.

- `defineFixedHarnessAgentTool()` returns a tool that invokes `HarnessAgent` with most parameters controlled at the tool definition; it allows easy composition of various specific tools that rely on `HarnessAgent`
- `defineDynamicHarnessAgentTool()` returns a tool that invokes `HarnessAgent` with most parameters controlled via the tool call itself; by definition it makes the tool extremely flexible, but it will probably be less efficient with most agents because it's complicated to use for them

A new `coding-agent` fixture allows testing the tools. Prompt e.g.:
> Add different locale support to `ms`, e.g. so that strings like "2 Tage" work just as well as "2 days".

This PR is **experimental**. It's built as part of eve for now so that I would not be restricted by any potential gaps in the extension API, but once we're good with a first version, it should become an eve extension instead.

### Misc Notes

- The implementation doesn't handle AI Gateway credentials safely yet, they get forwarded to the sandbox. `HarnessAgent` supports a safer behavior already, but that cannot be used here because eve orchestrates its own sandbox and therefore is responsible on its own. https://github.com/vercel/ai/pull/19415 provides the necessary `HarnessAgent` surface for eve to fully take control and prevent credential forwarding.
- `packages/eve/src/internal/nitro/host/dev-workflow-world-setup.ts` in this PR has temporary "fixes" so that local workflows can run long enough to allow for agentic turns to pass.

### Checklist


felixarntz:
for some reason, when I run `pnpm --filter coding-agent dev`, the agent appears to run in the sandbox as `root` user. this is unexpected because the eve Dockerfile explicitly sets the user to `vercel-sandbox`.

I'm not quite sure why this is happening. even asking it to run `whoami` in bash confirms that it runs as `root`.

for now, the commit https://github.com/vercel/eve/pull/2509/changes/4d4201e0643480c4c950a03dea0bb1d94ede7e45 is a temporary workaround to invoke `HarnessAgent` specifically as the `vercel-sandbox` user, but that really shouldn't be necessary once we've figured out what's going wrong here.

cc @cmpadden

felixarntz:
this is now being implemented in a separate project https://github.com/vercel-labs/harness-agent-eve-extension.

A more proper implementation, with durable workflow and progress stream emissions, is being explored in #3156.

Issue #2347, opened 2026-08-20
runtime: generic owner-scoped runtime-tool contribution seam


AndrewBarba:
## Problem

Framework-owned runtime code sometimes needs to contribute a qualified set of branded dynamic tools after preamble resolution (today: `connection_search`; next: memory provider tools). No generic contributor seam owns validation, durable capture, replacement, collision handling, and provenance — so each contributor either reimplements those behaviors or bypasses them.

## Reproduction

Contribute a branded tool set from runtime code outside the authored `defineDynamic` path. There is no single internal API that validates entries, records durable metadata through the F1 descriptor path, attaches owner/source provenance, replaces a prior contribution atomically per owner, handles name collisions by the established precedence rule, and cleans up on runtime-revision change.

## Desired invariant

One internal runtime-tool contribution API: owner-scoped atomic contribution transactions feeding the same validation/merge path as authored dynamic tools, with qualification through the existing name-validation primitive, durable capture via F1, call origins via F2, and centralized replacement/cleanup. `connection_search` migrates as the first consumer without changing its public behavior.

## Scope

Internal runtime API built entirely on F1/F2. No memory types.

Part of the first-class memory restructure (#1510); foundation PR F3.

AndrewBarba:
Closing as superseded by the canonical source-graph research in #2404 and implementation in #2516. Runtime contribution is too late to own source selection, bindings, hydration, and inspection. Memory provider tools will compile as ordinary programmatic defineDynamic wrappers through that graph, so no runtime-tool contribution seam remains to implement.

PR #2516, opened 2026-08-24
feat(eve): sources - compile one canonical graph

AndrewBarba:
### Summary

Related to #2347. Implements PR 1 from the programmatic agent sources plan: eve now compiles framework defaults, authored files, extensions, overrides, and in-memory modules through one source-neutral candidate graph and one authoritative binding table. Each local node is projected and composed exactly once; selection projects every winning filesystem, resource, or programmatic source into the same `{ candidate, kind, source }` shape without executing a loader. The selected config still loads first, its external dependencies finalize bindings over the same typed state, and every remaining primitive runs through one normalization switch without provider-specific dispatch.

Application files at a mounted extension's final qualified path now exercise the documented highest-precedence override, with packaged and flat skill forms collapsing onto the same canonical slot. Module-backed instructions, skills, and schedules keep ownership only in their binding and recover it when resolved; markdown and resource records retain inline ownership when no binding exists. Development-server reuse probes the internal development control route instead of the replaceable public health channel, and the superseded health helper is gone. In-memory and production extension hydration both derive scope from the stable package namespace, while `load_skill` reads authored skills from the active compiled bundle instead of carrying a second serialized skill catalog.

Review follow-up hardens three independently reproduced failure modes: remote subagent overrides now remain active selection winners, overlapping channel routes reject incompatible CORS preflights while allowing identical policies, and the redundant `subagentEdges` graph was removed in favor of validated `parentNodeId` ownership with parent-existence, reachability, and cycle checks. Canonical tool, eve-channel, sandbox, instrumentation, and skill contracts now live outside public adapters; public subpaths are thin re-exports, and framework sources consume the same canonical definition values without importing through the public surface.

Manifest v42, diagnostics v2, and agent-info v3 reject older serialized shapes instead of repairing them, while per-tool exports replace `eve/tools/defaults`. This deliberately preserves the existing name-based effect dispatch seam and the transitional `localSubagentExecutors` identity classifier; PR 2 replaces both with declared dispatch effects.

Review should focus on source selection and binding completeness, extension/subagent provenance, and filesystem/in-memory loader parity.

### Validation

Focused regressions cover root and local-subagent single-pass projection, application-over-extension precedence including local-to-remote and remote-to-remote subagent replacement, packaged-skill replacement by a flat module, binding-owned module provenance, overlapping CORS preflight behavior, subagent tree integrity, internal dev readiness, stable extension hydration scope, and loading authored skills from the active compiled bundle.

All non-TUI eve unit tests passed with 6,134 assertions (one skipped), integration passed with 665 assertions, and scenario passed with 414 assertions (15 skipped). Workspace typecheck, formatting, lint, invariant guards, and docs checks passed. `pnpm test:tui` was not rerun for this follow-up; fixture-owned e2e suites run only in CI. The existing `agent-tools-sandbox` redeploy eval remains the CI gate for fresh-deployment skill adoption.

### Checklist

### Diff size

**Docs** — 14 files · `+90 / -51`

Updates the affected config, tool, channel, health, sandbox, client, extension, and inspection documentation, records the PR 2 dispatch commitment, and includes the required minor changeset.

**Implementation** — 350 files · `+8,013 / -7,333`

The cross-cutting compiler/runtime migration replaces parallel source catalogs and graph authorities with the canonical graph, moves shared contracts behind source-neutral boundaries, and carries the generated channel capability report required by the public route contract.

<details>
<summary>Why the implementation spans many files</summary>

Source identity and ownership previously leaked across compiler, runtime, channel, harness, setup, and client layers. Making the manifest authoritative requires those consumers to read selected bindings and provenance directly; keeping compatibility adapters would preserve the second source of truth this PR removes.

</details>

**Tests** — 148 files · `+2,386 / -5,811`

Replaces catalog-, fallback-, and agent-info-v2 assertions with compiler, serialization, loading, route-plan, ownership, adapter-readiness, and parity coverage. The large net deletion comes from removing tests for superseded framework tool/channel catalogs and the dead public-health probe instead of retaining duplicate coverage.


packages/eve/src/compiler/validate-artifact.ts
```diff
@@ -0,0 +1,390 @@
...
+    if (binding === undefined) fail(`module source "${sourceId}" has no compiled binding`);
+    validateBinding(sourceId, reference.logicalPath, binding);
+  }
+  for (const sourceId of Object.keys(node.bindings)) {
+    if (!referencedModuleSources.has(sourceId)) {
+      fail(`compiled binding "${sourceId}" is not referenced by its node manifest`);
+    }
+  }
+
+  const activeSourceIds = new Set([
```
ruiconti, line 190: An application override of an extension subagent can validly replace the package definition with a remote agent, but this validator rejects the resulting artifact. Repro: the package contributes `subagents/weather`; `extensions/crm/subagents/weather` overrides it with `defineRemoteAgent(...)`; compilation ends with `shadowed source "ext:crm:subagents/weather" has dangling winner "ext-override:crm:subagents/weather"`. The remote winner is present in `node.remoteAgents` and its binding is validated below, but `activeSourceIds` never includes `remoteAgents[].sourceId`. Please include remote-agent sources in the active set and cover local-to-remote and remote-to-remote overrides.

packages/eve/src/compiler/channel-route-plan.ts
```diff
@@ -0,0 +1,149 @@
...
+  };
+}
+
+function derivePreflights(
+  effective: readonly CompiledChannelDefinition[],
+): CompiledChannelPreflightDefinition[] {
+  const byPath = new Map<string, CompiledChannelDefinition[]>();
+  for (const channel of effective) {
+    if (channel.method === "WEBSOCKET" || channel.cors === undefined) continue;
+    const identity = normalizeRoutePattern(channel.urlPath);
```
ruiconti, line 95: `byPath` uses normalized identity, not route overlap. That leaves two generated `OPTIONS` handlers for `GET /accounts/:id` with CORS A and `POST /accounts/me` with CORS B. Both ordinary routes survive because their methods differ, but H3 resolves `OPTIONS /accounts/me` to the static handler, so a GET preflight receives policy B. I verified that with the installed H3 router in both registration orders. The research contract already defines a static path inside a parameter route's match-space as overlapping. Please compare CORS policies across overlapping selected patterns and reject differing policies. Identical policies are safe whichever preflight handler wins.

packages/eve/src/compiler/validate-artifact.ts
```diff
@@ -0,0 +1,390 @@
...
+  for (const edge of manifest.subagentEdges) {
+    if (!nodes.has(edge.parentNodeId))
+      fail(`subagent edge parent "${edge.parentNodeId}" is missing`);
+    if (!nodes.has(edge.childNodeId)) fail(`subagent edge child "${edge.childNodeId}" is missing`);
+    const key = `${edge.parentNodeId}\0${edge.childNodeId}`;
+    if (edges.has(key))
+      fail(`subagent edge "${edge.parentNodeId}" -> "${edge.childNodeId}" is duplicated`);
+    edges.add(key);
+  }
+  for (const subagent of manifest.subagents) {
```
ruiconti: We now serialize parentage twice: `CompiledSubagentNode.parentNodeId` and `subagentEdges`. This only proves that the declared edge exists; it does not reject an extra incoming edge or a disconnected cycle. An extra `root -> child` edge validates, then runtime traverses the child twice and throws `Found multiple runtime agent nodes`. A disconnected `A <-> B` cycle validates and is never reached from root. Since `parentNodeId` already contains the full tree, can we remove `subagentEdges` and derive adjacency? If the field must remain, validation needs exactly one incoming edge per child, root reachability, and acyclicity.

packages/eve/src/public/tools/bash.ts
```diff
@@ -36,20 +34,13 @@ export const BASH_OUTPUT_SCHEMA = z.strictObject({
...
-    input as BashInput,
-  );
-}
-
-export const BASH_TOOL_DEFINITION: ResolvedToolDefinition = {
+export const bash = defineTool({
   description: "Execute a shell command in the shared workspace environment.",
-  execute: executeBash,
+  async execute(input, ctx) {
+    return await executeBashOnSandbox(await ctx.getSandbox(), input as BashInput);
```
ruiconti: I don't understand why these are still `#execution`-owned. `bash` is already expressed entirely through the authored-tool contract: `defineTool`, schemas, and `ctx.getSandbox()`. The helper only operates on `SandboxSession`. Static bundle reachability requires a static import; it does not require that import to point into `execution/`. The same is true for `glob`, `grep`, `read_file`, `write_file`, and `web_fetch` (`ctx.abortSignal`). These can be canonical publicly authored tools, with private helpers colocated under a primitive-owned/public boundary; the framework source and package subpath can still import the same exact value. Please either move these implementations out of `execution/`, or document the concrete dependency that prevents each one from being publicly authored. This bundler comment is not that dependency. `load_skill`, `connection_search`, and the compaction side of `todo` are real exceptions; they need narrow seams rather than defining the ownership of the otherwise isolated tools.

ruiconti (changes requested):
Requesting changes on four issues: two reproducible compiler/route-plan failures, one redundant graph authority that validation does not make safe, and the unresolved public/execution boundary. Details inline.

packages/eve/src/compiler/validate-artifact.ts
AndrewBarba: Confirmed locally before changing the code. A real mounted extension with a local weather subagent overridden by an application remote reproduced validate/source-selection: dangling winner; the remote-to-remote form exposed the same missing active-source accounting. validateCompiledAgentResources now treats remote-agent source IDs as active, and the mounted-extension scenario covers both override directions. Full scenario suite: 414 passed, 15 skipped.

packages/eve/src/compiler/validate-artifact.ts
AndrewBarba: Confirmed locally before changing the code by schema-parsing a disconnected parent cycle plus an extra incoming edge; the old validator accepted it. I removed subagentEdges from the manifest, compiler, and runtime, derive adjacency from each node's canonical parentNodeId, and validate parent existence, root reachability, and acyclicity. The regression passes.

packages/eve/src/compiler/channel-route-plan.ts
AndrewBarba: Confirmed locally before changing the code with GET /accounts/:id using CORS policy A plus POST /accounts/me using policy B: the compiler emitted overlapping incompatible OPTIONS entries. Preflight derivation now compares every overlapping selected route pattern and rejects differing CORS policies; identical policies remain valid. The focused regressions and all non-TUI eve unit tests pass.

AndrewBarba:
Also addressed the review-summary boundary concern. Canonical tool, eve-channel, sandbox, instrumentation, and skill contracts now live in primitive-owned or shared modules; public subpaths are thin re-exports, framework sources import the same canonical values, and production code under src/public has no imports from runtime, harness, or execution. Structural boundary and file-size checks, public portability coverage, and framework module-evaluation coverage pass.

ruiconti (approved):

PR #2523, opened 2026-08-25
fix(eve): resolve -thinking suffix models correctly

Krish120003:
### Summary
The existing `findCatalogModelBySlug` implementation strips `-thinking` before every catalog lookup, which broke models whose canonical slug always had the prefix (e.g. `arcee-ai/trinity-large-thinking`). 

This PR changes slug lookups to try exact slug before stripping the `-thinking` prefix. 

### Validation

Added tests to confirm model slug resolving behaves as expected. 

### Checklist


ruiconti (approved):

packages/eve/src/internal/model-catalog.ts
```diff
@@ -41,7 +41,15 @@ export function findCatalogModelBySlug(
...
-  return models.find((model) => model.slug === normalizeCatalogModelId(slug));
+  // Some models publish `-thinking` as their own canonical slug, so the
+  // verbatim id must win before the suffix is stripped.
+  const exact = models.find((model) => model.slug === slug);
+  if (exact !== undefined) {
+    return exact;
+  }
+
+  const normalized = normalizeCatalogModelId(slug);
+  return normalized === slug ? undefined : models.find((model) => model.slug === normalized);
```
ruiconti, line 52: @!e0 sanity-check this that this is fully backcompat

ruiconti: @!e0 that cache miss is unavoidable; should be safe to merge then

ctgowrie (approved):

PR #2524, opened 2026-08-25
chore(docs): update geistdocs to 1.24.0

molebox:
Bumps the docs site to `@vercel/geistdocs` 1.24.0 and opts into its new `markdownNotFound` proxy behavior: agent and `Accept: text/markdown` requests to unmatched paths now get a concise Markdown 404 with links to `/docs`, `/sitemap.md`, and `/llms.txt`, instead of the HTML not-found page. `/templates`, `/benchmarks`, and `/resources` stay HTML-only (they have no Markdown mapping), and browsers are unaffected everywhere.

1.24.0 also brings the "When to use eve" section (driven by the existing `agent.product` metadata) into the generated `/llms-full.txt`, and a package-internal navbar accessibility fix.

Verified with `test:unit` (165/165), `docs:check`, `check:deps`, oxlint/oxfmt, a production build, and `next start` smoke checks of the new 404 negotiation and the unchanged HTML families.

## Review notes

- One review candidate claimed the predicate would Markdown-404 `/` and `/integrations` for agents; refuted — both are matched by package/consumer Markdown routes before the `markdownNotFound` fallthrough runs, and runtime checks confirm they return 200 Markdown.

### DIFF
--- a/apps/docs/lib/geistdocs/markdown-routes.test.ts
+++ b/apps/docs/lib/geistdocs/markdown-routes.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { markdownRoutes } from "./markdown-routes";
+import { markdownNotFound, markdownRoutes } from "./markdown-routes";
 
 const integrationsRoute = markdownRoutes.find(({ from }) => from === "/integrations/*path");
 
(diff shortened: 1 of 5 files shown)


molebox:
Closing as im not happy with the API from geistdocs side having thought about it some more

PR #2529, opened 2026-08-25
docs(research): propose per-subagent background execution

ruiconti:
### Summary

Related to #1084. The current task experiment chooses background execution for every subagent from one root flag. This proposal moves that choice to each local or remote subagent definition while keeping `experimental.tasks` as the root capability gate.

Before → after: one root-wide execution mode → mixed blocking and background definitions in the same model response, with execution pinned to each durable `agentId`.

The plan preserves legacy helper behavior for one compatibility epoch and keeps the built-in `agent` background under the task gate. It also specifies mixed provider settlement, unified `[Agents]` projection, fresh-start token fanout, dynamic selection, Workflow policy boundaries, and durable replay. Generic Workflow tool execution, authored background tools, task-mode graduation, and task waiting remain separate work.

### Validation

`pnpm docs:check` passes against current `main`; the proposal also completed multiple source-grounded independent reviews with no unresolved findings.

### Checklist

### Diff size

**Docs** — 1 file · `+261 / -0`

One research proposal defines the authoring API, runtime semantics, compatibility rules, delivery sequence, and required verification.

Not applicable.

Not applicable; this PR changes no runtime behavior.

### DIFF
--- a/research/subagent-background-execution.md
+++ b/research/subagent-background-execution.md
@@ -1,0 +1,261 @@
+---
+issue: https://github.com/vercel/eve/issues/1084
+status: draft
+last_updated: "2026-08-25"
+---
+
+# Per-subagent background execution
+
+## Summary
+
+Let each local or remote subagent definition choose background task execution instead of making
+`experimental.tasks` rewrite every subagent owned by an agent. Availability and execution remain
+separate: `defineDynamic` decides whether a subagent is exposed for a session or turn; the returned
+definition's `background` field decides whether calls return child output or a task receipt.
+
+Background subagents use the existing generic `defineTool({ execution: "background" })` path.
+They do not become a new runtime-action variant. Workflow/code programs execute those same tool
+definitions through the normal tool executor, with the tool's existing visibility, approval,
+authorization, and session capability rules. Code execution adds no separate permission gate.
+
+This is a follow-up redesign of the global behavior shipped for #1084. Definition-level execution
+is preferred over per-call model election because authors, not models, own whether work may outlive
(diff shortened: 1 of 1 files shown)


De-ASI-INTERFACE (approved):

ruiconti:
Implementation is now open independently from current `main`: #2596. It includes the public helpers, per-definition static/dynamic lowering, root capability gate, mixed dispatch and fanout, durable mode/target continuation checks, compatibility warnings, docs, changeset, and deterministic e2e coverage. Background execution inside `Workflow` remains scoped to #2542 because it needs the generic invocation and approval adapter.

ruiconti:
#2596 was closed after identifying a simpler implementation boundary in #2477.

Once workflow-backed tool `execute` bodies land, local and remote subagents can share one durable executor that waits on the child-result hook. Blocking calls wait for that executor's result; background calls admit it as a task and return a receipt. The replacement implementation will start fresh from `main` after #2477 rather than extending the legacy parked runtime-action path.

ruiconti:
Withdrawing this proposal with #2645.

The proposal preserves `experimental.tasks` as a root capability gate and adds per-subagent execution before the task runtime is generic enough to own it. That sequencing produced exactly the layering problem exposed in #2645: subagent-specific relay, cancellation, HITL, provenance, and settlement leaked into task abstractions.

The replacement direction is not another per-subagent implementation. Land the generic Workflow invocation/executor seams in #2542 and the other prerequisites named by #2537, then graduate task mode as a deletion-focused change. Per-definition subagent execution can consume that generic surface once it exists.

PR #2145, opened 2026-08-14
chore(eve): memory - cover file-provider recall

AndrewBarba:
### Summary

Related to #1510 and stacked on #2144. This adds fixture-owned end-to-end coverage for `fileMemory()`: a described personal-memory slot saves one exact fact through `user__save_memory`, and a distinct session receives it at turn start as user-role recalled context without a retrieval tool. A second turn in that session must still contain exactly one copy of the document, catching append-only implementations that duplicate unchanged recall. The same eval is registered for the live-model matrix and deterministic Postgres and Vercel worlds; the authored scope includes deployment identity so concurrent Preview deployments cannot share a memory document.

### Validation

The fixture's production build and TypeScript check pass locally. On the final pushed head, the eval passes with OpenAI, Anthropic, and ZAI models and in the deterministic Postgres and Vercel worlds, covering both cross-session persistence and unchanged-recall deduplication.

### Checklist

- [x] I linked an issue with prior discussion confirming this change is wanted
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

### Diff size

**Docs** — 0 files · `+0 / -0`

Not applicable.

**Implementation** — 1 file · `+19 / -0`

The workspace lockfile registers the new fixture package.

**Tests** — 10 files · `+159 / -0`

The self-contained fixture defines the agent, described scoped memory slot, deterministic behavior, persistence eval, duplicate-recall assertion, and deployment metadata needed for the E2E matrix.

PR #2144, opened 2026-08-14
feat(eve): memory - add bounded file provider

AndrewBarba:
### Status

M2 salvage branch for #1510. This draft is intentionally not reviewable or mergeable in its current form.

The derived-source prerequisite #2539 is merged, and the first-class memory core in #2534 has been rebased and simplified on current `main`. After #2534 merges, this branch will be rebased directly onto the then-current `main` and rebuilt against the shipped `eve/memory` provider contract. Do not rebase it onto the pre-M1 branch or preserve its current stack on #2142.

### What will be preserved

- The bounded document parser and serializer.
- Monotonically allocated, never-reused indexes.
- The conditional `MemoryDocumentBackend` contract.
- Explicit in-memory and Vercel Blob backends.
- Lazy environment selection, optimistic concurrency, limits, and focused provider/backend tests.

### What will be discarded during the rebase

- All compiler, discovery, manifest, runtime, harness, and memory-core changes.
- History scanning and public recall-attribution helpers.
- The pre-source-graph module-map and provider-tool lifecycle machinery.
- Any assumptions inherited from superseded #2142.

The rebuilt provider will recall one stable whole-document keyed message so M1 supersession removes stale content, including after the last entry is removed. File-provider E2E coverage from superseded #2145 will be recreated against the final API in this PR.

### Validation

The checks and deployments on the current branch are historical only. Validation, checklist status, diff accounting, docs, and the changeset will be rewritten after the post-#2534 rebase.

PR #2534, opened 2026-08-25
feat(eve): memory - add first-class provider lifecycle

AndrewBarba:
### Summary

Part of #1510 and implements the M1 core specified by the proposal merged in #1581.

This adds first-class, path-authored memory to eve. An agent can define `memory.ts` or named `memory/<slot>.ts` providers with nested recall handlers, optional nested capture handlers, and optional provider-owned tools. The lifecycle keys live under the operation they perform—`recall["turn.started"]`, `recall["compaction.completed"]`, `capture["compaction.requested"]`, and `capture["turn.completed"]`—so recall and capture remain distinct provider responsibilities without a phase discriminator or top-level event map.

Memory participates in the existing source graph, durable session history, dynamic-tool lifecycle, compaction, subagent compilation, and agent inspection surfaces. Each selected memory source instantiates its `tools/<slot>.ts` wrapper from a registered programmatic template anchored to that source. The wrapper inherits the anchor's node and layer, records the memory binding as a dependency, and yields to a direct tool in the same slot. There is no memory-specific source layer, registration mode, source-ID convention, or post-resolution tool merge.

Definition factories are materialized once per module-map load. That generic rule lets a direct slot and any derived slots share one definition/provider instance while keeping compile, generated, bundled, and hydrated module maps isolated from one another; it does not introduce a global provider registry.

### Behavior

- Discovers mutually exclusive flat and named memory slots, including subagent-owned slots, with path-derived identities and application override semantics.
- Exposes `defineMemory`, `defineMemoryProvider`, `defaultNamespace`, phase-specific lifecycle contexts, and the `byPrincipal` scope helper through `eve/memory` and `eve/memory/scope`.
- Resolves and locks namespace/scope before recall; `null` disables the slot for the operation.
- Attributes recalled user-role records in durable history, projects only records visible to the current locks, and supports keyed supersession and idempotent replay.
- Dispatches provider handlers by their nested lifecycle keys: recall at turn start and optionally after compaction; capture optionally before compaction and after a completed turn.
- Canonicalizes memory records outside ordinary summarization when compaction runs or the raw record log crosses its maintenance trigger.
- Exposes provider tools through the ordinary dynamic-tool lifecycle, including approvals, callback metadata, and cold-resume rebinding.
- Adds agent-info v4 memory inspection, published documentation, a changeset, unit/integration/scenario coverage, extension-contract fixtures, and a deterministic E2E fixture.

### Scope

This is the provider-neutral core only. It intentionally does not add storage or a bundled file provider. The completed `fileMemory()` follow-up is #2580, stacked directly on this branch; its old stack-locked PR record #2144 and the earlier core/E2E attempts #2142 and #2145 are superseded.

Extensions cannot contribute memory slots at their root because the consuming application owns scope and lifecycle state. Extension-owned subagents can define their own memory like any other declared subagent.

### Validation

- `pnpm typecheck`
- `pnpm docs:check`
- `pnpm lint` (three pre-existing warnings)
- `pnpm guard:invariants`
- Focused memory unit and authored-definition tests
- Focused compile-agent scenario (25 tests)
- Full unit run: 7,127 passed, one skipped; two unrelated five-second timeouts passed together on immediate focused rerun

CI is the official merge gate for pushed head `b15d8bbba`.

### Checklist

- [x] I added a changeset for the published `eve` package

### Diff size

**Docs** — 14 files · `+501 / -225`

Adds the memory guide, updates related concept and pattern docs, records the research contract, and includes the package changeset.

**Implementation** — 51 files · `+2,167 / -87`

Adds provider-neutral memory discovery, compilation, lifecycle, durable projection, compaction, inspection, and the registered provider-tool wrapper. Module maps share zero-argument definition-factory results within one load; provider tools otherwise use the existing dynamic-tool machinery.

**Tests** — 28 files · `+1,545 / -13`

Covers provider validation, lifecycle dispatch, projection, state, source precedence, generated/bundled/hydrated module maps, extension contracts, fresh-process compilation, cold rebinding, and a fixture-owned E2E eval.


packages/eve/extension-contracts/compatibility/subagent/v3.ts
```diff
@@ -0,0 +1,7 @@
+import { defineAgent } from "#public/index.js";
```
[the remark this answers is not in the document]
AndrewBarba: False positive: main has no compatibility/subagent/v2.ts fixture. #2506 made v3 current and explicitly dropped v1/v2. This PR advances the contract to v4 while retaining v3, so the new v3.ts compatibility fixture is required. No immutable history is deleted, and guard:invariants passes.

docs/patterns/multi-tenant-memory.md
```diff
@@ -1,157 +1,112 @@
...
+      };
+    },
 
-```ts title="agent/tools/list_memories.ts"
-import { defineTool } from "eve/tools";
-import { z } from "zod";
-import { memoryStore } from "../lib/memory-store";
-import { requireTenantCaller } from "../lib/tenant";
+    async capture(ctx) {
+      if (ctx.phase !== "turn.completed") return;
```
ctgowrie: consider explicit functions per phase would be more consistent with the rest of the framework

docs/memory.md
```diff
@@ -0,0 +1,259 @@
...
+The canonical in-session memory state is limited to 128 KiB by default. Raise
+or lower that bound independently of the model's token threshold:
+
+```ts title="agent/agent.ts"
+import { defineAgent } from "eve";
+
+export default defineAgent({
+  model: "anthropic/claude-opus-4.8",
+  compaction: {
+    memoryMaxBytes: 256 * 1024,
```
ctgowrie: This config feels eager to me. I don't quite understand when I'd want to use it

AndrewBarba: Already in the process of removing it

docs/patterns/multi-tenant-memory.md
AndrewBarba: Would you prefer this shape?

```ts
interface MemoryProvider {
  readonly recall: {
    readonly "turn.started": MemoryRecallHandler<MemoryTurnStartedContext>;
    readonly "compaction.completed"?: MemoryRecallHandler<MemoryCompactionCompletedContext>;
  };

  readonly capture?: {
    readonly "compaction.requested"?: MemoryCaptureHandler<MemoryCompactionRequestedContext>;
    readonly "turn.completed"?: MemoryCaptureHandler<MemoryTurnCompletedContext>;
  };

  readonly tools?: (context: MemoryToolsContext) =>
    MemoryToolSet | null | Promise<MemoryToolSet | null>;
}
```

I feel strongly that recall and capture should remain separate top-level concepts rather than becoming one `events` map. eve should be opinionated about when each operation runs, and recall has distinct output semantics because eve owns how its result is injected into context. The phase-specific handlers would remove `ctx.phase` branching and give each handler a narrowed context without erasing that boundary.

My main concern with this exact shape is that it makes the common case repetitive. Most providers should define one recall function that eve runs at both recall points and one capture function that eve runs at both capture points. This shape requires authors to put the same functions in two places and understand the lifecycle phases just to get the default behavior. Switching behavior based on phase should be a deliberate advanced case for a provider that knows why it needs it, not something every provider has to think about.

Does this match what you meant by explicit functions per phase? If so, do you prefer the explicit duplication here, or should we preserve single `recall` and `capture` defaults and expose phase-specific behavior only as an advanced override?

ctgowrie: Yes. I prefer the api with the event names since it's more consistent across the whole framework

ctgowrie (approved):

docs/patterns/multi-tenant-memory.md
ctgowrie:
```
interface MemoryProvider {
  readonly recall: {
    readonly "turn.started": MemoryRecallHandler<MemoryTurnStartedContext>;
    readonly "compaction.completed"?: MemoryRecallHandler<MemoryCompactionCompletedContext>;
  };

  readonly capture?: {
    readonly "compaction.requested"?: MemoryCaptureHandler<MemoryCompactionRequestedContext>;
    readonly "turn.completed"?: MemoryCaptureHandler<MemoryTurnCompletedContext>;
  };

  readonly tools?: (context: MemoryToolsContext) =>
    MemoryToolSet | null | Promise<MemoryToolSet | null>;
}
```
this shape

AndrewBarba: updated

AndrewBarba:
Implemented the nested lifecycle-key shape Casey and Andrew agreed on in `1f6f632c3`: required `recall["turn.started"]`, optional `recall["compaction.completed"]`, optional capture handlers under `capture`, and `tools` unchanged. The public phase discriminator, union recall/capture contexts, and old handler code paths are gone; runtime dispatch and authored-definition validation now use the nested keys directly. The file-provider follow-up is stacked on this exact API in #2580.

ctgowrie (approved):

PR #2537, opened 2026-08-25
docs(research): propose task-mode graduation

ruiconti:
### Summary

Related to #1084. Removing `experimental.tasks` is currently coupled to unresolved behavior in subagent execution, Workflow calls, authored background tools, task controls, remote callbacks, retention, and discovery. This proposal turns graduation into a deletion-focused final release by making each of those contracts an explicit prerequisite.

Before → after: one root flag selects task infrastructure and subagent mode → definition/session-owned execution, controls, callbacks, and discovery land first; graduation then removes the flag while preserving the built-in `agent` background behavior and a versioned legacy durable-state reader.

The plan defines callback negotiation before remote creation, lost-response replay, task retention/expiry, child finalization, effective session discovery, extension epoch classification, legacy helper removal, and an executable compatibility matrix. It adds no runtime behavior.

### Validation

`pnpm docs:check` passes on current `main`; the proposal completed repeated source-grounded review with no unresolved findings.

### Checklist

### Diff size

**Docs** — 1 file · `+224 / -0`

One research proposal defines prerequisites, externally observable graduation behavior, compatibility, deletion criteria, and verification.

Not applicable.

Not applicable; this PR changes no runtime behavior.

### DIFF
--- a/research/task-mode-graduation.md
+++ b/research/task-mode-graduation.md
@@ -1,0 +1,224 @@
+---
+issue: https://github.com/vercel/eve/issues/1084
+status: draft
+last_updated: "2026-08-25"
+---
+
+# Task-mode graduation
+
+## Summary
+
+Remove `experimental.tasks` after every behavior it currently selects has an ordinary definition-
+or session-owned authority. Graduation becomes a deletion-focused minor release, not the place to
+invent Workflow execution, task-control effects, callback negotiation, retention, dynamic metadata,
+or discovery contracts.
+
+Declared local/remote subagents use their definition-level `background` choice. The built-in
+`agent` remains background by default to preserve self-delegation receipts, persistent copies, and
+`task_update` behavior used by existing task agents. Blocking self-delegation is intentionally not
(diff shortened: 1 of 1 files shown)


ruiconti:
The #2645 review exposed one prerequisite that needs to be explicit before implementation: the task kernel itself must become executor-neutral before graduation.

Today `tasks/` still contains `SubagentTaskMetadata`, `SubagentExecutorData`, child-session/turn state, subagent input and authorization payloads, handle reconciliation, and hardcoded subagent cancellation. Graduation cannot merely remove the root flag around that design. Those concepts need to move behind a generic executor controller and executor-neutral progress/interaction/outcome wire; the subagent adapter should own agent IDs, local/remote transport, child provenance, answer delivery, and child cancellation.

The rejected alternative is #2645\047s shape: a framework subagent relay plus new `bind-relay`/`TaskRelayData`/subagent run-report variants in the task workflow. That adds a second subagent protocol instead of deleting the first one.

PR #2567, opened 2026-08-25
chore(eve): update Workflow SDK to 5.0.0-beta.44

VaguelySerious:
Updates eve's bundled Workflow SDK packages to the current 5.0.0 beta line.

| Package | From | To |
| --- | --- | --- |
| `@workflow/core` | beta.43 | beta.44 |
| `@workflow/errors` | beta.17 | beta.18 |
| `@workflow/utils` | beta.8 | beta.9 |
| `@workflow/world` | beta.28 | beta.29 |
| `@workflow/world-local` | beta.37 | beta.38 |
| `@workflow/world-vercel` | beta.39 | beta.40 |
| `@workflow/world-postgres` (catalog) | beta.35 | beta.36 |

Sibling versions are taken from what `@workflow/core@5.0.0-beta.44` itself depends on, so the lockfile keeps exactly one copy of each and the vendored bundle can't mix two.

## What comes in

- Sealed-log event identity, `specVersion` 7. The backend assigns event positions and closes dead ones with `noop` events. A run's version is fixed at creation, and `WORKFLOW_SEALED_LOG=0` is the kill switch.
- Draw order is pinned to event-log order rather than replay order (`WORKFLOW_LOG_ORDER_DRAWS=0` to opt out).
- `world.runs.waitForTerminalStatus`, and `await run.returnValue` now long-polls instead of spinning.
- `SerializationError` fails the step it happened in, so it is catchable.
- Bundled packages read their state off `globalThis`, and the queue handler comes from `getWorld()`.
- A wait that received an early continuation could sleep indefinitely; it now re-arms under a fresh idempotency key.

## Code change

`createDevelopmentWorkflowWorld()` advertised `specVersion: 6`. `@workflow/world@beta.29` moves `SPEC_VERSION_CURRENT` to 7, so `development-world-spec-version.drift.test.ts` fails until the literal matches. Bumped to 7, which is exactly what that test's docstring prescribes.

The other `specVersion: 6` occurrences are arbitrary fixtures in `validate-world.test.ts` and `development-world-server.integration.test.ts`, not claims about the current spec, so they are left alone.

## Validation

Everything CI runs, locally: `pnpm lint`, `pnpm check:deps`, `pnpm guard:fixtures`, `pnpm guard:invariants`, `pnpm docs:check`, `pnpm typecheck` (41/41), `pnpm test:unit` (7070 passed, 1 skipped), `pnpm build`, and `pnpm build:compiled` re-vendored against the new pins (stamp records core beta.44 / world beta.29 / world-vercel beta.40).

`pnpm test:integration` passed 5 of 6 runs. The one failure was the first run of the session: `workflow-entry.integration.test.ts > ignores stale and duplicate callbacks after a challenge is replaced`, timed out at 30s with cold-cache timings (152s setup, 30s import). It passed alone, passed the next 5 consecutive full-suite runs on this branch, and a pristine `origin/main` worktree ran the suite 4 more times green. I could not reproduce it on either side, so I can't call it pre-existing with confidence. Flagging it because that test is hook- and wait-heavy, which is directly in the path of the wait re-arm and sealed-log changes above.


AndrewBarba (approved):

AndrewBarba:
Superseded by #2570, which carries the Workflow beta.44 dependency alignment and adds the required scenario-test compatibility update for the new generated declaration text. Closing this PR in favor of the replacement.

PR #2570, opened 2026-08-25
chore(eve): workflow - upgrade to beta.44

AndrewBarba:
### Summary

Supersedes #2567. Upgrades eve's vendored Workflow SDK to beta.44 and aligns every `@workflow/*` sibling with the versions required by core, including the Postgres world used by e2e fixtures. The development World now advertises sealed-log spec version 7; no eve authoring API changes. Validation exposed timing-sensitive fixtures: task coverage relied on a five-second delay, while beta.44 roughly doubled sequential Vercel turn latency and exhausted the stress eval's old 10-minute budget. The task evals now synchronize through explicit approval and release state, and the 100-turn stress assertion keeps its full behavioral scope with a 15-minute safety deadline.

### Validation

After `pnpm build`, the unit suite passed 7,071 tests with 1 skipped, the integration suite passed 665 tests, and the scenario suite passed 415 tests with 15 skipped. The updated fixture builds, task transition validation, and full workspace typecheck pass locally. On CI, the task fixture passed all 17 evals in both Postgres (201 gates) and Vercel (210 gates), while the Vercel stress fixture completed all 100 sequential turns and passed both evals with 361 gates; all PR checks are green.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the release-facing changeset for the published `eve` package.

**Implementation** — 4 files · `+83 / -82`

Aligns the vendored Workflow dependency family and development World spec version. The lockfile accounts for most of this mechanical churn.

**Tests** — 5 files · `+85 / -19`

Updates the compiled-vendor scenario assertion, replaces wall-clock ordering in two task evals with protocol-controlled approval gates, and gives the unchanged 100-turn Vercel stress assertion enough time to complete under beta.44.


VaguelySerious:
(AI) CI triage from the duplicate bump I had open as #2567 (same pins), in case it saves you time. Two different conclusions for the two red suites.

**`e2e-postgres (fixture-tasks)` looks like a beta.44 regression, not flake.**

`E2E Tests (Postgres)` on the last 20 runs: green on every completed non-bump branch (`test/pending-approval-projection`, `rui/expose-channel-bot-identity`, `e0/f379cedd…`, `barba/openai-safety-identifier` ×4, `feat/tui-startup-card` ×2, `prha/selfmod-packagever`), and red on all three beta.44 runs: your job `97953874669` and my two, `97948130899` and the run before it. Identical signature in all three:

```
✗  task.agent.continue.rejected-agent-busy  gates 54/54
  Task task_… did not reach "completed" after 20 verification attempts.
```

All 54 gates pass, and the `AGENT_BUSY` rejection the eval is actually testing works. What fails is the wait afterwards: `e2e/fixtures/fixture-tasks/evals/shared.ts:170` polls 20 times with `t.sleep(100)` between attempts, so the task has 2s to move off `working`, and under postgres it no longer does. No world-level error in the log, no `Could not allocate an event slot`, no 503. The rest of the suite is fast (50.1s for 17 evals), so it is that task specifically, not a slow runner.

Worth a look before merge: `@workflow/world-postgres@5.0.0-beta.36` stamps `mintedSpecVersion()`, which is now 7 (sealed log) by default. `WORKFLOW_SEALED_LOG=0` on that lane drops minting back to 6 and would confirm or clear it in one run. `world-postgres` allocates each slot inside the INSERT that takes it (`SLOT_INSERT_MAX_ATTEMPTS = 40`, backoff after 8 immediate attempts), so it should be spec-7 compliant by construction with no sealing obligation, but the busy-agent eval is the highest-contention case in the suite and that allocator is new in the path.

**The three `e2e-vercel` lanes are pre-existing instability, safe to ignore.**

All three fail on unrelated branches concurrently, each on a different eval than mine:

| lane | branch | failing eval |
|---|---|---|
| fixture-tasks | `e0/f379cedd…` | `task.input.answer.accepted-complete.local` |
| fixture-tasks | `barba/openai-safety-identifier` | `rejected-agent-busy`, `calledTool(busy-worker)` found 0 |
| fixture-tasks | mine | `task.authorization.callback.accepted-current-attempt`, aborted due to timeout |
| agent-tools-sandbox | `feat/tui-startup-card` | `sandbox/redeploy`, `loadedSkill(deploy-note)` |
| agent-workflow-stress | `prha/selfmod-packagever` | `concurrent-workflow-turns` |

Mine were `sandbox/redeploy` (`loadedSkill(deploy-note)`, observed tools `[]`) and `sequential-workflow-turns` (22/22 gates, then aborted due to timeout). Rotating single evals per lane, no overlap, so nothing bump-specific.

**One nit on the scenario test.** `expect(indexDts).toContain("Core utilities intended for import by user")` will break again on the next upstream reword, which is exactly what it just did. Deriving it means the assertion never needs touching, and it checks the whole rewrite rather than one line, using the helper already in the file:

```ts
const upstreamIndexDts = await readFile(
  join(dirname(dirname(require.resolve("@workflow/core"))), "dist/index.d.ts"),
  "utf8",
);

expect(indexDts).toBe(
  rewriteDeclarationImports(upstreamIndexDts, {
    "@workflow/errors": "#compiled/@workflow/errors/index.js",
  }),
);
```

Same shape as the AI SDK assertion at line 402. `dirname(dirname(require.resolve(...)))` because `@workflow/core/package.json` is not an export.

Also: your `specVersion: 6 → 7` in `development-world-client.ts` is right and I had missed it. `SPEC_VERSION_CURRENT` is 7 in `@workflow/world@5.0.0-beta.29`.

VaguelySerious:
^ I'm having an agent dig into the potential postgres issues, sorry for the spam

PR #2580, opened 2026-08-26
feat(eve): memory - add bounded file provider

AndrewBarba:
### Summary

Part of #1510 and stacked directly on the first-class memory core in #2534.

This adds eve's bounded `fileMemory()` provider. Each authored memory slot stores one versioned, indexed document per resolved scope and exposes scope-bound `save_memory` and `remove_memory` tools. The provider registers recall at turn start and after compaction, returns one stable keyed whole-document message, and intentionally omits automatic capture or extraction.

The branch was rebuilt on #2534. It retains the file provider, storage contract, backends, concurrency handling, tests, docs, and E2E coverage; it does not carry any compiler, source-graph, runtime-lifecycle, or history-attribution implementation from the superseded memory stack.

### Behavior

- Persists normalized `index: text` entries under a versioned header with a `lastAllocatedIndex` high-water mark. Indexes are monotonic and never renumbered or reused, including after the highest live entry is removed.
- Recalls one stable keyed message containing the slot name, exact qualified removal tool, and current document. A missing document recalls nothing; an empty persisted document recalls an explicit empty state so the final removed entry cannot remain stale in context.
- Exposes `save_memory({ text })` and `remove_memory({ index })`. Both tools finish after a conditional write and return no model output. Duplicate saves and missing-index removals are no-ops.
- Enforces 2,048 UTF-8 bytes per normalized entry, 65,536 bytes per serialized document, and `maxEntries` live entries (100 by default). It rejects instead of truncating or evicting.
- Publishes a portable optimistic `MemoryDocumentBackend`, an isolated `inMemory()` backend, and a private Vercel Blob backend. Stale writes are retried against the latest document through `MemoryDocumentConflictError`.
- Selects the default backend lazily and fails closed: configured Vercel deployments use Blob, `eve dev` uses one shared process-local backend, and every other environment requires an explicit backend. `NODE_ENV` alone and Blob credentials outside Vercel do not alter that selection.
- Adds public package exports, user-facing memory documentation, a patch changeset, packaged portability coverage, a real workflow integration, and deterministic file-provider coverage in the existing memory E2E fixture.

### Scope

This is the bounded, model-maintained reference provider. It does not add semantic search, conversation extraction, provider-managed capture, silent eviction, or a general-purpose filesystem API. The Vercel Blob client slice is vendored into eve; it is not a new runtime dependency.

#2534 must merge first. Once it lands, this branch can be rebased from `barba/memory-core` onto `main` without changing the file-provider architecture. The separate E2E tail in #2145 and the earlier core in #2142 remain superseded.

### Validation

- `pnpm --filter agent-memory exec tsc --noEmit`
- `pnpm docs:check`
- `pnpm lint` (three pre-existing warnings)
- `pnpm guard:invariants`
- 32 focused provider/backend unit tests
- File-memory workflow integration test
- Packaged file-provider scenario test
- File-memory public API portability scenario test

E2E suites run only in CI; this PR adds the deterministic eval to `e2e/fixtures/agent-memory` for that gate.

### Checklist

- [x] I added a changeset for the published `eve` package


packages/eve/src/public/memory/file/backends/vercel-blob.ts
```diff
@@ -0,0 +1,88 @@
...
+    read,
+    async write(input) {
+      try {
+        const result = await put(pathname(prefix, input.key), input.content, {
+          ...credentials,
+          abortSignal: input.signal,
+          access: "private",
+          addRandomSuffix: false,
+          allowOverwrite: input.expectedVersion !== null,
+          cacheControlMaxAge: 60,
```
AndrewBarba:
```suggestion
          cacheControlMaxAge: 0,
```

ctgowrie (approved):

PR #2585, opened 2026-08-26
feat(eve): agents - support direct invocation

### LINKED ISSUE #393: Allow some setup for subagents to be called directly or as subagents

subagents should be able to be called either directly or as subagents 

### Proposed solution

this could be done either by supporting multiple root level agent directories and have subagents refer to them, or allow for direct calling of subagents

### Alternatives considered

_No response_

AndrewBarba:
### Summary

Closes #393. This adds one root-relative `agent` selector across the eve HTTP API, TypeScript client, fixed sessions, and channel sends so applications can invoke a statically declared descendant without asking the parent model to delegate.

Creating a session with `agent` makes that descendant the session default; setting it on an existing session selects the descendant for one turn and then returns to the original default:

```ts
const { session } = await client.sessions.create({
  agent: "researcher",
  message: "Investigate this report.",
});

await session.send("Critique the evidence.", {
  agent: "researcher/critic",
});
```

The same selector is available to channels, including `from(threadId).send(message, { agent })` and cross-channel sends, which supports adapters implementing commands such as `/researcher investigate this`. Selected turns keep the same session ID, continuation, auth, channel state, and model history while using the descendant's model, instructions, tools, skills, hooks, connections, and sandbox; HITL responses resume the selected descendant automatically. V1 intentionally accepts only entirely static local paths, returns stable errors for malformed, missing, dynamic, or remote targets, and emits ordinary session/channel events rather than synthetic subagent lifecycle events.

### Reviewer focus

The durable boundary now carries an internal node ID in versioned session-inbox and turn-workflow payloads. Buffered deliveries are partitioned by target, pending HITL retains the selected node, and sandbox snapshots are keyed by sandbox owner so alternating root and descendant turns do not overwrite one another; the session's original bundle remains its durable default.

### Validation

The eve unit suite passed with 7,123 tests and one skip. A built-app scenario covers shared history, targeted defaults, nested paths, target-specific instructions/tools/hooks/sandbox, HITL recovery, mixed-target queueing and steering, channel event flow, and rejection cases; the expanded `agent-subagents` E2E fixture typechecks, while its real-model eval will run only in CI.

### Checklist

### Diff size

**Docs** — 6 files · `+223 / -3`

Documents the HTTP, client, subagent, and channel authoring surfaces, records the implementation plan for #393, and adds release metadata.

**Implementation** — 51 files · `+741 / -76`

The selector crosses the public boundaries, durable workflow inputs, runtime bundle selection, and sandbox persistence. The implementation reuses one resolver and the existing session workflow; this category also includes the three generated extension-capability reports required by the additive API changes.

**Tests** — 23 files · `+1195 / -5`

Most of the test diff is the built-app scenario and E2E fixture needed to exercise routing across real HTTP, client, channel, HITL, queue/steer, and sandbox lifecycles. Unit coverage owns parsing, resolver errors, wire migration, operation identity, delivery partitioning, pending-agent recovery, and sandbox ownership.


docs/channels/eve.mdx
```diff
@@ -47,12 +47,34 @@ curl -X POST https://<deployment>/eve/v1/session \
...
 
+Set `agent` to start the session with a statically declared local descendant as its default agent. Paths are root-relative and may address nested descendants:
+
+```bash
+curl -X POST https://<deployment>/eve/v1/session \
+  -H "Content-Type: application/json" \
+  -d '{"agent":"researcher/critic","message":"Review this evidence."}'
+```
+
+Every later message without `agent` continues as `researcher/critic`. To select a descendant for only one turn in an existing session, include `agent` on that message:
```
ruiconti, line 58: This is non-obvious semantic that is indicative of coupling in the wrong place. It introduces implicit statefulness that it's not clear we want? 

- when you create a session targeting a subagent, will you never want to share with other agents? it seems like yes
- do you want an a) set of agent capabilities coordinated by an orchestrator or b) a shared session-bound storage that a set of independent agents operate against?

Issue #1287, opened 2026-07-28
expose a session handle so a remote agent can be steered


shoelessness:
### What problem are you trying to solve?

I'm building a gateway/orchestrator agent on eve — a chat front door that never does the work itself. It classifies an ask, authorizes it, dispatches it to an execution backend, and relays progress back to the humans in the thread. Backends are heterogeneous: Cursor Cloud Agents today, other eve agents alongside them, more later.

Every backend has to satisfy one executor contract:

```ts
interface ExecutorAdapter {
  launch(descriptor): Promise<RunRef>;
  steer(runId, message): Promise<{ delivered: boolean }>;   // send another turn
  status(runId): Promise<{ status, url?, ... }>;
  abort(runId): Promise<{ aborted: boolean }>;
}
```

Cursor Cloud Agents satisfy all four. **`defineRemoteAgent` satisfies only `launch`** — a remote call is one turn in, one result out, and no handle/id allows for follow-ups, so the other three verbs have nowhere to go.

```
                    ┌─────────────────────────────────┐
                    │   gateway agent  (eve)          │
                    │   classify · authorize ·        │
                    │   dispatch · relay              │
                    └──────┬───────────────────┬──────┘
                           │                   │
              same contract│                   │same contract
                           ▼                   ▼
        ┌──────────────────────────┐  ┌──────────────────────────┐
        │  Cursor Cloud Agent      │  │  eve agent               │
        │  (HTTP API + run ids)    │  │  (defineRemoteAgent)     │
        ├──────────────────────────┤  ├──────────────────────────┤
        │  launch   ✅             │  │  launch   ✅             │
        │  steer    ✅             │  │  steer    ❌  no handle  │
        │  status   ✅             │  │  status   ❌  no handle  │
        │  abort    ✅             │  │  abort    ❌  no handle  │
        └──────────────────────────┘  └──────────────────────────┘
                    the eve-native backend is the one
                    my gateway can't fully address
```

In the orchestrator code, I have a union whose second variant can't do three of the four verbs:

```ts
type ProfileEntry =
  | { kind: "cloud";    adapter: ExecutorAdapter }   // steerable, checkable, abortable
  | { kind: "subagent"; tool: string };              // fire, wait, done
```

Everything downstream special-cases the second kind, including prose telling the model to complete the call itself because authored code can't.

### Why this isn't the background-tasks story

[...]

PR #2586, opened 2026-08-26
docs(eve): research - propose steering for running subagents

EfeDurmaz16:
### Summary

Related to #786 and #867. Today a message to a busy subagent fails with `AGENT_BUSY`, and channel `"steer"` cancels the active turn and replays the message as a new turn, discarding in-flight work. This research doc proposes the missing primitive behind both: deliver a message into an active turn at its next safe step boundary, same turn, no cancellation.

Main decisions proposed:

- `TurnPolicy` becomes `"queue" | "steer" | "interrupt"`: steer gains its honest meaning (in-turn injection), today's cancel-and-replace behavior keeps existing under the name `"interrupt"`.
- Steering a subagent needs no new tool or schema: `agentId` + `message` to a `working` child becomes a steer routed through the task run, which referees the race against task settlement. Delivery reuses the existing parent-to-running-child lane that input responses already travel.
- Steering never changes task status; a task records the delegation, not the goal. No component classifies message intent.
- Every steer leaves a durable record in the parent conversation (tool call + receipt); fire-and-forget steering is ruled out.
- Direct child addressing stays out of scope, with a note on forward compatibility with the worker-handle ask in #1287.

Context: this responds to the public call for a champion on subagent steering (and the maintainer reply that continuation, background execution, and cancel-based steering are shipped, with in-turn steering as the remaining swing).

### Validation

Docs-only. `pnpm guard:invariants` passes locally (including the research frontmatter rule). No behavior changes.

### Checklist

Tests and a changeset are not applicable to a research document; the maintainer-approval box is left for a maintainer, since this proposal responds to a public call rather than an explicit request.

### Diff size

**Docs** — 1 file · `+167 / -0`

One research document; length matches the scope of the proposed contract (policies, semantics, delivery stages, alternatives).

Not applicable.

Not applicable.

### DIFF
--- a/research/subagent-steering.md
+++ b/research/subagent-steering.md
@@ -1,0 +1,167 @@
+---
+issue: https://github.com/vercel/eve/issues/786
+status: proposed
+last_updated: "2026-08-26"
+---
+
+# Steering running subagents
+
+Send a message to an agent that is already working, and have it take the
+message into account without losing the work in progress.
+
+In this document, **must**, **should**, and **may** are normative.
+
+## Problem
+
+Three surfaces refuse or mishandle a message to a busy agent today:
+
+| Surface | Today | Consequence |
+| --- | --- | --- |
+| Subagent tool, busy child | `AGENT_BUSY` tool error | The parent can only wait or cancel; redirection is impossible |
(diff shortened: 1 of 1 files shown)


EfeDurmaz16:
@AndrewBarba happy to implement this after your review, starting with the core injection. If parts are already in flight on your side, tell me where to plug in instead.

PR #2588, opened 2026-08-26
docs: propose multi-agent workspace subagents

OwenKephart:
### Summary

Related to #2043. Adds a research proposal for the next layer of agent workspaces: workspace-unique member names, build-time metadata, and `defineWorkspaceSubagents()` expansion into existing remote-subagent dispatch. Nested member paths organize glob selectors, while unique final directory names preserve the current service and route model. It also proposes map-returning dynamic subagents for multiple runtime-selected remote targets, with a durable resolver-node-plus-key identity shared by static workspace expansion. The proposal documents Vercel routing and explicit transport/forwarded-identity boundaries, plus an operator-owned target-resolution model for self-hosting.

### Validation

Reviewed the proposal against the current remote-agent implementation and the workspace deployment stack in #2043. The research document passed direct oxfmt validation; no runtime behavior changes in this PR.

### Checklist

No changeset is needed: this is a research-only proposal.

### Diff size

**Docs** — 1 file · `+479 / -0`

The research document records the combined workspace, dynamic-subagent-map, and workspace-subagent proposal, implementation boundaries, and validation plan for review before code changes begin.

Not applicable.

Not applicable.

### DIFF
--- a/research/multi-agent-workspaces.md
+++ b/research/multi-agent-workspaces.md
@@ -20,6 +20,7 @@ customer-operations/
 ├── pnpm-workspace.yaml
 └── agents/
     ├── foreman/
+    │   ├── package.json (optional)
     │   └── agent/
     │       ├── agent.ts
     │       └── subagents/
@@ -46,37 +47,34 @@ customer-operations/
 }
 ```
 
-A member's root `defineAgent` description is its workspace delegation
-description. The workspace compiler records the ID, description, and deployment
-locator for each member in a build-time catalog. It never asks a running agent
-for this information.
+When deploying an eve workspace to vercel using `eve deploy`, services config
+is automatically generated to allow each agent to be deployed behind separate
+routes.
 
-A member opts in to model-visible delegation candidates by authoring a normal
-subagent source that expands its workspace selection:
+Agents within are able to easily declare other members of the workspace as
+subagents. A member opts in by adding a file underneath their `subagents/`
+directory containing a `defineWorkspaceSubagents()`.
+
+Without arguments, this adds all other members of the workspace as subagents.
+Users can filter which agents are included using path filtering.
 
 ```ts title="agents/foreman/agent/subagents/workspace.ts"
 import { defineWorkspaceSubagents } from "eve";
 
 export default defineWorkspaceSubagents({
   include: ["agents/tasks/*", "agents/utilities/*"],
-  exclude: ["triage"],
+  exclude: ["agents/triage"],
 });
 ```
 
-`defineWorkspaceSubagents()` compiles the selected members into ordinary remote
-subagent entries. To the calling model they are named subagent tools; at runtime
+`defineWorkspaceSubagents()` functions by compiling the selected members into
+remote subagent entries. To the calling model, they are named subagent tools; at runtime
 they use the existing `defineRemoteAgent` execution protocol, including isolated
 child sessions, durable callbacks, cancellation and reset, output schemas,
-trace propagation, and optional forwarded caller identity. The workspace
-feature owns topology discovery and target resolution; it does not introduce a
-second subagent execution path.
+trace propagation, and optional forwarded caller identity.
 
-This plan builds on the agent-workspace deployment stack beginning with
-[PR #2043](https://github.com/vercel/eve/pull/2043). That stack establishes
-explicit workspace membership, optional member packages, workspace-root builds,
-and independently deployed member services. Workspace subagents extend that
-foundation; they do not replace independent `defineRemoteAgent()` declarations
-or make every multi-package repository an eve workspace.
+The description of the agent is sourced from the original agent's description field --
+it does not need to be duplicated in the consumer's code.
 
 ## Goals
 
@@ -145,8 +143,9 @@ monorepo marker.
 ### Names and paths
 
 A workspace member's final directory name is its **agent name**. Names must be
-unique across the entire workspace, including nested groups. The existing
-workspace deployment stack already uses this name for its public route and
+unique across the entire workspace, including nested groups.
+
+The existing workspace deployment stack already uses this name for its public route and
 service identity, so this proposal preserves that model:
 
 ```text
@@ -160,7 +159,7 @@ products/escalation                     escalation       /eve/agents/escalation/
 
 If two members would otherwise be named `review`, authors must choose distinct
 names such as `security-review` and `content-review`. This is an intentional
-workspace constraint, not an authored `name` field.
+workspace constraint, as it simplifies routing and identity logic considerably.
 
 The agent name is the shared identity in catalog entries, routes, generated
 tool names, diagnostics, and trace attributes. The member's workspace-relative
@@ -169,7 +168,7 @@ does not change its name, public route, or model-facing tool name.
 
 ### Descriptions and the catalog
 
-Every workspace member must author a non-empty root description:
+Every workspace member that is declared as a subagent must author a non-empty root description:
 
 ```ts title="agents/utilities/company-knowledge/agent/agent.ts"
 import { defineAgent } from "eve";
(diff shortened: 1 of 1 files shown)


research/multi-agent-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+directory containing a `defineWorkspaceSubagents()`.
+
+Without arguments, this adds all other members of the workspace as subagents.
+Users can filter which agents are included using path filtering.
+
+```ts title="agents/foreman/agent/subagents/workspace.ts"
+import { defineWorkspaceSubagents } from "eve";
+
+export default defineWorkspaceSubagents({
+  include: ["agents/tasks/*", "agents/utilities/*"],
```
ruiconti, line 65: Wouldn't this introduce a split into how a "subagent workspace" is defined? What happens to `package.json["eve"]`? Wouldn't want to have two ways to do the same thing

ruiconti: Or is the idea that `include` here is to add _additional_ filtering of the set returned by the `package.json["eve"]["agents"]` entry?

research/multi-agent-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+unique across the entire workspace, including nested groups.
+
+The existing workspace deployment stack already uses this name for its public route and
+service identity, so this proposal preserves that model:
+
+```text
+member directory                       agent name       public route
+─────────────────────────────────────  ───────────────  ──────────────────────────────────
+agents/foreman                          foreman          /eve/agents/foreman/eve/v1/*
+agents/tasks/triage                     triage           /eve/agents/triage/eve/v1/*
```
ruiconti, line 153: Can we anchor this to the `agent.ts` module instead of dir? It's unclear (though I wouldn't expect) whether e.g., `tasks` also have an `agent.ts`?

research/multi-agent-workspaces.md
ruiconti: @!e0 confirm this please. I'm assuming the latter after reading the doc, but the glob pattern made it a bit confusing e.g what is the pwd we're running `glob` from? is it the (pnpm) workspace root or the (agents) workspace root?

```suggestion
   include: ["tasks/*", "utilities/*"] // actually includes `$project_root/agents/tasks/*`
```

Also, not clear why not have all configuration in `package.json`? What do we gain from moving this to a separate `.ts` and splitting it?

research/multi-agent-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+  deployment.
+
+### Generated subagent names
+
+Each selected member's workspace-unique agent name becomes its model tool name:
+
+```text
+content-review       → content-review
+company-knowledge    → company-knowledge
+```
```
ruiconti: Haha

research/multi-agent-workspaces.md
```diff
@@ -0,0 +1,479 @@
...
+          ▼
+subagent tool registry → remote session dispatch → callback / continuation
+```
+
+The resulting tools have the normal remote-agent contract.
+
+`defineWorkspaceSubagents()` does not itself need to expose a URL, `auth`, or
+`headers` field. It describes a same-workspace target. Direct
+`defineRemoteAgent()` remains the explicit escape hatch for a different
+endpoint or transport policy.
```
ruiconti, line 333: Worth expanding here a bit on how the lowering will take advantage of this set-up being within the confines of the same service identity (or Vercel project).

research/multi-agent-workspaces.md
OwenKephart: @ruiconti I was being a bit lazy when writing up the example paths here which made things confusing, but yep the intention was for this to be additional filtering, and for paths to be relative to the project root.

So a bare `defineWorkspaceSubagents()` would mean "all agents defined by package.json["eve"]["agents"]", and if you defined a path in the `include` or `exclude` param for `defineWorkspaceSubagents()` that matched no agent in the workspace (e.g. `foo/bar/fake-agent`), then we'd probably want to error at compile time.

OwenKephart: The robot is correct here in terms of the reasoning for the split between package.json and the .ts files, but the heart of the issue is basically "each agent needs to be able to individually determine which other agents it can delegate to". The "foreman" agent might want to delegate to every single other agent, while other agents may only want to be able to delegate to some shared utility agents.

research/multi-agent-workspaces.md
OwenKephart: ok you caught me this one slipped through my anti-slop edit lol

research/multi-agent-workspaces.md
OwenKephart: correct that `tasks` would not have an `agent.ts` -- mental model is essentially that all agents in the workspace have the format

```
<some path>/
  <agent name>/
    package.json (optional)
    agent/
      agent.ts
      <etc.>
```

Typically, `<some path>` will be `<root>/agents/`, or `<root>/agents/<category>/`, but we don't enforce that specific structure, users are able to organize their directories however they want, as long as they have the correct metadata in their root `package.json`

research/multi-agent-workspaces.md
OwenKephart: added a little more detail here, the gist is basically "we know the exact target URLs and can guarantee that vercelOidc will work for auth".

separate from service identity specifically, it's also nice that we have access to the agent description, meaning we don't have to separately define descriptions for each remoteAgent wrapper of the underlying agents

research/multi-agent-workspaces.md
```diff
@@ -0,0 +1,444 @@
...
+customer-operations/
+├── package.json
+├── pnpm-workspace.yaml
+└── agents/
+    ├── foreman/
+    │   ├── package.json (optional)
+    │   └── agent/
+    │       ├── agent.ts
+    │       └── subagents/
+    │           └── workspace.ts
```
ruiconti, line 27: Still not a fan of the current authoring shape as it blurries the existing `subagents/[name]` named-subagent slot semantics splits workspace configuration between `package.json` and a `.ts` module. What do you think of `agent/workspace.ts` or within `agent.ts`?

ruiconti: FWIW I think we should be playing with these ideas as I can also appreciate the convenience of side-stepping the one-entity-per-file semantics e.g.`subagents/[internal-workspace].ts` for a collection of subagents; analogous to `tools/[disabled].ts` for a collection of `defineTools`.

But I think we can defer it

research/multi-agent-workspaces.md
ruiconti: How could we make this convention crispy clear?

- Local agent workspace is defined in `package.json`
- Specific agent enrollment defined in `workspace.ts`

research/multi-agent-workspaces.md
```diff
@@ -0,0 +1,444 @@
...
+          ▼
+workspace catalog + selector expansion
+          │
+          ▼
+compiled remote subagent entries
+          │
+          ▼
+subagent tool registry → remote session dispatch → callback / continuation
+```
+
```
ruiconti, line 321: @!e0 can you confirm that this semantics matches this proposal's? https://github.com/vercel/eve/pull/2612

research/multi-agent-workspaces.md
ruiconti: Wouldn't allowing this to cross package.json packages necessarily incur the risk of attempting to link cross-project agents? e.g this wouldn't work as a drop-in replacement in `internal-agents`

research/multi-agent-workspaces.md
OwenKephart: hm in my mind this isn't workspace configuration

the workspace is configured purely in `package.json`: that is definitive source of truth for what agents are considered part of the workspace, and no .ts files have any impact on that.

this file is just defining additional subagents that may be taken advantage of by a given agent, hence it going in `subagents/`. the `workspace.ts` filename is just an arbitrary choice, the user could instead call it `experts.ts` or `my-project-subagents.ts` etc. 

the file just conceptually contains sugar on top of a `defineDynamic` that expands into multiple subagents, e.g. you could achieve a similar result manually by having workspace.ts look like:

```
  events: {
    "session.started": () => ({
      triage: defineRemoteAgent({/* … */}),
      "company-knowledge": defineRemoteAgent({/* … */}),
    }),
  },
});
```

the subagent names are expanded with prefixes (same as tool calls), so the `subagents/[name]` pattern is still valid here. for workspace.ts, you would end up with `workspace__triage` and `workspace__company_knowledge` as the subagent names.

currently defineDynamic for subagents only supports a single subagent at a time, but that's a simple change to allow multiple from a single file

OwenKephart: oops commented before I saw the second comment here, but I do think that the current proposal adds less conceptual overhead than a new file type with new semantics

research/multi-agent-workspaces.md
OwenKephart: ah I didn't specify this clearly in the comment, but all packages (if specified) need to be part of the same overall workspace.  "A member may have a package.json, but its package manager workspace must claim that package; otherwise the root installation and member build cannot be made reliable."

definitely would not work as a drop-in replacement for `internal-agents`, but I don't think there's a way to accidentally mess this up in a way that doesn't result in clear errors

e.g. if I tried to use this in internal-agents, `defineWorkspaceSubagents()` would initially fail because the agent is not in an eve workspace. if I added eve metadata to the root package.json, it'd then complain that the subpackages are not part of the same package manager workspace

you wouldn't want to use this regardless in internal-agents regardless though, as it assumes we're deploying everything as a single project (which is not the right model for internal-agents)

research/multi-agent-workspaces.md
ruiconti: @!e0 can you build up a table with clear comparison axis? also interested in session ownership, what's the difference in your "local reference edges" and "defineRemoteAgent-style nodes" as to me these are equivalents.

Execution-wise I do believe that we want this PR's model - separate services.

research/multi-agent-workspaces.md
ruiconti: I get the distinction you're making but these statements are sort of contradicting:

> hm in my mind this isn't workspace configuration

> that is definitive source of truth for what agents are considered part of the workspace

In essence, think we can agree that `workspace.ts` is controlling how [the root agent] consumes the available local agent registry. And it's not clear to me why we'd want that to borrow from similar semantics as the named subagent slot (even though knowing that this is already "violatable" by `defineDynamic`).

In other words: Do we expect users to define >1 modules that define `defineWorkspaceSubagents`? What happens if I now have two modules e.g., `workspace-a` and `workspace-b` with conflicting specification?

To me, the agent-specific configuration (ie `workspace.ts`) should be granularly aligned to the workspace (`package.json`). Thus, we wouldn't want to have one workspace configuration yielding N agent-specific configurations.

research/multi-agent-workspaces.md
ruiconti: Agreed. This proposal's execution model is clearly superior for the problem.

PR #2542, opened 2026-08-25
docs(research): propose generic Workflow tool execution

ruiconti:
### Summary

Related to #1084. Workflow programs currently orchestrate only execute-less delegation runtime actions. Exposing ordinary/background tools by passing their `execute` functions directly would bypass approval, mishandle authorization pauses, leak raw outputs past `toModelOutput`, and violate background-task batch admission.

This proposal defines one eve-owned Workflow call adapter and canonical effective tool inventory. Blocking subagents retain runtime-action interrupts; execute-capable tools share model-loop visibility, approval, authorization, task, event, instrumentation, and disclosure policy. QuickJS receives a tagged model projection rather than raw output.

The plan also specifies heterogeneous interruptions, definition revisions that fail closed across source/code changes, concurrent background cohort admission, exact nested call IDs, independent resource limits, and durable continuation compatibility. It changes no runtime behavior.

### Validation

`pnpm docs:check` passes on current `main`; the proposal completed repeated source-grounded review with no unresolved findings.

### Checklist

- [ ] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

### Diff size

**Docs** — 1 file · `+262 / -0`

One proposal defines the security boundary, execution lifecycle, compatibility, delivery sequence, limits, and required coverage.

**Implementation** — 0 files · `+0 / -0`

Not applicable.

**Tests** — 0 files · `+0 / -0`

Not applicable; this PR changes no runtime behavior.

PR #2538, opened 2026-08-25
refactor(eve): add subagent execution compatibility substrate

ruiconti:
### Summary

Related to #1084. Per-subagent blocking/background selection needs durable execution metadata before the public helper API can safely change. Today task capability is read from whichever agent node is active, and existing handles/bindings do not expose a normalized execution or target contract.

Before → after: active nodes re-read `config.experimental.tasks`, and handle mode/target are implicit in phase/address → manifest v43 projects one root-owned task capability to every runtime node, new starts persist execution/target metadata, and legacy handles infer the same values from their existing durable shape.

This preserves current behavior, including the built-in `agent` remaining background under root `experimental.tasks`. Task executor bindings keep their established byte shape so old/new bind replay remains idempotent. No public subagent helpers or per-definition lowering ship here.

### Validation

The full unit tier passes 659 files and 7,066 tests (1 skipped). The runtime-action dispatch integration file passes all 27 tests, all 41 workspace typecheck tasks pass, and mechanical invariants pass. Independent review found no remaining issues.

### Checklist

- [ ] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

### Diff size

**Docs** — 1 file · `+5 / -0`

A patch changeset records the behavior-preserving compatibility substrate.

**Implementation** — 21 files · `+153 / -52`

Adds root capability projection, manifest validation, normalized durable handle metadata, and compatibility-preserving task-binding readers/writers.

**Tests** — 19 files · `+349 / -19`

Updates synthetic manifests and exact handle assertions, and covers nested capability propagation, legacy phase/address inference, and old/new binding replay.

PR #2596, opened 2026-08-26
feat(eve): add per-subagent background execution

ruiconti:
### Summary

Related to #1084 and #2529. `experimental.tasks` currently changes every subagent call on an agent from blocking to background, so authors cannot wait for a focused child while starting independent long-running work in the same response.

This adds `defineLocalSubagent` and `defineRemoteSubagent`, each with a per-definition `background` switch. `experimental.tasks` remains the root capability gate; the built-in `agent` and legacy subagent helpers keep their existing root-wide behavior for one compatibility epoch and emit migration diagnostics.

Before → after:

- Authored definition → compiler/runtime: the root task flag selected every subagent's mode → explicit definitions persist `blocking` or `background`, including dynamic selections.
- Model response → child dispatch: all subagents used one dispatch lifecycle → blocking calls park for child results while background calls return task receipts, with one shared local token-fanout denominator.
- `agentId` continuation → durable handle: execution and target kind were implicit → handles pin both and reject changed definitions with `AGENT_MISMATCH`, while task executor bindings retain their previous byte shape for replay compatibility.
- Agent context → model prompt: blocking and task agents rendered separately → one `<agents>` projection and per-tool guidance describe mixed execution truthfully.
- Nested task ownership → controls: task cancellation was root-only → every task-enabled local node receives `task_cancel`, while `task_update` remains visible only to task-owned sessions.

Background subagents run through the model tool loop. Workflow programs continue to expose only blocking subagents until the generic Workflow invocation and approval adapter proposed in #2542 exists; this change does not call background executors directly from the Workflow sandbox.

This supersedes #2538: the root capability propagation and durable mode/target pinning needed from that branch are included here without its standalone compatibility layer.

### Validation

The full unit tier passes 7,169 tests with one skipped. The affected dispatch, cancellation, HITL, hook, and task-execution integration cluster passes 34 tests; the complete runtime-action dispatch file passes 28 tests. The deterministic mixed-execution fixture builds, typechecks, and passes its local eval with all 3 gates. Documentation, TypeScript, lint, extension contracts, and mechanical invariants pass locally.

### Checklist

### Diff size

**Docs** — 9 files · `+137 / -89`

Updates the subagent, dynamic capability, remote agent, Workflow, agent config, getting-started, and API guidance, plus the release changeset. The docs also state the Workflow and cancellation boundaries that differ between blocking children and admitted tasks.

**Implementation** — 60 files · `+954 / -192`

The runtime portion crosses compiler artifacts, dynamic selection, prepared tools, dispatch, durable handles, prompts, and extension contracts. Most of this surface preserves replay and mixed-mode invariants: byte-compatible task bindings, root capability propagation, unified dispatch, shared token fanout, and mode/target-pinned continuations.

**Tests** — 36 files · `+979 / -44`

Covers static and dynamic local/remote lowering, root-gate failures, mixed dispatch and prompts, legacy persisted selections and handles, continuation mismatch precedence, shared fanout, extension compatibility, and deterministic fixture-level mixed execution.


ruiconti:
Closing this implementation because it targets the pre-#2477 execution architecture.

The public API and per-definition semantics remain the goal, but blocking and background subagents should share one workflow-backed `defineTool.execute` body:

- blocking ownership waits for the durable execute workflow's child-result hook;
- background ownership returns a task receipt while that same execute workflow waits and later settles the task.

That removes the need to preserve and extend a separate parked runtime-action path for blocking subagents. Rebuilding after #2477 lands should substantially reduce the dispatch, turn-driver, and compatibility plumbing in this PR.

The replacement will start fresh from current `main`, not rebase or continue this branch. #2538 remains superseded; #2542 remains the Workflow sandbox policy boundary.

PR #2601, opened 2026-08-26
fix(eve): compile authored imports from installed packages

prha:
### Summary

Authored package imports previously rejected published TypeScript entrypoints outside the monorepo; package-boundary compilation now allows declared installed-package source exports while preserving undeclared-package rejection.

`@eve/self-modification` declares just-bash as a runtime external:

 ```json
   "externalDependencies": ["just-bash"]
 ```

When eve compiled the package, it resolved just-bash to an absolute path under the build machine’s 
node_modules, then emitted that path into the generated runtime module.

 That was incorrect because:

 - the generated module could outlive or move away from the build environment;
 - an absolute node_modules path is not portable into deployment output;
 - it bypasses normal Node package resolution and package export conditions;
 - consequently, a packaged self-modification agent could build from workspace sources but fail when installed from npm or a tarball.

 ### What changes

 For configured external dependencies imported from an installed package, eve now
 preserves the authored package specifier:

 ```ts
   import "just-bash";
 ```

 instead of emitting something equivalent to:

 ```ts
   import "/build/path/node_modules/just-bash/...";
 ```

 At runtime, Node resolves just-bash from the deployed application’s dependency tree
 and applies its normal export conditions.

### Validation

Not rerun while packaging the stack; focused packed-package scenario coverage is included for the installed-source path.

### Checklist

### Diff size

**Docs** — 2 files · `+10 / -0`

Two changesets describe the observable compiler and packaged self-modification fixes.

**Implementation** — 1 file · `+16 / -1`

The package-boundary exception stays local to authored import resolution.

**Tests** — 1 file · `+1 / -0`

Extends the existing packed-package scenario to cover the source export.

### DIFF
--- a/.changeset/authored-installed-source-exports.md
+++ b/.changeset/authored-installed-source-exports.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Keep direct authored imports of configured external packages as package specifiers so installed packages with compiler-only TypeScript source exports load their runtime entrypoints correctly.
--- a/.changeset/selfmod-packed-source.md
+++ b/.changeset/selfmod-packed-source.md
@@ -1,0 +1,5 @@
+---
+"@eve/self-modification": patch
+---
+
+Include authored TypeScript entrypoints in the published package so installed self-modification scaffolds compile without workspace source, and clean generated scaffold output before each build.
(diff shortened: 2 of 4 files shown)


AndrewBarba (approved):

prha:
This in fact did get fixed by #2614

Issue #2425, opened 2026-08-22
world-local 30s hook timeouts redeliver and re-bill a long model call; process death mid-call does not always replay


os0xi:
## Summary

Two related world-local / workflow-sdk failures around a single in-flight model call:

1. The default 30s `WORKFLOW_LOCAL_BODY_TIMEOUT_MS` / `WORKFLOW_LOCAL_HEADERS_TIMEOUT_MS` on the world-local Undici hook redelivers a long Responses generation and OpenRouter (or any provider) bills twice for one visible answer.
2. A model call interrupted by process death does not always replay. Sometimes the stream stops at `step.started` after restart even though the run is re-enqueued.

Raising the timeouts to 600s **narrows** the double-bill window. It does not close it. A call that stays open past the timeout still redelivers and still re-bills, because the recovered inline step is not idempotent.

eve 0.44.0. Local `eve start` / world-local, `WORKFLOW_NODE_HTTP` unset.

## Repro 1 — 30s hook timeout double-bills a long generation

Installed world-local (`eve/dist/src/compiled/@workflow/world-local/index.js`) resolves:

```text
bodyTimeout: yn(`WORKFLOW_LOCAL_BODY_TIMEOUT_MS`, 3e4)
headersTimeout: yn(`WORKFLOW_LOCAL_HEADERS_TIMEOUT_MS`, 3e4)
```

`Cn()` builds `new mn.Agent(t)` from those values when `WORKFLOW_NODE_HTTP` is unset. The Node HTTP path reads the same pair.

A Responses stream that stays open past 30s trips the hook fetch. Core then logs:

```text
[world-local] Queue delivery failed at the transport (loop 2), retrying
[workflow-sdk] Re-executing inline steps owned by this queue message — a previous delivery crashed mid-body...
```

and re-executes the inline step. The provider completes both requests.

Forced both env vars to `30000` on an otherwise identical server. One generation produced two identical request bodies 35.263s apart and two charges:

| seq | cost |
|---|---|
| 2 | $0.031748 |
| 3 | $0.026858 |

One visible assistant completion. Two `message.received` executions.

Setting both to `600000` on the same code, a 59.734s generation made **one** request and incurred **one** `$0.015660` charge.

That is causal for the knob. It is not a fix for non-idempotent execution: any generation longer than the configured timeout still duplicates.

Tradeoff of 600s: a genuinely stuck hook now stalls ~10 minutes instead of ~30 seconds. Recorded, accepted.

[...]

PR #2609, opened 2026-08-26
fix(eve): workflow - prevent local delivery replays

AndrewBarba:
### Summary

Related to #2425. Supersedes the closed bot-authored draft #2427. Long local agent steps can exceed world-local's 30-second transport deadline, causing the same healthy queue delivery to be retried while its original provider or tool call remains active. This change defaults local delivery timeouts to unbounded before eve constructs either the parent-owned development world or the generated self-hosted world, while preserving explicit timeout overrides and leaving custom and Vercel worlds unchanged. A genuinely stuck local delivery will now remain in flight until the local server exits, which avoids automatically duplicating provider charges or tool side effects.

### Validation

The eve 0.45.0 reporter log reproduced the 30-second deadline plus 5-second backoff cadence. Focused unit coverage passed 7 tests, the parent-world integration suite passed 10 tests, and `pnpm test:unit` passed 7,148 tests with one skipped.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the required patch changeset describing the user-visible local delivery fix.

**Implementation** — 3 files · `+27 / -0`

Keeps the timeout policy in one eve-owned helper and applies it at both local-world construction boundaries.

**Tests** — 3 files · `+76 / -0`

Covers default and explicit timeout behavior, the development-world integration, and generated self-hosted world wiring.

### DIFF
--- a/.changeset/calm-local-turns.md
+++ b/.changeset/calm-local-turns.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Prevent long local Workflow deliveries from timing out and replaying an in-flight turn. Explicit local delivery timeout overrides continue to take precedence.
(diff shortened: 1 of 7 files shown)


ctgowrie (approved):

AndrewBarba:
Follow-up #2615 fixes the scenario-test regression exposed by this change. The existing recovery test had previously settled only because world-local's 30-second headers timeout eventually released its graceful Undici agent close. The follow-up actively cancels that process-owned transport during local-world shutdown and keeps the recovery assertion unchanged.

PR #2611, opened 2026-08-26
chore(eve): workflow - update beta dependencies

karthikscale3:
### Summary

Updates Eve's Workflow SDK packages to the latest aligned 5.0.0 beta releases, including `@workflow/core@5.0.0-beta.47` and `@workflow/builders@5.0.0-beta.47`.

Keeps Eve on Workflow spec version 7 and vendors the attributes-validation declarations required by the current beta family. In beta.47, durable hook resumption is provided by the public `resumeHook` API again, so the beta.46-specific `resumeHookDurable` call sites, mocks, re-export, standalone vendor entry, and bundler-warning suppression have been reverted.

### Validation

- `pnpm typecheck`
- `pnpm run build:js`
- Focused unit tests: 120 passed
- Session command inbox integration tests: 6 passed
- Compiled vendor asset scenario tests: 12 passed

### Checklist

- [x] I ran the relevant checks from CONTRIBUTING
- [x] I added a changeset if published eve
- [x] DCO: every rebased commit includes its matching author signoff

### Diff size

- Docs: 1 file — updates the patch changeset for beta.47.
- Implementation: 5 files — updates the aligned Workflow versions, lockfile, spec version, and vendored declarations.
- Tests: 1 file — updates the compiled declaration assertion for the current SDK text.


.changeset/update-workflow-beta-46.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": patch
+---
+
+Update eve's bundled Workflow SDK packages to the latest 5.0.0 beta releases. The development Workflow world now advertises spec version 7, and session inbox aliases preserve payloads committed during rekeying when Workflow retains the VM.
```
VaguelySerious:
```suggestion
Update eve's bundled Workflow SDK packages to the latest 5.0.0 beta releases, which includes performance and stability improvements.
```

pranaygp:
### Lockfile: confirmed fixed (independent verification)

I was asked to rebuild the stale lockfile blocking CI here. `26bd7d7b` ("fix: repair workflow postgres lockfile entry") landed while I was pushing, so I've verified it instead of duplicating it.

**Root cause.** Merging `main` (`716864b3`) left `pnpm-lock.yaml` internally inconsistent. The `e2e/fixtures/agent-pending-approval-projection` importer — new on `main` via #2569 — kept `main`'s resolved `@workflow/world-postgres@5.0.0-beta.35`, while the catalog and every other importer moved to `5.0.0-beta.38`. No snapshot entry existed for the stale resolution, so every job died at `pnpm install --frozen-lockfile`:

```
[ERR_PNPM_LOCKFILE_MISSING_DEPENDENCY] Broken lockfile: no entry for
'@workflow/world-postgres@5.0.0-beta.35(...)' in pnpm-lock.yaml
```

**Verification** (pnpm 11.21.0, Node 24, matching `.nvmrc`):

- I regenerated the lockfile independently with `pnpm install --lockfile-only` off `716864b3`. The result is **byte-identical** to `26bd7d7b` — so the hand-repair matches a true rebuild exactly.
- On the current head, `pnpm install --lockfile-only` is a **no-op** (empty diff, no `Could not find preferred package` warning). The lockfile is genuinely not stale.
- `pnpm install --frozen-lockfile` exits 0.
- `pnpm check:deps` (syncpack) — no issues.
- Intended bumps all preserved as resolved on disk: `core@5.0.0-beta.46`, `errors@beta.18`, `utils@beta.9`, `world@beta.31`, `world-local@beta.40`, `world-vercel@beta.42`, `world-postgres@beta.38` (catalog), `serde@beta.2` unchanged. No other resolution changed.
- `build:compiled` reports vendored modules already up to date, so the vendored `@workflow/core.mjs` matches the lockfile-resolved tree.
- The 6 unit test files this PR touches pass (102 tests), as does `compiled-vendor-assets.scenario` (11 tests).

Since the fix on the branch is byte-identical to my rebuild, I deliberately did **not** push a redundant commit — it would have been an empty diff that retriggers the full e2e matrix.

### Heads-up: `fixture-tasks` now fails in all three worlds

With install unblocked, e2e ran for the first time on this PR. `fixture-tasks` fails in **local, postgres, and vercel** — and it passed on other branches at the same time, so it looks specific to this PR rather than infra:

| World | Failing eval | Symptom |
| --- | --- | --- |
| local | `task.parent.wake.emitted-ready` | `Stream for session "wrun_…" closed before a turn boundary.` |
| postgres | `task.agent.continue.rejected-agent-busy` | `Task task_… did not reach "completed" after 20 verification attempts.` |
| vercel | `task.parent-interaction.send.accepted-live-children`, `task.parent.wake.emitted-ready` | same stream/turn-boundary shape |

`e2e-local (agent-subagents, openai-sol)` is also red.

The tell: **every gate passes** in each failing eval (54/54, 3/3) — it's the final settle that misses, i.e. the task never reaches a terminal state or the stream closes before a turn boundary. The failing eval rotates by world, which points at one systemic settling issue on the parent-wake / session-resume path rather than three separate bugs. That's the area this PR adapts for beta.46's retained-VM and lazy-resume behavior, so it's worth a look before merge.

I've left that one alone — it's behavioral work on your durable-resume changes, not lockfile staleness, and I didn't want to guess at the intended semantics. Everything else is green (106 checks).

AndrewBarba (approved):

PR #2612, opened 2026-08-26
docs(eve): propose local agent references

ruiconti:
Proposal doc only — no code. Adds `research/local-agents.md`, following the `research/` doc conventions.

## What

A public `defineLocalAgent` — the local counterpart of `defineRemoteAgent` — letting a subagent mount file reference another top-level agent in the same workspace and deployment, by importing its `agent.ts`:

```ts
// foreman/agent/subagents/reviewer.ts
import reviewer from "#agents/reviewer/agent/agent.ts";

export default defineLocalAgent(reviewer, {
  description: "Delegate code reviews here.",
});
```

The mount is an address, not a config carrier. The import is a statically traceable link (discovery resolves the specifier the way extension mounts do, never importing authored code) plus a type-level connection for editors; the value it binds is never read. Config, instructions, tools, sandbox, and nested subagents all come from the referenced directory, compiled once per deployment. Delegation dispatches in-process; the creating session owns the child session, unchanged from today. No per-agent `package.json` required — one `imports` map entry at the workspace root covers every agent.

## Why

The [software factory template](https://github.com/vercel-labs/eve-software-factory-template) is built on "one agent per task, each with its own prompts, context, and evals" — but the grammar can only express task agents as nested limbs of one root: no standalone entry point, no channels, no reuse across parents. This proposal makes the factory a workspace of peer agents with the orchestrator referencing them.

## How the semantics resolve

- Mount today: nested directory owned by one parent → mount under this proposal: reference edge to a top-level agent, one compiled graph per deployment, shared by all referencing parents.
- Cycles/diamonds: infinite-expansion and duplicate-copy problems under embedding → legal mutual recursion and shared graphs under references (same shape as remote loopbacks today).
- Root surfaces: channels/schedules are inert when referenced, active standalone — a runtime activation policy, not a grammar restriction, so one directory serves both roles.
- Session ownership: creator owns the child session, exactly as today; the reference only selects which agent graph defines behavior.

An earlier draft proposed embedding (compiling the sibling into the parent's tree); it survives in the doc as a rejected alternative, with the comparison table showing what references dissolve.

## The honest costs

Called out in the doc rather than glossed: multi-graph compiled manifests are the real lift; the mount must match a statically recognizable shape (inherited from extension mounts; non-resolvable mounts fail discovery loudly); referenced directories become build inputs (watching, fingerprinting); on platforms with a project root directory the sibling may not exist in the build container (must fail at build with an actionable error, not at runtime); secrets centralize in the referencing deployment; referenced agents gain multiple build-time consumers.

## Open questions for review

Specifier forms at v1, workspace boundary definition, whether eve should bless an `agents/` layout convention, opt-in channel activation (multi-agent deployments), workspace-aware `eve dev`, and naming.

## Example

`examples/factory/` is the [software factory template](https://github.com/vercel-labs/eve-software-factory-template) (Foreman + stations, the system behind ai-sdk-factory) restructured as a workspace of peer agents. The full diff against the template: each `agent/subagents/<station>/` becomes a top-level `agents/<station>/agent/` verbatim, five one-call mount files (`defineLocalAgent(station)`) appear under Foreman's `subagents/`, shared `lib/` moves to the workspace root behind `#lib/*`, and the classifier gains a station-level eval that runs without Foreman — the concrete payoff. Aspirational (marked as proposed API in-file); excluded from workspace globs and the root tsconfig so CI never compiles it.


ruiconti:
I sign off on the work in this pull request.

PR #2615, opened 2026-08-26
fix(eve): workflow - abort local deliveries on shutdown

AndrewBarba:
### Summary

Related to #2425. Follow-up to #2609. Unbounded local delivery deadlines exposed that world-local shuts down its undici agent with `close()`, which waits for an active request to drain; `eve dev` could therefore hang while crashing or stopping during a turn. eve's vendored world-local now destroys that agent on shutdown, aborting only the process-owned transport request while preserving the durable run for restart. The vendoring transform fails loudly if upstream's shutdown contract changes.

### Validation

Reproduced the CI failure in `dev-server-workflow-generations.scenario.test.ts`: it failed with the unbounded defaults and passed when the former 30-second timeout was restored. After the fix, all three workflow-generation scenarios pass with unbounded defaults, including the interrupted-turn recovery case.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the patch changeset for prompt shutdown of active local deliveries.

**Implementation** — 1 file · `+30 / -1`

Adds a guarded vendoring transform that changes only world-local's undici shutdown from graceful draining to active cancellation.

The existing workflow-generation scenario already covers the regression end to end.

### DIFF
--- a/.changeset/fast-local-shutdown.md
+++ b/.changeset/fast-local-shutdown.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Allow local Workflow worlds with unbounded deliveries to shut down promptly while a turn is still running.
--- a/packages/eve/scripts/vendor-compiled/@workflow/world-local.mjs
+++ b/packages/eve/scripts/vendor-compiled/@workflow/world-local.mjs
@@ -56,10 +56,39 @@ const workflowWorldLocalVersionPlugin = {
   },
 };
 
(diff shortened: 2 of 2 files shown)


AndrewBarba:
The durable fix belongs upstream in `@workflow/world-local`: [vercel/workflow#3824](https://github.com/vercel/workflow/pull/3824) wires the queue's existing shutdown signal into both delivery transports and adds a regression test that fails on Workflow `main`.

I recommend not merging this vendoring transform. Once the upstream patch is released, eve can replace this PR with a normal `@workflow/world-local` version bump.

PR #2625, opened 2026-08-26
fix(eve): surface turn version skew

benpankow:
### Summary

Resuming a parked session can run its next turn on a newer deployment. If that turn returns a control message or driver action the pinned driver does not understand, the driver can currently keep waiting with no acknowledgement to the user.

This makes those detectable compatibility breaks fail through the existing terminal session error path instead. Additive fields and newer compatible session-state versions remain supported, and this does not change the latest-deployment behavior from #2378.

### Validation

Adds regression coverage for unknown controls and actions, pre-turn migration failures, and the resulting terminal caller notification. Also covers preserving additive fields and compatible newer session-state versions without false positives. Built the package and the weather and TUI fixtures to exercise workflow bundling.

### Diff size

**Docs** — 1 file · `+5 / -0`

Patch changeset for the user-visible failure behavior.

**Implementation** — 4 files · `+84 / -8`

Keeps the checks at the existing turn control send/receive boundaries and input migration error path. The decoder is separate from the step module so it remains available inside workflow bundles.

**Tests** — 4 files · `+98 / -1`

Covers each detectable skew case, compatible state evolution, and the terminal session failure path.

### DIFF
--- a/packages/eve/src/execution/turn-control-codec.ts
+++ b/packages/eve/src/execution/turn-control-codec.ts
@@ -1,0 +1,63 @@
+import type { NextDriverAction } from "#execution/next-driver-action.js";
+import type { TurnControlPayload } from "#execution/turn-control-protocol.js";
+import { isObject } from "#shared/guards.js";
+
+/** Rejects control values a pinned driver cannot safely interpret. */
+export function decodeTurnControlPayload(value: unknown): TurnControlPayload {
+  if (!isObject(value)) return unsupportedControl(value);
+
+  const payload = value as TurnControlPayload;
+  switch (payload.kind) {
+    case "turn-error":
+    case "turn-continuation-token":
+    case "turn-delivery-request":
+    case "turn-delivery-accepted":
+    case "turn-delivery-cancelled":
+      return payload;
+    case "turn-result":
+      decodeDriverAction(payload.action);
+      return payload;
+    default: {
+      const unsupported: never = payload;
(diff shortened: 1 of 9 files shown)


packages/eve/src/execution/turn-control-codec.ts
```diff
@@ -0,0 +1,63 @@
+import type { NextDriverAction } from "#execution/next-driver-action.js";
+import type { TurnControlPayload } from "#execution/turn-control-protocol.js";
+import { isObject } from "#shared/guards.js";
+
+/** Rejects control values a pinned driver cannot safely interpret. */
+export function decodeTurnControlPayload(value: unknown): TurnControlPayload {
```
ruiconti, line 6: @!e0 my understanding is that this would only be beneficial if we were able to backport eve code to old deployments (ie before versioning). Can you outline real scenarios that `decodeTurnControlPayload` will materially change things granted that hooks in eve deployments past 0.40.0 will ensure proper versioning on _delivery_?

ruiconti: @!e0 ah, this is turn wire versioning, not session-inbox? Shouldn't we be extending session- versioning to turn instead?

ruiconti: @!e0 `TurnControlPayload` the sender of `TurnWorkflowInput` for the turn hook? If so, they should be using the same codec?

PR #2628, opened 2026-08-27
feat(eve): add trace capture decisions

chadhietala:
### Summary

The boolean `tracePolicy` can only drop or admit a trace, leaving content capture implicit. Extend it with an explicit `{ emit: false }` or `{ emit: true, recordInputs, recordOutputs }` decision while keeping boolean policies source- and behavior-compatible.

Eve normalizes the result once at session creation and carries that internal decision on the existing durable trace context across retries and child sessions. It is a capture ceiling for AI SDK and Eve OTel spans: each delivery's audience and each destination may narrow it further. Existing lifecycle providers remain active when a trace is dropped and retain their per-delivery audience projection, preserving boolean policy behavior. This adds no exporter or processor abstraction.

Related to #2335 and #2619.

### Validation

The full Eve unit suite passes 7,223 tests across 672 files with one existing skip. Focused coverage exercises legacy booleans, explicit emission and all four directional record combinations, lifecycle delivery under drop, mixed-audience delivery redaction, destination narrowing, child-session propagation, legacy seeds, directional event projection, AI SDK telemetry, and agent spans. Eve package typecheck, invariants, formatting, linting, and documentation validation pass.

### Checklist

### Diff size

**Docs** — 2 files · `+30 / -12`

Updates the existing research contract and adds the minor release note without documenting the flagged API in the published guide.

**Implementation** — 13 files · `+257 / -66`

Adds the public decision vocabulary, normalized capture state, durable parent/child propagation, and directional capture at the existing telemetry and provider boundaries while preserving lifecycle delivery.

**Tests** — 7 files · `+463 / -7`

Covers the complete decision matrix and the compatibility/privacy boundaries called out above, including event-type direction semantics and in-flight legacy seeds.

### DIFF
--- a/packages/eve/src/tracing/sampled-trace.test.ts
+++ b/packages/eve/src/tracing/sampled-trace.test.ts
@@ -6,13 +6,25 @@ describe("resolveTracePolicyDecision", () => {
   afterEach(() => vi.unstubAllEnvs());
 
   it.each([
-    { action: "drop" },
-    { action: "record", recordInputs: false, recordOutputs: false },
-    { action: "record", recordInputs: true, recordOutputs: false },
-    { action: "record", recordInputs: false, recordOutputs: true },
-    { action: "record", recordInputs: true, recordOutputs: true },
-  ] as const)("preserves the explicit $action decision", (decision) => {
-    expect(resolveTracePolicyDecision(decision, "private")).toEqual(decision);
+    [{ emit: false }, { action: "drop" }],
+    [
+      { emit: true, recordInputs: false, recordOutputs: false },
+      { action: "record", recordInputs: false, recordOutputs: false },
+    ],
+    [
+      { emit: true, recordInputs: true, recordOutputs: false },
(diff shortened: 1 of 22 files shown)


packages/eve/src/tracing/sampled-trace.test.ts
```diff
@@ -0,0 +1,53 @@
...
+
+describe("resolveTracePolicyDecision", () => {
+  afterEach(() => vi.unstubAllEnvs());
+
+  it.each([
+    ["drop", { action: "drop" }],
+    ["metadata", { action: "record", recordInputs: false, recordOutputs: false }],
+    ["inputs", { action: "record", recordInputs: true, recordOutputs: false }],
+    ["outputs", { action: "record", recordInputs: false, recordOutputs: true }],
+    ["content", { action: "record", recordInputs: true, recordOutputs: true }],
```
ctgowrie: A string feels like a worse description of behavior than the object produced here

ctgowrie (approved):

PR #2645, opened 2026-08-27
feat(eve): configure background subagents per definition

ruiconti:
### Summary

Related to #1084 and implements the proposal in #2529. This PR is stacked on #2477 and must merge after it.

`defineLocalSubagent` and `defineRemoteSubagent` now accept `background?: boolean`, so one agent can mix blocking and background subagents in the same model response. `background` defaults to `false`; background definitions require `experimental.tasks: true` on the root agent, while the built-in `agent` keeps its existing root-wide behavior.

Before -> after: the root task flag selected one execution mode for every subagent -> each declared local or remote edge carries its own durable execution policy.

Both modes use the stable subagent tool workflow introduced by #2477. The turn or task owner remains the only session writer: it starts or resumes the child, the workflow owns the child callback hook, and the workflow relays progress, input requests, and the exact terminal result to the existing owner inbox. This avoids a second background-only runtime path and gives cancellation one binding containing both relay and child coordinates.

### Behavior

- Blocking calls park the current turn until the child returns. Background calls return a task receipt and continue through the task lifecycle.
- Static and dynamic local and remote definitions support the same policy. Dynamic background selections are omitted when the root task capability is unavailable.
- Mixed blocking and background local starts share one fresh-start fanout denominator.
- Background definitions are not exposed inside the model-authored `Workflow` sandbox. Generic Workflow background support remains separate in #2542.
- The subagent extension capability advances to v5 while retaining the v4 contract, and the manifest advances to v44.

### Validation

The complete unit tier passes 7,124 tests and the complete integration tier passes 674 tests. Focused compiler and mounted-extension scenarios, affected fixture builds, and contract checks cover static and dynamic selection, mixed blocking/background execution, ordered relay settlement, replay ownership, cancellation, task tokens, named-child task controls, and extension compatibility. Fixture-owned evals will run in CI.

### Checklist

### Diff size

**Docs** — 9 files · `+99 / -77`

Updates the local, remote, dynamic, Workflow, cancellation, and TypeScript API guidance together with the release changeset so the per-definition default, task gate, and scope boundaries are explicit.

**Implementation** — 68 files · `+1862 / -551`

The compiler carries execution policy from authored definitions into the manifest and runtime graph. Execution converges both modes on one ordered, owner-released workflow relay, removes the superseded task dispatcher, and extends cancellation to the composite relay/child binding. This category also includes the generated v5 extension capability report.

**Tests** — 38 files · `+1503 / -482`

Covers static and dynamic definitions, nested root capability inheritance, mixed-mode lowering and fanout, blocking and background local and remote execution, relay-owned failure settlement, cancellation, compatibility, scenario compilation, and fixture-owned e2e paths.


ruiconti:
Closing this implementation rather than revising it in place.

The branch put subagent-specific relay, provenance, cancellation, and settlement concepts into the task kernel. That is the wrong boundary. Per-definition execution should consume generic background-tool/task infrastructure, and parity should remove the root-wide `experimental.tasks` mode and legacy subagent path rather than add another one.

#2477 supplies generic workflow-tool owner channels, but this stack still lacks the generic executor controller, executor-neutral interaction wire, task-control effect seam, Workflow invocation/cohort adapter, and durable execution metadata required to graduate tasks cleanly. Those prerequisites are already described in #2542 and #2537. Recreating them here under subagent names would repeat the same mistake.

The local branch now contains history-preserving reverts of both implementation commits and has no net code diff from the #2477 base. No replacement implementation will be pushed from this branch.

PR #2652, opened 2026-08-27
research: unify HITL request lifecycle

ruiconti:
Research doc and compiling prototype only. No runtime path is wired in this PR.

## What

Proposes one durable lifecycle for eve's existing HITL paths before adding a public HITL API.

The persisted model has three facts:

- **Request**: one durable question awaiting an outcome.
- **Group**: Requests created by one parked operation, its Owner, and completion-delivery state.
- **ResponseAttempt**: an Approval response waiting for internal Authorization, identified by `{requestId, deliveryId}`.

Approval, Question, and Limit are framework request kinds. Authorization is internal while checking an Approval response. Each kind owns only `resolve(request, input) -> verdict`; the interpreter owns ordering, stale responses, races, state transitions, forced closure, and Group completion.

## Before → after

| Flow | Today | Proposed |
| --- | --- | --- |
| Request state | pending batches, limit state, pending authorization, and approval-attempt state | Requests, Groups, and waiting ResponseAttempts in one store |
| Resolution | coordinator + domain router + three resolvers | one interpreter + one reducer per request kind |
| Stale response | separate stale-response conversion | retained terminal Request means `stale`; unknown id means `invalid` |
| Resume | withheld continuation can be lost or wedged across separate state | Group becomes `ready`, retries idempotent Owner delivery, then becomes `delivered` after acknowledgement |
| Approval authorization | separate Challenge/candidate state machine | internal Authorization Request linked to a waiting ResponseAttempt |

Owner lives once on Group, not on every Request. There is no separate tombstone type and no in-memory continuation claim.

## Why this shape

The authoritative cut must move request creation, persistence, resolution, migration, and Group completion together. Reconstructing temporary Requests from `PendingInputBatch` would leave another policy engine responsible for scheduling and closure. Dual writes would leave two authorities.

The implementation sequence is:

1. characterize current behavior and #1224 target transitions;
2. make the store, `createRequests`, interpreter, Approval/Question/Limit reducers, Groups, and migration authoritative together;
3. move ResponseAttempts, live approval policies, and internal Authorization into the interpreter;
4. migrate the framework approval gate as the final existing Owner;
5. only then build workflow-backed tools and public `ctx.request` / `ctx.auth`.

This reverses the earlier public-first sequence. The framework must prove and fully migrate to the model before users depend on it.

## Scope

This PR includes:

- the research doc;
- a strict-TypeScript prototype of the model and reducers;
- legacy-state migration mapping;
- late binding for dynamic approval policies;
- retryable `waiting -> ready -> delivered` Group completion;
- mapping to #1224's transition catalog.

It does not wire runtime behavior, change the public input wire, alter AI SDK approval presentation, change projector routes, or change scheduler admission.

Open decisions before implementation are terminal-Request retention, delivered-Group retention, Owner delivery idempotency/acknowledgement, async policy execution, and the exact framework-gate delivery path.


research/hitl-requests.md
```diff
@@ -0,0 +1,324 @@
...
+reducers** — and adds an owner axis that turns the same machine into a
+user-facing capability: tools that park on human input mid-task.
+
+## The factoring
+
+Two independent axes, entangled today:
+
+- **Variant** — _what is owed_: response-handling rules, outcome names,
+  supersession. Approval, Question, Limit, Challenge.
+- **Owner** — _who waits_: the parked session turn, the framework approval
```
benpankow, line 45: `resume_target` might be a little clearer here, since this really describes the piece of code that resumes once the approval completes, not necessarily the piece of code which initiated the wait (e.g. does it resume the turn directly, or route to approval gate etc)

research/hitl-requests.md
```diff
@@ -0,0 +1,324 @@
...
+  | { kind: "message"; actor: "originating" | "other" | "anonymous" }
+  | { kind: "callback"; params: JsonObject }
+  | { kind: "deadline" }
+  | { kind: "linked"; outcome: string }; // a row this one blocked on completed
+
+/** Every verdict a reducer can return — the closed set. */
+type Verdict<Outcome> =
+  | "ignore"
+  | { settle: Outcome }
+  | { reject: "unauthorized" | "invalid" | "policy-failed" | "candidate-cancelled" }
```
benpankow: somewhat unclear by name if this rejects the HITL _response_ or rejects/closes the HITL _request_

research/hitl-requests.md
```diff
@@ -0,0 +1,324 @@
...
+`DurableSessionSnapshot`. Free atomicity, but: no read path without
+hydrating a run, full-snapshot rewrite per step, unbounded journal growth
+from the withheld blobs.
+
+The ledger moves to a dedicated store (prototype
+[`store.ts`](./hitl-requests/store.ts)) — the same shape as
+`MemoryDocumentBackend`: `read` / conditional `write` / conflict error.
+Derived reads (`openRows`) are pure functions over the ledger, not store
+methods, so backends cannot diverge on query semantics. Default backend via
+nitro (`db0` locally, the deployment's database in production); the
```
benpankow: what does "the deployment's database" mean here? is this workflow state / session snapshot?

benpankow (commented):
directionally very aligned, couple inline qs/naming feedback

research/hitl-requests.md
ruiconti: > not necessarily the piece of code which initiated the wait

the code that initiated the wait should be the one owning its resumption i.e it is not expected for a session turn request to be resumed by a tool-body, or a framework gate.

@!e0 pressure-test the assumption above

PR #2691, opened 2026-08-28
fix(eve): deliver subagent.called events to hooks

### LINKED ISSUE #2223: subagent.called is in the hook event map but never delivered to hooks
### Version

eve 0.38.3 (also observed on 0.38.x earlier), self-hosted (`eve start`), Node 24.

### What happens

`subagent.called` is part of the hook event map — `HookEventMap` in
`dist/src/public/definitions/hook.d.ts` declares
`readonly "subagent.called": ProtocolEvent<"subagent.called">` — so a hook can
register a handler for it and the project typechecks. The handler is never
invoked.

Other events registered in the same `defineHook` call, in the same file, on the
same runs, are delivered normally: `message.received`, `actions.requested`,
`message.completed` and `turn.failed` all fire.

The event itself is not missing from the system. It appears in the message
stream (`createSubagentCalledEvent` in `dist/src/protocol/message.js`), the dev
TUI consumes it, and `docs/evals/assertions.mdx` documents asserting on it in
eval runs. It is specifically the hook delivery path that skips it — grepping
the published bundle, `dist/src/public/hooks/index.js` contains no reference to
`subagent` at all.

### Why it is easy to lose time on

The failure is silent in both directions. Nothing warns at startup that a
registered handler can never run, and the handler's absence looks exactly like
the subagent never having been called. In my case an orchestrator with five
subagent stations drove an external progress board from this hook; the board
simply never moved, while the run itself completed normally and the delivery
landed. The natural conclusion is that the orchestrator skipped its stations,
so the time goes into the prompt and the subagent definitions rather than into
the hook.

### Reproduce

```ts
// agent/hooks/probe.ts
import { defineHook } from "eve/hooks";

export default defineHook({
  events: {
    "actions.requested": (event) => {
      for (const action of event.data.actions) {
        if (action.kind === "subagent-call") console.log("REQUESTED", action.subagentName);
      }
    },
    "subagent.called": (event) => {
      console.log("CALLED", event.data.name); // never printed
    },
  },
});
```

Run any turn that calls a subagent. `REQUESTED` prints; `CALLED` does not.

### Expected

Either `subagent.called` is delivered to hooks, or it is removed from the hook
event map so that registering it is a type error. The current state — typed,
documented elsewhere, silently undeliverable — is the expensive one.

### Workaround, for anyone who lands here first

`actions.requested`, narrowed on the discriminant, carries the same information
at the moment the call is requested:

```ts
"actions.requested": async (event, ctx) => {
  for (const action of event.data.actions) {
    if (action.kind !== "subagent-call") continue;
    // action.subagentName is the station being called
  }
}
```

For progress reporting this is equivalent. It fires marginally earlier — at the
request rather than the dispatch — which for a status board is not a meaningful
difference.

NicoMoli:
### Summary

Closes #2223. Parent `subagent.called` events reached the stream but bypassed authored hooks in plain and task dispatch modes. The shared emitter now sends the same stamped event to the stream and hooks inside the framework context, while logging post-start failures to prevent child dispatch replay.

### Validation

Focused dispatch and hook lifecycle integration coverage passes (31 tests), along with eve package typecheck, formatting, lint, invariant, and docs checks. The full unit run reached 6,619 passing tests but could not complete because generated `#compiled` modules were unavailable locally and the sandbox blocked one port-binding test; CI will cover the complete suite.

### Checklist

### Diff size

**Docs** — 2 files · `+8 / -1`

Documents `subagent.called` delivery and its replay-safe failure behavior, with the required patch changeset.

**Implementation** — 3 files · `+66 / -44`

Keeps the lifecycle change in the shared emitter and threads provider-committed session state through both callers.

**Tests** — 1 file · `+114 / -24`

Uses a real context container to cover both dispatch modes, event identity, provider commits, and the throwing-hook replay boundary.


NicoMoli:
@ruiconti could you review this when you have a chance? This fixes the accepted P2 bug in #2223 and now includes the latest main. The focused integration suite is running without issues.
Thanks in advance!

PR #2697, opened 2026-08-28
fix(eve): isolate provider capture from OTel policy

chadhietala:
### Summary

OpenTelemetry `tracePolicy` was applied to the shared lifecycle bus, allowing an OTel decision to redact or suppress content requested by unrelated instrumentation providers. Provider delivery now depends only on each provider's `capture`, while eve's internal OTel provider independently projects events through the persisted trace decision and audience ceiling. Dropped eve traces retain metadata-only AI spans under the ambient Workflow trace, including sanitized failure metadata, without emitting `agent.session`; emitted traces continue to honor directional input/output capture. Persisted decisions are reused across replacement workers and pre-seed workflow replays rather than being widened by policy re-evaluation.

### Validation

The full eve unit suite passed with 7,298 tests across 679 files. Contract-level regression coverage co-registers the public `otel({ tracePolicy })` and `defineInstrumentation({ capture: "content" })` declarations for both emitted/redacted and dropped decisions, then verifies that the authored provider receives the full private model input. A real `@ai-sdk/otel` integration backed by `InMemorySpanExporter` verifies that metadata-only operation and tool failures contain neither sentinel error content nor raw status messages, and direct fallback tests cover stored, sampled, unsampled, and missing decision sources. A manual weather-agent run exported to local Jaeger confirmed that an emitted/redacted policy retained both `agent.session` and Workflow AI topology with no content keys, while `tracePolicy: () => false` retained metadata-only `invoke_agent` and `chat` spans under the Workflow root with no `agent.session`; neither marker reached Jaeger, and the custom content provider received every full model/tool payload in both runs.

### Jaeger proof

**Emitted, redacted `agent.session` trace.** The selected `chat` span retains provider, model, token, operation, and finish-reason metadata. There are no prompt or output sections.

Jaeger agent.session trace with a metadata-only chat span selected

**Ambient Workflow trace for the same emitted/redacted run.** AI and tool topology remains nested under `workflow.route.flow`, while the selected `chat` span exposes only metadata.

Jaeger Workflow trace with metadata-only AI spans

**Ambient Workflow trace with `tracePolicy: () => false`.** There is no `agent.session` trace. The selected `chat` span remains under the Workflow trace with model, timing, and token metadata but no input or output content.

Jaeger Workflow trace for a dropped eve trace with metadata-only AI spans

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds release metadata for the corrected trace-policy behavior.

**Implementation** — 16 files · `+309 / -157`

Moves content projection from the shared bus into the OTel provider, preserves decisions across durable boundaries, and protects metadata-only error paths.

**Tests** — 8 files · `+663 / -145`

Reworks the content-policy matrix around OTel-only projection and adds coverage for public declaration composition, provider independence, metadata-only integrations and compaction, real OTel error sanitization, fallback decisions, directional content, and persisted replay decisions.

### DIFF
--- a/.changeset/preserve-rejected-trace-bridge.md
+++ b/.changeset/preserve-rejected-trace-bridge.md
@@ -2,4 +2,4 @@
 "eve": patch
 ---
 
-Make instrumentation-provider content capture independent from OpenTelemetry `tracePolicy`. Rejected eve traces retain metadata-only AI spans in the ambient Workflow trace, while each provider receives only the metadata or content it requested.
+Instrumentation-provider content delivery now depends only on each provider's `capture` declaration: providers that declare `capture: "content"` receive full event content regardless of channel audience or OpenTelemetry `tracePolicy`. Separately, a `tracePolicy` that drops the eve trace no longer disables AI SDK telemetry: metadata-only AI spans (model, tokens, and timing, without message content) are still emitted into the ambient Workflow trace, and `agent.session` is not emitted.
--- a/.changeset/preserve-rejected-trace-bridge.md
+++ b/.changeset/preserve-rejected-trace-bridge.md
@@ -2,4 +2,4 @@
 "eve": patch
 ---
 
-Keep instrumentation lifecycle events and content flowing to provider integrations when `tracePolicy` rejects eve's trace, while limiting AI SDK OpenTelemetry spans in the ambient Workflow trace to metadata only.
+Make instrumentation-provider content capture independent from OpenTelemetry `tracePolicy`. Rejected eve traces retain metadata-only AI spans in the ambient Workflow trace, while each provider receives only the metadata or content it requested.
--- a/.changeset/preserve-rejected-trace-bridge.md
+++ b/.changeset/preserve-rejected-trace-bridge.md
@@ -2,4 +2,4 @@
 "eve": patch
 ---
 
-Keep instrumentation lifecycle events and content flowing to provider integrations when `tracePolicy` rejects eve's trace, without creating AI SDK OpenTelemetry spans in the ambient Workflow trace.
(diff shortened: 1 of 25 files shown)


chadhietala:
Addressed the follow-up review in `8c4bfc31d`:

- **Task 1:** Error sanitization and the explicit-integration-list gate now apply unless both effective input and output capture are approved. The `false/true` and `true/true` cases are pinned in the telemetry matrix.
- **Task 2:** Added a real `@ai-sdk/otel` integration backed by `InMemorySpanExporter`; failing operation and tool hooks verify `SECRET-ERROR-BODY` appears in no attributes, exception events, or status messages.
- **Task 3:** A failed eve-integration identity match now emits one warning and returns the registry unchanged; the requested unregistered/foreign case verifies it warns once.
- **Task 4:** Added direct tests for stored decisions, sampled legacy session state, unsampled legacy session state, and missing-state fail-closed behavior.
- **Task 5:** Expanded the changeset to state both provider-capture independence and dropped-trace ambient metadata behavior.
- **Task 6:** Documented the deterministic-policy requirement on `OtelOptions.tracePolicy` and noted the first-session double-evaluation window at both call sites without restructuring the flow.

One Task 4 request does not match the current contract: `InstrumentationTurnStartedEvent` has no `channelAudience` field, and neither does the protocol `turn.started` event. `audienceForEvent` therefore cannot read it; falling back to the persisted session audience is the intended path. Adding `channelAudience` would be a separate lifecycle API change rather than coverage for an existing arm.

Validation: `pnpm --filter eve typecheck`; full unit suite, 679 files and 7,298 passing tests (1 skipped); focused oxlint.

ctgowrie (approved):

PR #2711, opened 2026-08-28
[1/2] feat(eve): project task milestones into activity

benpankow:
### Summary

Related to #1673. Shows accepted `task_update` messages as the latest status for a background task.

The lifecycle is:

```text
task_update -> task workflow -> parent inbox (notifies/wakes the parent)
            -> action.result in the child session event log
            -> activity observer -> activity collector -> channel renderer
```

The parent notification and activity rendering are separate: activity can fail or be disabled without affecting delivery to the parent. Task settlement follows the same event-log rule, projecting terminal views from the task workflow's single-writer stream rather than posting to the collector from task code.

#2712 builds the authored-tool API on this state.

### Validation

Adds focused coverage for accepted task updates, event replay, updates arriving before work starts, terminal task views, and parent workflow forwarding.

### Diff size

**Docs** — 3 files · `+91 / -0`

Records the event-log rule, the implementation decision, and the release note.

**Implementation** — 8 files · `+203 / -36`

Moves task update and settlement rendering behind observers of the existing durable session and task streams. The invariant guard accounts for 31 added lines.

**Tests** — 6 files · `+218 / -44`

Replaces direct collector-submission tests with event projection and task-view observer coverage.

### DIFF
--- a/packages/eve/src/execution/tasks/child/steps.ts
+++ b/packages/eve/src/execution/tasks/child/steps.ts
@@ -6,12 +6,10 @@ import type {
   SubagentAuthorizationEventHookPayload,
   SubagentInputRequestHookPayload,
 } from "#channel/types.js";
-import { normalizeActivityText } from "#execution/activity-text.js";
-import { submitActivity } from "#execution/submit-activity.js";
+import { observeTaskActivity } from "#execution/task-activity-projection.js";
 import { isTaskWorkflowTargetGone } from "#execution/tasks/workflow-target.js";
 import { resumeSessionInbox } from "#execution/wire/session-inbox-resume.js";
 import { createLogger } from "#internal/logging.js";
-import type { ActivityEventV1 } from "#protocol/activity.js";
 import type { JsonValue } from "#shared/json.js";
 import {
   isTerminalTaskStatus,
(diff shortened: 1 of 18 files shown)


packages/eve/src/execution/tasks/child/steps.ts
```diff
@@ -155,12 +156,23 @@ export async function wakeTaskParentStep(input: {
...
 
+  void submitActivity({
+    events: projectTaskUpdateActivity({
+      activityObserver: input.activityObserver,
+      update: input.update,
+      updatedAt: new Date().toISOString(),
+      view: input.view,
+    }),
+    sink: input.activityObserver?.sink,
+  });
```
ruiconti: My existing mental model is that the only input for activity is the event stream ie it is single-writer. Thus, ad-hoc writes to it violates it and makes the model harder to reason about as consequently to maintain; see current HITL stack which allows unbounded writes to its pending batch.

benpankow: I updated `task_update` to emit an event; the observer now forwards this event rather than adjusting when we emit to the activity stream

ruiconti (approved):

benpankow:
Closing because task and tool activity rendering continued through the newer activity stack, including #2717, #2856, #2896, and #2916. If task milestone semantics are still missing after that stack lands, they should return as a focused follow-up.

PR #2712, opened 2026-08-28
[2/2] feat(eve): let tools report activity updates

benpankow:
### Summary

Related to #1673. Adds `ctx.activity.update(message)` for reporting progress from inside a long-running authored tool.

The lifecycle is:

```text
ctx.activity.update(message)
  -> action.updated in the session event log
  -> activity observer
  -> activity collector
  -> channel renderer
```

The update is durable and available to stream consumers and hooks, but remains presentation-only: it does not enter model context, wake a parent session, or change the tool result. An activity renderer failure also does not fail the tool.

Built on #2711. MCP progress remains a follow-up.

### Validation

Adds integration coverage through the real authored tool context, plus coverage for event emission, replay-safe projection, updates around action startup and settlement, and failure isolation.

### Diff size

**Docs** — 4 files · `+24 / -1`

Documents the tool API and the new `action.updated` stream event, plus the release note.

**Implementation** — 20 files · `+298 / -23`

Adds the tool emitter, stream event, activity projection, and context wiring. Generated extension metadata accounts for 121 added lines because extension callbacks share the event union.

**Tests** — 15 files · `+317 / -4`

Covers the public type and end-to-end emission path. Compatibility fixtures account for 97 added lines.

### DIFF
--- a/.changeset/calm-tool-activity.md
+++ b/.changeset/calm-tool-activity.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Allow authored tools to publish presentation-only progress through `ctx.activity.update()`. Updates flow to configured channel activity renderers without entering model context or waking a parent session.
--- a/.changeset/calm-tool-activity.md
+++ b/.changeset/calm-tool-activity.md
@@ -2,4 +2,4 @@
 "eve": patch
 ---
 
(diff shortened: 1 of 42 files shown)


benpankow:
Closing in favor of a smaller definition callback built on #2717. The follow-up will project typed action partials into bounded activity text without adding an imperative tool-context event path.

PR #2721, opened 2026-08-28
feat(eve): add provider trace policies

chadhietala:
### Summary

Instrumentation providers receive events about agent runs, such as model calls, tool calls, and session changes. This PR adds `tracePolicy`, which lets each provider decide whether to receive those events and whether inputs or outputs should include their full content.

By default, providers receive content from public channels and metadata only from private or unknown audiences. Existing `capture` settings continue to work, and an explicit `tracePolicy` takes precedence. Provider policies are independent from `otel({ tracePolicy })`, so disabling one provider does not affect other instrumentation. If a policy throws, only that provider is skipped and eve logs a warning.

### Validation

- Focused instrumentation, runtime, OTel, and tool-loop tests: 9 files, 386 passed
- `pnpm --filter eve test:unit`: 7,421 passed, 1 skipped; one unrelated TUI test exceeded the 5-second default and passed alone with a 15-second timeout
- `pnpm test:integration`: 93 files, 678 passed
- Compiled-artifacts bootstrap scenario: 1 file, 6 passed
- `pnpm --filter eve build`
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `oxfmt` on changed files

### Checklist

### Diff size

**Docs**: 3 files, `+47 / -17`

Updates internal research and release metadata. Published docs remain unchanged while the API is behind its experimental flag.

**Implementation**: 14 files, `+337 / -214`

Evaluates each provider policy through the centralized instrumentation runtime, removes content before delivery when it is not allowed, and keeps start and terminal events paired.

**Tests**: 12 files, `+490 / -41`

Covers public and private defaults, custom input and output rules, failing policies, existing `capture` behavior, OpenTelemetry independence, runtime binding, delivery completion, and compiled providers.

### DIFF
--- a/packages/eve/src/execution/settle-cancelled-turn-step.ts
+++ b/packages/eve/src/execution/settle-cancelled-turn-step.ts
@@ -73,7 +73,7 @@ export async function settleCancelledTurnStep(input: {
     turnAgent: effectiveAgent.turnAgent,
   });
   const instrumentation = bindSessionInstrumentation({
-    agentName: effectiveAgent.turnAgent.id,
+    agentName: bundle.graph.root.agent?.config?.name,
     ctx,
     rootSessionId: session.rootSessionId ?? session.sessionId,
     sessionId: session.sessionId,
--- a/.changeset/add-provider-trace-policy.md
+++ b/.changeset/add-provider-trace-policy.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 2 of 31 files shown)


packages/eve/src/execution/settle-cancelled-turn-step.ts
```diff
@@ -73,7 +73,7 @@ export async function settleCancelledTurnStep(input: {
     turnAgent: effectiveAgent.turnAgent,
   });
   const instrumentation = bindSessionInstrumentation({
-    agentName: effectiveAgent.turnAgent.id,
+    agentName: bundle.graph.root.agent?.config?.name,
```
ctgowrie, line 76: Why are these args optional?

ctgowrie (approved):

PR #2736, opened 2026-08-29
feat(eve): preserve trusted remote trace ceilings

chadhietala:
### Summary

Related to #2331. Remote agent hops previously lost the origin audience and upstream input/output capture limits, allowing downstream trace policy to evaluate content without constraints applied earlier in the trace. Sampled remote starts using `forwardPrincipal` now carry the immutable origin audience and current effective ceiling in W3C Baggage; receivers honor it only for callback requests with a valid sampled `traceparent` after route authentication and `trustedForwarders` accepts the calling deployment's forwarded-principal assertion. Each receiver intersects its own policy with the incoming ceiling and forwards only that result, so capture can narrow but never widen across remote or local-subagent hops.

Missing, malformed, duplicate, unsampled, and mixed-version assertions fall back to metadata-only capture; rejected principal forwarding retains its existing `403` behavior, and wire-level drop remains encoded only by unsampled trace flags. Origin policy stays separate from the live delivery audience: unknown or matching callback delivery uses the accepted decision, while a different explicit audience can restrict it further. Accepted policy reuses the existing durable session trace seed without a new top-level request field or context key, and Workflow content visibility remains fail-closed for directional ceilings.

### Validation

Coverage spans Baggage parsing and version skew, trust and sampling gates, durable local and remote propagation, directional redaction, malformed state, and three-hop monotonic narrowing. Full package unit and integration suites passed; after the final delivery-ceiling adjustment, 156 affected unit tests and 6 integration tests were rerun.

### Checklist

### Diff size

**Docs** — 4 files · `+42 / -6`

Updates the remote-agent and authorization guidance, records the security contract in research, and adds the patch changeset.

**Implementation** — 18 files · `+541 / -68`

Adds the Baggage protocol and trusted policy intersection, then carries the effective decision through existing dispatch, durable trace state, and instrumentation boundaries.

**Tests** — 14 files · `+1177 / -18`

The larger test diff covers the security matrix across protocol parsing, trust, persistence, redaction, relay behavior, and monotonic multi-hop narrowing.


.changeset/bright-spans-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": minor
```
AndrewBarba: **[P2] Use a patch changeset**

The repository policy says pre-1.0 features and fixes use `patch`; `minor` is reserved for breaking public API changes. This PR adds optional state and behavior without an identified breaking API, so this should be `patch`.

packages/eve/src/execution/remote-agent-dispatch.ts
```diff
@@ -166,6 +177,28 @@ export async function startRemoteAgentSession(input: {
   return { sessionId: parsed.data.sessionId };
 }
 
+function buildForwardedTraceAssertion(input: {
+  readonly forwardedPrincipal: ForwardedPrincipal | undefined;
+  readonly originAudience: ChannelAudience | undefined;
+  readonly parentTraceContext: SessionTraceContext | undefined;
+  readonly traceparent: string | undefined;
+}) {
+  const ceiling = decisionToTraceContentCeiling(input.parentTraceContext?.decision);
```
AndrewBarba, line 186: **[P1] Relay the current effective ceiling**

This serializes the persisted parent decision, not the decision narrowed for the current live delivery. If a session began public and is now private, the sender can emit `originAudience=private` with `ceiling=i1o1`; a receiver that explicitly admits private content can then record it even though the sender redacted the current delivery. Please apply the live-audience ceiling before constructing Baggage while preserving already-intersected forwarded decisions.

packages/eve/src/instrumentation/runtime.ts
```diff
@@ -316,7 +336,13 @@ export function bindInstrumentationRuntime(
         const audience = normalizeChannelAudience(channel?.metadata.audience);
         const capturesContent = shouldCaptureInstrumentationContent(audience);
         const effectiveDecision =
-          decision === undefined ? undefined : applyAudienceCeiling(decision, audience);
+          decision === undefined || sessionContext.forwardedTracePolicy !== undefined
+            ? decision
+            : applyAudienceCeiling(decision, audience);
```
AndrewBarba: **[P1] Reapply the live audience ceiling**

Every forwarded trace bypasses `applyAudienceCeiling` here. A trace that began public but later has an explicitly private delivery therefore keeps `recordInputs` and `recordOutputs` enabled, contrary to the documented hard ceiling. I reproduced this with a forwarded public seed plus a private live audience: telemetry returned `true/true` instead of `false/false`. Only an unknown or origin-matching audience should reuse the session decision; `setChannelContext` also needs to retain the live adapter audience rather than replacing it with the origin.

AndrewBarba (approved):
Reviewed the latest commits. The previously raised findings are addressed, and I found no remaining code-level blockers.

PR #2738, opened 2026-08-29
feat(eve): connections - add defineDynamic support

### LINKED ISSUE #1711: Allow session-scoped connection registration for multi-account SSO (e.g. AWS)

We're building an internal ops agent (Next.js + Eve) that connects to AWS via the managed AWS MCP after IAM Identity Center login. Each user can be assigned many AWS accounts dynamically from SSO (`ListAccountRoles` / account assignment APIs).

Today Eve MCP/OpenAPI connections are filesystem-discovered at build time (`agent/connections/<slug>.ts` → one connection name). That forces us to either:

1. Hardcode a static roster of accounts (one connection file each) and ignore SSO accounts outside that list, or  
2. Collapse everything into a single shared MCP connection and select the account only inside `getToken` (losing per-account connection names/descriptions for `connection_search`).

For multi-account cloud agents, the set of usable accounts is **caller-dependent and not known at `eve build`**. `defineDynamic` already covers tools, skills, instructions, subagents, and models — but not connections — so we can't expose "the accounts this user enabled" as first-class Eve connections without a PR/redeploy per account.

### Proposed solution

Support runtime connection resolution analogous to `defineDynamic` for tools, e.g.:

- `defineDynamic` (or equivalent) for MCP/OpenAPI connections resolved on `session.started` / `turn.started` from the authenticated principal, **or**
- An official pattern + APIs for "one MCP endpoint, many logical account connections" that still participate cleanly in `connection_search` (distinct descriptions / tool qualification per account).

Ideal sketch (illustrative):

```ts
// agent/connections/aws.ts
  events: {
    "session.started": async (_e, ctx) => {
      const accounts = await listEnabledAwsAccounts(ctx.session.auth.current);
      return Object.fromEntries(
        accounts.map((a) => [
          a.slug, // aws-acme-prod, etc.
          defineMcpClientConnection({
            url: AWS_MCP_URL,
            description: `AWS ${a.label} (${a.accountId})`,
            auth: {
              principalType: "user",
              getToken: ({ principal }) => mintAwsMcpToken(principal, a),
            },
            approval: always(),
          }),
        ]),
      );
    },
  },
});
```

Even without that exact API, clear guidance on whether dynamic connection registration is on the roadmap vs. "always use one connection + dynamic `getToken`" would help.

### Alternatives considered

1. **Static roster ∩ SSO** (what we do today) — works, but every new account needs a connection file + rebuild; can't offer user-specific account sets.  
2. **Single `aws` MCP connection + account selection in `getToken` / app store** — Eve-aligned with multi-tenant auth patterns; workable, but weak for model routing (`connection_search` sees one blob, not "prod vs stg").  
3. **Authored tools wrapping boto3** — bypasses MCP connections; duplicates AWS MCP / skills.  
4. **Dynamic tools that proxy MCP** — possible with `defineDynamic` tools, but reinventing connection auth, discovery, and tool schemas.

Happy to adjust to whatever design you prefer; main ask is a supported path for **principal-dependent connection sets** without rebuilds.

AndrewBarba:
### Summary

Closes #1711. Modules under `connections/` can now export `defineDynamic` and return caller-specific MCP/OpenAPI connections at session and turn boundaries while retaining ordinary discovery, approval, filtering, and qualified-tool behavior. Authenticated dynamic entries require a stable, non-secret `instanceKey`; eve hashes it with the resolved endpoint and source identity so token caches and parked OAuth callbacks cannot cross connection instances. Resolver failures abort the lifecycle without rebuilding the registry, preventing shadowed static connections from reappearing, and connection resolvers now receive only authenticated session identity plus `channel.kind` rather than model-controlled conversation or continuation data. The narrower resolver contract deliberately drops connection extension epoch 9; static connection behavior and the callback URL's path-derived name remain unchanged.

### Validation

The full unit suite passed 7,418 tests with one skipped, and the full integration suite passed 678 tests. Focused lifecycle and authorization coverage verifies fail-closed shadowing, exact-instance callback rejection, token-cache isolation, model-safe challenge redaction, and same-named tool/connection callback isolation; deterministic fixture evals cover dynamic OpenAPI discovery and model-visible dynamic MCP registration. Extension contract generation and invariant checks pass with tool epoch 21 retained and connection epoch 9 explicitly dropped.

### Checklist

### Diff size

**Docs** — 7 files · `+255 / -33`

Documents the public authoring API, restricted resolver context, instance identity contract, fail-closed semantics, issue-backed design, and release metadata.

**Implementation** — 33 files · `+844 / -82`

Adds compiler/runtime resolution, registry lifecycle integration, exact-instance authorization state, and generated extension capability metadata. The extension reports are required to make the intentional resolver-context compatibility break explicit.

**Tests** — 19 files · `+876 / -2`

Covers compiler and runtime return shapes, lifecycle precedence and failure modes, authorization pinning and cache isolation, durable reconstruction, retained extension compatibility, and deterministic OpenAPI/MCP end-to-end fixtures.


ctgowrie (commented):
Add e2e for dynamic MCP as well

ctgowrie:
1. Pin auth to the resolved connection instance

On resume, this reruns the session and turn resolvers before authorization completes. Because auth is still scoped only by `connectionName`, a parked OAuth callback for connection A can be passed to a newly resolved same-named connection B. Please pin an instance/auth identity when authorization starts and verify or reconstruct that exact instance before completion.

2. Do not reveal a shadowed static connection on resolver failure

A failed resolver becomes an empty dynamic result, after which the registry rebuild starts from static connections. If the dynamic result normally shadows a same-named static connection, failure silently restores the static connection—potentially with broader auth. Please fail the lifecycle or retain ownership tombstones so this path is actually fail-closed.

3. Remove model-controlled messages from connection resolver context

This passes the shared dynamic context, including conversation history, into code that selects endpoints, connectors, accounts, and headers. That makes model-directed routing and SSRF an easy foot gun. Dynamic connections should receive a narrower context with authenticated session identity and minimal channel metadata, but no messages, delivery, tool input, or model output.

ctgowrie (approved):

mrzmyr:
so exciting this is coming. 🥳

Issue #2747, opened 2026-08-30
Allow per-call subagent output schemas to narrow a declared schema floor


rcsandza:
## What problem are you trying to solve?

A declared subagent `outputSchema` currently acts as a fallback, not a floor. If a caller supplies any non-empty per-call `outputSchema`, eve replaces the declared schema with it. A loose or half-formed caller schema can therefore admit output that the subagent author explicitly rejected.

This is observable in `packages/eve/src/execution/subagent-tool.ts`, where dispatch selects `requestedOutputSchema ?? source.outputSchema`, and is pinned by the test “lets a per-call outputSchema override the local child's declared schema.” The generic resolution path likewise documents that a run-scoped schema always wins.

The concrete production use case is a sharded extraction worker. Its declared schema supplies a stable artifact-envelope floor: required wrapper and fields, closed objects, status enum, and text bounds. A per-dispatch caller knows the assigned ids and shard size and should be able to add an id enum plus exact array cardinality. Today that caller must repeat the entire declared floor correctly. If it emits only the shard-specific constraints—or an accidentally loose schema—the declared safety and context bounds disappear.

Omitting the caller schema is not an acceptable dependency for tightening: model-generated subagent arguments may normalize to `{}` under strict tool calling. The declared schema must remain sufficient by itself, while a successfully supplied caller schema should only narrow it.

## Proposed solution

When a declared local subagent schema and a non-empty per-call schema are both present, resolve their logical intersection instead of letting the per-call schema replace the declaration.

Desired semantics:

- neither schema → no output schema
- declared only → declared schema unchanged
- caller only → caller schema unchanged
- both → output must satisfy both schemas
- empty `{}` caller input keeps its existing “not requested” normalization

An `allOf: [declared, requested]` wrapper is the clearest JSON Schema representation if every structured-output backend eve supports accepts it. If the provider-facing subset does not reliably support `allOf`, a deliberately limited structural intersection should fail closed on unsupported or contradictory combinations rather than falling back to caller-wins behavior.

This could be global subagent semantics or an explicit authoring policy if maintainers need to preserve replacement for existing callers. The important guarantee is that an agent author can declare a non-bypassable floor and allow per-call narrowing.

## Suggested implementation prompt

Implement a declared-output-schema floor for local subagents.

Requirements:

- Preserve declared-only, caller-only, and empty-caller behavior.
- When both declared and requested schemas exist under the floor/narrowing policy, enforce their logical intersection without mutating either input.
[...]

PR #2748, opened 2026-08-30
fix(eve): preserve declared subagent output schema floors

rcsandza:
### Summary

Related to #2747. A per-call subagent output schema currently replaces the agent's declared schema, so a loose or malformed caller can silently remove the declared validation floor. Fresh local and remote subagent dispatches now fail closed on unsupported schema constructs and structurally intersect supported caller constraints with the declared schema; direct runs keep their existing override semantics. An incompatible local caller returns a per-call `SUBAGENT_START_FAILED` without creating a child or ownership record.

### Validation

Regression coverage first reproduced local and remote caller replacement, including extraction-shaped malformed schemas, then proved a weak caller cannot broaden a closed object or bypass nested validation. The focused schema suite passes 65 tests, the full unit suite passes 7,404 tests with one existing skip, and the full integration suite passes 679 tests, including the no-child/no-handle failure path.

### Checklist

### Diff size

**Docs** — 2 files · `+7 / -0`

Documents the narrowing rule and carries the patch release metadata.

**Implementation** — 5 files · `+564 / -37`

The recursive intersection is deliberately explicit and fail-closed: every supported keyword is validated and flattened for provider portability, while unsupported composition or property-evaluation constructs are rejected instead of falling through to permissive conversion. The local start adjustment keeps those expected rejections inside the existing per-call failure boundary.

**Tests** — 4 files · `+414 / -6`

Covers local and remote dispatch, incompatible constraints, closed-object behavior, nested bypass attempts, malformed schemas, the no-caller fallback, and failure before child ownership.

### DIFF
--- a/.changeset/firm-schema-floors.md
+++ b/.changeset/firm-schema-floors.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Make per-call subagent output schemas narrow declared output schemas instead of replacing them, so caller constraints cannot weaken an agent's declared structured-output contract.
--- a/docs/subagents/index.mdx
+++ b/docs/subagents/index.mdx
@@ -149,6 +149,8 @@ The root built-in `agent` tool is the exception. Its children share the root's s
 
 eve lowers every subagent visible to the current agent (the root built-in copy, declared, or [remote](./guides/remote-agents)) into a model-visible tool with the same `{ message, agentId?, outputSchema? }` shape. The parent packs `message` with everything the child needs, since the child never sees the parent's history. Set `outputSchema` to require structured output for that turn; the child remains available for follow-up messages afterward.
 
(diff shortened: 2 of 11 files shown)


rcsandza:
Addressed the automated error-handling suggestion in d505c6fdc: local schema-resolution failures are now caught before ownership is prepared and return a per-call `SUBAGENT_START_FAILED`; the integration regression proves no child session and no handle-store entry are created. Independent Class-R re-review found no unresolved Critical, High, or Medium issue, and the exact head passes 7,404 unit tests plus 679 integration tests.

rcsandza:
Closing after ownership clarification. This change belongs to an application-specific integration we control, not to Vercel Eve upstream. We will retain Eve’s current caller-wins behavior as an explicit external limitation and enforce the available shard and reconciliation controls in HomebaseOS instead. Thanks to the automated review for the useful error-handling finding.

PR #2763, opened 2026-08-31
feat(eve): add batch setup planner

OwenKephart:
### Summary

This reworks the `/add` flow to allow users to select several integrations at once, then review their selections before being put into the actual installation flow. Each step in the flow will be independently cancelable / preserve error information.

For now, instead of the previous breakout into several different categories, I've just broken things down into "Channels" and "Integrations". Users who are new to the product may not initially appreciate the differences between all of these different categories and forcing them to choose early is a bit aggressive.

Once selfmod / memory are fully landed, we can add another tab for "Evolution" or something that encompasses both of those categories

Screenshot 2026-08-31 at 2 51 24 PM
. 

Screenshot 2026-08-31 at 2 51 34 PM

### Validation

- Focused registry and TUI unit tests: 251 passing
- `pnpm lint`
- `pnpm --filter eve typecheck` (on the stack tip)
- `pnpm docs:check` (on the stack tip)
- `pnpm guard:invariants` (on the stack tip)

### Checklist

### DIFF
--- a/.changeset/gentle-onboarding-flow.md
+++ b/.changeset/gentle-onboarding-flow.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Replace the Dev TUI's single-item `/add` browser with a channel and integration planner that installs ordered batches, preserves completed results and failures, and installs explicitly addressed items directly.
--- a/apps/docs/registry.json
+++ b/apps/docs/registry.json
@@ -261,7 +261,15 @@
             "command": "eve",
             "package": "eve",
             "bin": "eve",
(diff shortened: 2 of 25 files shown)


benpankow (commented):
General UX navigation comments:
- "Space" for toggle makes some sense, but it is not a key I typically think about when dealing with TUI (e.g. `pi` multiselect uses enter)
- "Enter" to advance has potential of being very confusing here, e.g. I hover "slack", hit "enter", and it jumps to the next screen. I assume I picked slack, but in actuality I've selected nothing.

I wonder if using only the arrow keys is discoverable enough?

benpankow (approved):

PR #2802, opened 2026-08-31
fix(eve): harden legacy inbox caller projection

benpankow:
### Summary

Hardens encoding for persistent subagent inboxes created by older eve versions. Instead of removing the known `activityObserver` field ad hoc, legacy targets now project caller metadata through a stripping schema derived from the immutable v1 caller schema, preventing current or future caller fields from leaking into v0/v1 payloads.

### Validation

Adds a unit matrix covering both legacy v0 variants and v1 with omitted and populated activity observers, plus a simulated future caller field. The original encoder rejects the undefined case and would reject the future field; the schema projection emits only v1 caller fields.

### Diff size

**Docs** — 1 file · `+5 / -0`

Patch changeset for the published `eve` package.

**Implementation** — 1 file · `+15 / -9`

Replaces the field-specific downgrade shim with a projection derived from the existing v1 schema.

**Tests** — 1 file · `+44 / -0`

Focused matrix for all legacy targets and both current and prospective caller fields.

### DIFF
--- a/packages/eve/src/execution/wire/session-inbox-encoder.ts
+++ b/packages/eve/src/execution/wire/session-inbox-encoder.ts
@@ -2,10 +2,12 @@ import type {
   DeliverHookPayload,
   SessionCommand,
   SessionTimeoutHookPayload,
+  TurnCaller,
 } from "#channel/types.js";
 import {
   encodeSessionCommandV1,
   type SessionInboxWireV1,
+  sessionInboxWireV1Schema,
 } from "#execution/wire/session-inbox-wire.v1.js";
 import {
   encodeSessionCommandV2,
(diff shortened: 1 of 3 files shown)


packages/eve/src/execution/wire/session-inbox-encoder.ts
```diff
@@ -52,10 +54,7 @@ function encode(
   target: SessionInboxWireTarget,
 ): SessionInboxWireV1 | SessionInboxWireV2 | Record<string, unknown> {
   if (target.version === 0) {
-    return encodeSessionCommandV0(
-      encodeSessionCommandV1(withoutActivityObserver(command)),
-      target.variant,
-    );
+    return encodeSessionCommandV0(encodeSessionCommandV1(toV1Command(command)), target.variant);
```
ruiconti, line 57: Can we move toV1Command to inside encodeSessionCommandV1? That way we can later add guards that prevents changes to this module

ruiconti:
Superseded by #2840.

PR #2804, opened 2026-08-31
[2/2] test(eve): cover undefined fields across wire targets

benpankow:
### Summary

Adds a general session-inbox encoder invariant: explicitly `undefined` current fields must encode identically to omitted fields for every supported wire target. The matrix derives versioned targets from `SESSION_INBOX_WIRE_VERSIONS` and also includes both legacy v0 variants, so future wire versions join the test automatically.

Depends on #2802.

### Validation

The test fails for all three legacy targets without #2802 and passes for the complete target matrix with the fix.

### Diff size

Not applicable.

Not applicable.

**Tests** — 1 file · `+35 / -0`

One table-driven invariant covers both v0 variants and all declared versioned targets.

### DIFF
--- a/.changeset/fix-legacy-subagent-inboxes.md
+++ b/.changeset/fix-legacy-subagent-inboxes.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Fix deliveries to persistent subagent inboxes created by older eve versions when no activity observer is configured.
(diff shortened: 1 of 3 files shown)


benpankow:
Closing in favor of keeping the focused regression test with #2802.

PR #2815, opened 2026-09-01
test(eve): characterize HITL request lifecycle

ruiconti:
## What

Add one conformance suite for the existing Approval, Question, and Limit lifecycle before replacing its implementation.

The suite locks down the cross-kind behavior that is currently spread across the pending-input routers:

- a mixed Question/Approval batch stays parked until its Approval is answered;
- settling that Approval dismisses an unanswered Question in the same batch;
- independent groups remain independently answerable;
- duplicate responses in one delivery use the last response;
- an open Limit prompt takes precedence over responses for other groups.

Before → after: these behaviors were covered indirectly across domain-specific tests → they now have one executable boundary suite that can run unchanged against the request interpreter.

This PR changes no runtime behavior or persisted state. It is the first implementation increment after #2652 and is the base for the authoritative request-ledger PR.

### DIFF
--- a/packages/eve/src/harness/hitl/request-lifecycle.conformance.test.ts
+++ b/packages/eve/src/harness/hitl/request-lifecycle.conformance.test.ts
@@ -1,0 +1,188 @@
+import type { ModelMessage } from "ai";
+import { describe, expect, it } from "vitest";
+
+import {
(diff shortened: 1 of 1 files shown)


ruiconti:

PR #2816, opened 2026-09-01
refactor(eve): store HITL requests in one ledger

ruiconti:
## What

Replace persisted pending-input batches with one versioned Request ledger and route Approval, Question, and Limit deliveries through one interpreter.

Before → after: pending batches and a separate stale-response pass each interpreted part of the lifecycle → retained Requests and Groups are the persisted authority, and the HITL interpreter owns delivery ordering, kind routing, duplicate-response canonicalization, and stale/invalid identity classification.

The ledger:

- stores one session-turn owner per Group;
- enforces request-id uniqueness across retained Requests;
- retains terminal Requests and delivered Groups for stale classification;
- uses conditional writes with a dedicated conflict error;
- imports legacy singleton and collection batch state, then removes those keys on first mutation.

Approval, Question, and Limit resolution remain kind-specific reducers under the interpreter. Existing batch APIs now project open Groups for compatibility with call sites; they no longer own persisted state.

Backend selection stays out of this change. The ledger lives in the existing durable session state.

## Stack

Base: #2815. Approval response attempts and internal Authorization follow in the next stacked PR.


ruiconti:

PR #2821, opened 2026-09-01
refactor(eve): interpret approval authorization requests

ruiconti:
## What

Move Approval response attempts and internal Authorization under the HITL interpreter.

Before → after: approval candidates, pending Authorization, and their coordinator were separate durable state machines → response attempts and Authorization challenges now persist beside Requests and Groups in the versioned ledger, and the interpreter-owned approval-response pass advances them.

The two-phase crash boundary remains:

1. ingest and commit the responder-bound attempt;
2. on a later pass, run the live response policy;
3. persist an Authorization challenge when required;
4. consume its matching callback and re-feed the held Approval response.

Competing responders, duplicate delivery, cancellation, policy failure, timeout, and audit events retain their existing behavior. Legacy approval and Authorization keys are read once and removed on the next mutation.

The old `approval-candidates.ts` and `approval-delivery-coordinator.ts` modules are gone. Their tests moved with the interpreter-owned implementation.

## Stack

Base: #2816. The framework approval-gate owner follows in PR 4.


ruiconti:

PR #2825, opened 2026-09-01
fix(eve): require exact Zod peer version

muratsu:
### Summary

eve vendors Zod for schema conversion but previously allowed consumer projects to provide live schemas from any Zod version, which can cross incompatible private conversion contexts and crash before a model call. This makes Zod a required exact peer sourced from the same catalog entry as the vendored bundle, so incompatible installs are caught during dependency resolution. No runtime schema behavior changes for projects already on Zod 4.5.4.

### Validation

The focused version-token suite passes all 6 tests, including new coverage that the Zod catalog entry is exact and eve's peer consumes it. The eve package typecheck and dependency-policy check also pass.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Patch changeset describing the new install-time compatibility contract.

**Implementation** — 1 file · `+2 / -1`

Declares Zod as a required catalog-backed peer without changing runtime code.

**Tests** — 1 file · `+10 / -0`

Regression coverage keeps the peer declaration and exact vendored catalog version aligned.

### DIFF
--- a/.changeset/tidy-zod-peers.md
+++ b/.changeset/tidy-zod-peers.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Require projects to install the exact Zod version used by eve so live schemas cannot cross incompatible Zod conversion contexts.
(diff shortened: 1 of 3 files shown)


AndrewBarba:
This should not be required. zod is an implementation detail of eve and end users should be able to use whatever validation library they want, that includes a different zod version.

PR #2833, opened 2026-09-01
fix(eve): performance - accelerate session inbox resumes

AndrewBarba:
### Summary

Related to #876. This is PR 1 of 3 in the turn-performance stack.

Fixed-session delivery currently reads and decrypts hook metadata before every command, even though every stable-inbox consumer accepts eve's historical `send` envelope. Compatible fixed-session commands now use that envelope and resume by token, avoiding the metadata lookup while allowing Workflow to select its optimized durable-write and queue-publish path.

Continuation aliases still negotiate the persisted consumer version and resume the exact inspected Hook. Activity-observer deliveries retain that path, and undefined observer keys are stripped before legacy encoding. This layer deliberately preserves child workflows and deployment routing so its hosted result isolates session ingress.

### Hosted timing

Isolated Vercel Workflow stress run: [attempt 2](https://github.com/vercel/eve/actions/runs/33524878948/attempts/2), artifact `workflow-stress-performance`.

| Scenario | Samples | Mean | p50 | p95 |
| --- | ---: | ---: | ---: | ---: |
| Sequential turns | 100 | 2.058s | 2.049s | 2.483s |
| Concurrent first turn | 50 | 4.674s | 4.467s | 5.523s |
| Concurrent second turn | 50 | 1.773s | 1.558s | 2.995s |

The first ten warm turns had a 1.727s p50 and the last ten had a 2.166s p50. This layer does not materially remove eve's approximately two-second sequential turn floor.

The isolated artifact predates the final undefined-observer correction. The stress fixture does not send a delegated caller, so that correction does not touch the measured path. Exact-head Vercel and Postgres E2E matrices passed, and the local prompt-cache retry also passed.

### Validation

- `pnpm guard:invariants`
- `pnpm --filter eve typecheck`
- Focused session-inbox wire suites (28 passed)
- Hosted Workflow stress fixture (100 sequential turns and 50 two-turn concurrent sessions)
- Exact-head Vercel and Postgres E2E matrices

CI note: the required `test-scenario` job remains red because `dev-server-drain.scenario.test.ts` twice observed a new client port across worker replacement ([PR retry](https://github.com/vercel/eve/actions/runs/33529625808/attempts/2)). The same assertion is failing on current `main` ([main run](https://github.com/vercel/eve/actions/runs/33527178967)); this PR does not touch dev-server draining.

### Checklist

- [x] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
[...]

Issue #876, opened 2026-07-16
Hosted Workflow: ~3.5 s of dispatch/orchestration between channel webhook and the model step, on a warm instance


Lukavyi:
## Summary

On a production Vercel deployment (hosted Workflow world), **~3.5 seconds elapse between the channel webhook arriving and the model step starting**, on a warm instance with no cold start involved. For an interactive chat agent this is a fixed floor added to every turn before the first token can possibly appear — native draft streaming or client streaming cannot mask it, because nothing is being generated yet.

This is distinct from #455 (intermittent ~300 s visibility-timeout stalls — this is the *steady-state* cost) and from #859 (late delivery *after* the turn completes — this is the segment *before* the turn does any work).

## Environment

- eve **0.22.3**, hosted Workflow world, Node 24.x
- Telegram channel (`telegramChannel`), production deployment
- Warm instance: an earlier turn ran one minute before; cold start is not a factor

## Measured turn (single-word message, 26 output tokens)

Agent Run `wrun_01KXP73WMB0Q8W9XYZQSJ8N5ED`, project `prj_aDUXbr43e4bApMx9x4LzXS6Chc5f`:

| event | t | Δ from webhook |
|---|---|---|
| `POST /eve/v1/telegram` (webhook, from runtime logs) | 19:40:33.x | 0 |
| `run_created` | 19:40:34.896 | **~1.5–1.9 s** |
| `run_started` | 19:40:35.657 | ~2.6 s |
| step 1 created → completed | 35.702 → 35.830 | |
| hooks created | 35.987 → 36.393 | |
| step 2 (model) created → **started** | 36.518 → **36.548** | **~3.5 s** |
| step 2 completed (reply ready) | 36.994 | ~4.0 s |

So the model work itself is ~0.4 s of a ~4 s wall-clock to "reply ready" — the other ~3.5 s is queue dispatch to `workflowEntry` (~1.5–2 s) plus the step-1/hooks/step-2 orchestration hops (~1.5 s).

The same shape shows on every turn we sampled across two projects/agents in this team, not just this run.

## Impact

A resident-process agent (same model, same account, long-polling Telegram) answers in ~1.5–2 s to first token; the eve/Vercel deployment of the same agent cannot go below ~3.5 s + model TTFT for structural reasons. That gap is the whole difference users notice.

## Questions / asks

[...]

PR #2834, opened 2026-09-01
fix(eve): performance - route turns to accepted deployments

AndrewBarba:
### Summary

Related to #876. This is PR 2 of 3, stacked on #2833.

Production sessions currently resolve `deploymentId: "latest"` before every per-turn child workflow, even when the Nitro request and the pinned session driver are already executing on that exact deployment. This change stamps the accepting Vercel deployment onto framework-owned channel delivery metadata, carries it through a new append-only session inbox wire version, and starts the child directly on that exact deployment when the dispatch step observes a trusted match.

Missing, legacy, mixed, and cross-deployment deliveries retain the existing latest-deployment behavior. The stable raw-send path from #2833 carries the additive provenance field without inspecting hook metadata; older consumers ignore it, while versioned v1/v2 consumers continue receiving their frozen schemas unchanged.

### Hosted timing

Isolated Vercel Workflow stress run: [attempt 2](https://github.com/vercel/eve/actions/runs/33524880760/attempts/2), artifact `workflow-stress-performance`.

| Scenario | Samples | Mean | p50 | p95 |
| Sequential turns | 100 | 2.078s | 2.023s | 2.473s |
| Concurrent first turn | 50 | 5.204s | 5.445s | 6.010s |
| Concurrent second turn | 50 | 1.614s | 1.593s | 1.863s |

Compared with #2833's isolated run, sequential mean was 1.0% slower, p50 was 1.3% faster, and p95 was 0.4% faster. Those are noise-scale changes: exact-deployment routing improves locality and removes an avoidable lookup, but does not materially change the steady sequential turn floor by itself.

The isolated artifact predates the final undefined-observer correction inherited from #2833. The stress fixture does not send a delegated caller, so that correction does not touch the measured path. Exact-head Vercel and Postgres E2E matrices passed after a cancellation retry; the local prompt-cache suite passed on its final retry after empty provider responses.

### Validation

- `pnpm guard:invariants`
- `pnpm lint` (passes with existing unrelated warnings)
- `pnpm --filter eve test:unit` (7,528 passed, 1 skipped before the lower-stack regression addition)
- Focused wire, routing, runtime, and Nitro route suites (161 passed)
- Hosted Workflow stress fixture (100 sequential turns and 50 two-turn concurrent sessions)
- Exact-head Vercel and Postgres E2E matrices

### Checklist

### Diff size

- Docs: 1 file, +5/-0
- Implementation: 10 files, +224/-25
- Tests: 4 files, +200/-0


AndrewBarba:
Superseded by #2836, which now targets main and contains the complete turn-performance implementation and benchmark results.

PR #2836, opened 2026-09-01
fix(eve): turn performance - remove workflow round trips

AndrewBarba:
### Summary

Related to #876. Interactive channel turns currently pay for a session-hook metadata lookup, latest-deployment resolution, and a child-workflow handshake before ordinary model work can begin. Fixed-session commands now use Workflow's optimized token resume, trusted delivery metadata carries the deployment that accepted the request, and ordinary same-deployment turns execute directly in the session workflow. Turns that need child-owned behavior continue from the already-completed step in the existing child workflow, preserving eve's feature set without repeating work.

### Behavior and safety

- Continuation aliases and activity-observer deliveries retain version negotiation; only compatible stable-inbox commands use the raw token-resume path.
- Nitro stamps deployment provenance in framework-owned metadata. Missing, legacy, mixed, and cross-deployment deliveries fail closed to the existing latest-deployment dispatch.
- Inline steps verify `VERCEL_DEPLOYMENT_ID` before reading or mutating durable state, so deployment promotion races fall back to child dispatch.
- Sleep, background work, runtime actions, task-mode parking, cancellation settlement, and other complex paths hand their completed checkpoint to the existing child workflow.
- A late cancellation during handoff rolls back only the in-flight step and preserves state committed by earlier inline steps.
- Concurrent cancel, steer, reset, queue, clear, compact, and timeout commands retain their inbox semantics while an inline step is running.

### Hosted timing

Staged isolated Vercel Workflow stress runs measured each layer with 100 sequential turns and 50 two-turn concurrent sessions:

| Runtime shape | Mean | p50 | p95 | Concurrent second-turn p50 |
| [Optimized session resume](https://github.com/vercel/eve/actions/runs/33524878948/attempts/2) | 2.058s | 2.049s | 2.483s | 1.558s |
| [Plus direct deployment routing](https://github.com/vercel/eve/actions/runs/33524880760/attempts/2) | 2.078s | 2.023s | 2.473s | 1.593s |
| [Complete inline path](https://github.com/vercel/eve/actions/runs/33526586836/attempts/1) | 0.985s | 0.983s | 1.234s | 0.885s |

The first two layers do not materially move the two-second floor by themselves; they establish the trusted, compatible path required for inline execution. The complete path improves isolated sequential mean by 52.6%, p50 by 51.4%, and p95 by 50.1% over direct routing.

A separate [final implementation run](https://github.com/vercel/eve/actions/runs/33529804765/attempts/1) confirmed a 0.950s sequential p50 and 0.782s concurrent second-turn p50. Each run publishes the `workflow-stress-performance` artifact with the full distribution.

### Validation

The current branch passed 7,709 unit tests with one skip, including a 103-test focused review suite. The staged implementation also passed all 682 integration tests and its complete Vercel E2E matrix, including workflow stress, fixture tasks, and Postgres coverage. Focused coverage proves consecutive same-deployment turns execute as root steps without child dispatch, late cancellation preserves the prior inline checkpoint, and stable raw sends remain readable through the previous v2 migration chain.

The consolidated CI rerun is in progress. A prior required scenario run failed only the dev-server keep-alive port assertion that is also failing on current `main` ([main run](https://github.com/vercel/eve/actions/runs/33527178967)); this change does not touch dev-server draining.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

A single patch changeset describes the complete latency improvement as one release unit.

**Implementation** — 21 files · `+648 / -36`

The implementation combines the durable inbox wire change, trusted deployment routing, and the feature-complete inline state machine. Most new logic is kept in the dedicated inline-turn module so the checkpoint and fallback boundary can be reviewed in one place.

**Tests** — 12 files · `+748 / -28`

Coverage spans frozen wire compatibility, trusted and untrusted deployment routing, durable migration, concurrent commands, cancellation, and both inline and child-workflow execution paths.


packages/eve/src/channel/types.ts
```diff
@@ -95,6 +95,8 @@ export interface SessionTraceContext {
 
 /** Framework-owned identity for one inbound channel operation. */
 export interface ChannelDeliveryMetadata {
+  /** Vercel deployment that accepted this delivery, when known. */
+  readonly acceptedDeploymentId?: string;
```
chadhietala: Remove comment?

packages/eve/src/execution/wire/session-inbox-wire.ts
```diff
@@ -72,14 +74,20 @@ function decode(value: unknown): DecodedSessionInbox {
...
       `${WIRE_LABEL} declares version ${JSON.stringify(wire.version)}, expected ${SESSION_INBOX_WIRE_VERSION}.`,
     );
   }
-  if (declaredVersion === 2 && wire.kind === "deliver" && !("payload" in wire)) {
-    throw new SessionInboxWireError(`${WIRE_LABEL} does not match wire version 2.`);
+  if (
+    (declaredVersion === 2 || declaredVersion === 3) &&
+    wire.kind === "deliver" &&
+    !("payload" in wire)
+  ) {
```
chadhietala, line 91: Is this just going to continue to grow with versions and duck typing?

packages/eve/src/execution/wire/session-inbox-wire.v3-migration.ts
```diff
@@ -0,0 +1,16 @@
+import type { VersionMigration } from "#execution/durable-session-migrations/chain.js";
+import { normalizeSessionInboxWireV2 } from "#execution/wire/session-inbox-wire.v2-migration.js";
+
+export const sessionInboxWireV2Migration: VersionMigration = {
```
chadhietala: Not for this PR but I think there is a way through construction to get all of these migration, encoder, decoder etc pieces into a DSL that manages the versioning for us.


```ts
// root 0 version
const sessionStateV0 = durable.legacy({
  decode: decodeSessionStateV0,
});

// v1 derived from v0
const sessionStateV1 = sessionStateV0.next({
  decode: decodeSessionStateV1,
  encode: encodeSessionStateV1,
  migrate: migrateSessionStateV0ToV1,
  schema: sessionStateV1Schema,
});

// use v1 value
export const sessionState = durable.value({
  name: "sessionState",
  current: sessionStateV1,
});
```

packages/eve/src/execution/wire/session-inbox-encoder.ts
```diff
@@ -46,21 +52,64 @@ function encode(
...
+      acceptedDeploymentId === undefined ||
+      legacyRecord.kind !== "send" ||
+      typeof delivery !== "object" ||
+      delivery === null
+    ) {
+      return legacy;
+    }
+    return {
+      ...legacy,
+      delivery: { ...(delivery as Record<string, unknown>), acceptedDeploymentId },
```
ctgowrie, line 79: Could this avoid adding `acceptedDeploymentId` to the legacy wire shape? A session pinned to the previous v2 consumer rejects this unknown metadata field and drops the follow-up. We should negotiate when this field is present, then retain it only for v3 consumers.

packages/eve/src/execution/turn-workflow.ts
```diff
@@ -114,15 +114,20 @@ async function runTurnOwnedWorkflow(input: TurnWorkflowInput): Promise<void> {
...
+    let initialStep = input.initialStep;
     while (true) {
-      const beforeStep = {
+      const beforeStep = initialStep?.beforeStep ?? {
         serializedContext: cursor.serializedContext,
         sessionState: cursor.sessionState,
       };
-      const result = await turnStep(cursor.createStepInput(nextStepInput, cancellation?.signal));
+      const result =
+        initialStep?.result ??
```
ctgowrie, line 129: Before consuming this result, should the cursor adopt `initialStep.beforeStep`? After multiple inline steps the cursor still starts from the original turn state, so a late cancellation can drop state committed by earlier inline steps.

packages/eve/src/channel/types.ts
AndrewBarba: Removed in 2e0fcfc7d. The field name is sufficient here.

packages/eve/src/execution/wire/session-inbox-wire.ts
AndrewBarba: Yes, the explicit list would have grown. I changed this to a numeric v2+ policy in 2e0fcfc7d: v1 is the one historical shape that permits the missing `payload` mirror, while v2 and later require it. The decoder remains schema-free so it does not pull Zod into the workflow bundle.

packages/eve/src/execution/wire/session-inbox-wire.v3-migration.ts
AndrewBarba: Agreed—this construction would be a useful direction once more durable wire families adopt the pattern. I am keeping this PR scoped to the existing append-only chain and not introducing the DSL here.

packages/eve/src/execution/wire/session-inbox-encoder.ts
AndrewBarba: I traced this through both the previous v2 decoder and the older raw-send consumer. The v2 path does not schema-reject this value: its v0 migration validates that `delivery` is an object and then spreads its fields, normalization retains unknown properties, and decode does not run a strict schema parse. The older raw-send consumer simply ignores the unknown `delivery` property. Negotiating v3 before adding the field would restore the `getHookByToken` round trip that this fast path removes. I kept the additive field and added a regression in 2e0fcfc7d that runs the raw send through the previous v2 migration chain.

packages/eve/src/execution/turn-workflow.ts
AndrewBarba: Good catch. Fixed in 2e0fcfc7d by initializing the child cursor from `initialStep.beforeStep` before cancellation control or the completed result is consumed. The regression now uses distinct original, inline-checkpoint, and completed states and verifies a late cancellation settles from the inline checkpoint.

AndrewBarba:
Addressed the Vercel Agent review’s late-cancellation rollback finding in 2e0fcfc7d. The child workflow now begins at the inline step’s pre-step checkpoint, and regression coverage verifies that a late cancel preserves earlier inline state instead of reverting to the original pre-turn state.

ctgowrie (approved):

.changeset/inline-same-deployment-turns.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": patch
```
AndrewBarba:
```suggestion
"eve": minor
```

PR #2822, opened 2026-09-01
refactor(eve): centralize durable HITL request state

ruiconti:
### Summary

Related to #1224. Approval, Question, Limit, and Authorization currently persist overlapping park/resume state through separate modules. This centralizes their durable facts in one versioned Request ledger and routes pending input through one interpreter.

Before → after: pending batches, approval candidates, and pending Authorization blobs each carried lifecycle identity → Requests, Groups, ResponseAttempts, and internal Authorization Requests retain that identity in the ledger. Delivery IDs now reach approval responses, so replay deduplicates by `{requestId, deliveryId}` while separate deliveries remain competing attempts.

This PR stops at durable state and interpretation. Retryable Group completion, owner acknowledgement, forced closure, and explicit framework-gate delivery will land in a stacked PR. Workflow-backed tools and public `ctx.request` / `ctx.auth` remain out of scope.

### Validation

Package typechecking passed. Focused ledger, Authorization callback, approval-response, workflow-step, and full tool-loop coverage passed with 302 tests; delivery-attribution coverage passed in the earlier focused 307-test run.

### Checklist

- [x] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

### Diff size

**Docs** — 1 file · `+5 / -0`

One changeset describes the durable HITL state migration.

**Implementation** — 18 files · `+1892 / -1402`

The interpreter, ledger, ResponseAttempt records, and internal Authorization Requests replace several parallel state modules. Much of the deletion is the former approval-candidate and stale-response machinery.

**Tests** — 6 files · `+585 / -45`

Adds migration, stale-versus-invalid, delivery-identity, competing-attempt, and cross-kind conformance coverage while moving approval tests beside the new implementation.

PR #2853, opened 2026-09-01
feat(eve): deliver HITL group completion durably

ruiconti:
### Summary

Stacks on #2822. That PR centralizes durable HITL state, but a resolved Group could still advance directly to delivered while its transcript, lifecycle events, and framework-gate continuation ran in the same step. A process failure between those operations could lose the owner delivery.

Before → after: resolution terminalized Requests and completed the Group in one harness pass → resolution now persists `ready` with an idempotency key and owner payload, returns across the durable step boundary, then replays the payload and acknowledges `delivered` only after owner effects succeed. Retrying from the ready snapshot produces the same delivery.

Turn cancellation now force-closes waiting and ready Groups, internal Authorization Requests, and active approval ResponseAttempts. Direct harness callers keep same-invocation behavior; the checkpoint is enabled by the workflow execution boundary.

### Validation

Package typechecking passed. Seven focused unit files passed with 319 tests, covering the ready/delivered protocol, retry from a ready snapshot, existing HITL behavior, active-attempt closure, and turnStep cancellation persistence. The cancellation-settle integration test could not start because this isolated worktree lacks the generated `#internal/workflow-bundle/workflow-builders.js` module required by the integration config; CI builds that prerequisite normally.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

A separate changeset describes the observable durability and cancellation behavior added by this stacked PR.

**Implementation** — 9 files · `+326 / -64`

Most additions define the explicit Group delivery protocol and adapt the interpreter to store and replay its existing resolution result. Cancellation wiring closes the same ledger state from both in-step and driver-observed cancellation paths.

**Tests** — 4 files · `+204 / -3`

Covers state transitions, idempotent acknowledgement, crash-boundary replay, ResponseAttempt closure, and persisted turnStep cancellation behavior.

### DIFF
--- a/.changeset/tidy-pears-wait.md
+++ b/.changeset/tidy-pears-wait.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Persist completed HITL Groups before owner effects, retry unacknowledged delivery, and close pending Groups, Authorization requests, and response attempts when a turn is cancelled.
--- a/packages/eve/src/execution/node-step.ts
+++ b/packages/eve/src/execution/node-step.ts
@@ -106,6 +106,7 @@ export function createExecutionNodeStep(input: CreateExecutionNodeStepInput): St
     capabilities: input.capabilities,
     clearOnly: input.clearOnly,
     compactOnly: input.compactOnly,
+    durableGroupCompletionDelivery: true,
     workflow: input.node.agent.workflowTool !== undefined,
     workflowMaxSubagents: input.workflowMaxSubagents,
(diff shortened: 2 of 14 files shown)


ruiconti:
Folded into #2863, which now stacks directly on #2822. This PR moved response-attempt and Authorization state into the ledger without changing which modules could write it; reviewed alone it shows one ledger with three writers, which is not a meaningful boundary. #2863 contains these commits unchanged and finishes the cut in the same review.

PR #2854, opened 2026-09-01
fix(eve): mcp connections - accept structured-only tool results

AndrewBarba:
### Summary

Shopify's catalog MCP returns tool results with `structuredContent` but no `content`, which causes the strict MCP result parser to reject otherwise usable output. MCP connections now recognize that shape in successful JSON `tools/call` responses and synthesize the spec-recommended text fallback before validation. Existing content remains untouched, and the compatibility behavior stays inside eve's MCP transport boundary without changing the public connection API.

### Validation

Reproduced the failure through the real vendored MCP client with an in-memory Streamable HTTP response containing only `structuredContent`. The focused integration confirms that result now parses with mirrored text, the compliant-result case confirms existing content is preserved, and the full unit suite passed with 7,718 tests and 1 skip.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the patch changeset for the published `eve` package.

**Implementation** — 1 file · `+98 / -2`

Keeps the compatibility normalization at the MCP fetch boundary and scopes it to matching JSON-RPC tool results while preserving response metadata.

**Tests** — 2 files · `+76 / -0`

Covers the regression through the real MCP client, preserves compliant results, and verifies both transports receive the compatibility fetch.

### DIFF
--- a/.changeset/calm-catalogs-listen.md
+++ b/.changeset/calm-catalogs-listen.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Accept MCP tool results that return only structured content by synthesizing the protocol's text fallback before strict validation.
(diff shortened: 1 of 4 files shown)


AndrewBarba:
Closing this in favor of the upstream AI SDK parser fix: https://github.com/vercel/ai/pull/20173. eve should wait for that change to land and update its dependency rather than maintain a transport-level fetch workaround.

Issue #1476, opened 2026-07-31
New-session bootstrap costs ~4s inside the workflow run before session.started (queue dispatch is only ~0.2s)


ryantbrown:
### Summary

On a production Vercel deployment, every **new session** spends ~4–6s between the user's message reaching the eve server and `session.started` being emitted. We instrumented the gap end to end: the Vercel Workflow queue dispatch accounts for only ~0.2s of it — the remaining ~4s happens **inside the workflow function, after the run starts, before `session.started`** — i.e. eve's session bootstrap (compiled-bundle load, durable stream init, park-hook commit).

For a chat product this is the dominant share of time-to-first-reasoning on every new conversation (~10s total in our case), and it precedes any user code, so there is nothing an app author can do about it.

### Environment

- eve `0.29.1`, deployed on Vercel (`WORKFLOW_TARGET_WORLD=vercel`), Node 24
- Agent model: `openai/gpt-5.6-luna`
- Session entry: HTTP chat channel proxied from a same-origin dashboard

### Measurements

**Cohort (40 new-thread first turns over one day, production).** We ledger every eve stream event with its `occurred_at`, and stamp thread creation in our DB at send time:

| Segment | Typical (p50-ish) | Range |
| --- | --- | --- |
| send → `session.started` | **4–6s** | 3.5s – 17.6s (plus rare 40–110s outliers) |
| `session.started` → `turn.started` | 1.5–2s | (dynamic skill resolution — ours, being fixed) |
| first model step → first output | 3–5s | |

**Split of the send → `session.started` segment (two production runs sampled today via the Workflow API):**

| Segment | Run 1 | Run 2 |
| --- | --- | --- |
| workflow run `createdAt` → `startedAt` (queue dispatch) | 0.19s | 0.21s |
| run `startedAt` → eve `session.started` event | **3.66s** | **4.42s** |

So the platform dispatch is fast; the cost is eve's own serial boot work per new session before the first turn begins processing.

For contrast, local dev against the filesystem world: warm new-session pickup is 0.3–0.8s (first session after a dev-server boot pays ~15s one-time init).

### Expected

New-session bootstrap overhead in the low hundreds of milliseconds, or at least overlappable with the first turn's model call — the docs' "commits the park hook before processing its first turn" ordering suggests several sequential state-store round-trips that could plausibly be parallelized, deferred, or pipelined.

### Ask

1. Can the session bootstrap (bundle load, stream init, hook commit) be parallelized or moved off the first-turn critical path?
[...]

PR #2859, opened 2026-09-01
fix(eve): channels - remove session-start hook polling

AndrewBarba:
### Summary

Related to #1476 and #876.

Session creation no longer polls Workflow hook state from the HTTP process after durable start acceptance. eve, MCP, and authored-channel starts return the accepted candidate immediately; concurrent authored-channel cold starts settle inside durable execution, where the loser forwards its distinct message to the owner and retires its activity collector.

Workflow starts no longer resolve a `"latest"` deployment sentinel. Fresh sessions and nested work target the deployment executing the call, while turn dispatch targets the trusted deployment ID stamped by ingress and falls back to the driver deployment when no unanimous ingress stamp exists.

Racing `operationId` creates can return different candidate IDs, so callers that need the canonical owner must resolve after startup. Reset still waits for inbox release to preserve lossless follow-up delivery, and replay-idempotent task, tool, and subagent ownership checks remain inside durable steps.

### Validation

The full unit suite passes with 7,717 tests across 716 files. Another 45 focused integration tests cover workflow entry, durable conflict handoff, and replayed subagent dispatch, while focused unit coverage verifies current, ingress-stamped, mixed, and development deployment routing.

### Checklist

### Diff size

**Docs** — 4 files · `+25 / -9`

Documents immediate start acceptance and the candidate-ID race, and updates the release note to include exact deployment routing while preserving reset's existing contract.

**Implementation** — 14 files · `+223 / -188`

Removes request-side ownership arbitration and deployment resolution, moves distinct-message handoff and collector cleanup into durable execution, and routes new work using current or ingress-stamped deployment IDs. Reset's release wait remains unchanged.

**Tests** — 9 files · `+254 / -175`

Covers polling-free starts, durable conflict handoff, legacy owner recovery, replay-idempotent subagent ownership, and exact deployment selection without broad fixture changes.


packages/eve/src/execution/workflow-runtime.ts
```diff
@@ -353,10 +342,6 @@ async function dispatchWorkflowCommand<TCommand extends SessionCommand>(
     throw error;
   }
 
-  if (command.kind === "reset") {
-    await waitForCommandHookRelease(sessionCommandHookToken(hook.runId), hook.runId);
-  }
-
   return activeCommandResult(command, hook.runId);
```
AndrewBarba, line 345: **Blocking.** Removing `waitForCommandHookRelease` here reopens a silent message-drop window on every alias-addressed channel.

`channel-address.ts` `deliver()` tries `dispatchContinuation` first and only falls through to `createSession` on `session_not_active`. The retiring run's continuation and stable hooks stay live until `commandInbox.dispose()` in the `finally` at `workflow-entry.ts:~277`, which runs only after `runDriverLoop` observes the reset at its next parked wait. So between `reset` returning and that dispose committing, `from(alias).send()` resumes the retiring run's inbox, gets `accepted`, and the payload is discarded when the run terminates.

Concrete repro: Slack `/reset` followed immediately by a message in the same thread. Today the poll guarantees the second message creates a fresh session; with this change it is lost, and the caller was told `accepted`.

This is the one request-path poll with no cheap replacement. Suggest restoring `waitForCommandHookRelease` (and the three reset doc paragraphs) in this PR, and keeping the create-path removal, which is the part that actually pays off. A durable fix (dying run disposes its hooks in a step, then re-delivers already-queued sends via `from(alias).deliver(...)` before finalizing) is a separate PR.

packages/eve/src/execution/continuation-conflict-step.ts
```diff
@@ -0,0 +1,72 @@
...
+
+async function forwardCommand(input: {
+  readonly command: Extract<SessionCommand, { readonly kind: "send" }>;
+  readonly continuationToken: string;
+  readonly ownerSessionId?: string;
+}): Promise<void> {
+  try {
+    await resumeSessionInbox(input.continuationToken, input.command);
+  } catch (error) {
+    if (!HookNotFoundError.is(error) || input.ownerSessionId === undefined) throw error;
```
AndrewBarba: `ownerSessionId` is `undefined` whenever the world does not populate `conflictingRunId` (`hook-ownership.ts:normalizeHookClaimError` explicitly allows this for legacy worlds). In that case a `HookNotFoundError` from the alias resume (winner already re-keyed its continuation) rethrows, the step retries, and hits the same `HookNotFoundError` again forever until the step's retry budget fails the run — with the losing candidate's message never delivered.

Two things:
1. When `ownerSessionId === undefined` and the alias resume is `HookNotFound`, resolve the current alias owner once with `getHookByToken(continuationToken)` (single lookup, not a loop) and fall back to its stable inbox — or fail loudly with a message that names the dropped delivery so it shows up in logs.
2. Add a test for `ownerSessionId === undefined` + `HookNotFoundError`. The two tests in `continuation-conflict-step.test.ts` both supply an owner.

docs/concepts/sessions-runs-and-streaming.md
```diff
@@ -210,7 +210,10 @@ Compaction summarizes context without adding a user message. User-role instructi
 
 Clear removes model-message history in place, including static and dynamic user-role instructions and recalled memory records, while preserving the session identity, system-role instructions, tools, skills, application-defined durable state, limits, and sandbox. It clears framework memory locks and replay bookkeeping but does not delete data from a provider's external store. It does not rerun instruction definitions or resolvers. It emits `context.cleared` followed by `session.waiting`.
 
-Reset terminally retires the exact session ID. A reset ID never becomes a new session; create another session explicitly for a fresh conversation. Compact, clear, and reset return `"no_active_session"` when the target is already inactive.
+Reset queues terminal retirement for the exact session ID and returns as soon as the durable inbox
```
AndrewBarba: If the reset wait is restored (see comment on `workflow-runtime.ts`), revert this paragraph to the original wording. If it is not, this paragraph needs to say the unpleasant part out loud: a `send()` to the same address in the window before the retiring run releases its inbox is accepted and discarded. Right now the text implies the follow-up creates a fresh session, which is only true if the caller is slow.

docs/channels/custom.mdx
```diff
@@ -155,8 +159,10 @@ address's current owner as a fixed handle.
   success emits `compaction.requested`, `compaction.completed`, then `session.waiting`.
 - `clear` removes model-message history in place. It preserves the system prompt,
   skills, tools, durable state, limits, address ownership, and session sandbox.
-- `reset` terminally retires the current session. A later address `send()` creates
-  a fresh session; a fixed `Session` handle remains pinned to the retired ID.
+- `reset` queues terminal retirement for the current session and returns once the
```
AndrewBarba: Same as the concepts page: either revert with the reset wait, or document that an address `send()` racing the retirement can be accepted by the retiring session and lost.

packages/eve/src/execution/subagent-start-local.ts
```diff
@@ -99,42 +98,36 @@ export async function startLocalSubagent(input: {
...
   try {
-    const handle = await contextStorage.run(
-      new ContextContainer({ localDevRequest: input.localDevRequest }),
-      () => childRuntime.createSession(runInput),
+    await contextStorage.run(new ContextContainer({ localDevRequest: input.localDevRequest }), () =>
+      childRuntime.createSession(runInput),
     );
-    childSessionId = handle.sessionId;
+    // This runs inside the replayed dispatch step, which must persist the
+    // canonical child rather than the candidate accepted by start().
```
AndrewBarba, line 105: Non-blocking. This is the correct place for the remaining poll (inside a replayed step, and the persisted child ID must be the canonical owner), so no change requested. Just noting for the PR description: this moves the poll rather than removing it, and the same `waitForCommandHookOwner` remains for tasks and tool runs. The title's "request-path" qualifier covers it; the summary's "Replay-idempotent ... checks remain inside durable steps" is accurate.

.changeset/tidy-rivers-return.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": minor
+---
+
+Return eve, MCP, and authored-channel session starts after durable acceptance without polling continuation ownership, and return reset requests without polling for inbox release.
```
AndrewBarba: Nit: release-notes readers need the behavior change, not the mechanism. Suggest something like: "Session creation on eve, MCP, and authored channels returns as soon as Workflow accepts the run. Concurrent first messages on one channel address are settled inside the workflow, and `operationId` create-once requests may now return different candidate IDs while racing; re-resolve after startup for the canonical owner." Drop the reset sentence if the reset wait is restored.

AndrewBarba (commented):
The create-path changes are good: ownership arbitration belongs in the workflow, and `settleContinuationConflictStep` is the right shape for it. Requesting changes only for the reset half, which reintroduces a message-drop window that the docs currently promise does not exist. Restore the reset wait (or land the durable-release design separately), tighten the forward fallback, and this is approvable.

packages/eve/src/execution/workflow-runtime.ts
AndrewBarba: Restored waitForCommandHookRelease and its reset-release test, and reverted the three reset docs to the original terminal-retirement contract. Reset remains the deliberate request-path exception until a durable release design lands separately.

packages/eve/src/execution/subagent-start-local.ts
AndrewBarba: Kept this durable-step poll unchanged. The PR title now scopes the removal to session starts, and the summary explicitly preserves replay-idempotent task, tool, and subagent ownership checks inside durable steps.

packages/eve/src/execution/continuation-conflict-step.ts
AndrewBarba: Added a single getHookByToken lookup when conflictingRunId is absent, then forwards through the resolved stable inbox. If that lookup still cannot resolve an owner, the step now throws an explicit error containing the delivery request ID. Added coverage for both successful legacy recovery and the unresolved-owner failure.

.changeset/tidy-rivers-return.md
AndrewBarba: Rewrote the changeset around observable start behavior and removed the reset sentence.

docs/concepts/sessions-runs-and-streaming.md
AndrewBarba: Reverted this paragraph to the original reset contract now that the release wait is restored.

docs/channels/custom.mdx
AndrewBarba: Reverted this paragraph to the original reset contract now that the release wait is restored.

AndrewBarba (commented):
`1afab81` resolves everything from the previous review.

- Reset wait restored; `workflow-runtime.ts` and all reset doc paragraphs are identical to `main`, so the reset→send drop window is closed.
- `resolveOwnerSessionId` does a single `getHookByToken` lookup when the world omits `conflictingRunId` and fails loudly (naming the `requestId`) when the owner is gone, instead of retrying into the same `HookNotFound`. Both undefined-owner paths are tested.
- Changeset and body now describe the behavior change and state that reset still waits.

LGTM. Net effect: the create-path poll is gone for every channel and MCP, the reset poll stays until it has a durable replacement, and the remaining `waitForCommandHookOwner` sites are all inside replayed steps.

ctgowrie (approved):

PR #2867, opened 2026-09-02
fix(eve): persist text streams as deltas

ruiconti:
### Summary

**Before**

```json
{ "messageDelta": "Revenue", "messageSoFar": "Revenue" }
{ "messageDelta": " increased", "messageSoFar": "Revenue increased" }
{ "messageDelta": " by 12%", "messageSoFar": "Revenue increased by 12%" }
```

**After**

```json
{ "messageDelta": "Revenue" }
{ "messageDelta": " increased" }
{ "messageDelta": " by 12%" }
```

`reasoning.appended` and `action.input.appended` now have the same shape: each event persists only the new chunk plus its existing turn, step, and call coordinates. Empty tool-input starts no longer emit an append event; the first append contains actual input text.

Consumers that need text-so-far accumulate chunks in stream order. The default reducer, TUI, Slack, Chat SDK, and Next.js trace consumer now do that locally. Slack still posts final answers from `message.completed`; its progressive reasoning accumulator resets at reasoning completion and turn/step boundaries.

The durable event index remains the reconnect cursor, and stable event IDs deduplicate replay overlap. The protocol does not add a second text position or block-boundary mechanism.

The stream version advances from 24 to 25. The client and server normalize v21-v24 cumulative message/reasoning events and v24 tool-input offsets into the current delta-only shape. Missing deltas and append fields that contradict the declared version fail at the decode boundary. Published extensions whose capability contract changed must be rebuilt and republished.

Provider retries remain append-only: events emitted by an interrupted attempt stay durable. A later completed event can replace provisional text in replaceable consumers; the TUI now replaces its still-live block before committing the completed text. A block completed before its attempt later fails cannot be retracted without adding attempt identity.

### Reproduction

The same 159-call, 414,436-event sequence was encoded with the cumulative and delta-only formats:

| Measurement | Cumulative format | Delta-only format | Difference |
| --- | ---: | ---: | ---: |
| Total serialized eve protocol | 11,381,164,093 bytes | 85,280,426 bytes | -11,295,883,667 bytes (-99.25%) |
| `reasoningSoFar` payload | 9,666,376,952 bytes | 0 bytes | -9,666,376,952 bytes |
| `messageSoFar` payload | 1,629,506,715 bytes | 0 bytes | -1,629,506,715 bytes |

The cumulative stream was 133.46 times the delta-only size; the new stream is 0.75% of the previous size. These are serialized eve protocol bytes before Workflow framing, not billed Workflow Data Written.

### Validation

The matched replay sends the same chunk sequence through the cumulative and current eve emitters and protocol encoders. The delta full branch passed 7,758 unit tests and 711 integration tests. The final TUI repair passes 298 focused TUI tests, both provider-retry integration tests, and eve typecheck; CI remains the authoritative full-commit result.

Tests cover delta accumulation and coalescing, provider retry output, legacy replay, reconnects across stream versions, malformed v25 events, and built-in client, Chat SDK, Slack, TUI, ACP, and trace consumers. The TUI regression asserts that discarded retry text is absent from the committed transcript and the recovered answer appears once.

### Checklist

### Diff size

**Docs** — 4 files · `+112 / -8`

Explains the delta-only wire format, consumer migration, reconnect semantics, and retry limitation.

**Implementation** — 29 files · `+627 / -190`

Removes cumulative snapshots and tool-input offsets, normalizes supported historical streams, validates the v25 shape, and migrates first-party consumers. Eight generated capability reports record the breaking author-facing event contract.

**Tests** — 30 files · `+933 / -129`

Covers byte scaling, accumulation, coalescing, retry behavior, legacy replay, reconnects, malformed events, and first-party consumer behavior.


ctgowrie:
Why do we have an offset vs just using the chunk which is closer to what providers and ai sdk use?

ruiconti:
> Why do we have an offset vs just using the chunk which is closer to what providers and ai sdk use?

The reasoning was to provide a way for clients, during reconnect midstream, to distinguish if the current `reasoning-delta` event:
- should be appended or not e.g `offset !== 0`
- has skewed from current local state e.g last known offset was 13 and it received 45

But these are theoretical and not consistent with other `-delta` usage, so looks like a theoretical issue.

ctgowrie (approved):

PR #2888, opened 2026-09-02
chore(eve): containers - move publishing to VCR

AndrewBarba:
### Summary

eve's standalone container workflow published to GHCR and copied to VCR with a long-lived token, independently of npm releases. The existing release job now authenticates to VCR through GitHub OIDC using `VCR_TEAM_ID` and publishes `vcr.vercel.com/vercel/eve/base:<version>` plus `:latest` before Changesets can publish npm; the standalone GHCR workflow is removed. The Dockerfile and its ignore file now live with the eve package under `packages/eve`.

All built-in container backends now consume one internal `DEFAULT_EVE_SANDBOX_IMAGE`, which temporarily remains `ghcr.io/vercel/eve:latest`. The same module includes the installed-version VCR resolver and a TODO, so the follow-up after the first VCR publish only changes the default initializer to call that resolver. The login action is pinned to its current verified commit because it does not yet publish a versioned ref.

### Validation

The focused Docker, microsandbox, and Vercel Sandbox suites pass all 63 tests. The moved image definition is byte-for-byte unchanged and both affected workflows parse as valid YAML; the actual image build and push remain covered by CI because the local Docker daemon is unavailable and VCR publication requires the main-branch GitHub OIDC context.

### Checklist

### Diff size

Not applicable; the shared runtime default remains on the existing GHCR image until the follow-up.

**Implementation** — 10 files · `+78 / -193`

Most deletions remove the standalone GHCR and token-based VCR workflow. The replacement keeps OIDC-authenticated VCR publication ordered inside the release job, relocates the unchanged image definition, and prepares the shared version-aware image resolver without enabling it yet.

**Tests** — 3 files · `+8 / -4`

Verifies Docker, microsandbox, and hosted Vercel Sandbox all consume the shared temporary image default.


cmpadden (approved):
very nice

PR #2894, opened 2026-09-02
refactor(eve): share Gateway model catalog parsing

prha:
### Summary

The self-modification model-search tool is already on `main` after the package move. This PR now shares its AI Gateway model-listing parser with `/model`, so both surfaces validate catalog entries and derive their model shape consistently.

### Validation

Focused `/model` and self-modification model-search coverage passed (18 tests), along with `pnpm --filter eve typecheck` and `pnpm --filter eve build`.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Records the patch release note for the shared catalog validation.

**Implementation** — 3 files · `+53 / -69`

A small shared parser replaces duplicate catalog-validation code in the setup picker and bundled self-modification tool.

**Tests** — 1 file · `+11 / -8`

Updates self-modification coverage to exercise the shared parser while preserving model-search behavior.

### DIFF
--- a/.changeset/shared-gateway-model-catalog.md
+++ b/.changeset/shared-gateway-model-catalog.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+`eve/self-modification` now uses the same AI Gateway model catalog validation as `/model`.
--- a/packages/eve/src/self-modification/extension/tools/search_models.ts
+++ b/packages/eve/src/self-modification/extension/tools/search_models.ts
@@ -1,7 +1,11 @@
 import { defineTool } from "eve/tools";
 
-const AI_GATEWAY_MODELS_URL = "https://ai-gateway.vercel.sh/v1/models";
-const DEFAULT_AGENT_MODEL_ID = "openai/gpt-5.6-luna-fast";
+import { AI_GATEWAY_MODELS_URL } from "#internal/gateway.js";
+import { DEFAULT_AGENT_MODEL_ID } from "#shared/default-agent-model.js";
+import {
+  parseGatewayModelCatalog,
+  type GatewayCatalogModel,
+} from "#shared/gateway-model-catalog.js";
 const FETCH_TIMEOUT_MS = 5_000;
 const MAX_RESULTS = 20;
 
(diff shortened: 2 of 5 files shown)


packages/eve-self-modification/extension/tools/search_models.ts
```diff
@@ -0,0 +1,108 @@
...
+    .map((model) => ({
+      id: model.id,
+      name: model.name,
+      provider: model.id.split("/")[0] ?? "",
+    }));
+}
+
+export default defineTool({
+  description:
+    "Search the list of available models. Call this to get the list of available models or to resolve a model to its exact ID.",
```
OwenKephart: should the description here make it clear that it's only looking through the vercel AI Gateway? i.e. if you have a different model provider we shouldn't advertise this as a generic model search

packages/eve-self-modification/extension/tools/search_models.ts
```diff
@@ -0,0 +1,108 @@
...
+  async execute(input, ctx) {
+    if (typeof input.query !== "string") throw new Error("Expected a model search query.");
+    const controller = new AbortController();
+    const timeout = setTimeout(() => controller.abort(), FETCH_TIMEOUT_MS);
+    try {
+      const signal = AbortSignal.any([ctx.abortSignal, controller.signal]);
+      const response = await fetch(AI_GATEWAY_MODELS_URL, { signal });
+      if (!response.ok) {
+        throw new Error(`AI Gateway model catalog request failed (${response.status}).`);
+      }
```
OwenKephart, line 78: could probably share code with some of the existing model fetching logic (e.g. the model select flow in the CLI also does a similar thing)

OwenKephart (approved):
approved but two comments

packages/eve-self-modification/extension/tools/search_models.ts
prha: yeah, considered it, but it'd require opening up a public eve export or creating a shared package for what feels like a pretty thin layer.

packages/eve-self-modification/extension/tools/search_models.ts
prha: 👍

PR #2902, opened 2026-09-02
feat(eve): sandbox - use versioned VCR image

AndrewBarba:
### Summary

The sandbox default still points at the temporary `ghcr.io/vercel/eve:latest` fallback even though the first VCR image has been published. Docker, microsandbox, and Vercel sandboxes now share `vcr.vercel.com/vercel/eve/base:<eve-version>`, resolved from the installed eve version. The minor changeset will produce 0.50.0, whose release workflow publishes the matching VCR image before npm. This depends on [vercel/api#89572](https://github.com/vercel/api/pull/89572) being deployed so local runtimes can pull the public image without credentials; keep this PR in draft until then.

### Validation

The four focused sandbox suites pass with 64 tests, including coverage that the resolver and exported default use the installed package version. The docs compile and workspace typechecking passes. Anonymous registry access is covered by the dependent API PR and must be verified against production after that change deploys.

### Checklist

- [x] Tests added or updated for behavior changes
- [x] Documentation updated for user-facing changes
- [x] Changeset added for the published `eve` package

### Diff size

Docs — 5 files · +18 / -10
Implementation — 1 file · +2 / -3
Tests — 3 files · +24 / -4

Total — 9 files · +44 / -17

### DIFF
--- a/.changeset/tidy-owls-build.md
+++ b/.changeset/tidy-owls-build.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
+Use the version-matched Vercel Container Registry image as the default base for Docker, microsandbox, and Vercel sandboxes. The backend environment must have pull access to the `vercel/eve` VCR project.
--- a/.changeset/tidy-owls-build.md
+++ b/.changeset/tidy-owls-build.md
@@ -2,4 +2,4 @@
 "eve": minor
 ---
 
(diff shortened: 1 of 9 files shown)


AndrewBarba:
Superseded by #2905, which keeps the release/version plumbing but targets the anonymously pullable GHCR package.

PR #2911, opened 2026-09-02
fix(eve): dual-publish sandbox images

AndrewBarba:
### Summary

eve sandbox images are currently published only to VCR, which hosted Vercel Sandbox can pull but Docker and microsandbox cannot. This restores dual publishing to VCR and GHCR with immutable version tags and `latest`, while independently preserving tags that are already published. Docker and microsandbox now resolve the version-matched GHCR image; Vercel Sandbox resolves the version-matched VCR image. GHCR remains multi-architecture, while VCR remains `linux/amd64` for Vercel.

### Validation

Focused image resolver, Docker options, microsandbox, and Vercel unit suites passed with 65 tests, and `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts test/scenarios/public-api-portability.scenario.test.ts` passed with 9 tests. Registry pushes require release credentials and remain CI-only.

### Checklist

### Diff size

**Docs** — 2 files · `+10 / -3`

Updates the sandbox guide and release metadata for the registry-specific defaults.

**Implementation** — 6 files · `+82 / -22`

Adds independent VCR and GHCR publishing while keeping the runtime resolver split small and explicit.

**Tests** — 3 files · `+32 / -7`

Covers both versioned image resolvers and verifies that Vercel Sandbox receives the VCR image.

### DIFF
--- a/.changeset/dual-images-publish.md
+++ b/.changeset/dual-images-publish.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Use version-matched GHCR images for Docker and microsandbox while Vercel Sandbox uses VCR. Publish each eve sandbox image version to both registries.
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,7 @@ concurrency:
   cancel-in-progress: false
 
 env:
+  GHCR_IMAGE: ghcr.io/vercel/eve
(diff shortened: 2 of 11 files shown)


ijjk (approved):
Rollout note: make sure the release workflow is triggered after merge so it publishes the current version tag, ghcr.io/vercel/eve:0.49.0, before the versioned image is used by Docker and the local sandbox E2E.

PR #2917, opened 2026-09-02
feat(eve): sleep - run as a workflow tool

AndrewBarba:
### Summary

Follow-up to #2477. The provided `sleep` tool now runs its execute body as a durable workflow tool, removing the custom state projection and wait handling from the turn workflow. To keep this source-neutral, eve's package build preserves `"use workflow"` bodies while stamping their generated callback metadata, and the compiler derives workflow-tool handling from the loaded executor instead of rescanning source or requiring a sleep-specific reference. Concurrent calls still run in parallel and resume the turn after the longest wait, with cancellation and execution handled by the standard workflow-tool lifecycle. The review-sensitive boundary is the metadata-only transform: it must leave the workflow body intact for the downstream driver build.

### Validation

Validated with the full unit suite (7,733 passed), full integration suite (711 passed), workflow bundle scenarios (11 passed), and focused compiler scenarios (2 passed); inspected the package build output to confirm the sleep executor retains its workflow body and carries the generated versioned workflow ID.

### Checklist

### Diff size

**Docs** — 2 files · `+6 / -1`

Updates the built-in tool behavior and adds release metadata for the published package.

**Implementation** — 16 files · `+65 / -139`

The net deletion removes sleep-specific turn plumbing and source rescanning; the replacement is a small package-build metadata transform plus source-neutral executor inspection.

**Tests** — 8 files · `+147 / -158`

Replaces turn-projection coverage with focused metadata-transform, compiler, package-build, and real workflow-tool lifecycle coverage.


packages/eve/src/execution/stable-workflow-names.ts
```diff
@@ -8,6 +8,7 @@ export const TURN_WORKFLOW_NAME = "turnWorkflow";
 export const SESSION_TIMEOUT_WORKFLOW_NAME = "sessionTimeoutWorkflow";
 export const TASK_RUN_WORKFLOW_NAME = "taskRunWorkflow";
 export const TOOL_RUN_WORKFLOW_NAME = "toolRunWorkflow";
+export const SLEEP_TOOL_WORKFLOW_NAME = "executeSleepTool";
```
ruiconti: Why is this needed? This should track internal workflow names, tool-specific shouldn't need to be added here as they'd run as a `TOOL_RUN_WORKFLOW_NAME`

AndrewBarba: Agreed — this was leftover special casing from the original implementation. I removed `executeSleepTool` from `STABLE_WORKFLOW_NAMES`; only the enclosing `toolRunWorkflow` needs the cross-deployment identity. The package metadata transform now stamps ordinary package workflows with their versioned package identity, so the executor metadata still matches the downstream workflow registry without making the sleep workflow stable. The package build, compiler scenario, workflow bundle scenario, and focused lifecycle integration test all pass with that versioned ID.

ruiconti (approved):

ctgowrie (approved):

PR #2923, opened 2026-09-02
fix(eve): approvals - keep pending prompts independent

AndrewBarba:
### Summary

When separate approval batches shared a `once()` approval key, settling one recorded the reusable grant immediately. A resumed step could therefore auto-authorize matching calls while their original prompts remained pending, splitting execution state from visible approval cards. This change masks the recorded grant until all already-pending requests with that key are settled, while preserving independent compound keys and the existing behavior for later calls.

### Validation

Reproduced the current-main failure with three separately pending calls sharing one key, then confirmed the fix with the focused approval unit suites (79 tests) and approval-resume integration suites (10 tests), including compound-key coverage.

### Checklist

### Diff size

**Docs** — 2 files · `+7 / -0`

Documents the pending-prompt semantics and adds the required patch release note.

**Implementation** — 2 files · `+21 / -4`

Keeps the change local to approval-key projection and its tool-loop caller.

**Tests** — 1 file · `+75 / -0`

Covers multiple pending calls with the same key and verifies that compound keys remain independently reusable.


benpankow (approved):
🙏 good catch

PR #2959, opened 2026-09-03
feat(eve): isolate local delegated traces

chadhietala:
### Summary

Gives local delegated sessions replay-stable child traces before dispatch and records a child trace on the parent caller only after the local runtime acknowledges the same coordinates. The child evaluates its own trace policy and sampler while retaining a separate parent link context.

This is the local dispatch layer. Remote and channel transport of the strict trace extension remains in the next stacked PR.

Depends on #2957.

### Validation

Focused local dispatch coverage passed with 78 tests. The full workspace typecheck passed all 44 tasks, the `eve` package typecheck passed, and `pnpm guard:invariants` passed.

### Checklist

- [x] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the required patch changeset.

**Implementation** — 12 files · `+239 / -84`

Moves session seed evaluation into a dedicated helper, threads internal child seeds through local workflow creation, stores confirmed local trace IDs on handles, and persists acknowledged coordinates on caller state.

**Tests** — 4 files · `+127 / -0`

Covers child policy evaluation, trace-seed propagation, local acknowledgement, handle persistence, and context flow.

PR #2958, opened 2026-09-03
fix(eve): propagate agent trace context and conversation identity

chadhietala:
### Summary

Delegated agent traces must retain parent context and a stable conversation ID so remote callbacks correlate correctly without widening trace-content policy across agent boundaries. This PR threads that context through local and remote subagent startup, adds bounded shared parsing for eve's trace and conversation baggage, and accepts inbound conversation correlation only on callback requests. Authentication, root-session limits, and the distinct activation traces with causal links remain unchanged; trace topology is covered by #2966.

### Validation

- Squashed this PR to one signed commit; removed the unreachable trace-root/link branch, pass-through result context, obsolete assertions, and unrelated workflow test token.
- All ten changed unit-test files passed: 217 tests, using `vitest run --config vitest.unit.config.ts` with the changed-file paths.
- Base-PR integration validation: 22 files / 144 tests passed with `vitest run --config vitest.integration.config.ts src/subagents src/eve-channel src/instrumentation src/tracing src/execution`.
- Base-PR callback and OTel registration scenarios: 2 files / 18 tests passed with `vitest run --config vitest.scenario.config.ts src/execution/multi-agent-callback-routing.scenario.test.ts src/tracing/otel-registration.scenario.test.ts`.
- Final stack-tip full unit suite: 8,268 passed, 1 skipped, 2 failed. Both failures are macOS path expectations in unchanged `src/cli/telemetry/preference.test.ts`; that test and implementation match `main`.
- Final stack-tip relevant integration suites: 23 files / 154 tests passed.
- Final stack-tip callback and OTel registration scenarios: 2 files / 22 tests passed.
- TypeScript `--project packages/eve/tsconfig.json --noEmit`, `oxlint`, `oxfmt --check`, `node scripts/guard-invariants.mjs`, and all three documentation checks passed.
- The previously reported docs previews and `e2e-vercel (fixture-tasks)` passed on preceding revision `59eb379f8`; the rewritten head requires fresh CI.
- Existing changeset retained. No new authentication mechanism or local E2E run.

### Checklist

- [x] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

PR #2962, opened 2026-09-03
feat(eve): transport remote delegated traces

chadhietala:
### Summary

Carries schema 4 delegated trace coordinates through remote eve session creation. The sender transmits explicit invocation lineage and a distinct child seed, the receiver validates the binding and echoes only accepted coordinates, and the caller records a trace ID only when the acknowledgement exactly matches.

A 400 response from an older receiver triggers one compatibility retry with the new invocation and trace extension removed; the fallback does not join the child to the parent trace.

Depends on #2959.

### Validation

Focused protocol, remote dispatch, continuation metadata, and session-route coverage passed with 62 tests. The full workspace typecheck passed all 44 tasks, the `eve` package typecheck passed, and `pnpm guard:invariants` passed.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

Adds the required patch changeset.

**Implementation** — 17 files · `+397 / -106`

Extends the strict trace protocol, transports it through remote create requests and the Eve channel, persists replay acknowledgements on continuation metadata, and stores confirmed trace IDs on remote handles.

**Tests** — 3 files · `+171 / -0`

Covers forwarded policy validation, exact acknowledgement, mismatched responses, compatibility fallback, and continuation metadata.

### DIFF
--- a/.changeset/trace-remote-delegated-sessions.md
+++ b/.changeset/trace-remote-delegated-sessions.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Carry delegated trace coordinates across remote eve session creation and acknowledge only exact child traces accepted by the receiver.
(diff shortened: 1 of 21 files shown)


chadhietala:
Superseded by the replacement delegated-tracing stack rooted at #2958.

PR #2969, opened 2026-09-03
fix(eve): release stopped stream response bodies

OwenKephart:
### Summary

This is the first PR in a four-layer CI reliability stack. Closing a client stream now also cancels its opened response body, so reconnecting or stopped consumers do not retain server-side stream listeners. The cleanup remains idempotent and preserves existing abort behavior.

### Validation

Focused unit coverage verifies the body is cancelled and the request signal is aborted exactly once. `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/client/open-stream.test.ts` and `pnpm --filter eve typecheck` passed. The stream-follow integration suite could not start because its Vitest config cannot resolve the pre-existing `#internal/workflow-bundle/workflow-builders.js` import.

### Checklist

### Diff size

**Docs** — 1 file · `+5 / -0`

The changeset records the observable resource-lifecycle fix for release notes.

**Implementation** — 1 file · `+9 / -2`

Keeps cleanup at the shared stream connection boundary.

**Tests** — 1 file · `+36 / -0`

Covers body cancellation and request abort idempotence.

### DIFF
--- a/.changeset/release-stream-follower-response-bodies.md
+++ b/.changeset/release-stream-follower-response-bodies.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Release response bodies when a client stops following a session stream, preventing retained stream listeners during repeated reconnects.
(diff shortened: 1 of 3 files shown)


OwenKephart:
Superseded by the equivalent, already-green #2928.

PR #2977, opened 2026-09-03
[3/3] feat(eve): add configuration options to slackChannel to populate manifest

benpankow:
### Summary

Allows Connect-backed `slackChannel()` definitions to configure the inline Slack app manifest: bot presentation, required and optional OAuth scopes, and Events API subscriptions. eve merges authored events and scopes with its baseline requirements; Connect adds the callback and trigger URLs when assembling the final requirement.

```ts
export default slackChannel({
  bot: {
    backgroundColor: "#000000",
    description: "Answers support questions.",
    name: "Support agent",
  },
  eventSubscriptions: ["message.channels", "message.groups"],
  scopes: ["channels:history", "groups:history"],
  optionalScopes: ["reactions:write"],
});
```

A scope listed in both `scopes` and `optionalScopes` remains required. eve does not infer requirements from open-ended handlers, so authors still declare their additional events and corresponding scopes. Built on #2496.

### Validation

Adds compiler and Slack manifest unit coverage for bot presentation, required and optional scope merging, and event subscriptions. Updates the Slack guide with the Connect-managed manifest flow.

🤖


allenzhou101:
looks like we remove `features.bot_user` from the comparison hash on the Connect side, but dont separately compare `always_online`.

if i only change `bot.alwaysOnline`, wouldnt this show in sync and skip the update? should include that in the comparison since we're exposing it in eve now.

https://github.com/vercel/api/blob/db6fa93fdb9c5e78f43521b5de1953afbfd70878/packages/connex/src/client-types/slack/provider-configuration.ts#L365-L367

packages/eve/src/internal/vercel-connect-manifest.test.ts
```diff
@@ -159,14 +159,23 @@ describe("buildVercelConnectRequirements", () => {
...
       channelRoutes: {
         effective: [
           {
             adapterKind: "slack",
             name: "support",
             logicalPath: "channels/support.ts",
+            slackAppManifest: {
+              alwaysOnline: true,
+              backgroundColor: "#000000",
+              botEvents: ["app_mention", "message.channels"],
```
dvoytenko: Why doesn't this just use the exact slack manifest format? It'd be hard to track all differences across different SDKs this way. We can just use the manifest schema directly and avoid translation layer.

benpankow: Addressed in the latest stack update: the compiler now preserves the completed Slack-native manifest rather than an eve-specific metadata shape, and the emitter writes that object unchanged. The Slack-owned builder is type-checked against the official SDK manifest request type.

packages/eve/src/public/channels/slack/app-manifest.ts
```diff
@@ -2,30 +2,79 @@ import type { AppsManifestCreateArguments } from "@slack/web-api";
...
           app_home: {
             home_tab_enabled: false,
             messages_tab_enabled: true,
             messages_tab_read_only_enabled: false,
           },
-          bot_user: { display_name: name, always_online: false },
+          bot_user: botUser,
         },
-        oauth_config: { scopes: { bot: ["app_mentions:read", "chat:write"] } },
+        oauth_config: { scopes: oauthScopes },
```
dvoytenko, line 65: redirect url is missing. or is it added elsewhere? Same for eventSubscriptions and interactivity

packages/eve/src/public/channels/slack/app-manifest.ts
```diff
@@ -2,30 +2,79 @@ import type { AppsManifestCreateArguments } from "@slack/web-api";
...
+        description?: string;
+        long_description?: string;
+        name: string;
+      } = { name };
+      if (input.backgroundColor !== undefined) {
+        displayInformation.background_color = input.backgroundColor;
+      }
+      if (input.description !== undefined) displayInformation.description = input.description;
+      if (input.longDescription !== undefined) {
+        displayInformation.long_description = input.longDescription;
```
dvoytenko: FYI: this one is always a pain with the rules like:

> The minimum length is 174 characters and the maximum length is 4,000 characters

dvoytenko (changes requested):

packages/eve/src/public/channels/slack/app-manifest.ts
benpankow: Added in #2496, `buildConnectSlackManifest` adds redirect URL, event subscription and interactivity URLs

packages/eve/src/public/channels/slack/app-manifest.ts
benpankow: Agreed; dropped `longDescription` for now

benpankow:
Superseded by #3470 on the replacement resource-snapshot stack.

PR #2986, opened 2026-09-03
test(e2e): make test theming less likely to trigger refusals

OwenKephart:
### Summary

The remote principal-forwarding eval used private-data and OAuth wording that caused a GPT-5.6 Sol provider-policy rejection before the behavior could run. Reframe the same per-turn authorization test around workspace labels, and fail immediately when a parent turn fails rather than waiting for a subagent event that cannot arrive.

The authorization invariant is unchanged: a durable remote child resolves the current caller's membership on each continuation and does not retain a previous caller's access.

### Validation

Confirmed with `pnpm --filter agent-subagents typecheck`; the full Local and Postgres E2E suites passed, including the OpenAI and Opus agent-subagents matrix jobs. Vercel E2E is still running.

### Checklist

No changeset is needed: this changes E2E fixtures only.

### Diff size

Not applicable.

Not applicable.

**Tests** — 5 files · `+65 / -66`

A focused fixture rename preserves the authorization scenario while avoiding policy-sensitive terminology; the eval now surfaces unrecoverable parent-turn failures directly.


ruiconti (approved):
Lol I fixed some of this for prompt caches too.

PR #2988, opened 2026-09-03
fix(eve): publish the MCP invocation contract to clients

cmpadden:
### Summary

Hosted MCP clients such as Grok Bot had to infer eve's durable invocation protocol from four terse tool descriptions, and the guide did not spell out the deployment choices that decide whether a public demo works at all. This PR publishes one concise operating contract through MCP server `instructions` (returned on both legacy `initialize` and modern `server/discover`) and tightened tool descriptions, then aligns `docs/channels/mcp.mdx` with it.

The contract covers: keep the invocation ID; poll no faster than `pollAfterMs`; branch on every `status`; answer the complete pending input batch in one `agent_update`; surface authorization instructions to a user; stop on terminal states; and treat a terminal `failed.error` differently from a tool-level `isError`. It also states two durability guarantees that previously lived nowhere: `agent_start` is not idempotent, so on an ambiguous transport failure the client should re-check with `agent_get` rather than retry blindly; and once `agent_start` has returned, a dropped connection does not cancel the work — only `agent_cancel` does.

The docs gain a "Durability guarantees" section, the recommendation to use `oauthResource()` over `none()` for any public demo, and a note that most hosted MCP clients require Dynamic Client Registration (RFC 7591) at the authorization server — eve does not run an AS, so the issuer must support DCR or the client must be pre-registered. Consolidates #2993, #2994, and the contract portion of #2995. Instructions are kept short because hosted clients truncate or ignore long ones.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/mcp.test.ts src/internal/mcp` — 30 passed; new assertions cover `instructions` on `initialize` and `server/discover`, and the four description contracts via `tools/list`.
- `pnpm fmt`, `pnpm lint`, `pnpm --filter eve typecheck`, `pnpm guard:invariants`, `pnpm docs:check` — all pass.
- Not run: scenario/e2e tiers (no behavioral change to invocation execution).

### Checklist

### DIFF
--- a/docs/channels/mcp.mdx
+++ b/docs/channels/mcp.mdx
@@ -67,6 +67,8 @@ The authorization server identified by `issuer` must issue tokens accepted by th
 
 `oauthResource()` does not issue tokens or run an authorization server. It decorates an ordinary inbound `AuthFn` with OAuth protected-resource metadata so `mcpChannel()` can publish discovery and add `resource_metadata` to Bearer challenges. Client registration, consent, token issuance, and authorization-server metadata remain the identity provider's responsibility.
 
+Most hosted MCP clients (Claude, ChatGPT, Grok, and similar) register themselves with the authorization server through Dynamic Client Registration ([RFC 7591](https://www.rfc-editor.org/rfc/rfc7591)) the first time a user connects. Choose an issuer that supports it, or pre-register each client with the issuer before the demo. If registration fails, the client stops before it ever reaches eve.
+
 ### Verify the bearer token yourself
 
 Use a custom `AuthFn` when token verification needs application-specific logic:
@@ -144,7 +146,7 @@ export default mcpChannel({
 });
 ```
 
-This allows anyone to invoke the agent. Every caller shares the anonymous principal, so invocation IDs become bearer capabilities until workflow retention expires. Use public access only when anonymous invocation is intentional.
+This allows anyone to invoke the agent. Every caller shares the anonymous principal, so invocation IDs become bearer capabilities until workflow retention expires: anyone holding an ID can read, answer, or cancel that invocation. Use public access only when anonymous invocation is intentional. For a public demo, prefer [Interactive OAuth](#interactive-oauth) so each caller owns their own invocations.
 
 ## HTTP security
 
@@ -171,6 +173,10 @@ MCP clients receive four tools:
 | `agent_update` | `{ invocationId, responses }` | Answer the complete pending human-input batch.              |
 | `agent_cancel` | `{ invocationId }`            | Request cooperative cancellation of non-terminal work.      |
 
+The server also returns `instructions` from `initialize` and `server/discover` that summarize this
+protocol for the connecting model, so a hosted client does not have to infer it from the tool
+schemas alone.
+
 `agent_start` creates one task-mode eve session and returns after durable acceptance without
 waiting for the session continuation hook to become readable. Keep its `invocationId`, then call
 `agent_get` until the invocation reaches a terminal state. While the status is `working`, wait at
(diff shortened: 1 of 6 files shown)


packages/eve/src/public/channels/mcp.ts
```diff
@@ -376,6 +377,22 @@ async function handleMcpRequest(
...
+ * drop long instructions, and the details live in each tool description.
+ */
+const MCP_SERVER_INSTRUCTIONS = [
+  "This server runs one durable agent task per invocation.",
+  "Call agent_start once per task and keep the returned invocationId.",
+  "Poll agent_get, waiting at least pollAfterMs between calls, until status is completed, failed, or cancelled.",
+  "If status is input_required, answer every entry in inputRequests with one agent_update call.",
+  "If status is authorization_required, show the user the authorization url or instructions and keep polling.",
+  "A tool result with isError describes a problem with your call; a failed status describes a problem with the task.",
+  "agent_start is not idempotent, so on an ambiguous transport failure re-check with agent_get instead of retrying blindly.",
```
allenzhou101: how would the client call `agent_get` here if the response containing the server-generated `invocationId` was lost? seems like we should tell it to ask the user instead, unless we add an idempotency key or client-provided id.

packages/eve/src/public/channels/mcp.ts
```diff
@@ -475,7 +501,8 @@ function createInvocationTools(
           readOnlyHint: false,
         },
         description:
-          "Requests cancellation of a durable invocation. Read it again to observe acknowledgement.",
+          "Requests cooperative cancellation of a non-terminal invocation. Cancellation is asynchronous: " +
+          "poll agent_get until status is cancelled. Safe to call repeatedly.",
```
allenzhou101: should this say to poll until any terminal state? cancellation can race with completion, so this could stay in a loop if the status becomes `completed` or `failed` instead of `cancelled`.

docs/channels/mcp.mdx
```diff
@@ -185,15 +191,20 @@ The invocation response is discriminated by `status`:
...
 
-## Invocation ownership
-
-Every MCP operation reruns the configured auth policy. With authenticated policies, an invocation belongs to the principal that started it; knowing its ID is not sufficient. Bearer tokens are not stored with the invocation or forwarded to the agent's tools.
+### Durability guarantees
 
-With `none()`, every caller shares the anonymous principal, so the random invocation ID becomes a bearer capability. Keep it out of logs and URLs and treat it as usable until workflow retention expires. Responses include `expiresAt` when the workflow backend reports a retention deadline.
+- Once `agent_start` returns, the work is durable. A dropped HTTP connection, a client restart, or a closed MCP session does not cancel it. Only `agent_cancel` stops work.
+- `agent_start` is not idempotent. If the response is lost, do not retry blindly: a second call starts a second task. Re-check state with `agent_get` or ask the user before starting again.
+- `agent_update` answers one pending batch. A second identical answer is rejected as a conflict; read the current state with `agent_get` instead.
```
allenzhou101: can we rebase this on #2990? that PR makes identical `agent_update` retries succeed, but this still says they’re rejected and that updates always return `working`. the retry should return whatever the current invocation state is.

allenzhou101 (approved):

docs/channels/mcp.mdx
```diff
@@ -185,15 +191,20 @@ The invocation response is discriminated by `status`:
...
+report the task failure in the second.
+
 Cancellation is cooperative, so call `agent_get` after `agent_cancel` until the state becomes terminal.
 
 Optional output schemas are limited to 64 KiB, 32 levels, and 2,048 nodes. External `$ref` values are rejected.
 
-## Invocation ownership
-
-Every MCP operation reruns the configured auth policy. With authenticated policies, an invocation belongs to the principal that started it; knowing its ID is not sufficient. Bearer tokens are not stored with the invocation or forwarded to the agent's tools.
+### Durability guarantees
```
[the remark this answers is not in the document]
cmpadden: Confirmed and fixed in 1a1e7bd43. The section removal in the amend was over-broad and took `## Invocation ownership` with it. Restored verbatim from main (per-request auth ownership, tokens not stored/forwarded, `none()` bearer-capability warning, `expiresAt`).

packages/eve/src/public/channels/mcp.ts
cmpadden: Good catch — a lost response means no `invocationId`, so `agent_get` isn't available. Reworded in 1a1e7bd43 to: ask the user before starting again rather than retrying. Applied to the server instructions, the `agent_start` description, and the docs durability bullet. (An idempotency key was proposed in #2995 and closed for now; the Workflow World has no attribute index, so it needs a deterministic-run-ID design first.)

packages/eve/src/public/channels/mcp.ts
cmpadden: Yes — fixed in 1a1e7bd43. `agent_cancel` description and docs now say to poll until any terminal status (`cancelled`, `completed`, or `failed`) and note that cancellation can race with completion.

docs/channels/mcp.mdx
cmpadden: Agreed on the contradiction. Rather than stack #2988 on #2990's branch, I aligned the wording here to #2990's behavior in 1a1e7bd43: repeating the same accepted answers returns the current invocation state; different answers for an answered batch are a conflict; and "returns fresh working state" is now "returns the current invocation state". #2990 carries the same sentence in the `input_required` bullet, so the two PRs no longer disagree — whichever lands second is a trivial textual rebase.

PR #2992, opened 2026-09-03
fix(eve): bound MCP request sizes

cmpadden:
### Summary

Modern MCP POSTs bypassed the 4 MiB inspection cap that only guarded the missing-`MCP-Protocol-Version` compatibility path, and tool fields such as `message` and input-response `text` had no explicit bounds, so a remote caller could push arbitrarily large bodies into the SDK and into durable storage. This PR applies one body limit before SDK processing and adds semantic bounds on the caller-supplied tool fields.

Rather than add a second inspection path, every POST body is now read once, bounded at 1 MiB, and the parsed value is always handed to the SDK. A declared `content-length` above the cap fails fast without reading; the streamed byte count remains the real guard. Over-limit bodies return a JSON-RPC-compatible `413` and malformed JSON a `-32700` parse error on every path, before auth-dependent tools run. `agent_start.message` is capped at 64 KiB, input-response `text` at 16 KiB, IDs at 256 chars, and responses per `agent_update` at 64, all via the tool input schemas so oversized input is rejected before `createSession` is called. Server-generated request IDs are not separately bounded.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/mcp.test.ts src/internal/mcp` — 33 passed. New tests cover oversized bodies on legacy-without-header, legacy-with-header, and modern requests; the `content-length` fast path; malformed JSON with the modern header; and oversized `message`/`text` never reaching `createSession`.
- `pnpm build` then `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/internal/invocation/mcp-agent-channel.scenario.test.ts` — passed (full modern + legacy lifecycle through a real `eve dev` process), since this changes the request path for all POSTs.
- `pnpm fmt`, `pnpm lint`, `pnpm --filter eve typecheck`, `pnpm guard:invariants`, `pnpm docs:check` — all pass.

### Checklist

### DIFF
--- a/packages/eve/src/public/channels/mcp.ts
+++ b/packages/eve/src/public/channels/mcp.ts
@@ -420,7 +420,7 @@ function createInvocationTools(
             ? startDescription
             : `${agentDescription} ${startDescription}`,
         inputSchema: z.strictObject({
-          message: z.string().min(1),
+          message: utf8Bounded(MAX_MESSAGE_BYTES).min(1),
           outputSchema: z.looseObject({}).optional(),
         }),
         name: "agent_start",
@@ -447,7 +447,7 @@ function createInvocationTools(
           "Reads complete durable invocation state. Wait at least pollAfterMs between calls. " +
           "Terminal statuses are completed, failed, and cancelled. input_required needs agent_update; " +
           `authorization_required needs the user to follow the returned authorization.${publicHandleDescription}`,
-        inputSchema: z.strictObject({ invocationId: z.string().min(1) }),
+        inputSchema: z.strictObject({ invocationId: z.string().min(1).max(MAX_ID_CHARS) }),
         name: "agent_get",
         outputSchema: AGENT_INVOCATION_OUTPUT_SCHEMA,
       },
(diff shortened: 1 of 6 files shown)


packages/eve/src/public/channels/mcp.ts
```diff
@@ -541,6 +541,19 @@ const AUTHORIZATION_REQUEST_SCHEMA = z.strictObject({
   webhookUrl: z.url().optional(),
 });
 
+// Bounds on caller-supplied text. The transport already caps the whole body
+// at MCP_REQUEST_BODY_MAX_BYTES; these keep individual durable fields sane.
+const MAX_MESSAGE_CHARS = 64 * 1_024;
```
allenzhou101: these look like character limits, but the docs describe them as KiB limits. multibyte input can be quite a bit larger than 64/16 KiB. should we validate the encoded byte length, or call these character limits in the docs?

allenzhou101 (approved):

packages/eve/src/public/channels/mcp.ts
cmpadden: Right — those were UTF-16 code-unit bounds mislabeled as KiB. Switched to encoded byte length in b6e814625: `message` and `text` now use a `.refine()` on `TextEncoder().encode(...).byteLength`, constants renamed `*_BYTES`, and the docs say "measured as UTF-8 bytes, not characters". The test now uses a CJK string that's well under the character count but over the byte limit. IDs stay as character bounds since they're opaque ASCII hashes.

PR #2995, opened 2026-09-03
feat(eve): make MCP agent starts idempotent

cmpadden:
### Summary

A retry of `agent_start` after an ambiguous HTTP failure can create duplicate durable work, which is unsafe for a hosted MCP client executing side-effectful tasks. This proposal adds an optional owner-scoped client idempotency key: equivalent retries return the original invocation, while a reused key with different input reports a stable conflict. Calls without a key retain today’s behavior.

### Validation

Plan-only draft; implementation validation is not yet run. The implementation will cover retry-after-acceptance, concurrent duplicate starts, conflicting key reuse, ownership isolation, retention cleanup, and existing no-key behavior.

### Checklist


cmpadden:
Closing for now. The Workflow World API only indexes runs by status/workflowName, so an owner-scoped idempotency-key lookup needs either an unbounded scan or a new durable store — far larger than this PR implied, and it would touch session creation in execution/ (requiring a research doc). Hosted MCP clients also do not auto-retry tools/call, so the demo value is low. The contract ("agent_start is not idempotent; re-decide on ambiguous failure rather than blindly retry") is folded into #2988. Can be reopened with a deterministic (ownerKey, idempotencyKey) run-ID design if needed.

Issue #874, opened 2026-07-16
Telegram: group mentions start a new session per bot message; key groups chat-wide like private chats


Lukavyi:
### Bug

A Telegram bot in a **group** has no memory between mentions. Every `@bot` mention that is not a reply to a bot message starts a new session with one message of context.

Cause: group and supergroup inbound messages key the session to a bot message id (`telegram:<chatId>:<topicId>:<botMessageId>`, see `conversationIdForMessage` in `telegram/state.ts`). Only a reply-chain to one specific bot message keeps continuity. Private chats already key chat-wide (`telegram:<chatId>:<topicId>:`) since #377; groups are the only surface that resets.

Timeline in one group:

1. U2: `@bot` question A → session `chat::101`
2. U2 replies to the bot's answer → same session, fine
3. U1: `@bot` question B (fresh mention, not a reply) → **new session**, blind to A
4. U2 replies to the old answer A a day later → resumes `chat::101`, blind to everything since

### Fix

Key groups the same way as private chats: one session per chat, or per forum topic. Mentions, replies to bot messages, callback queries and outbound sends all resume it.

- **PR #3020** does exactly this. It also treats `message_thread_id` as a session discriminator only when `is_topic_message` is set, because non-forum supergroups stamp reply-chain ids on replies and those must not fragment the chat-wide session.
- Per-bot-message anchoring is removed rather than kept behind an option: it stays reproducible via `conversationId` on a proactive `receive` target (and #671 would cover inbound), while the reverse was not possible without patching channel internals. Slack and Teams already key sessions to the platform's visible container (a thread); see the channel comparison in the PR.

We have run chat-wide keying as a `patchedDependencies` patch in production on two group bots since July.

### Follow-up, separate PR

Unaddressed group messages are dropped by the dispatch gate and reach the session nowhere, so even with one chat-wide session the bot does not see what people said between mentions. Slack has `threadContext` for this. The plan is an `observe` result from `onMessage` that appends a message to the session as history without running a turn; that touches core delivery and will come as a stacked PR after #3020.

### Related

#364 / #377 (where group anchoring was kept), #671 (host-controlled continuation token), #897 (fixed, hit while building the observe buffer), #211 (same class of bug in Teams, fixed as a bug), #223 (Slack: respond without mentions).

<details>
<summary>Original report (long version with diagrams, proposed API shape and edge cases)</summary>

### What problem are you trying to solve?

[...]

Issue #671, opened 2026-07-09
Let the host pick the session for an inbound turn (continuation token override)


ishaan812:
### What problem are you trying to solve?

I’m building a personal agent where a single Telegram chat hosts multiple
sessions and the user can run /new-session, /list-sessions, /resume 2
and the same chat switches between them (same idea as “new chat” in any
chat product, just on a messaging surface). I keep the session registry on
my side; all I need from eve is to control which session an inbound message
lands on.

Turns out that’s not possible right now. Every channel computes the
continuation token internally and there’s no way for the host to influence
it. For a private Telegram chat the token is pinned to ${chatId}::
(continuationTokenFromState drops the conversationId), so the chat is
welded to one session forever. The new chatSdkChannel has the same
property, the token is hardcoded to the serialized thread id at both
call sites (bridgeSend and receive), and ChatSdkSendOptions doesn’t
let you override it even though I’m the one writing the handler and calling
send.

Things I tried before filing this:

session.setContinuationToken() - only works on the live session while
it’s running a step. getSession(id) returns the inert Session, which
per the docs can’t mutate the token, so I can’t re-key a parked session
from the outside.
Writing a custom channel. The Telegram inbound helpers are all exported,
but buildTelegramHandle isn’t, so owning dispatch means re-implementing
~100 lines of sends/splitting/typing/HITL that you’ve already
battle-tested. Didn’t want to maintain that fork.
Passing a different thread to chatSdkChannel’s send (“thread == session,
so mint one thread per session”). Doesn’t survive contact with adapters:
@chat-adapter/telegram encodes telegram:<chatId>[:<topicId>], validates
the shape on decode, and parses the delivery target out of the id — DMs
have exactly one possible thread.
What I ended up doing is patching eve via pnpm patchedDependencies (in
production since 0.12.0): a resolveContinuationToken config hook on
telegramChannel, consulted at both dispatch send-sites, null falls back
to the default token. It’s tiny and it’s been solid, but I’d love to delete
the patch.

### Proposed solution

Two options, happy to PR either or both:

1. continuationToken on ChatSdkSendOptions. In the chat-sdk channel
I already author the handler, so an explicit override is the natural shape:

bot.onDirectMessage(async (thread, message) => {
  const sessionKey = await controlPlane.activeSession(thread.id); // mine
  await send(messageToUserContent(message), { thread, continuationToken: sessionKey });
});
Implementation is basically options.continuationToken ?? serialized.id in
[...]

Issue #211, opened 2026-06-23
Teams Channel - Calls from a Channel thread lose the context from earlier messages in the same thread


andrewgreenh:
### eve version

eve@0.13.0

### Node.js version

v24.16.0

### Where does the bug occur?

Production (`eve start` / deployed)

### Describe the bug

In the Microsoft Teams channel, a follow-up `@mention` inside the same Teams thread can start a new eve session instead of continuing the existing one.

The Teams webhook payloads I observed keep the same `conversation.id` across the root message and the follow-up reply, but the follow-up activity does not include `replyToId`. Teams appears to encode the root message id in the `conversation.id` as `;messageid=<rootActivityId>`.

Because eve currently derives the Teams thread root from `replyToId ?? id`, the first message and follow-up message get different `replyToActivityId` values. That changes the Teams continuation token and causes eve to treat the follow-up as a new conversation, so the agent loses prior context.

The code path that appears to trigger this is:

- `packages/eve/src/public/channels/teams/inbound.ts`
  - `teamsThreadRootActivityId(...)` currently returns `activity.replyToId ?? activity.id` for non-personal messages.
- `packages/eve/src/public/channels/teams/teamsChannel.ts`
  - `stateFromActivity(...)` stores that value as `replyToActivityId`.
  - `stateToken(...)` includes `replyToActivityId` in `teamsContinuationToken(...)`.

Relevant source reference:
https://github.com/vercel/eve/blob/main/packages/eve/src/public/channels/teams/inbound.ts

Observed first Teams message:

```json
{
  activityId: "1782227500445",
  botMentioned: true,
  channelId: "19:f8f4483275ae45f79540448023eb9164@thread.skype",
  conversationId: "19:f8f4483275ae45f79540448023eb9164@thread.skype;messageid=1782227500445",
  conversationType: "channel",
  replyToId: undefined,
  rootActivityId: "1782227500445",
  scope: "channel",
  teamId: "19:d3039cb8cc7a425daf753c8b91aa9035@thread.skype",
  tenantId: "874d7388-04b4-4847-b873-68164db3f140"
}
```

Observed follow-up message in the same Teams thread:

```json
{
  activityId: "1782227547644",
  botMentioned: true,
  channelId: "19:f8f4483275ae45f79540448023eb9164@thread.skype",
  conversationId: "19:f8f4483275ae45f79540448023eb9164@thread.skype;messageid=1782227500445",
  conversationType: "channel",
  replyToId: undefined,
[...]

Issue #897, opened 2026-07-17
coalesceDeliverPayloads silently drops earlier messages in a batched delivery (last-key-wins for message/context)


Lukavyi:
### Summary

`coalesceDeliverPayloads` (`src/execution/deliver-payloads.js`) merges a multi-payload delivery batch with **last-key-wins** semantics for every field except `inputResponses`:

```js
for (let r of e) {
  for (let [k, v] of Object.entries(r))
    k !== `inputResponses` && v !== void 0 && (t[k] = v);
  ...
}
```

So when a parked session receives several deliveries that get batched into one turn, only the **last** payload's `message` and `context` survive — every earlier message in the batch is silently discarded after being accepted.

### Where it bites in stock eve (0.22.3)

The driver loop's `waitForNextDeliver` (`workflow-entry.js`) deliberately drains all ready deliveries and coalesces them into one delivery with multiple payloads (`coalesceDeliveries` concatenates `payloads`). That batch then goes through `routeDeliverToChildren` → `coalesceDeliverPayloads` → a single payload for `dispatchAndAwaitTurn`.

Repro: a user sends two messages in quick succession to a parked session (or two users hit the same group session at once). Both deliveries resolve while the workflow is waking, get batched, and the turn only contains the second message. The first one is gone — no log, no error; the model just never sees it. The same happens to `context` entries (e.g. the Telegram channel's per-message `<telegram_context>` block: only the last message's block survives).

The asymmetry is confusing too: `inputResponses` are correctly concatenated across the batch, and the harness-level `coalesceTurnInputs` (`harness/messages.js`) already implements proper merge semantics for exactly these fields — string messages joined with a blank line, content arrays concatenated, contexts concatenated. It's only this merge point that loses data.

### Suggested fix

Give `coalesceDeliverPayloads` the same merge semantics as `coalesceTurnInputs`:

- `message`: reduce across payloads — `a + "\n\n" + b` when both are strings, otherwise merge into one user-content array;
- `context`: concatenate in payload order;
- `inputResponses`: unchanged (already concatenated);
- other keys: last-wins is fine (they're scalars like `outputSchema`).

We hit this in production (Telegram group bot, patched eve via `patchedDependencies`) and have been running exactly that change; happy to PR it.

### Related

- #874 — the batching path gets much hotter with chat-wide group sessions, which is how we found it, but the loss itself reproduces in unpatched eve with any coalesced delivery batch.

NicoMoli:
[...]

PR #3020, opened 2026-09-04
[1/2] feat(eve)!: key Telegram group sessions chat-wide

Lukavyi:
### Summary

A Telegram bot in a group has no memory between mentions. Group and supergroup inbound messages key the session to the bot message being replied to, or to the mention itself when it is not a reply (`conversationIdForMessage` in `telegram/state.ts`), and outbound group sends re-key the session to the posted message id. Every fresh `@bot` mention therefore starts a new session with one message of context, while private chats already keep one continuous session per chat since #377.

This makes groups behave like private chats: one session per chat, or per forum topic, keyed to `telegram:<chatId>:<messageThreadId>:`. Mentions, replies to bot messages, HITL callback queries, and outbound sends all resume that session, so the bot remembers what it said earlier in the chat.

- Non-forum supergroups stamp `message_thread_id` on replies (the reply-chain root, not a topic). The thread id now counts as a session discriminator only when Telegram marks the message with `is_topic_message`, so a reply chain does not fragment the chat-wide session. `isTopicMessage` is exposed on `TelegramMessage` and `TelegramMessageReference` for this.
- The per-bot-message anchoring is removed rather than kept behind an option, following the reasoning in #874: it is reproducible in userland (`conversationId` on a proactive `receive` target still pins a thread, and #671 would cover inbound), while the reverse was not possible without patching channel internals. A late reply to an old bot message also resumed a session blind to everything since, so the anchoring was not correct even for the multi-thread case it served.

- Second commit: a reply to a bot message is delivered with its text as `message` alongside the synthetic `telegram_reply:<id>` response again. The initial channel did this; #1597 switched the reply path to `respond()` with only the response, so a reply that was not an answer to a pending ForceReply prompt resolved to nothing in the adapter and the session re-parked without a turn. With chat-wide sessions a reply to the bot is the ordinary follow-up, so this matters more than before. The adapter already had the fallback and the test for it.

**Behavior change** for existing group bots, hence the `minor` changeset and the `!` in the commit. Existing parked group sessions keep their old anchored tokens; new mentions start on the chat-wide token.

We have run chat-wide group keying as a `patchedDependencies` patch in production on two group bots since July.

Related to #874 (option 1 there). #671 is the more general mechanism; this is the smaller change that fixes the group use case now.

### How this lines up with the other channels

| Channel | Session key | Boundary |
|---|---|---|
| Slack | `channel:threadTs` | a thread, visible to the user |
| Teams | `tenant:conversation:rootActivityId` | a thread, visible to the user |
| Discord | `channel:interactionId` | one slash command; the channel has no inbound messages |
| Telegram (before) | `chat:topic:botMessageId` | a reply chain to one bot message; not visible to the user |
| Telegram (after) | `chat:topic:` | the chat or forum topic, visible to the user |

Slack and Teams key a session to the smallest container the platform itself shows: a thread. Telegram has no threads outside forum topics, and instead of taking the nearest visible container (chat or topic) the channel synthesized its own boundary from the reply chain to a bot message. Users cannot see that boundary, which is why a plain `@bot` mention that is not a reply reads as "the same conversation" to them and as "a new session" to eve. Teams had the same class of bug in #211 (a follow-up mention in a thread lost context) and it was fixed as a bug, not made an option.

The Telegram and Discord channels share one shape from the initial commit (`conversationId` state, `anchor()` after outbound posts, the same `conversationId`/`initialMessage` guard). Discord is interaction-only, so "one session per bot post" is the natural unit there. Telegram inherited that unit for groups; private chats were exempted from the start, and #377 later kept the group anchoring as-is while fixing proactive private chats. After this change Telegram behaves like Slack and Teams, and Discord stays the justified exception.

### Validation

```
pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/telegram   # 63 passed
pnpm --filter eve exec tsc -p tsconfig.json --noEmit
pnpm run docs:check
pnpm run guard:invariants
```

New coverage: a fresh mention, a reply to a bot message (freeform HITL path), a mention inside a forum topic (`is_topic_message: true` keeps the topic), a reply-chain `message_thread_id` in a non-forum supergroup (stays chat-wide), a HITL callback query, and an outbound post that must not re-key the session. The existing anchoring tests were rewritten to assert chat-wide tokens; the one that asserted re-keying to the posted message id was dropped, and the explicit-`conversationId` group test now asserts the id is preserved instead of overwritten.

### Checklist

- [ ] This change was requested or approved by a maintainer (asked in #874, awaiting an answer)

https://claude.ai/code/session_015P6RBuagWTMzBmJLCiYKW9


Lukavyi:
Pushed 304b1ea: a reply to a bot message now carries its text as `message` on the same delivery as the synthetic `telegram_reply:<id>` response. The adapter still resolves a pending freeform answer first; when no prompt is registered, the text runs as a normal turn instead of the session re-parking without a turn. Also corrected the `TelegramReceiveTarget` docstring (`conversationId` pins an id that inbound never routes to; `initialMessage` stays chat-wide in groups). Telegram unit suite: 62 passed; tsc, guard:invariants, docs:check clean.

Lukavyi:
Ready for review. @benpankow (recent Telegram HITL work) @ruiconti (merged #897, which this relies on): could one of you take a look? #3023 stacks on top with the observe half.

PR #3063, opened 2026-09-04
feat(eve): sessions - replace parent orchestration with holding workflows

AndrewBarba:
### Summary

Keeping stable session streams in a parent workflow pins turn execution to long-lived orchestration code. A small holding workflow now owns session resources and continuation aliases, while independent terminating workflows execute turns through unified owner inboxes. This draft focuses on the eve architecture and its interfaces; the cross-workflow stream writer remains a declaration stub pending SDK support. Existing-session migration is deferred.

### Validation

- Reset to the architecture and cleanup revision before the Workflow SDK backport. No dependency patches or subsequent SDK performance experiments are included.
- `pnpm install --frozen-lockfile` and `pnpm --filter eve build` passed after clearing the generated SDK and package outputs. The installed SDK is unpatched; `Run.getWritable`, `Run.getStreamReference`, and `Run.getReturnValue` are absent.
- Full stream lifecycle integration and end-to-end behavior remain blocked by the missing `Run.getWritable()` API. The stub fails explicitly when that API is unavailable.
- Performance results from the later patched revisions do not describe this draft revision. The architecture remains under review and this draft is not ready to merge.

### Checklist


pranaygp:
Suggestion: `@workflow/core@5.0.0-beta.48` (published Sep 3) makes the "replay a durable hook claim in-process" amendment unnecessary, and is cheaper than it.

[vercel/workflow#3938](https://github.com/vercel/workflow/pull/3938) adds `continueOverHookWrite` in `runtime.ts`. After a suspension commits the `hook_created` (or `hook_conflict`) that an awaited `hook.getConflict()` is parked on, it resumes the retained VM in the same invocation instead of `reinvoke(0)`. The hook create sends `sinceCursor`, and when the response carries the complete delta (`eventLogCarriedForward`) the next iteration reads nothing. The local patch instead sets `retainedSession = null` and does `nextEventLogLoad` before a cold replay, so beta.48 removes the cold replay everywhere and the incremental `events.list` wherever the delta applies.

Conditions:

- Zero reads only when the suspension created exactly one hook (`hookDeltaCursor` is unset for two), the loaded log had a cursor, no wait is open, and the World returns the delta. Hosted server support is [workflow-server#913](https://github.com/vercel/workflow-server/pull/913), merged Sep 3.
- Otherwise beta.48 does one incremental read from the cursor, still in-process on the retained VM. `WORKFLOW_RETAINED_VM=0` restores the re-invocation.

How that lands on eve's two claim shapes:

- Holder control and alias claims are sequential single-hook suspensions, so they qualify for the zero-read path.
- The turn claim creates two hooks in one suspension: `eve:turn:<sessionId>` plus the `abrt_*` system hook, because the workflow VM's `AbortController` is hook-backed (the native storage test asserts `hookCreations` of `[1, 1]`). So the turn still pays one incremental `events.list` under beta.48, but no cold replay. Constructing the `AbortController` after the claim resolves, or carrying cancellation on the inbox hook, would get the turn to zero reads too.

The other amendments (cache immutable stream ownership, immutable stream references, resume on individual step completion, 1 ms stream group-commit window) are not in beta.48 and would need rebasing onto it, plus bumping `@workflow/world-vercel` to beta.44. A paired hosted stress run would confirm the saving.

I also pushed `research/session-wire-budget.html` to this branch: a per-path map of the workflow-server and VQS round trips from ingress to the first model call, with the hosted phase timings from the stress run. Open it locally in a browser.

PR #3065, opened 2026-09-04
fix(eve): distinguish background task wake-ups

OwenKephart:
### Summary

Much smaller change than it appears. When a background task wakes up the outer agent, that is done with a `message.received` event, which is the same as the event type used for user messages. A recent addition added discriminators in to these events (kind tags) and so at the model history layer, we could distinguish between a user message and a background task completion.

However, that kind was not carried through the session stream projection (emitTurnPreamble). This updates the MessageReceivedEvent type to include the `execution.background_task` source so that consumers of the stream can differentiate between regular usage messages and background task completions, and adds a quick filter in the reducer used by the frontend to not treat those completions as user messages.

#### Before

https://github.com/user-attachments/assets/f11b1586-27ee-4db1-9c59-75cd769d1652

#### After

https://github.com/user-attachments/assets/f095e8b3-e12f-4c3e-8b1b-bc80a6f01535

### Validation

- `pnpm --filter eve test:unit` — 8,909 passed, 1 skipped
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/tools/workflow/workflow-tool-run.integration.test.ts` — 24 passed
- `pnpm update:extension-contracts`
- `pnpm guard:invariants`
- `pnpm fmt`
- `pnpm lint` — passed with existing warnings
- `pnpm docs:check`
- `pnpm --filter agent-background-tools typecheck`
- `cd e2e/fixtures/agent-background-tools && EVE_E2E_MODEL=mock pnpm exec eve eval --strict --exclude-tag real-model` — 3 evals passed, 41 gates passed

### Checklist


packages/eve/src/harness/emission.ts
```diff
@@ -100,8 +101,10 @@ export async function emitTurnPreamble(
   }
 
   if (input.message !== undefined) {
+    const kind = frameworkMessageKindForStepInput(input);
     await emitFn(
       createMessageReceivedEvent({
+        kind: kind === "execution.background_task" ? kind : undefined,
```
ruiconti, line 107: Thought we were already exposing `kind` in events

ruiconti: cc @e0 fact check me

OwenKephart: bot is right yeah

ruiconti: @e0 so this means we strip the user-role model messages' `.kind` property when the model messages are emitted as events? just want to confirm we are not doing something like

```ts
{
 {
   // ...
   kind: 'background_task',
   data: {
     message: 'foo',
     parts: [
       { role: 'user', kind: 'background_task', text: 'foo' }
     ]
   }
}
```

ruiconti (approved):

PR #3076, opened 2026-09-05
fix(eve): point eve-source exports at dist

### LINKED ISSUE #3031: Tool with workflow broken

https://github.com/dummdidumm/eve-workflow-bug-repro

### Steps to reproduce

1. `pnpm install`
2. `pnpx eve dev`
3. observe error

Alternatively

1. `pnpm build && pnpm verify:broken-bundle`
2. (you could also publish it to Vercel and then see it fail there once you try to talk to it)

Tools with `"use workflow"` should work - right now they just don't, because `eve-source` condition is requested by `WorkflowBundleBuilder`, which resolves to nothing because the eve npm package does not include the `src` folder but some of the export conditions in `package.json` point to it.

A local patch of pointing to dist instead of src fixes the issue.

### eve version

eve@0.51.1

### Environment

```shell
OS: Linux
Node: v24.18.0
Package manager: pnpm@11
```

### Where does the bug occur?

### Deployment

_No response_

### Build and runtime logs

```shell
irrelevant
```

_No response_

### Additional context

_No response_

ShobhanKarthish:
Closes #3031

## Summary
- Repoint published eve-source export conditions from ./src to ./dist/src (workflow-modules, self-modification, setup).
- Leave package-local #*.js imports on ./src/*.ts for the eve build.

## Test plan
- [x] Confirmed new eve-source targets exist in the eve@0.52.1 tarball
- [ ] Reproduce with dummdidumm/eve-workflow-bug-repro against this change

### DIFF
--- a/.changeset/eve-source-dist-exports.md
+++ b/.changeset/eve-source-dist-exports.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Point published `eve-source` export conditions at `dist` so Workflow bundle resolution works without the unpublished `src` tree.
--- a/packages/eve/package.json
+++ b/packages/eve/package.json
@@ -182,7 +182,7 @@
       "default": "./dist/src/public/tools/workflow.js"
     },
     "./workflow-modules": {
-      "eve-source": "./src/public/workflow-modules.d.ts",
+      "eve-source": "./dist/src/public/workflow-modules.d.ts",
       "types": "./dist/src/public/workflow-modules.d.ts"
     },
     "./models/openai": {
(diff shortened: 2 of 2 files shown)


AndrewBarba (approved):

AndrewBarba (approved):
Verified against latest main (1807ff9c5); the updated PR head a02e7c0 has exactly the same tree as the locally tested merge.

Confirmed the original 0.51.1 reproduction emits require("eve/workflow") and fails while evaluating the generated driver. Correcting that old export to dist removes the require and the driver evaluates successfully. Current main has intentionally removed eve/workflow in favor of defineWorkflowTool and ctx.agent; the migrated reproduction works on main already.

Separately confirmed a live regression covered by this PR: a defineWorkflowTool executor returning DEFAULT_SELF_MODIFICATION_MODEL imported from eve/self-modification/agent builds a driver containing require("eve/self-modification/agent") with main's exports, and evaluating that driver throws ReferenceError: require is not defined. Applying exactly this PR's export changes removes that require and the generated driver evaluates successfully.

All five changed JavaScript exports bundle without unresolved imports using eve's actual Nitro Rolldown wrapper, createEvePackageImportsPlugin, and workflow step external-package policy. Main's export map fails all five probes. The package build and pack succeed, and the workflow-bundle plus packed self-modification consumption scenarios pass (13 tests).

Approved on code correctness; merge remains contingent on required CI passing.

AndrewBarba (approved):
Verified against latest main (1807ff9c5); the updated PR head a02e7c0 has exactly the same tree as the locally tested merge.

Confirmed the original 0.51.1 reproduction emits require("eve/workflow") and fails while evaluating the generated driver. Correcting that old export to dist removes the require and the driver evaluates successfully. Current main has intentionally removed eve/workflow in favor of defineWorkflowTool and ctx.agent; the migrated reproduction works on main already.

Separately confirmed a live regression covered by this PR: a defineWorkflowTool executor returning DEFAULT_SELF_MODIFICATION_MODEL imported from eve/self-modification/agent builds a driver containing require("eve/self-modification/agent") with main's exports, and evaluating that driver throws ReferenceError: require is not defined. Applying exactly this PR's export changes removes that require and the generated driver evaluates successfully.

All five changed JavaScript exports bundle without unresolved imports using eve's actual Nitro Rolldown wrapper, createEvePackageImportsPlugin, and workflow step external-package policy. Main's export map fails all five probes. The package build and pack succeed, and the workflow-bundle plus packed self-modification consumption scenarios pass (13 tests).

Approved on code correctness; merge remains contingent on required CI passing.

PR #3082, opened 2026-09-05
fix(eve): resolve ten session, tool, and integration failures

### LINKED ISSUE #2195: Title: ESM compat banner double-declares __dirname/__filename for comma-list CJS-global declarations → SyntaxError: Identifier '__dirname' has already been declared at load

not needed :) 

### Steps to reproduce

# Minimal reproduction (no agent needed)

``` 
import { buildNodeEsmCompatBanner } from "eve/dist/src/internal/node-esm-compat-banner.js";

// comma-list form, as emitted by the vendored @vercel/otel build:
const chunk = `var _x=(import.meta.url),__filename=fileURLToPath(import.meta.url),__dirname=_nPath.dirname(__filename);`;

console.log(buildNodeEsmCompatBanner(chunk));
// Actual:   "import { dirname as __eveDirname } from \"node:path\";\nconst __dirname = …"
//           → prepended banner re-declares __dirname/__filename → SyntaxError once concatenated.
// Expected: "" (chunk already binds both globals)
```

# Summary

`createNodeEsmCompatBannerPlugin` (`internal/node-esm-compat-banner.js`) prepends a `const __dirname = …` / `const __filename = …` banner to each bundled chunk, and skips its own line when the chunk already declares the binding. The "already declared" detection only recognizes a declaration that starts a line in `const|let|var __dirname` form:

```
bindingPattern: /^(?:const|let|var)\s+__filename(?![\w$])/m
bindingPattern: /^(?:const|let|var)\s+__dirname(?![\w$])/m
```

It does not match a comma-list declaration (`__filename=…`,`__dirname=…`), which is exactly what eve's own vendored `@vercel/otel` build emits:

```
// node_modules/eve/dist/src/compiled/@vercel/otel/index.js  (single-line/minified)
…(import.meta.url),__filename=fileURLToPath(import.meta.url),__dirname=_nPath.dirname(__filename)…
```

When a chunk that inlines this code is bannered, the plugin fails to detect the existing `__dirname`/`__filename` and prepends its own `const __dirname = …` → two top-level `__dirname` bindings → the runtime error, matching the exact failure mode the banner's own doc comment warns about.

# Impact

Any agent whose bundle pulls in `@vercel/otel` (i.e. the vendored Workflow DevKit / tracing path) fails to load in `eve dev`. Every route that loads the agent graph 500s:
```
ERROR [request error] [GET] /eve/v1/info
HTTPError: Identifier '__dirname' has already been declared
    at compileSourceTextModule (node:internal/modules/esm/utils)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators)
    …
```

### eve version

eve@0.38.3

### Environment

```shell
OS: MacOS Sonoma 14.7.5
Node: 24.19.0
```

### Where does the bug occur?

Local dev (`eve dev`)

### Deployment

_No response_

### Build and runtime logs

```shell
{
  "error": true,
  "stack": [
    "Identifier '__dirname' has already been declared",
    "at compileSourceTextModule (node:internal/modules/esm/utils:318:16)",
    "at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:89:18)",
    "at #translate (node:internal/modules/esm/loader:434:20)",
    "at afterLoad (node:internal/modules/esm/loader:502:29)",
    "at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:507:12)",
    "at #getOrCreateModuleJobAfterResolve (node:internal/modules/esm/loader:560:36)",
    "at afterResolve (node:internal/modules/esm/loader:607:52)",
    "at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:613:12)",
    "at node:internal/modules/esm/loader:632:32",
    "at TracingChannel.tracePromise (node:diagnostics_channel:362:14)"
  ],
  "status": 500,
  "unhandled": true,
  "message": "HTTPError"
}
```

### Additional context

eve version: 0.38.3 (latest) · Node: ESM · OS: macOS (platform-independent)

# Summary

`createNodeEsmCompatBannerPlugin` (`internal/node-esm-compat-banner.js`) prepends a `const __dirname = …` / `const __filename = …` banner to each bundled chunk, and skips its own line when the chunk already declares the binding. The "already declared" detection only recognizes a declaration that starts a line in `const|let|var __dirname` form:

```
bindingPattern: /^(?:const|let|var)\s+__filename(?![\w$])/m
bindingPattern: /^(?:const|let|var)\s+__dirname(?![\w$])/m
```

It does not match a comma-list declaration (`__filename=…`,`__dirname=…`), which is exactly what eve's own vendored `@vercel/otel` build emits:

```
// node_modules/eve/dist/src/compiled/@vercel/otel/index.js  (single-line/minified)
…(import.meta.url),__filename=fileURLToPath(import.meta.url),__dirname=_nPath.dirname(__filename)…
```

When a chunk that inlines this code is bannered, the plugin fails to detect the existing `__dirname`/`__filename` and prepends its own `const __dirname = …` → two top-level `__dirname` bindings → the runtime error, matching the exact failure mode the banner's own doc comment warns about.

# Impact

Any agent whose bundle pulls in `@vercel/otel` (i.e. the vendored Workflow DevKit / tracing path) fails to load in `eve dev`. Every route that loads the agent graph 500s:
```
ERROR [request error] [GET] /eve/v1/info
HTTPError: Identifier '__dirname' has already been declared
    at compileSourceTextModule (node:internal/modules/esm/utils)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators)
    …
```

# Minimal reproduction (no agent needed)

``` 
import { buildNodeEsmCompatBanner } from "eve/dist/src/internal/node-esm-compat-banner.js";

// comma-list form, as emitted by the vendored @vercel/otel build:
const chunk = `var _x=(import.meta.url),__filename=fileURLToPath(import.meta.url),__dirname=_nPath.dirname(__filename);`;

console.log(buildNodeEsmCompatBanner(chunk));
// Actual:   "import { dirname as __eveDirname } from \"node:path\";\nconst __dirname = …"
//           → prepended banner re-declares __dirname/__filename → SyntaxError once concatenated.
// Expected: "" (chunk already binds both globals)
```

# Root cause

`bindingPattern`'s `^…/m` anchor only detects line-leading `const|let|var __dirname` and misses comma-list / non-line-start declarations produced by bundlers (esbuild-style `var a=…,__filename=…,__dirname=…`).

# Suggested fix

Broaden `bindingPattern` to detect a binding/assignment of the global anywhere it isn't a property access or a comparison:
```
// __filename
bindingPattern: /(?<![\w$.])__filename\s*=(?!=)/
// __dirname
bindingPattern: /(?<![\w$.])__dirname\s*=(?!=)/
// require
bindingPattern: /(?<![\w$.])require\s*=(?!=)/
```

- `(?<![\w$.])` excludes `foo__dirname` / `x.__dirname`.
- `\s*=(?!=)` requires an actual assignment (still shims read-only usage), excludes `==`/`===`.

Verified against the plugin: comma-list chunks now produce an empty banner (no double-declare), chunks that only use `__dirname` still receive the shim, and line-leading `const __dirname = …` chunks are unchanged.

### LINKED ISSUE #2319: Dynamic tools: approvalKey is dropped by toHarnessToolDefinition, silently downgrading scoped approvals to bare tool names
**Version:** eve 0.39.1 (also 0.39.0)

## What happens

`ToolDefinition.approvalKey` is documented as letting the runtime record a compound, input-derived key into the session's approved-tools set instead of the bare tool name (`runtime/types.d.ts`), and the harness honors it: `resolveApprovalKeyFromTools` in `harness/tool-loop.js` reads `approvalKey` off the resolved tool map, and approval recording resolves through it.

But for tools produced by a **dynamic resolver** (e.g. `step.started`), the conversion in `context/dynamic-tool-lifecycle.js` strips the field:

```js
function toHarnessToolDefinition(e, t) {
  return {
    description: t.description,
    execute: createToolExecuteWithAuth({...}),
    inputSchema: toInputSchema(t.inputSchema),
    name: e,
    approval: t.approval,
    // approvalKey is not copied
    outputSchema: toOutputSchema(t.outputSchema),
    ...(t.toModelOutput === void 0 ? {} : { toModelOutput: t.toModelOutput }),
  };
}
```

So a dynamic tool that declares `approvalKey` gets its approvals recorded under the bare tool name. The failure is silent: an approval policy that checks `ctx.approvedTools` for the scoped key never matches, and the UX degrades to re-prompting on every call — or worse, a policy written against bare names grants more than the author scoped.

## Repro

1. Return a tool with `approvalKey: (input) => \`mytool@\${input.scope}\`` from a `step.started` dynamic resolver.
2. Trigger it, approve the request.
3. Inspect session state / `ctx.approvedTools` on the next call: it contains `"mytool"`, not `"mytool@..."`.

## Expected

`approvalKey` survives the dynamic-tool conversion, matching the behavior documented on `ToolDefinition` and already implemented in the harness read path.

## Suggested fix

One field-passthrough in `toHarnessToolDefinition` — we run this as a patch in production and it restores the documented behavior:

```js
approval: t.approval,
...(t.approvalKey === void 0 ? {} : { approvalKey: t.approvalKey }),
```

Possibly related: `replayDynamicSessionTools` in the same module also rebuilds tools without `approval`/`approvalKey` — unreachable for step-resolved tools, but worth aligning while in there.

Context: we use input-scoped approvals to let one human approval cover a chain of GitHub-authoring writes bound to a (repo, branch) target; without the passthrough the scoping silently never engages.

### LINKED ISSUE #2369: OpenAPI connections pass through `default`/`example` values that contradict the schema's own `type`, causing models to emit invalid tool input
### Summary

`defineOpenAPIConnection` copies each property's `default` and `example` into the model-facing JSON Schema verbatim, even when those values contradict the property's declared `type`. Models follow the example, produce input that fails the generated zod schema, and the tool call dies.

Real specs do this. Xero's official `xero_accounting.yaml` has **10** boolean properties whose `default` and `example` are quoted strings.

### Reproduction

Minimal spec exhibiting the pattern:

```ts
export default defineOpenAPIConnection({
  baseUrl: "https://api.example.com",
  description: "Repro.",
  spec: {
    openapi: "3.0.3",
    info: { title: "Repro", version: "1.0.0" },
    paths: {
      "/things": {
        post: {
          operationId: "createThing",
          requestBody: {
            content: {
              "application/json": {
                schema: {
                  type: "object",
                  properties: {
                    HasAttachments: {
                      type: "boolean",
                      default: "false",   // string, contradicts type
                      example: "false",   // string, contradicts type
                    },
                  },
                },
              },
            },
          },
          responses: { "200": { description: "OK" } },
        },
      },
    },
  },
});
```

The model reliably emits `"HasAttachments": "false"` and the call fails:

```
AI_InvalidToolInputError: Invalid input for tool example__createThing:
AI_TypeValidationError: Type validation failed
  path: ["body","HasAttachments"]
  "Invalid input: expected boolean, received string"
```

### Real-world case

Xero's spec, pinned at commit `853dc01`:

```yaml
HasAttachments:
  description: boolean to indicate if an invoice has an attachment
  readOnly: true
  type: boolean
  default: "false"
  example: "false"
```

Scanning that spec for booleans carrying string `default`/`example`:

```
ApprovedForSending, FromIsReconciled, HasAccount, HasAttachments,
HasErrors, HasValidationErrors, IncludePDF, MarkAsSent, SendCopy,
ToIsReconciled
```

`MarkAsSent`, `IncludePDF` and `SendCopy` are `emailInvoice` parameters — a model is *meant* to set those, so it cannot simply be told to omit them.

This bit us in production on a `createInvoices` call. Four fields failed at once, all from copied examples.

### Where

`dist/src/runtime/connections/openapi-schema.js`. `derefSchema` walks every key and recurses:

```js
let a={};
for(let[e,o]of Object.entries(n)) a[e]=derefSchema(t,o,r+1,i);
return normalizeSchemaType(a), normalizeNullable(a), a
```

`normalizeSchemaType` strips `type` values that aren't valid JSON Schema types, and `normalizeNullable` folds `nullable` into `type` — so there is already a normalization pass here, it just doesn't look at `default` or `example`.

### Suggested fix

In that same pass, drop `default` and `example` when they don't validate against the node's own `type`. A schema that contradicts itself is a spec bug, and the annotation is the wrong half to trust — dropping it costs the model nothing, while keeping it actively teaches the wrong type.

`readOnly: true` properties appearing in request schemas at all is arguably a second, separate issue.

### Versions

- `eve` 0.39.0
- `ai` 7.0.66
- Node 24, deployed on Vercel

### LINKED ISSUE #2532: isStreamDisconnectError misclassifies post-response TypeErrors, bypassing both client reconnect policies

https://github.com/dblandin/eve-stream-disconnect-repro

Created with `npx eve@latest init`. It doesn't need real Cloudflare/nginx flakiness to reproduce — the bug is deterministic at the unit level. The repro mocks `fetch` and drives a real `eve/client` `Client`/session through three scenarios, each delivering one real NDJSON event and then failing the next body read with a `TypeError`: Safari's real `"Load failed"`, Chrome's observed `"network error"`, and — as a control — `"failed to fetch"`, the one message the classifier does recognize. Only the control reconnects; the other two propagate the error and never attempt a second request. Full output in the repo's README.

### Steps to reproduce

```
git clone https://github.com/dblandin/eve-stream-disconnect-repro
cd eve-stream-disconnect-repro
npm install
node repro.mjs
```

No server needed — `repro.mjs` is a standalone script against the installed `eve` package. Observe: for the `"Load failed"` and `"network error"` scenarios, only one stream-open request is ever made and the error propagates, even though `streamIdleReconnectPolicy` exists specifically to handle this case. The `"failed to fetch"` control scenario reconnects (a second request with `?startIndex=1`) and completes cleanly — proving the difference is the error message, not the harness.

**Current:** When a live session stream dies mid-body *after* the request already returned `200`, the body read throws a `TypeError` that `isStreamDisconnectError` classifies as fatal instead of recoverable. It propagates out of `open-stream.ts` and kills the turn — neither `streamOpenReconnectPolicy` nor `streamIdleReconnectPolicy` is ever consulted, and zero retry requests are made.

**Expected:** A `TypeError` thrown while reading an already-open response body is, by construction, a network failure (see "Why this happens" below) and should be treated as a recoverable disconnect, triggering the existing reconnect policies.

**Why:** the `TypeError` branch in `isStreamDisconnectError` is anchored to the exact messages `failed to fetch` / `fetch failed` — what a browser or Node uses when `fetch()` never got a response at all. It doesn't match messages used for a *post-response* body-read failure, e.g. Safari's `TypeError: Load failed` ([axios#6766](https://github.com/axios/axios/issues/6766), [hotwired/stimulus#782](https://github.com/hotwired/stimulus/issues/782), multiple Apple Developer Forums threads) or Chrome's `TypeError: network error` for an HTTP/2 reset (`net::ERR_HTTP2_PROTOCOL_ERROR`) — the Chrome string is our own production observation, not verified against Chromium source, and the Safari case below doesn't depend on it. Neither string contains `abort`, `cancel`, `disconnect`, `premature close`, `socket`, or `terminated`, so the generic keyword regex doesn't rescue them either.

```ts
// packages/eve/src/client/ndjson.ts — https://github.com/vercel/eve/blob/ae83a086e590baed57eabfc48a87f43a236fb816/packages/eve/src/client/ndjson.ts#L3-L25
export function isStreamDisconnectError(error: unknown): boolean {
  if (error instanceof DOMException) {
    return error.name === "AbortError";
  }
  if (!(error instanceof Error)) {
    return false;
  }
  const errorCode = "code" in error && typeof error.code === "string" ? error.code : undefined;
  return (
    error.name === "AbortError" ||
    error.message === "terminated" ||
    errorCode === "UND_ERR_SOCKET" ||
    (error instanceof TypeError && /^(?:failed to fetch|fetch failed)$/i.test(error.message)) ||
    /abort|cancel|disconnect|premature close|socket|terminated/i.test(error.message)
  );
}
```

**The structural reason this is the wrong shape of check:** once `fetch()` has resolved with a response, a `TypeError` thrown while reading `response.body` can only mean a network failure — the errors that mean something else (bad URL, CORS, invalid init) all throw from the `fetch()` call itself, before a response exists. So a finite allowlist of known messages will always be incomplete for this position in the code.

**This classifier gates two independent call sites** in `packages/eve/src/client/open-stream.ts`:

1. Mid-stream reconnect, inside the read loop: `catch (error) { if (!isStreamDisconnectError(error)) throw error; ... }` — decides whether `streamIdleReconnectPolicy` / backoff-and-reopen ever runs after a body-read error. This is the site our production failure hits.
2. Initial stream-open retry, in `openStreamBody` — same classifier gates whether a failed `fetch()` to open the stream is retried under `streamOpenReconnectPolicy`.

Only site 2 can legitimately see a `TypeError` that isn't a network failure (bad URL/CORS/invalid init happen there, never at site 1) — relevant to the suggested fix below.

**Not a duplicate of #1159 or #1769** — the distinction is throws vs. doesn't-throw:
- **#1159** ("Live session stream can stall open indefinitely without erroring, evading all reconnect logic (Vercel CDN compression)") is a stream that goes silent and *never throws*, so no error ever reaches `isStreamDisconnectError`. This issue is the complementary case: the stream *does* throw, and the throw is misclassified as fatal.
- **#1769** ("fix(eve): recover stalled client stream reads") adds an idle-timeout race around `reader.read()`, already on `main`. An idle timeout never fires when `reader.read()` rejects immediately, so it doesn't help here.

### eve version

`eve@0.44.0` (exact, from `package-lock.json`). Confirmed not a regression: `isStreamDisconnectError` is byte-identical in `0.38.3`, `0.44.0`, `0.44.4` (latest published), and on `main` at `ae83a086e590baed57eabfc48a87f43a236fb816`.

### Environment

```
OS (production runtime): Debian "trixie" (node:24.18.1-trixie-slim)
Node: v24.18.1
Package manager: npm
Client: browser (Safari and Chrome affected; first reproduced on Brave, Chromium-based)
Server: Next.js app using eve, deployed behind an nginx reverse proxy (Cloudflare-proxied in front of that)
```

### Where does the bug occur?

- [x] Production (`eve start` / deployed)

(The failure is client-side and network-condition-dependent — it's not tied to `eve dev`/`eve build`/`eve init`, though the deterministic mock repro above should reproduce it in any of those too.)

### Deployment

Not included — our deployment sits behind an internal SSO gate, so a link wouldn't be independently reproducible for a maintainer. Happy to provide log correlation IDs privately if useful.

### Build and runtime logs

We did not capture this with `EVE_LOG_LEVEL=debug` — the original investigation was black-box (diffing installed `eve` versions, reading nginx/browser evidence), not a debug-log capture session. What we do have, verbatim or by field-shape:

nginx access/error log (verbatim, standard nginx message):
```
upstream prematurely closed connection
```

Our own stream-instrumentation logging (custom, not eve's) — field shape from `eve.stream.closed`, logged by our proxy route:
```
eve.stream.closed  durationMs=<n>  bytes=<n>  heartbeats=<n>  lastBoundary=null  upstreamError=ResponseAborted
```

Browser console: `TypeError: Load failed` (Safari) / `TypeError: network error` (Chrome), thrown from the body-read loop.

*If `EVE_LOG_LEVEL=debug` output specifically would help, we can capture it — say so and we'll reproduce again with it enabled.*

> Fix `isStreamDisconnectError` (`packages/eve/src/client/ndjson.ts`) so a `TypeError` thrown while reading an already-open response body is treated as a recoverable disconnect, not a fatal error — regardless of its message text.
>
> **Requirements:**
> - A `TypeError` surfacing from the mid-stream read path (`open-stream.ts`'s read loop, not the initial `fetch()` in `openStreamBody`) should be treated as a disconnect without requiring its message to match a known string, since no other kind of error can occur at that point once a response already exists.
> - The initial `fetch()` in `openStreamBody` should keep its current message-based behavior (or something equally conservative), since a `TypeError` there can legitimately mean something other than a network failure (bad URL, CORS misconfiguration, invalid init).
> - Don't regress the existing `AbortError` / `terminated` / `UND_ERR_SOCKET` / keyword-regex branches — they should still work at both call sites.
>
> **Acceptance criteria:**
> - `isStreamDisconnectError`-equivalent logic returns a disconnect classification for a mid-stream-read `TypeError` regardless of message (e.g. `Load failed`, `network error`, or any other vendor string), verified with a test per browser-observed message.
> - The stream-open path's existing behavior for `TypeError` is unchanged (still message-gated).
> - A test demonstrates: reader rejects with an unrecognized `TypeError` mid-body → `streamIdleReconnectPolicy` runs and a new stream-open request is made, where today the error propagates and no retry happens.

### Additional context

- **What we tried first:** raised nginx's `proxy-send-timeout`/`proxy-read-timeout`, which moved the failure from ~30s to ~60s but didn't eliminate it — evidence this isn't a timeout problem at the root, just a classifier that discards a real disconnect regardless of when the connection is cut.
- **Our workaround** (application-level, doesn't fix the underlying bug): we proxy the stream ourselves and emit a heartbeat to keep the connection warm, and always close the response cleanly on upstream end. This only removes one *trigger* (a connection idling for tens of seconds); it does nothing for a plain mid-stream network blip, which is why we're filing this rather than relying on the workaround alone.
- Easily reproducible on longer agent turns — specifically ones involving multiple tool calls against data connections. Anything running past roughly 30s of wall clock is at risk.
- Happy to open a PR for the suggested fix if that shape sounds right.

### LINKED ISSUE #2716: Regression since #2525: composer stays disabled for the full 15s stream idle timeout when resuming a settled session
### Summary

Since #2525, opening an existing chat session leaves the composer disabled for a full 15 seconds, every time. The session is healthy and parked; nothing is actually in flight. `send()` also throws `eve session is resuming.` for that whole window, so the user simply cannot send a message.

Bisected: **0.44.4 enables the composer at 0.1s, 0.47.2 at 15.1s** — same server, same session, measured below.

### Environment

- `eve@0.47.2` (latest published; regression present since `0.45.0`)
- Node v26.5.0, darwin arm64
- `eve dev` + Next.js 16.3.0-preview.6, scaffolded web chat
- Plain `localhost`, **no proxy involved**

Related: #2527 (server-side root cause: no close or heartbeat at settle), #1159 (origin of the 15s idle abort this path now depends on).

### Steps to reproduce

1. `eve dev` with the scaffolded Next chat UI.
2. Send one message, let the turn complete. The session parks on a `session.waiting` tail.
3. Reload `/s/<sessionId>`.

**Expected:** the composer is usable once history has loaded (~0.1s).
**Actual:** `agent.status` stays `"resuming"` for ~15s and the scaffold's `disabled={isResuming}` keeps the textarea dead for that entire window.

### What changed in #2525

The live-follow pass itself predates the PR — the entry condition is untouched:

```js
(!replaySettled || replayTail?.type === "session.waiting")
```

What changed is what that pass costs the user. Before #2525, resume ran with `status: "submitted"` and the scaffold gated on `isRestoring = sessionId !== undefined && agent.events.length === 0 && isBusy`. Once the bounded history replay published (31 events), `isRestoring` went false and the composer re-enabled immediately — the follow-stream wait was invisible to input. The visible defect was the cosmetic Stop/Thinking state on a finished conversation, which #2525 correctly fixed.

After #2525 the status is `"resuming"` for the whole pass, and note the asymmetry:

```js
if (!replaySettled) this.#status = "streaming";
```

A **non**-settled tail transitions to `streaming` and leaves the composer usable. A **settled** tail — the case where no further event can possibly arrive — keeps `"resuming"` for the entire loop. The branch that provably has nothing to wait for is the one that locks input.

### Why the stream never delivers

Bounded history returns immediately: 31 events, indices 0–30, `x-eve-stream-tail-index: 30`, tail `session.waiting`. The follow stream then opens one index past the tail and produces nothing. Measured byte-by-byte:

```
GET /eve/v1/session/<id>/stream?startIndex=31

   0.0s byte=b'\n'
  50.0s STREAM CLOSED     <- my own timeout; the server sent nothing further
```

With `replaySettled === true` the policy is `{reconnect: false}` → `streamIdleReconnectPolicy.maxAttempts === 0`, so the generator returns after the first idle timeout in `readNdjsonStream` (`idleTimeoutMs: 15e3`) — hence exactly 15s.

### Headless reproduction

Against a running dev server, using the shipped store:

```js
import { EveAgentStore, defaultMessageReducer } from "eve/client";

const store = new EveAgentStore({
  host: "http://localhost:3001",
  initialSession: { sessionId: "<parked session id>", streamIndex: 0 },
  reducer: defaultMessageReducer(),
});

const t0 = Date.now();
const at = () => `${((Date.now() - t0) / 1000).toFixed(1)}s`;
let last;
store.subscribe(() => {
  const { status, events } = store.snapshot;
  if (status !== last) { last = status; console.log(`${at()}  status -> ${status} (events: ${events.length})`); }
});
await store.resume();
console.log(`${at()}  resume() resolved`);
```

**`eve@0.47.2`:**

```
0.0s   status -> resuming (events: 0)
15.1s  status -> ready    (events: 31)
15.1s  resume() resolved
```

**`eve@0.44.4`,** same server, same session — the composer's `isRestoring` clears as soon as history lands:

```
0.0s   status -> submitted (events: 0)
0.1s   status -> streaming (events: 31)
0.1s   >>> composer ENABLED
15.1s  status -> ready     (events: 31)
```

The 15s wait still happens on 0.44.4; it just no longer blocks input.

### Suggested fix

Skip the live-follow pass entirely when the tail is settled and `collectPendingAuthorizations` is empty — the replayed history already carries the terminal event, so there is nothing to catch up on. (I inferred from reading the code that pending authorizations are the only reason to keep the pass on a settled tail; please sanity-check that.)

Fixing #2527 server-side would also mask the symptom, but the client would still be opening a round trip it does not need.

Separately: `followStreamIterable` accepts `streamReadIdleTimeoutMs` (`client/open-stream.js`), but nothing passes it — zero occurrences in `client/session.js`, `client/eve-agent-store.ts` and `react/use-eve-agent.ts` — so consumers cannot shorten the window themselves.

### Workaround

Pinned to `eve@0.44.4` (the last release before #2525) and reverted the generated `agent-chat.tsx` to the `isRestoring` gate. Downgrading is disruptive, though: step identities embed the package version (`step//eve@0.47.2//createSessionStep`), so every session parked under the old version dies on the first resume with `CorruptedEventLogError: Workflow replay diverged`.

### LINKED ISSUE #2779: Slack authorization button exceeds Block Kit text limit for long connection display names
## Summary

`buildAuthEphemeralBlocks()` creates the OAuth link-button label as `Sign in with ${displayName}` without enforcing Slack's 75-character button-text limit. A display name of 63 characters produces a 76-character label, causing `chat.postEphemeral` to fail with `invalid_blocks` and leaving the authorization flow blocked.

## Reproduction

On `eve@0.47.2`, call `buildAuthEphemeralBlocks()` with any 63-character `displayName`, for example:

```ts
const displayName = "x".repeat(63);
const blocks = buildAuthEphemeralBlocks({
  displayName,
  url: "https://connect.example.com/authorize/sca_abc",
});

expect((blocks[0] as { elements: Array<{ text: { text: string } }> }).elements[0].text.text)
  .toHaveLength(76);
```

Slack documents a 75-character maximum for a button's `text`. Sending these blocks through `chat.postEphemeral` returns `invalid_blocks`.

## Expected

The Slack authorization renderer should always emit a valid button label, regardless of `authorization.displayName`.

## Suggested direction

Use a bounded static label (for example, `Sign in`) or truncate the derived label to Slack's 75-character limit, and add a regression test for a long display name. The connection display name can remain in the fallback text or public status; it does not need to be repeated in the button label.

## Environment

- `eve@0.47.2`
- Slack channel
- user-scoped connection authorization delivered by `authorization.required`

### LINKED ISSUE #2781: eve eval --json silently truncates its report when stdout is a pipe

https://github.com/scafati98/eve/tree/fix/eval-json-report-truncation

A note on the form of this reproduction, since it is not a scaffolded app: the bug lives in the CLI's own output plumbing and only exists when stdout is a pipe, which an `eve init` app cannot express on its own — it needs a parent process. So the link is a branch of eve itself carrying a scenario test, `packages/eve/src/evals/cli/eval-json-report.scenario.test.ts`, which materializes an inline app, runs the CLI with stdout piped, and asserts the report parses. Dropping that test onto `main` fails; the branch also carries a candidate fix that makes it pass. Happy to publish a scaffolded app instead if you would rather have one.

The mechanism itself reproduces with no eve involved at all:

```js
// repro.mjs
const payload = JSON.stringify({ pad: "A".repeat(200_000) });
console.log(payload);
process.exit(0);
```

```console
$ node -e 'console.log(JSON.stringify({pad:"A".repeat(200000)}).length+1)'
200011

$ node repro.mjs | wc -c
  131072          # truncated at the pipe buffer

$ node repro.mjs > out.txt; wc -c < out.txt
  200011          # a file redirect keeps everything
```

### Steps to reproduce

On `main`, bring across only the test and run it:

```sh
git clone https://github.com/scafati98/eve && cd eve
git fetch origin fix/eval-json-report-truncation
pnpm install && pnpm build

# the test alone, with main's eval.ts -> fails
git checkout origin/fix/eval-json-report-truncation -- \
  packages/eve/src/evals/cli/eval-json-report.scenario.test.ts
pnpm --filter eve exec vitest run --config vitest.scenario.config.ts \
  src/evals/cli/eval-json-report.scenario.test.ts

# the whole branch, fix included -> passes
git checkout origin/fix/eval-json-report-truncation
pnpm --filter eve build && pnpm --filter eve exec vitest run \
  --config vitest.scenario.config.ts src/evals/cli/eval-json-report.scenario.test.ts
```

In an ordinary app, the same thing shows up as: run `eve eval --json` on a suite whose report exceeds the pipe buffer (about one eval's worth), collect stdout through a pipe rather than a shell redirect, and parse it.

**Expected.** `eve eval --json | anything` delivers the whole report, or fails loudly.

**Current.** The report is cut off at the stdout pipe buffer boundary. Nothing announces it: exit code `0`, empty stderr, and a JSON prefix that looks intact, so a consumer that does not validate reads a truncated report as if it were complete. A consumer that does validate gets a parse error with no indication of the cause.

I hit this as `durationMs: null` on every eval in a downstream tool while verdicts, gates, and tool calls were all correct — the signature of my own artifacts fallback engaging, because per-eval `startedAt`/`completedAt` exist only in the `--json` document and not in `.eve/evals/<timestamp>/`. Instrumenting the runner gave:

```json
{
  "exitCode": 0,
  "source": "artifacts",
  "parseError": "Unterminated string in JSON at position 65487 (line 1406 column 49)",
  "stdoutBytes": 65487,
  "outcomes": 6
}
```

The full report for those six evals was 1.77 MB.

**Isolating it.** Same command, same arguments, four ways of collecting stdout:

| stdout | bytes | result |
| --- | --- | --- |
| buffered through execa (`maxBuffer` 256 MB) | 65,468 | unterminated JSON |
| execa `stdout: { file }` | 65,536 | unterminated JSON |
| shell redirect `> file` | 206,318 | parses |
| raw descriptor, `stdio: ["ignore", fd, "pipe"]` | 193,744 | parses |

`maxBuffer` is irrelevant because nothing overflowed — the tail never arrived. The two survivors are exactly the cases where Node's writes to stdout are synchronous, and `stdout: { file }` is not one of them because execa implements it by piping into a `WriteStream`.

**Cause.** `packages/eve/src/evals/cli/eval.ts` prints the report with `logger.log(...)` (`logger` is `console`, so an asynchronous write when stdout is a pipe) and then ends the command with `process.exit(exitCode)`, which discards anything still queued.

Two details worth stating precisely, because both cost me time:

- **The boundary is the pipe buffer, not a fixed 64 KiB.** I measured 64 KiB through execa and 128 KiB with a directly piped child on the same machine. It varies with platform and consumer.
- **How much of the tail survives depends on the consumer.** On the local path the `finally` block awaits `devServer.close()` and `shutdownActiveSandboxHandles()`, and those awaits let an eagerly draining reader flush the report incidentally — I could not truncate the local path even at an 8 MB report with such a reader. With `--url` there is no dev server to close, so nothing is awaited between the write and the exit and the truncation is deterministic. That incidental drainage is why this looks intermittent rather than broken.

### eve version

eve@0.47.3

### Environment

```shell
OS: macOS 15.2 (arm64), Darwin 24.2.0
Node: v25.8.1
Package manager: npm 11.11.0 (downstream app), pnpm 10.28.0 (eve checkout)
```

### Where does the bug occur?

Other — `eve eval --json`

### Build and runtime logs

```shell
# the failing scenario test on main, before the fix
$ pnpm --filter eve exec vitest run --config vitest.scenario.config.ts \
    src/evals/cli/eval-json-report.scenario.test.ts

 ❯ src/evals/cli/eval-json-report.scenario.test.ts (1 test | 1 failed)
   × writes the complete report to a piped stdout

 FAIL  eve eval --json > writes the complete report to a piped stdout
SyntaxError: Unterminated string in JSON at position 65536

 Test Files  1 failed (1)
```

Note that the run's own exit-code assertion passes: the command exits `0` while losing the report.

Make `eve eval --json` deliver its whole report on a pipe.

Requirements:

- Do not lose the queued stdout write when the command exits. Either flush before exiting or write the report synchronously.
- Keep the forced exit. The command owns a dev server and sandbox handles, and dropping `process.exit()` for a natural exit risks hanging instead — that is a worse failure than the one being fixed.
- The same exit discards queued stderr, which is where the sandbox-shutdown diagnostics in the `finally` block go; worth covering in the same change.
- Whatever waits on the stream must not be able to wedge the exit when the consumer closes early (`| head`), or when stdout is a closed descriptor.

Acceptance criteria:

- A report larger than the pipe buffer parses when stdout is a pipe.
- Exit codes are unchanged for pass, fail, `--strict` threshold misses, and configuration errors.
- Coverage targets `--url`, where nothing is awaited between the report write and the exit; the local path's drainage is incidental and will pass either way.

### Additional context

I have a candidate fix on the branch above: one awaited stdout/stderr flush before the existing `process.exit()`, plus the scenario test, plus a patch changeset — 3 files, `+222 / -0`, with the forced exit and every exit code preserved. Locally `typecheck`, `lint`, `fmt`, `build`, `guard:invariants`, `docs:check`, `test:unit` (7446), `test:integration` (678) and `test:scenario` all pass, with three scenario files failing identically on `main` in my environment for unrelated reasons (Node 25.8.1 against the repo's pinned Node 24: `dev-server-bun`, `framework-sveltekit-build`, `nitro-bundle-report`).

Following CONTRIBUTING.md I am opening this as an issue rather than a pull request, since this touches CLI/runtime behavior and I have not been invited to implement it. Glad to open the PR if a maintainer agrees with the direction — and equally glad to leave it to you and just keep the reproduction here.

I did not change any docs: `docs/evals/running.mdx` lists `--json` without promising or qualifying pipe behavior, so the fix restores what the page already implies rather than altering documented behavior. If you would rather document the constraint than fix it, the one-line version is that `--json` must be redirected to a file rather than piped — but that seems worth avoiding, since the failure is silent and the correct-looking prefix is what makes it dangerous.

### LINKED ISSUE #2952: Steering during fileMemory recall can turn AbortError into terminal session failure
## Link to a minimal reproduction

Public regression and workaround: https://github.com/Merit-Systems/OpenInstinct/pull/121

This is not a standalone Eve fixture. The failure was captured from a production Vercel Workflow run, and the linked PR contains the smallest application-level reproduction of the lost cancellation reason.

## Steps to reproduce

1. Configure an Eve agent with `fileMemory()` backed by Vercel Blob.
2. Start a turn through a channel using the default `turnPolicy: "steer"`.
3. While `recall["turn.started"]` is reading the memory document, send a follow-up message so Eve cancels the active turn.
4. Let the Blob request surface cancellation as a generic `AbortError` rather than the signal's `TurnCancelledError` reason.

A deterministic provider fixture can reproduce the boundary by waiting for its supplied `abortSignal`, then throwing `new DOMException("This operation was aborted", "AbortError")` from `recall["turn.started"]`.

## Current vs. expected behavior

Current behavior: `turnStep` checks `isTurnCancellation(error)`, does not recognize the generic `AbortError`, and allows the workflow step to fail. The root workflow then emits terminal `session.failed`, so the buffered steering message never becomes a replacement turn.

Expected behavior: when the turn's supplied `abortSignal` is aborted, an error escaping from turn work should settle through Eve's normal cancellation path. The stream should emit `turn.cancelled`, return to `session.waiting`, and process the buffered replacement message. The same provider error must remain a failure while the signal is active.

## eve version

Observed on `eve@0.49.0`. I did not reproduce this against a deployed `0.50.0` fixture, but the current `vercel/eve` `main` source still checks only `isTurnCancellation(error)` at the `turnStep` catch boundary.

## Environment

```shell
Deployment: Vercel production
Runtime: Node.js 24
Workflow region: iad1
Memory backend: private Vercel Blob through fileMemory()
Channel: Linq using turnPolicy "steer"
```

## Where does the bug occur?

## Deployment

Observed on deployment `dpl_E2WZ6X5Y7dQ2dYSkPd7mzf3GS3DK`.

## Build and runtime logs

```shell
turnStep attempt 1 started: 09:27:42.997
turnStep entered retry:     09:27:43.129
turn cancellation received: 09:27:44.258
turnStep attempt 2 started:  09:27:44.598
turnStep failed:             09:27:45.003

AbortError: This operation was aborted
  at fetch
  at Object.read
  at readDocument
  at recallMemory
  at dispatchMemoryTurnStarted
  at dispatchMemoryLifecycleEvent

[eve:execution.workflow-entry] workflow loop threw — emitting terminal session.failed {
  errorId: 'd6e23347-5195-4889-9105-72abb081021d',
  code: 'FatalError'
}
```

The first attempt's exception was not retained in the Workflow event payload; only the final attempt's stack was available. The cancellation hook was resolved before the final attempt failed.

## Suggested implementation prompt

Fix turn cancellation when an operation throws a generic abort-shaped error after the turn's signal has been cancelled.

Requirements:

- At the durable `turnStep` error boundary, allow the supplied aborted signal to establish cancellation even when the thrown error lost the signal's canonical reason.
- Do not classify generic `AbortError` values as cancellation while the turn signal remains active.
- Preserve the existing cancelled-turn history, background-task, and replacement-delivery behavior.
- Add an integration test where memory recall is interrupted by a steering message and throws a generic `AbortError`.

Acceptance criteria:

- The interrupted turn emits `turn.cancelled`, not `turn.failed` or `session.failed`.
- The buffered follow-up starts as the replacement turn.
- An identical recall error with an active signal still fails normally.

## Additional context

Linq can deliver message text and its link as adjacent inbound messages. Eve's native steering behavior is the right semantic fit because it preserves the interrupted user input and lets the link continue the request. OpenInstinct's workaround wraps only its profile-memory provider and rethrows `abortSignal.reason` when recall fails after cancellation; a framework fix would cover every provider and turn-owned operation.

### LINKED ISSUE #3062: Compaction splits surrogate pairs at the 2000-char cap → provider 400 "no low surrogate" → session dies after retries

```ts
import { createCompactionPrompt } from "eve/dist/src/harness/compaction-prompt.js";

// Any message whose whitespace-collapsed text has an astral character (emoji)
// straddling index 1999/2000 of the 2000-char cap.
const content = "x".repeat(1999) + "📥" + " tail ".repeat(400);
const { prompt } = createCompactionPrompt({
  messages: [{ role: "user", content }],
  previousCheckpoint: undefined,
  transcriptBudgetTokens: 100,
});
console.log(/[\uD800-\uDBFF](?![\uDC00-\uDFFF])/.test(prompt)); // true → lone high surrogate
```

### Steps to reproduce

1. Run a long session until compaction kicks in (`transcriptBudgetTokens` exceeded, or a tool result over `TRANSCRIPT_PAYLOAD_LIMIT`).
2. Have one message where a surrogate-pair character (any emoji) lands exactly on the 2000-char boundary. In our case it was a tool result containing a CodeRabbit PR comment (`<summary>📥 Commits</summary>`).
3. The next turn builds the compaction prompt, sends it to Anthropic and gets:

```
The request body is not valid JSON: no low surrogate in string: line 1 column 140501 (char 140500)
```

4. The turn step is retried 3 times. The cut is deterministic (the prompt is rebuilt from the same history each time), so every retry fails the same way:

```
FatalError: Step "step//eve@0.50.0//turnStep" failed after 3 retries: The request body is not valid JSON: no low surrogate in string ...
```

The session dies. The stored history is clean (the truncated prompt is never persisted), so nothing in the transcript hints at the cause.

**Current:** compaction truncates by UTF-16 code-unit index and can split a surrogate pair:

- `packages/eve/src/harness/compaction-prompt.ts:322` — `capText`: `normalized.slice(0, limit)`
- `packages/eve/src/harness/compaction.ts:292` — `capToolResults`: `serialized.slice(0, TRANSCRIPT_PAYLOAD_LIMIT)`

`JSON.stringify` serializes the lone surrogate as `\ud83d`; the model provider rejects the body; retries cannot recover; the whole session is lost.

**Expected:** truncation never produces a lone surrogate. E.g. step back one unit when the last kept unit is a high surrogate, or call `.toWellFormed()` on the sliced string (Node ≥ 20).

### eve version

0.50.0 (still present on `main` at the lines above).

### Environment

- Node 22 on Vercel Functions, `@workflow` runtime
- Model: Anthropic `claude-fable-5.1` through a custom AI SDK provider (the 400 comes from the Anthropic API JSON parser)

### Where does the bug occur?

Runtime (deployed agent), during a turn that triggers compaction.

### Deployment

Vercel, production. Session run `wrun_41M1P6AW3X0GG9RBXN8CQZWPMS`, turn run `wrun_41M1PCM75C0GGVK568QZQDA0VV`, 2026-09-04 14:24 UTC.

### Build and runtime logs

```
[workflow-sdk] Max retries reached, bubbling error to parent workflow
  run    wrun_41M1PCM75C0GGVK568QZQDA0VV
  step   — · turnStep (eve@0.50.0)
  retry  4 attempts · 3 max retries
[eve:execution.workflow-entry] workflow loop threw — emitting terminal session.failed { code: 'FatalError' }
```

Linear agent-session comment posted by eve:

```
This session could not recover from an error (FatalError: Step "step//eve@0.50.0//turnStep" failed after 3 retries:
The request body is not valid JSON: no low surrogate in string: line 1 column 140501 (char 140500)).
```

I confirmed the cause by decrypting the last successful step's session state and running the real `createCompactionPrompt` over its history: exactly one lone surrogate, at the cut of history message #145 (tool result, collapsed length 8596, `charCodeAt(1999) === 0xd83d`, `charCodeAt(2000) === 0xdce5` = 📥).

In `packages/eve/src/harness/compaction-prompt.ts` (`capText`) and `packages/eve/src/harness/compaction.ts` (`capToolResults`), replace the raw `.slice(0, limit)` with a surrogate-safe cut: if `value.charCodeAt(limit - 1)` is in `0xD800–0xDBFF`, cut at `limit - 1`. Add a unit test that places an emoji on the boundary and asserts the result matches `/^[\s\S]*$/` with no lone surrogate (`str === str.toWellFormed()`). Consider a final `toWellFormed()` guard where the compaction prompt is assembled, so future cappers cannot reintroduce this.

### Additional context

Any tool that surfaces user- or bot-written text (PR review comments, chat messages, issue bodies) will hit this eventually; emoji density in CodeRabbit/GitHub comments makes it a matter of time for coding agents. The failure mode is the worst kind: deterministic, retry-proof, and invisible in the persisted transcript.

### LINKED ISSUE #3081: createWebhook workflow bundle shim omits isWebhook, causing public callbacks to return 404
## Versions

- `eve@0.52.1`, the latest published registry version at reproduction time
- Bundled Workflow core `5.0.0-beta.48`
- Hosted Vercel preview runtime

Eve current main contains the same shim behavior described below.

## Minimal reproduction

This is the relevant shape; names and values are generic.

```ts
import { createWebhook } from "workflow";

async function waitForExternalEvent() {
  "use workflow";

  using webhook = createWebhook();
  await webhook.getConflict();
  await registerCallbackUrl({ callbackUrl: webhook.url });

  for await (const request of webhook) {
    return await handleCallback(request);
  }
}
```

1. Run the workflow until `registerCallbackUrl` completes and it waits on the webhook.
2. POST a valid payload to the generated callback URL.
3. Observe HTTP `404`. Bounded retries reproduce the same response.

## Expected behavior

`createWebhook()` should create a hook with `isWebhook: true`, allowing the generated public callback route to resume it.

## Actual behavior

A read-only lookup of the created hook returns `isWebhook: false` while the hook is active, unresolved, undisposed, and linked to the expected workflow resume context. The public callback handler explicitly rejects `isWebhook === false` and maps it to `HookNotFound`, producing HTTP `404` before a hook-received event is committed.

The generated workflow bundle resolves the `workflow` import to Eve's `workflow-core-shim`. Its implementation is equivalent to:

```ts
export function createWebhook(options?: unknown) {
  const hook = createHook(options);
  const metadata = getWorkflowMetadata();
  hook.url = `${metadata.url}/.well-known/workflow/v1/webhook/${encodeURIComponent(hook.token)}`;
  return hook;
}
```

That call does not pass `isWebhook: true`, so the runtime stores the hook using the regular-hook default of false. By comparison, the Workflow core wrapper removes webhook-only options and calls `createHook({ ...options, metadata, isWebhook: true })` before constructing the URL.

The affected Eve source path is `packages/eve/src/internal/workflow-bundle/workflow-core-shim.ts`; the published equivalent is `dist/src/internal/workflow-bundle/workflow-core-shim.js`. The workflow bundler deliberately aliases `workflow` and the Workflow core condition to this shim, so the application receives the shim implementation inside the generated bundle.

## Requested change

Please update Eve's `createWebhook` shim to create the hook with `isWebhook: true`, while retaining its existing URL construction and supported webhook options. A regression should build a workflow through Eve's actual bundle path, call `createWebhook`, and assert that the resulting hook is resumable through the public webhook handler.

This was reproduced with the unmodified published Eve package. No private callback URL, hook token, project identifier, deployment identifier, run identifier, customer name, or production data is included in this report.

Affected source at the inspected main revision: https://github.com/vercel/eve/blob/1807ff9c5bd15c06b549e35fc5064dc912661275/packages/eve/src/internal/workflow-bundle/workflow-core-shim.ts#L79-L85

AndrewBarba:
### Summary

Routine operations can currently terminate healthy sessions, lose approval scope, reject valid integrations, or leave clients and automation unable to recover. This fixes ten reported failures across compaction, steering cancellation, workflow callbacks, dynamic approvals, streaming, eval output, bundle loading, OpenAPI schemas, and Slack sign-in. Regression coverage includes real bundle/subprocess tests and deterministic fixture evals for public webhook callbacks and durable approval scoping.

Closes #3062, closes #2952, closes #3081, closes #2319, closes #2532, closes #2781, closes #2195, closes #2716, closes #2369, closes #2779.

Adapted from the existing proposals in #3075, #2954, #2321, #2533, #2783, #2196, #2719, #2370, and #2782. The dynamic approval and stream fixes are updated for current durable callback reconstruction and event validation rather than applying the older patches unchanged.

### Validation

- `pnpm typecheck`: all 43 workspace tasks passed.
- Focused unit suites for compaction, compatibility banners, webhook shims/routes, dynamic tools, OpenAPI, Slack, and client stores passed.
- `pnpm test:integration`: 722 tests passed before the final approval-settlement regression; the expanded approval-resume suite passed all 17 tests afterward, including structured and text responses across step-, turn-, and session-scoped tools.
- Focused scenario suites for actual Rolldown output and piped eval reports: 4 tests passed.
- `pnpm build`, `pnpm lint`, `pnpm fmt`, `pnpm check:deps`, `pnpm docs:check`, `pnpm guard:fixtures`, and `pnpm guard:invariants` passed.
- Local `pnpm test:unit`: 8,117 passed, 1 skipped, and five macOS/path-sensitive failures in unchanged tests under `/private/tmp`. CI runs the full unit suite on Linux.
- GitHub API-created commits have verified signatures and DCO sign-offs.
- Independent subagent reviews covered runtime behavior, request serialization, approval scopes, stream races, and public TypeScript authoring. CI exposed two additional defects in the new fixture paths, both addressed in follow-up changes.

CI is green on `34d1ef44d764bb5cb2d9e65c25a05ad8d2661885`: 116 successful checks, one intentionally skipped check, no failures or pending checks. All required checks pass. The real-model cancellation eval passed on a targeted retry after one instruction-following refusal.

### Checklist

### DIFF
--- a/.changeset/clean-tools-schema.md
+++ b/.changeset/clean-tools-schema.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Ignore OpenAPI `default` and `example` annotations whose values conflict with the declared schema type, so models are not prompted to submit invalid tool input.
--- a/.changeset/fast-settled-session-resume.md
+++ b/.changeset/fast-settled-session-resume.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Resume settled sessions without waiting for the live stream idle timeout while preserving catch-up for turns accepted during replay.
--- a/.changeset/fix-node-esm-comma-bindings.md
+++ b/.changeset/fix-node-esm-comma-bindings.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Prevent `eve dev` bundles that inline comma-separated CommonJS path globals from failing to load with duplicate declarations.
--- a/.changeset/fix-piped-eval-json.md
+++ b/.changeset/fix-piped-eval-json.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Ensure `eve eval --json` writes the complete report when stdout is piped to another process.
--- a/.changeset/fix-slack-auth-button-label.md
+++ b/.changeset/fix-slack-auth-button-label.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Keep Slack connection authorization buttons valid when a connection has a long display name.
--- a/.changeset/fix-workflow-webhook-registration.md
+++ b/.changeset/fix-workflow-webhook-registration.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Workflow tools now create publicly resumable webhooks, fixing callbacks that returned 404. Webhook response options are preserved, and authored webhook tokens are rejected consistently with Workflow.
--- a/.changeset/quiet-ravens-steer.md
+++ b/.changeset/quiet-ravens-steer.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Preserve steering replacements when cancelled turn work throws a generic `AbortError`, so the session returns to waiting instead of failing.
--- a/.changeset/safe-compaction-unicode.md
+++ b/.changeset/safe-compaction-unicode.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Preserve complete Unicode characters when truncating compaction transcripts and tool results, preventing emoji at the cutoff from causing provider request failures.
--- a/.changeset/strong-snails-approve.md
+++ b/.changeset/strong-snails-approve.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Preserve input-scoped approval keys on dynamic tools, including durable replay, so approvals record the intended key instead of the bare tool name.
--- a/.changeset/tender-streams-reconnect.md
+++ b/.changeset/tender-streams-reconnect.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Reconnect session streams when browser response-body reads fail with vendor-specific `TypeError` messages, while keeping initial fetch retries limited to known transport errors.
--- a/docs/channels/slack.mdx
+++ b/docs/channels/slack.mdx
@@ -451,6 +451,8 @@ events: {
 },
 ```
 
+The private authorization button uses the label **Sign in**, so long connection display names do not make the button invalid. The public status identifies the connection.
+
 ### Proactive sessions
 
 Start a session without an inbound message through `ctx.to(slack, target).send(message, { auth })` from another channel, or `to(slack, target).send(message, { auth })` from a schedule `run` handler. The proactive target shape is `{ channelId, installationTeamId?, threadTs?, initialMessage? }`. Set `installationTeamId` to the app installation workspace when a function-form bot token must select an installation. This sends turn input to the agent; it is not a direct Slack post.
--- a/docs/connections/openapi.mdx
+++ b/docs/connections/openapi.mdx
@@ -150,6 +150,10 @@ export default defineOpenAPIConnection({
 
 The parameter `name` must exactly match the placeholder inside the path. If the spec omits an `in: "path"` parameter, the generated tool has no input for that segment and eve cannot fill it in from query parameters.
 
+## Schema annotations
+
+eve removes `default` and `example` annotations when their values conflict with the declared schema type. For example, a boolean property with `default: "false"` loses that annotation, while `default: false` is retained. This prevents invalid examples from guiding model input; eve does not coerce annotation values.
+
 ## Approval gates
 
 Generated operations can mutate state like authored tools. Use the shared connection `approval` option for human approval, and combine it with `operations.allow` for the smallest practical surface. See [Per-connection approval](../connections#per-connection-approval) for the helpers and [Human-in-the-loop](../human-in-the-loop#approvals) for policy behavior.
--- a/docs/evals/running.mdx
+++ b/docs/evals/running.mdx
@@ -35,6 +35,8 @@ Tag filters compose: `--tag` keeps evals carrying at least one listed tag, then 
 
 An eval that calls `t.skip(reason)` is reported as skipped, does not count as passed or failed, and never changes the exit code.
 
+`eve eval --json` flushes the complete report before exiting, including when stdout is piped to another process or redirected to a file.
+
 ## Artifacts
 
 Each run drops artifacts under `.eve/evals/<timestamp>/`: a run `summary.json`, a `results.jsonl` index, and per-eval assertion results, verdicts, captured event streams, and `t.log` lines under `evals/`. The console output stays tight on purpose; when an eval fails, the artifact has the full story.
--- a/docs/guides/client/streaming.mdx
+++ b/docs/guides/client/streaming.mdx
@@ -153,6 +153,8 @@ The stream can emit an interim `session.waiting` while the authorization callbac
 
 HTTP connections can end before a run does. The client reconnects from the number of events already consumed, so long turns continue without replaying events. By default, a turn response keeps reconnecting until it reaches a turn boundary or is aborted, including while the turn is paused for authorization. A manually opened `session.stream()` eventually stops after repeated empty streams when it can no longer make progress.
 
+Browser failures while reading a response body, including `TypeError: Load failed` and `TypeError: network error`, use the same reconnect policy. Invalid stream events still fail the read.
+
 If your consumer persists events, key on `event.meta.id`. It is stable across reconnects and rewinds, so an overlapping replay is safe to ingest twice. See [the event envelope](../../concepts/sessions-runs-and-streaming#the-event-envelope).
 
 Set `streamReconnectPolicy: { reconnect: false }` when a relay or proxy owns the cursor and reconnection policy. This makes a single stream GET attempt and returns when that connection ends; it does not stop the server-side turn:
--- a/docs/guides/dynamic-capabilities.md
+++ b/docs/guides/dynamic-capabilities.md
@@ -219,7 +219,7 @@ the handler.
 
 ## Dynamic tools
 
-Pass `defineDynamic` an `events` object whose handlers return either a single `defineTool(...)`, a `Record<string, defineTool(...)>`, or `null` for no tools. Wrap every entry in `defineTool()`. eve records durable descriptors for `execute`, approval request and response policies, and `toModelOutput`, so a parked call can reconstruct the same callbacks in a fresh process.
+Pass `defineDynamic` an `events` object whose handlers return either a single `defineTool(...)`, a `Record<string, defineTool(...)>`, or `null` for no tools. Wrap every entry in `defineTool()`. eve records durable descriptors for `execute`, approval request and response policies, input-scoped `approvalKey` callbacks, and `toModelOutput`, so a parked call can reconstruct the same callbacks in a fresh process.
 
 Dynamic tool executors receive the same `ToolContext` as static authored tools, including inline provider auth through `ctx.getToken(provider)` and `ctx.requireAuth(provider)`.
 
--- a/docs/guides/frontend/overview.mdx
+++ b/docs/guides/frontend/overview.mdx
@@ -306,7 +306,7 @@ then save a final snapshot in `onFinish`.
 
 For multiple chat threads, keep one saved event log and session cursor per thread. `agent`, `host`, `reducer`, `session`, `initialEvents`, `initialSession`, `auth`, `headers`, `optimistic`, and `resume` are read when the hook creates its store, so remount the chat component when switching threads, for example with `key={chat.id}`.
 
-Pass `resume: true` with `initialSession` to rebuild the projection from the durable stream after mount. While `status` is `"resuming"`, render hydrated `data` but disable message and HITL submission; do not present cancellation or active-turn progress controls. If catch-up finds an in-flight turn, `status` changes to `"streaming"` before the binding follows it to a boundary. A settled tail changes directly to `"ready"`, and a terminal session failure changes to `"error"`.
+Pass `resume: true` with `initialSession` to rebuild the projection from the durable stream after mount. While `status` is `"resuming"`, render hydrated `data` but disable message and HITL submission; do not present cancellation or active-turn progress controls. If catch-up finds an in-flight turn, `status` changes to `"streaming"` before the binding follows it to a boundary. A settled tail changes directly to `"ready"` after a bounded catch-up check, without waiting for the live stream idle timeout. If that check finds a newly started turn or pending authorization, eve keeps following it. A terminal session failure changes to `"error"`.
 
 ```tsx
 const agent = useEveAgent({
--- a/docs/memory/custom-provider.md
+++ b/docs/memory/custom-provider.md
@@ -173,7 +173,10 @@ turn recalls the same data again.
 ## Failure behavior
 
 - A throwing or invalid `recall["turn.started"]` fails the turn before the
-  model call. No slot's recall results are committed.
+  model call. No slot's recall results are committed. If the turn's
+  `abortSignal` is already aborted, eve treats the error as cancellation and
+  continues with any queued steering replacement. An `AbortError` with an active
+  signal still fails the turn.
 - A throwing `capture["compaction.requested"]` leaves history unchanged.
 - A throwing `recall["compaction.completed"]` fails an automatic turn. For
   standalone compaction, eve logs the error and returns the session to waiting,
--- a/docs/tools/workflows.mdx
+++ b/docs/tools/workflows.mdx
@@ -317,7 +317,9 @@ export default defineWorkflowTool({
 ```
 
 `createWebhook` mints a URL under `/.well-known/workflow/v1/webhook/` that eve serves. The external
-system posts to it when it is done. Nothing runs in between.
+system posts to it when it is done. Nothing runs in between. Webhook tokens are generated for
+you; use `createHook` with `resumeHook` if you need a deterministic token. To customize the HTTP
+response, pass `respondWith: new Response(...)` to `createWebhook`.
 
 ### Ask now, act when answered
 
--- a/e2e/fixtures/agent-tools-hitl/agent/agent.ts
+++ b/e2e/fixtures/agent-tools-hitl/agent/agent.ts
@@ -3,6 +3,8 @@ import { defineAgent } from "eve";
 import type { MockModelRequest, MockModelResponse } from "eve/evals";
 
 const AUTH_PROBE_DIRECTIVE = /call the auth-probe tool exactly once with marker "([^"]+)"/iu;
+const SCOPED_APPROVAL_DIRECTIVE =
+  /call the dynamic_scoped_approval tool exactly once with scope "([^"]+)"/iu;
 const REPLY_DIRECTIVE = /reply with exactly ([A-Z0-9-]+)/iu;
 
 /**
@@ -17,6 +19,17 @@ function respond(request: MockModelRequest): MockModelResponse | string {
   const reply = REPLY_DIRECTIVE.exec(message);
   if (reply?.[1] !== undefined) {
     return reply[1];
+  }
+
+  const scopedApproval = SCOPED_APPROVAL_DIRECTIVE.exec(message);
+  if (scopedApproval?.[1] !== undefined) {
+    const roles = request.messages.map((entry) => entry.role);
+    if (roles.lastIndexOf("tool") < roles.lastIndexOf("user")) {
+      return {
+        toolCalls: [{ input: { scope: scopedApproval[1] }, name: "dynamic_scoped_approval" }],
+      };
+    }
+    return `Approved scope: ${scopedApproval[1]}`;
   }
 
   const authProbe = AUTH_PROBE_DIRECTIVE.exec(message);
--- a/e2e/fixtures/agent-tools-hitl/agent/tools/dynamic-scoped-approval.js
+++ b/e2e/fixtures/agent-tools-hitl/agent/tools/dynamic-scoped-approval.js
@@ -1,0 +1,24 @@
+import { defineDynamic, defineTool } from "eve/tools";
+
+export default defineDynamic({
+  events: {
+    "session.started": () => ({
+      dynamic_scoped_approval: defineTool({
+        description:
+          "Echoes a scope after approving that scope. Only call when explicitly asked for dynamic_scoped_approval.",
+        inputSchema: {
+          type: "object",
+          properties: { scope: { type: "string" } },
+          required: ["scope"],
+          additionalProperties: false,
+        },
+        approvalKey: (input) => `dynamic_scoped_approval:${input.scope}`,
+        approval: ({ approvedTools, toolInput }) =>
+          approvedTools.has(`dynamic_scoped_approval:${toolInput?.scope}`)
+            ? "not-applicable"
+            : "user-approval",
+        execute: (input) => ({ scope: input.scope }),
+      }),
+    }),
+  },
+});
--- a/e2e/fixtures/agent-tools-hitl/agent/tools/dynamic-scoped-approval.ts
+++ b/e2e/fixtures/agent-tools-hitl/agent/tools/dynamic-scoped-approval.ts
@@ -1,4 +1,5 @@
 import { defineDynamic, defineTool } from "eve/tools";
+import { z } from "zod";
 
 export default defineDynamic({
   events: {
@@ -6,12 +7,7 @@ export default defineDynamic({
       dynamic_scoped_approval: defineTool({
         description:
           "Echoes a scope after approving that scope. Only call when explicitly asked for dynamic_scoped_approval.",
-        inputSchema: {
-          type: "object",
-          properties: { scope: { type: "string" } },
-          required: ["scope"],
-          additionalProperties: false,
-        },
+        inputSchema: z.object({ scope: z.string() }),
         approvalKey: (input) => `dynamic_scoped_approval:${input.scope}`,
         approval: ({ approvedTools, toolInput }) =>
           approvedTools.has(`dynamic_scoped_approval:${toolInput?.scope}`)
--- a/e2e/fixtures/agent-tools-hitl/evals/hitl/dynamic-scoped-approval.eval.ts
+++ b/e2e/fixtures/agent-tools-hitl/evals/hitl/dynamic-scoped-approval.eval.ts
@@ -1,0 +1,28 @@
+import { defineEval } from "eve/evals";
+
+const TOOL_NAME = "dynamic_scoped_approval";
+
+export default defineEval({
+  description:
+    "Dynamic input-scoped approval grants survive replay without authorizing other scopes.",
+  async test(t) {
+    const first = await t.send(`Call the ${TOOL_NAME} tool exactly once with scope "repo-a".`);
+    first.calledTool(TOOL_NAME, { status: "pending", count: 1 });
+    t.requireInputRequest({ toolName: TOOL_NAME });
+    const approved = await t.respondAll("approve");
+    approved.expectOk();
+    approved.calledTool(TOOL_NAME, { status: "completed", count: 1 });
+
+    const repeated = await t.send(`Call the ${TOOL_NAME} tool exactly once with scope "repo-a".`);
+    repeated.succeeded();
+    repeated.calledTool(TOOL_NAME, { status: "completed", count: 1 });
+
+    const other = await t.send(`Call the ${TOOL_NAME} tool exactly once with scope "repo-b".`);
+    other.calledTool(TOOL_NAME, { status: "pending", count: 1 });
+    t.requireInputRequest({ toolName: TOOL_NAME });
+    const approvedOther = await t.respondAll("approve");
+    approvedOther.expectOk();
+    t.succeeded();
+    t.calledTool(TOOL_NAME, { status: "completed", count: 3 });
+  },
+});
--- a/e2e/fixtures/agent-workflow-tools/agent/agent.ts
+++ b/e2e/fixtures/agent-workflow-tools/agent/agent.ts
@@ -40,6 +40,7 @@ function respond(request: MockModelRequest): MockModelResponse | string {
     ["WORKFLOW-ESCALATE-START", "escalate_deploy"],
     ["WORKFLOW-HOLD-START", "hold_deploy"],
     ["WORKFLOW-FANOUT-START", "fanout_deploy"],
+    ["WORKFLOW-WEBHOOK-START", "webhook_deploy"],
     ["WORKFLOW-AGENT-FANOUT-START", "fanout_agents"],
   ] as const) {
     if (!message.includes(directive)) continue;
--- a/e2e/fixtures/agent-workflow-tools/agent/lib/webhook.ts
+++ b/e2e/fixtures/agent-workflow-tools/agent/lib/webhook.ts
@@ -1,0 +1,16 @@
+export async function postWorkflowCallback(url: string, service: string): Promise<void> {
+  "use step";
+
+  const headers: Record<string, string> = { "content-type": "application/json" };
+  const bypass = process.env.VERCEL_AUTOMATION_BYPASS_SECRET;
+  if (bypass) headers["x-vercel-protection-bypass"] = bypass;
+  const response = await fetch(url, {
+    method: "POST",
+    headers,
+    body: JSON.stringify({ service }),
+    signal: AbortSignal.timeout(30_000),
+  });
+  if (response.status !== 202 || (await response.text()) !== "callback accepted") {
+    throw new Error(`Workflow callback returned an unexpected response (${response.status}).`);
+  }
+}
--- a/e2e/fixtures/agent-workflow-tools/agent/tools/webhook_deploy.ts
+++ b/e2e/fixtures/agent-workflow-tools/agent/tools/webhook_deploy.ts
@@ -1,0 +1,21 @@
+import { defineWorkflowTool } from "eve/tools";
+import { createWebhook } from "workflow";
+import { z } from "zod";
+
+import { postWorkflowCallback } from "../lib/webhook.ts";
+
+export default defineWorkflowTool({
+  description: "Verify a deploy callback through the public workflow webhook route.",
+  inputSchema: z.strictObject({ service: z.string() }),
+  async execute({ service }) {
+    "use workflow";
+
+    using callback = createWebhook({
+      respondWith: new Response("callback accepted", { status: 202 }),
+    });
+    await callback.getConflict();
+    await postWorkflowCallback(callback.url, service);
+    const request = await callback;
+    return { callback: await request.json() };
+  },
+});
--- a/e2e/fixtures/agent-workflow-tools/evals/webhook-deploy.eval.ts
+++ b/e2e/fixtures/agent-workflow-tools/evals/webhook-deploy.eval.ts
@@ -1,0 +1,12 @@
+import { defineEval } from "eve/evals";
+
+export default defineEval({
+  description: "A bundled workflow webhook accepts a public callback and preserves its response.",
+  async test(t) {
+    const turn = await t.send("WORKFLOW-WEBHOOK-START");
+    turn.expectOk();
+    turn.calledTool("webhook_deploy", { output: /"callback":\{"service":"api"\}/u });
+    turn.messageIncludes("WORKFLOW-WEBHOOK-RESULT");
+    t.noFailedActions();
+  },
+});
--- a/packages/eve/extension-contracts/compatibility/tool/v29.ts
+++ b/packages/eve/extension-contracts/compatibility/tool/v29.ts
@@ -1,0 +1,20 @@
+import { z } from "zod";
+import { defineTool, defineWorkflowTool } from "#public/tools/index.js";
+
+export const write = defineTool({
+  description: "Write an approved message.",
+  inputSchema: z.object({ message: z.string() }),
+  outputSchema: z.object({ written: z.string() }),
+  approval: ({ toolInput }) => (toolInput?.message ? "user-approval" : "not-applicable"),
+  execute: (input) => ({ written: input.message }),
+  toModelOutput: (output) => ({ type: "text", value: output.written }),
+});
+
+export const workflow = defineWorkflowTool({
+  description: "Run a report workflow.",
+  inputSchema: z.object({ report: z.string() }),
+  async execute(input) {
+    "use workflow";
+    return { report: input.report };
+  },
+});
--- a/packages/eve/extension-contracts/reports/dynamicTool/v29.json
+++ b/packages/eve/extension-contracts/reports/dynamicTool/v29.json
@@ -1,0 +1,13 @@
+{
+  "kind": "eve-extension-capability-contract",
+  "capability": "dynamicTool",
+  "epoch": 29,
+  "sha256": "119a44f66829ba64537d008f89f18681d6636fa1343270f8f512966ae43ad1c1",
+  "exports": [
+    "DynamicToolEntry",
+    "DynamicToolEvents",
+    "DynamicToolResult",
+    "DynamicToolSet",
+    "defineDynamic"
+  ]
+}
--- a/packages/eve/extension-contracts/reports/tool/v30.json
+++ b/packages/eve/extension-contracts/reports/tool/v30.json
@@ -1,0 +1,20 @@
+{
+  "kind": "eve-extension-capability-contract",
+  "capability": "tool",
+  "epoch": 30,
+  "sha256": "9aa353a05fd889df5ea41d25f095b162e829f45c69dd6470dfb1cd54aa054545",
+  "exports": [
+    "defaultWebSearch",
+    "defineTool",
+    "defineWorkflowTool",
+    "disableTool",
+    "experimental_workflow",
+    "isDisabledToolSentinel",
+    "isExperimentalWorkflowToolDefinition",
+    "isWebSearchToolDefinition",
+    "toolOutput",
+    "toolOutputPart",
+    "toolResultFrom",
+    "webSearch"
+  ]
+}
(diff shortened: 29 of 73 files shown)


AndrewBarba (commented):
Reviewed the final implementation with independent subagent cross-reviews. No outstanding correctness findings remain.

The review checked scoped approval recording and reuse across static, step-, turn-, and session-defined tools; structured and text approval responses; authorization boundaries; stream-read failures versus protocol validation; bounded resume races; webhook registration and native request serialization; Unicode truncation; and safe CLI output draining.

The added regression tests caught and drove fixes for the initial webhook serialization and dynamic approval settlement defects. Public TypeScript authoring and retained extension contracts are covered as well.

Reviewed commit `34d1ef44d764bb5cb2d9e65c25a05ad8d2661885`. Remaining CI jobs are being monitored before handoff.

ctgowrie (approved):

PR #3108, opened 2026-09-07
fix(eve): retry model calls cut by an AI Gateway stream timeout

### LINKED ISSUE #3084: classifyModelCallError never retries the AI Gateway's gateway_stream_timeout
### Summary

`classifyModelCallError` never classifies the AI Gateway's `gateway_stream_timeout` as retryable, so a model call the gateway aborts mid-stream gets **zero** retries and the turn is parked. Reproduced on `eve@0.52.1`.

The error is transient by nature, and `runModelCallWithRetries` is right there ready to retry it, but the classifier does not recognise the shape.

### What the gateway sends

From a production Vercel log (`POST /.well-known/workflow/v1/flow`):

```
[eve:harness.tool-loop] tool-loop stream error {
  error: {
    message: 'Stream exceeded maximum duration before function timeout',
    detail: "{ code: 'gateway_stream_timeout',
               message: 'Stream exceeded maximum duration before function timeout',
               origin: 'gateway' }"
  }
}
```

Then, in the same turn:

```
[eve:harness.tool-loop] model call failed — parking session for retry by the user {
  error: { message: 'Error: Stream exceeded maximum duration before function timeout',
           detail: 'Error: Stream exceeded maximum duration before function timeout
    at toError (/var/task/_libs/eve.mjs)
    at emitStreamContent (/var/task/_libs/eve.mjs)
    at async executeModelCall (/var/task/_libs/eve.mjs)
    at async runModelCallWithRetries (/var/task/_libs/eve.mjs)' }
}
```

No `model call failed transiently — retrying` warning appears anywhere in the run, so it was not retried even once.

### Why it is not retried

`isRetryableGatewayType` matches on the gateway **`type`** field:

```js
function isRetryableGatewayType(e) {
  return e === `overloaded_error` || e === `rate_limit_exceeded` || e === `timeout_error`
}
```

but this failure arrives as **`code: 'gateway_stream_timeout'`** with `origin: 'gateway'`. `readModelCallErrorSignals` only reads `type`, never `code`, so `gatewayType` and `upstreamType` are both `undefined`.

It then falls through everything else: there is no HTTP status to match (the stream is aborted mid-flight rather than rejected), and the message contains none of the `isLikelyNetworkError` substrings (`econnreset`, `etimedout`, `socket hang up`, `network`, `fetch failed`). The default is `recoverable`, and `runModelCallWithRetries` only retries on `retry`:

```js
if (n === 3 || classifyModelCallError(e) !== `retry`) throw e
```

Grepping the published tarball, the string `gateway_stream_timeout` does not appear anywhere in `eve@0.52.1`.

### Reproduction

Against `eve@0.52.1`'s own classifier:

```js
import { classifyModelCallError } from "eve/dist/src/harness/model-call-error.js";

const bare = new Error("Stream exceeded maximum duration before function timeout");

const withPayload = new Error("Stream exceeded maximum duration before function timeout");
withPayload.cause = {
  code: "gateway_stream_timeout",
  message: "Stream exceeded maximum duration before function timeout",
  origin: "gateway",
};

const control = new Error("timeout");
control.cause = { name: "GatewayTimeoutError", type: "timeout_error" };
```

```
recoverable <- bare stream-timeout error (as thrown in prod)
recoverable <- stream-timeout + gateway payload {code:'gateway_stream_timeout'}
retry       <- control: gateway {type:'timeout_error'}
```

Same result on `0.19.0`, `0.51.1` and `0.52.1`.

### Impact

The user-visible result is a dead end. eve's default Slack `turn.failed` copy fires:

> I hit an error while handling your request (Stream exceeded maximum duration before function timeout).
> Please try again, rephrase, or reach out if it keeps failing.

"Try again" is the one thing that cannot work here, because the retry re-runs the same oversized pass. In our case a user asked an analyst agent for a two-week review, it streamed for about two and a half minutes, the gateway cut it, and sixteen minutes of work was lost with advice that could only fail again.

A single automatic retry would not have helped that particular request either, since it would time out the same way. But the classification is still wrong, and it matters for the ordinary case: a call that streams close to the ceiling and is cut once will usually complete on a second attempt, exactly what `runModelCallWithRetries` exists for.

### Suggested fix

Recognise the gateway `code` field alongside `type`, and treat `gateway_stream_timeout` as retryable:

```js
function isRetryableGatewayType(e) {
  return e === `overloaded_error`
    || e === `rate_limit_exceeded`
    || e === `timeout_error`
    || e === `gateway_stream_timeout`   // arrives as `code`, not `type`
}
```

with `readModelCallErrorSignals` also reading `code` into the signal it compares. Reading `code` generally seems worthwhile regardless: any other gateway error that reports itself that way is currently invisible to the classifier and silently lands on the `recoverable` default.

Happy to open a PR if that would help, though the signed-commit requirement in CONTRIBUTING is not something I can satisfy from this environment.

### Environment

- `eve` 0.51.1 in production, reproduced against 0.52.1
- `ai` 7.0.92, Node 24, Vercel Fluid, `maxDuration: "max"` on the workflow function
- model `anthropic/claude-sonnet-4.6` via AI Gateway

Benny-Gil:
### Summary

Model calls that AI Gateway aborts with `gateway_stream_timeout` now retry through the existing model-call retry loop instead of parking the session after the first cut stream.

- **The classifier fell through to `recoverable`.** The gateway sends the abort as a plain payload keyed on `code`, and the catalog only read the body `type`, so the turn parked with zero retries.
- **One transient catalog rule fixes it.** `gateway-stream-timeout` matches `codeIs("gateway_stream_timeout")`, and the classifier already returns `retry` for any rule tagged `transient`.
- **The rule lives in the catalog, not in `classifyModelCallError`.** The catalog is the documented growth point for new error shapes, and the payload reaches the cause chain through `normalizeModelStreamError`.
- **The summary copy matches its sibling rules.** Users see a one-line cause and a hint to split the task or switch models when retries do not help.

| Area         | Lines added |
| ------------ | ----------- |
| Catalog rule | 19          |
| Changeset    | 5           |
| Tests        | 24          |
| **Total**    | **48**      |

Closes #3084

The reporter also proposed reading `code` inside `classifyModelCallError` itself; that broader change is left for a maintainer decision.

### Validation

Commands run locally on macOS:

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/harness/model-call-error.test.ts src/harness/semantic-errors/semantic-errors.test.ts` passes with 84 tests.
- `pnpm lint`, `pnpm --filter eve run typecheck`, and `pnpm guard:invariants` pass.
- `pnpm test:unit` passes except two tests in `packages/eve/src/cli/telemetry/preference.test.ts`. They fail identically on `origin/main` in this shell because `XDG_CONFIG_HOME` is set, and this branch does not touch telemetry. Three unrelated files timed out under load in the full run and pass on rerun on both this branch and `origin/main`.

The integration, scenario, and e2e suites were not run. The change has no fixture, subprocess, or HTTP surface.

### Checklist

### DIFF
--- a/.changeset/retry-gateway-stream-timeout.md
+++ b/.changeset/retry-gateway-stream-timeout.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Retry model calls that AI Gateway aborts with `gateway_stream_timeout` instead of parking the session after the first cut stream.
--- a/packages/eve/src/harness/model-call-error.test.ts
+++ b/packages/eve/src/harness/model-call-error.test.ts
@@ -185,6 +185,19 @@ describe("classifyModelCallError", () => {
     expect(classifyModelCallError(new Error("stream failed", { cause: overloaded }))).toBe("retry");
   });
 
+  it("returns retry for an AI Gateway stream timeout payload", () => {
+    // The gateway aborts a long stream with a plain payload keyed on
+    // `code`; the harness wraps it through normalizeModelStreamError.
+    const payload = {
+      code: "gateway_stream_timeout",
+      message: "Stream exceeded maximum duration before function timeout",
+      origin: "gateway",
+    };
+
+    expect(classifyModelCallError(normalizeModelStreamError(payload))).toBe("retry");
+    expect(classifyModelCallError(payload)).toBe("retry");
+  });
+
   it("returns retry for HTTP statuses the AI SDK treats as retryable", () => {
     for (const statusCode of [408, 409, 429]) {
       const err = Object.assign(new Error("retryable"), { statusCode });
--- a/packages/eve/src/harness/semantic-errors/rules/gateway.ts
+++ b/packages/eve/src/harness/semantic-errors/rules/gateway.ts
@@ -1,4 +1,12 @@
-import { allOf, anyOf, messageMatches, nameIs, typeIs, type SemanticErrorRule } from "../rule.js";
+import {
+  allOf,
+  anyOf,
+  codeIs,
+  messageMatches,
+  nameIs,
+  typeIs,
+  type SemanticErrorRule,
+} from "../rule.js";
 
 /** The summary `name` shared by the gateway-auth rule variants. */
 const GATEWAY_AUTH_FAILURE_SUMMARY_NAME = "AI Gateway authentication failed";
@@ -104,5 +112,15 @@ export const GATEWAY_RULES: readonly SemanticErrorRule[] = [
     when: anyOf(nameIs("GatewayTimeoutError"), typeIs("timeout_error", "overloaded_error")),
     message: "The model provider is overloaded or timing out upstream of AI Gateway.",
     hint: "This is transient — retry shortly, or switch models with `/model` in `eve dev`.",
+  },
+  {
+    id: "gateway-stream-timeout",
+    name: "AI Gateway stream timed out",
+    tags: ["gateway", "transient"],
+    // The gateway aborts a long stream with a payload that carries `code`,
+    // not the body `type` the rule above reads, so it needs its own match.
+    when: codeIs("gateway_stream_timeout"),
+    message: "AI Gateway ended the model stream before the response finished.",
+    hint: "Retries are automatic. If it persists, split the task into smaller turns or switch models with `/model` in `eve dev`.",
   },
 ];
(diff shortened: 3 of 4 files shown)


Benny-Gil:
Closing until a maintainer invites the implementation, as CONTRIBUTING.md asks. The branch stays on my fork, and the discussion continues on #3084.

PR #3127, opened 2026-09-07
fix(eve): opt-in task context caching

bryan-hunter:
### Summary

Changing task status breaks the reusable prompt prefix before conversation history. Add `experimental.taskContextAtTail` to move it after history. Default behavior is unchanged.

```ts
defineAgent({
  // ...
  experimental: { taskContextAtTail: true },
});

// Before (simplified)
[system, "2 tasks pending", ...history]
[system, "1 task pending",  ...history]

// Opted in: only the tail changes
[system, ...history, "2 tasks pending"]
[system, ...history, "1 task pending"]
```

Applies to task status and task-delivery context. On the first model step this also changes that context from system-role to user-role. Authored system instructions, scheduled instructions, and tool-approval ordering are unchanged. Opt-in while we evaluate model behavior and provider cache reuse.

### Validation

- Focused harness, session, compiler, and runtime tests pass, covering default ordering, opted-in placement, and session hydration.
- Synthetic 28 KB request retained ~30 characters of common prefix before the change versus ~28 KB after it. This measures request construction, not billed cache savings.
- No live model savings measured yet.

### Checklist

### DIFF
--- a/.changeset/task-context-tail.md
+++ b/.changeset/task-context-tail.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add opt-in `experimental.taskContextAtTail` to place changing task context after conversation history for prompt cache reuse. Default message ordering remains unchanged.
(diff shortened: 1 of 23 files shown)


bryan-hunter:
Superseded by #3129. Moving the mutable task inventory to the tail improves placement but still resends the inventory and retains the launch/pending/settled reporting policy. The replacement opts into existing chronological task receipts and notifications, lets the parent decide when to continue or report, and restores task identities after compaction. Default behavior stays unchanged; notification batching remains separate in #3128.

PR #3129, opened 2026-09-08
feat(eve): preserve prompt cache prefixes with task event delivery

bryan-hunter:
## Withdrawn — replaced by #3135

The original cache explanation was incorrect: delivered task snapshots already append to durable history. They do not replace earlier snapshots.

The reproduced prefix change occurs in the separate temporary first-turn task-guidance path. #3135 addresses that placement while retaining default task reporting and dispatch.

This branch also proposed dispatching children between parent steps and changing reporting policy. Those broader semantics are not required for the narrow fix and are no longer being pursued here. Completion batching remains independent in #3128.

The earlier live runs did not establish a controlled cache-savings result for this PR.

### DIFF
--- a/.changeset/task-event-delivery.md
+++ b/.changeset/task-event-delivery.md
@@ -2,4 +2,4 @@
 "eve": patch
 ---
 
-Add opt-in `experimental.taskEventDelivery` for agents that orchestrate background work. Task notifications remain chronological instead of reinjecting mutable cohort state; the agent decides when to continue or report, with task identities restored after compaction.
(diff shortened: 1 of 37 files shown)


bryan-hunter:
Superseded by #3135. Correction: delivered task snapshots already append to durable history; they do not rewrite earlier snapshots. The confirmed prefix change is the separate temporary first-turn task guidance in system instructions. #3135 addresses that placement without this PR’s experimental API, dispatch changes, or reporting-policy changes. No controlled provider cache-savings claim is being made.

PR #3131, opened 2026-09-08
fix(eve): persist framework prompt announcements

ruiconti:
### Summary

Framework announcements reached the model but were omitted from `session.history`. When task or skill state changed, eve rebuilt that context on the next request, changing an earlier part of the prompt and reducing cache reuse. This change saves those announcements with the conversation so subsequent requests retain the existing prefix and append updates. It also removes the pending-task silence instruction, which could suppress later user replies when retained in history.

Related to #3129.

### Behavior

`promptMessages` now receives `currentMessages.history`, including task, skill, and delivery announcements as user-role messages from the first turn. Task snapshots are read from `session.state` before each model call, so they include newly started work.

For two background tasks, A and B:

```text
Before
Call 1: [system]             [user request]
Call 2: [system + A pending] [user request] [start A → accepted]
Call 3: [system + A pending] [user request] [start A → accepted] [start B → accepted]

After
Call 1: [system] [user request]
Call 2: [system] [user request] [start A → accepted] [A pending]
Call 3: [system] [user request] [start A → accepted] [A pending] [start B → accepted] [A pending, B pending]
```

Previously, starting A changed the system prompt, and the snapshot remained stale when B started. Each new request now extends the preceding one, with unchanged announcements omitted.

`HistoryStateKey` tracks the last announcements saved in history. It advances only when a step returns updated history and resets when history is cleared or compacted. Failed model calls do not advance it, and completed compaction survives a later model failure. Client context and temporary approval-resume instructions remain outside durable history.

The pending-task silence instruction is removed. Once retained in history, it could suppress a later user question: a partial completion followed by “what is 7×8?” produced `<eve-empty-delivery/>` instead of `56`. Pending wakes still accept empty responses, but no longer instruct the model to be silent. Existing batching combines completions already queued; it does not wait for unfinished siblings, so partial reports are allowed.

### Validation

The new real-model eval launches five background reviewers in parallel, covering both the first and a later user turn. It verifies overlapping work, all five completion notifications, and a final report. Every subsequent parent request must read at least **98% of the preceding input** from the provider cache; a single miss below that threshold fails the eval.

Both cases pass on [OpenAI](https://github.com/vercel/eve/actions/runs/34299614423/job/102303587114) and [Anthropic](https://github.com/vercel/eve/actions/runs/34299614423/job/102303587059) at `e8581ac88`, with all twelve cache checks passing per provider. The same scenario on [unpatched main](https://github.com/vercel/eve/actions/runs/34294439356) completes the reviews but fails every cache check, demonstrating that it catches the regression.

Unit and integration coverage verifies announcement persistence across serialized steps, clear, compaction, cancellation, and approval resume. The compaction fixtures now request the review and handoff their agent performs and return repository findings and handoff notes instead of repeated filler; their two-compaction and no-repeat assertions remain. Unit, integration, scenario, compaction, task-reporting, lint, and typecheck checks passed at `e8581ac88`.

An intermittent failure remains unexplained: earlier OpenAI runs reported an isolated zero-cache response despite an unchanged prefix in the captured trace, and [correctly failed CI](https://github.com/vercel/eve/actions/runs/34294031305/job/102288824722). The latest run had no such misses. Neither provider's latest eval log contains filtering or an empty-response retry.

The [comparison results](https://github.com/vercel/eve/pull/3131#issuecomment-5604736862) show why the pending-task silence instruction was removed.

The final regression uses the ordinary fixture model without the experiment wrapper. It requires the user answer before sibling tasks finish and a complete final report; the compaction case also retains its complete-report assertion. At `2e6c44266`, rebased onto main `d630b218a`, 390 focused unit tests, 55 harness integration tests, build, all 45 typecheck tasks, frozen-lockfile install, conflict-file lint/format checks, documentation checks, and invariant checks pass locally. Real-model CI for the rebased revision is pending; the [pre-rebase run](https://github.com/vercel/eve/actions/runs/34371424852) tests `75241a560`.

### Checklist


De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

De-ASI-INTERFACE (approved):

AndrewBarba (approved):

ruiconti:
The parent should answer a new user question while background work continues.

The test starts three warehouse lookups. After a partial result arrives, the user asks:

> Alice is packing seven boxes with eight jars in each. How many jars is that?

Expected: a reply containing **56**, before the remaining lookups finish.

Both groups receive the same silence instruction on the first completion wake. At the user question, we either keep that instruction in the model's history or remove it. We ran 20 interleaved trials per group, with the same batching code.

| Behavior being checked (Opus) | Expected | Instruction kept at user question | Instruction removed at user question |
| --- | --- | ---: | ---: |
| Answer the jar-count question before the remaining lookups finish | **20/20** replies containing `56` | **4/20 — fail** | **20/20 — pass** |
| Suppress the user's answer with `<eve-empty-delivery/>` | **Never (0/20)** | **16/20 — fail** | **0/20 — pass** |
| Report all three lookup results after the work finishes | **20/20** complete reports | **20/20 — pass** | **20/20 — pass** |

The first completion wake was silent in **20/20 trials in both groups**, before the user asked the question. That is a shared experiment setup check; the behavior under test is answering the subsequent user question and delivering the final results.

The instruction meant for a background notification was also suppressing the user's answer. [Opus results and traces](https://github.com/vercel/eve/actions/runs/34363146587/job/102505056779).

[Sol answered 20/20 in both groups](https://github.com/vercel/eve/actions/runs/34363146587/job/102505056842), so the effect depends on the model. These were separate conversations, not identical replays; this comparison measures the instruction's effect, not batching improvements.

PR #3155, opened 2026-09-08
feat(eve): add semantic model message kinds

chadhietala:
### Summary

eve sometimes adds `role: "user"` messages because a provider requires a user-final history, but traces previously could not say why that message was present. Brand framework input with a namespaced kind: `context.instruction`, `context.state`, `context.compaction`, `memory.load`, `execution.background_task`, `execution.continuation`, or `execution.retry`; genuine user input still has no `kind`. `context.*` covers injected context, `memory.load` recalled memory, and `execution.*` task and recovery lifecycle messages. Task messages are classified from their task-owned delivery and tagged before adapter results coalesce; cancelled deliveries retain their kind without changing the public channel contract.

### Validation

- `CI=true pnpm --filter eve build`
- `CI=true pnpm --filter eve test:unit` — 8,487 passed, 1 skipped
- `CI=true pnpm --filter eve typecheck`
- `CI=true pnpm exec oxfmt --check` and `CI=true pnpm exec oxlint` on changed TypeScript files
- `git diff --check`

### Checklist

### DIFF
--- a/packages/eve/src/execution/workflow-steps.ts
+++ b/packages/eve/src/execution/workflow-steps.ts
@@ -47,7 +47,7 @@ import { setChannelContext } from "#execution/channel-context.js";
 import { observeSessionActivity } from "#execution/session-activity-projection.js";
 import { hasPendingInputBatch } from "#harness/input-requests.js";
 import { activeTurnId } from "#harness/active-turn-id.js";
-import { coalesceTurnInputs } from "#harness/messages.js";
+import { coalesceTurnInputs, type UserModelMessage } from "#harness/messages.js";
 import { getWorkflowTaskCallIds, isWorkflowTaskInterrupt } from "#harness/workflow-task-state.js";
 import { getPendingWorkflowInterrupt } from "#harness/workflow-interrupt-state.js";
 import type { HandleEventFn, HarnessSession, StepInput, StepResult } from "#harness/types.js";
@@ -76,6 +76,8 @@ import { createDurableSessionState, readDurableSession } from "#execution/durabl
 import type { TurnStepInput } from "#execution/durable-session-migrations/turn-workflow.js";
 import { buildRuntimeIdentity, createExecutionNodeStep } from "#execution/node-step.js";
 import {
+  getBackgroundTaskDelivery,
+  markBackgroundTaskStepInput,
   resolveInitiatingTaskContext,
   resolveTaskDeliveryContext,
 } from "#tasks/delivery-context.js";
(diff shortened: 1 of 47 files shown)


packages/eve/src/execution/workflow-steps.ts
```diff
@@ -282,6 +282,9 @@ export async function turnStep(rawInput: TurnStepInput): Promise<DurableStepResu
         context: [...(resolved.context ?? []), taskContext.context],
       };
     }
+    if (resolved.message !== undefined) {
+      resolved = markFrameworkStepInput(resolved, "execution.background_task");
+    }
```
ruiconti: This seems like an indirect way of detecting background tasks

packages/eve/src/harness/compaction.ts
```diff
@@ -135,7 +135,7 @@ function toolResultCapHeuristic(input: CompactionHeuristicInput): CompactionHeur
     input.previousCheckpoint === undefined
       ? []
       : [
-          { content: COMPACTION_CHECKPOINT_MARKER, role: "user" },
+          createFrameworkUserMessage("context.compaction", COMPACTION_CHECKPOINT_MARKER),
```
ruiconti, line 138: Don't we also emit a compaction event? Worried about exposing redundant signals

chadhietala: We currently do not. The reason why I'm branding them is really just so we can treat them differently when we render.

packages/eve/src/harness/tool-loop.ts
```diff
@@ -962,20 +964,27 @@ export function createToolLoopHarness(config: ToolLoopHarnessConfig): StepFn {
...
     const ephemeralContextMessages: ModelMessage[] =
-      activeClientContext?.map((content) => ({ content, role: "user" })) ?? [];
+      activeClientContext?.map((content) =>
+        createFrameworkUserMessage("context.instruction", content),
+      ) ?? [];
     const preparedTurnInput: ModelMessage[] = [];
     if (effectiveStepInput?.context !== undefined && pending.deferredContext !== true) {
       for (const entry of effectiveStepInput.context) {
-        preparedTurnInput.push({ content: entry, role: "user" });
+        preparedTurnInput.push(createFrameworkUserMessage("context.instruction", entry));
```
ruiconti, line 980: @chadhietala one thought so this pattern is preserved using the type system would be to tag type the messages with `kind` and ensure that all abstractions expect said type. 

That way we can (at least somewhat) ensure event data is produced consistently when changes happen.

ruiconti (approved):

PR #3175, opened 2026-09-09
feat(eve): add name-addressed workspace agents

OwenKephart:
### Summary

An agent that knows one workspace peer should not need glob selection or a dynamic remote-agent map. This adds `defineWorkspaceAgent({ name })`, which resolves an exact member under `agents/`, compiles it into the existing remote-subagent lifecycle, and uses the peer's root agent description by default. Omitted `transport` selects Vercel routing with OIDC; an explicit `{ url, auth?, headers? }` transport replaces that default elsewhere. Local workspace routing, peer-set discovery, and selector expansion remain out of scope. Related to #2768 and #2786.

### Validation

- `pnpm build`
- `pnpm fmt`
- `pnpm fmt --check`
- `pnpm lint` (passes with two pre-existing `no-control-regex` warnings in unrelated files)
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/definitions/workspace-agent.test.ts src/internal/structural-tests/file-length.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/internal/nitro/host/build-application.scenario.test.ts src/compiler/workspace-agent.scenario.test.ts`
- `pnpm typecheck`
- `git diff --check`

### Checklist

### DIFF
--- a/packages/eve/src/compiler/extension-compatibility.ts
+++ b/packages/eve/src/compiler/extension-compatibility.ts
@@ -72,8 +72,8 @@ const EXTENSION_CAPABILITY_CONTRACTS = {
     },
   },
   subagent: {
-    current: 12,
-    supported: [3, 4, 6, 7, 8, 9, 10, 11, 12],
+    current: 13,
+    supported: [3, 4, 6, 7, 8, 9, 10, 11, 12, 13],
     dropped: {
       1: "Persistent subagent sessions are now the default and the experimental opt-in was removed",
       2: "Persistent subagent sessions are now the default and the experimental opt-in was removed",
--- a/packages/eve/src/compiler/resolve-workspace-subagent.ts
+++ b/packages/eve/src/compiler/resolve-workspace-subagent.ts
@@ -15,7 +15,6 @@ export async function resolveWorkspaceSubagentDefinition(input: {
   readonly registries: readonly AgentSourceRegistry[];
   readonly source: LocalSubagentSourceRef;
 }): Promise<Extract<NormalizedSubagentConfig, { readonly kind: "remote" }>> {
-  if (input.definition.description.trim().length > 0) return input.definition;
   const workspaceContext = await findEveProjectContext(input.source.rootPath);
   if (workspaceContext?.kind !== "workspace-member") {
     throw new Error(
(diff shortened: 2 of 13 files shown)


packages/eve/src/compiler/extension-compatibility.ts
```diff
@@ -70,8 +70,8 @@ const EXTENSION_CAPABILITY_CONTRACTS = {
     },
   },
   subagent: {
-    current: 10,
-    supported: [3, 4, 6, 7, 8, 9, 10],
+    current: 11,
```
AndrewBarba, line 73: This is currently blocking merge: the required scenario check is red because this assertion still expects subagent 10, while this change generates 11. Please update the fixture expectation (or explain why the generated contract should remain 10) and rerun CI.

packages/eve/src/compiler/resolve-workspace-subagent.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { NormalizedSubagentConfig } from "#compiler/normalize-subagent.js";
+import type { AgentSourceRegistry } from "#compiler/source-graph.js";
+
+export async function resolveWorkspaceSubagentDefinition(input: {
+  readonly definition: Extract<NormalizedSubagentConfig, { readonly kind: "remote" }>;
+  readonly path: string;
+  readonly registries: readonly AgentSourceRegistry[];
+  readonly source: LocalSubagentSourceRef;
+}): Promise<Extract<NormalizedSubagentConfig, { readonly kind: "remote" }>> {
+  if (input.definition.description.trim().length > 0) return input.definition;
```
AndrewBarba: Can we validate the workspace path before this early return? With a description override, a malformed or nonexistent path skips member resolution and can silently produce a plausible but wrong runtime URL.

packages/eve/src/compiler/resolve-workspace-subagent.ts
```diff
@@ -0,0 +1,67 @@
...
+  const workspaceContext = await findEveProjectContext(input.source.rootPath);
+  if (workspaceContext?.kind !== "workspace-member") {
+    throw new Error(
+      `Workspace subagent "${input.source.logicalPath}" must be authored by an agent workspace member.`,
+    );
+  }
+  const member = workspaceContext.workspace.members.find(
+    (candidate) => relative(workspaceContext.workspace.root, candidate.appRoot) === input.path,
```
AndrewBarba: Can we normalize this comparison to the repo logical-path format? `relative()` returns backslashes on Windows, while the authored API uses slash-separated paths like `agents/research`.

packages/eve/src/public/definitions/workspace-agent.ts
```diff
@@ -0,0 +1,95 @@
+import { getVercelOidcToken } from "#compiled/@vercel/oidc/index.js";
```
AndrewBarba: Can we reuse eve’s existing `vercelOidc()` auth wrapper here instead of importing `getVercelOidcToken` and duplicating the header logic? That keeps this transport aligned with the existing auth behavior.

chadhietala (approved):

PR #3183, opened 2026-09-09
fix(eve): trace background actions through task settlement

chadhietala:
### Summary

Background workflow tools and subagents return a task receipt before durable work settles, but `agent.action` previously ended at the receipt and could report success before a later task failure or cancellation. The only new durable bookkeeping is `taskId?` on the existing serialized action record beside its scope; it is removed when the terminal task view settles the action. `callId` on `backgroundTasks` is transient cancellation-race transport, and rollback only carries existing observability state forward; there is no separate task-to-action map, persisted span object, or new task registry. Error details use the ordinary output-redaction decision, while the AI SDK `execute_tool` span remains scoped to the receipt.

### Validation

- Focused action lifecycle, background settlement, turn rollback, and tracing unit suite passed: 13 files, 229 tests.
- Instrumentation layout integration test passed: 1 file, 13 tests.
- Session-driver workflow bundle scenario passed, confirming the workflow body does not import Node.js builtins.
- `node ./scripts/guard-invariants.mjs` passed: all mechanical lints passed.
- `tsc -p packages/eve/tsconfig.json --noEmit` passed.
- `runtime.ts` remains below the invariant limit at 687 lines.
- `oxlint` passed for the changed TypeScript files.
- `node ./scripts/check-docs.mjs` passed.
- `node ./scripts/check-doc-snippets.mjs` passed.
- `node ./scripts/check-doc-mdx.mjs` passed.
- `git diff --check` passed.

### Checklist

### DIFF
--- a/packages/eve/src/execution/turn-workflow.ts
+++ b/packages/eve/src/execution/turn-workflow.ts
@@ -143,7 +143,7 @@ export async function runTurnOwnedWorkflow(
           throw new Error("Background tasks were returned without their committed session state.");
         }
         await cursor.adopt({
-          serializedContext: beforeStep.serializedContext,
+          serializedContext: result.backgroundTaskContext ?? beforeStep.serializedContext,
           sessionState: result.backgroundTaskState,
         });
         await acknowledgeDelegatedTasksStep({ tasks: result.backgroundTasks ?? [] });
--- a/.changeset/trace-background-action-lifetime.md
+++ b/.changeset/trace-background-action-lifetime.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Keep `agent.action` spans for background tools and subagents open until their tasks complete, fail, or are cancelled, and record the task's final outcome and policy-controlled error details instead of treating its receipt as completion.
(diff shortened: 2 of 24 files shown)


packages/eve/src/execution/turn-workflow.ts
```diff
@@ -143,7 +144,10 @@ export async function runTurnOwnedWorkflow(
           throw new Error("Background tasks were returned without their committed session state.");
         }
         await cursor.adopt({
-          serializedContext: beforeStep.serializedContext,
+          serializedContext: preserveSerializedBackgroundTaskObservabilityState(
+            beforeStep.serializedContext,
+            result.serializedContext,
+          ),
```
ruiconti: This seems sus. Why do we need an additional state writing related to tasks? @e0 for preliminary analysis

ruiconti: @e0 could we perform the correlation during the _existing_ task store writes? my concern is with unnecessary writes sprawl

ruiconti: > fold correlation into the existing background-task commit abstraction and remove the whole-context preservation from turn-workflow.ts


This makes sense to me, wdyt @chadhietala?

ruiconti (approved):

PR #3184, opened 2026-09-09
[1/2] feat(eve): expose workflow history context

benpankow:
### Summary

Adds `ctx.history` so authored workflows can use the conversation that led to a tool call, rather than relying on the model to summarize it into tool arguments. This supports application-controlled handoffs and research with the actual prior context. The value is a detached, read-only copy of the completed prefix; it excludes the calling tool exchange and grants no live session access.

```ts
// Inside defineWorkflowTool({ ... }):
async execute(input, ctx) {
  "use workflow";
  return await prepareHandoff(input, ctx.history); // application-owned step
}
```

History mutation and session initialization follow in #3185 and #3186.

### Validation

Adds a real workflow integration test for the captured pre-tool-call prefix, plus copy-isolation tests for attachment values.

🤖

### DIFF
--- a/.changeset/workflow-history-context.md
+++ b/.changeset/workflow-history-context.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Expose an immutable completed conversation prefix to workflow tools as `ctx.history`. The captured history excludes the calling tool exchange, letting workflows carry prior context into controlled follow-up work without a session history endpoint.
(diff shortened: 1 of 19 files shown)


benpankow:
Closing this workflow-history API exploration. The surrounding stack and follow-ups did not continue, and exposing workflow history is a sensitive public API that should be reconsidered from current requirements rather than revived here.

PR #3185, opened 2026-09-09
[2/2] feat(eve): preload workflow subagent history

benpankow:
### Summary

Allows a workflow tool to seed a new local subagent with conversation history before its first message. This keeps the API narrow to workflow-owned delegation rather than exposing general session creation or history mutation.

```ts
const result = await ctx.agent({
  key: "private-research-session",
  target: "research",
  history: ctx.history,
  message: task,
});
```

`history` cannot be combined with `agentId` or used with a remote agent. Depends on #3184.

### Validation

Adds unit coverage for fresh local preloading, existing and remote agent rejection, child session history ordering, and durable inbox wire round-tripping with older-version rejection.

🤖

### DIFF
--- a/.changeset/workflow-subagent-history.md
+++ b/.changeset/workflow-subagent-history.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Allow workflow tools to preload a new local subagent with conversation history through `ctx.agent({ history })`.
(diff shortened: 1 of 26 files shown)


benpankow:
Closing with #3184. This depends on the abandoned workflow-history API and should not be revived independently; local subagent history behavior can be redesigned if the requirement returns.

PR #3211, opened 2026-09-09
fix(eve): do not fail turns expecting structured output when background work is active

benpankow:
### Summary

Prevents a conversation turn which must emit structured output as pending as long as it still has outstanding background work. This presents it from failing while the background work completes. Once background work completes, if the structured output is not fulfilled, the turn will fail as expected.

### Validation

Adds unit coverage that active background work preserves the schema, then settled background work without structured output follows the existing `OUTPUT_SCHEMA_NOT_FULFILLED` error path.

### DIFF
--- a/packages/eve/src/harness/tool-loop.ts
+++ b/packages/eve/src/harness/tool-loop.ts
@@ -2893,6 +2893,18 @@ async function finishConversationTurn(input: {
 
   const structured = extractFinalOutput(result);
   if (structured === undefined) {
+    const initiatingTasks = resolveInitiatingTaskContext({
+      state: session.state,
+      turnId: emissionState.turnId,
+    });
+    if (initiatingTasks !== undefined) {
+      if (emit) {
+        emissionState = await emitTurnEpilogue(emit, emissionState, "conversation");
+        session = setHarnessEmissionState(session, emissionState);
+      }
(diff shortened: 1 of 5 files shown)


packages/eve/src/harness/tool-loop.ts
```diff
@@ -2893,6 +2893,18 @@ async function finishConversationTurn(input: {
 
   const structured = extractFinalOutput(result);
   if (structured === undefined) {
+    const initiatingTasks = resolveInitiatingTaskContext({
+      state: session.state,
+      turnId: emissionState.turnId,
+    });
+    if (initiatingTasks !== undefined) {
```
ruiconti, line 2900: @e0 is `!structured && !initiatingTasks` sufficient to determine that this turn has ended with empty results due its background tasks launching?

ruiconti: @e0 can you clarify with a concrete use case of when this would happen? If it's using workflow tools, it should include the source. I'm trying to understand the exact conditions in which this issue would appear. Ideally, this would have an accompanying issue with these details

ruiconti: @e0 let's add a readable simple yet precise e2e case in task fixture that repros this.

Issue #905, opened 2026-07-17
Add a configurable model-facing output budget for tools


jbreite:
### What problem are you trying to solve?

Eve can currently put an arbitrarily large tool result into model context.
Authored tools can use toModelOutput to keep the complete execution result while giving the model a smaller representation. Native MCP and OpenAPI connections do not expose the same control. If a remote connection returns too much data, the agent author must change the server, replace Eve's native connection with a custom wrapper, or patch Eve.

This comes up with search results, logs, documents, database rows, and API responses. One large result can consume much of the context window, increase token usage, and trigger compaction sooner.
The current code appears to leave this path unbounded:
- Without an authored toModelOutput, Eve passes the complete string or JSON result to the model in [packages/eve/src/harness/tools.ts](https://github.com/vercel/eve/blob/main/packages/eve/src/harness/tools.ts).
- Native MCP tools return the remote result in [packages/eve/src/runtime/connections/mcp-client.ts](https://github.com/vercel/eve/blob/main/packages/eve/src/runtime/connections/mcp-client.ts), and the generated connection tool does not expose a public projection hook.
- The keeps tool results verbatim across steps so history is append-only test verifies that a 200,000-character result remains unchanged across a continuation step in [packages/eve/src/harness/tool-loop-compaction-accounting.test.ts](https://github.com/vercel/eve/blob/main/packages/eve/src/harness/tool-loop-compaction-accounting.test.ts).

Keeping history append-only makes sense for provider prompt caching. The missing capability is an optional bound applied before a result first enters model history.

### Proposed solution

Add an opt-in agent-level budget for text and JSON tool results sent to the model.

An omitted or `false` setting would preserve Eve's current behavior. When configured, Eve would:

1. Keep the complete executor output in `action.result` and application hooks.
2. Use Eve's normal model-output serialization.
3. Apply the configured limit before adding the result to model history.
4. Send generated MCP and OpenAPI results through the same bounded path.
5. Leave results under the limit unchanged.
6. Clearly mark truncated results and include size information when available.
7. Store the bounded result as-is so later steps do not rewrite existing history.

For a smaller first version, an authored `toModelOutput` could remain authoritative and bypass the generic limit. It already gives the author explicit control over the model-facing representation. Per-tool and per-connection overrides could be considered later if the agent-level setting proves useful.

[...]

PR #3231, opened 2026-09-10
feat(eve): project connection tool results with toolCall.toModelOutput

rajp33:
### Summary

Related to #905. Connection tool results enter model history verbatim and are re-sent on every subsequent model call in the turn, so one large OpenAPI operation (~40 KB per response) pushed a turn past 800k input tokens and a later run was rejected with `context_length_exceeded`. Authored tools solve this with `toModelOutput`; connections expose only `providedArguments`. This adds `toolCall.toModelOutput` to MCP and OpenAPI connections — projections keyed by remote tool or operation name, reusing the authored-tool contract. Only the model-facing representation changes: `action.result`, channel events, and hooks still receive the complete remote result, and an operation without an entry is byte-identical to today.

```ts
// before: block the operation, hand-reimplement it as an authored tool,
// and give up the connection's generated input schema
operations: { block: ["getReport"] },

// after
toolCall: {
  toModelOutput: {
    getReport: (result: ReportResponse) => ({
      type: "json",
      value: { id: result.body.id, rowCount: result.body.rows.length },
    }),
  },
},
```

Deliberately excluded: the agent-level output budget #905 opens with — no ceiling, no truncation, no new config surface, no compaction changes. A projection handles a *predictably* large, well-structured result and returns a valid smaller object; a ceiling is the backstop for an *unexpectedly* large one and would cut mid-payload. The two are complementary, and the ceiling is worth designing separately.

Two decisions worth review. **Unknown operation names are inert** — remote names are not known at build time (MCP publishes at runtime, an OpenAPI `spec` URL is fetched on first use), so a projection keyed to a nonexistent operation is unused, as a `providedArguments` key or `tools.allow` entry matching nothing is today; build-time validation checks shape only. **`output` is unchecked** — eve does not own the remote result shape, so the parameter is `any` and authors annotate it, mirroring `DynamicToolEntry`'s `TOutput`.

Offered for direction-setting on question 4, not as a claim on the issue. Happy to fold it into a `research/` proposal instead.

### Checklist

### DIFF
--- a/.changeset/connection-tool-model-output.md
+++ b/.changeset/connection-tool-model-output.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Add `toolCall.toModelOutput` to MCP and OpenAPI connections: a per-operation projection of the model-facing tool result, keyed by remote tool or operation name. It reuses the authored-tool `toModelOutput` contract, so `action.result`, channel events, and hooks still receive the complete remote result. Operations without an entry are unchanged.
--- a/docs/connections/mcp.mdx
+++ b/docs/connections/mcp.mdx
@@ -133,9 +133,28 @@ Return `"user-approval"` (or `true`) to pause for a person and `"not-applicable"
 
 ## Control result size
 
-eve sends an MCP tool's returned content to the model as the tool result. MCP connections do not expose a per-result transform equivalent to an authored tool's [`toModelOutput`](/docs/tools#shape-what-the-model-sees-with-tomodeloutput).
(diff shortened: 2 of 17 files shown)


rajp33:
@AndrewBarba  tagging because I see you have worked on this. Added this as a quick PR. Adds toolCall.toModelOutput reusing the custom tool's ToolModelOutput contract. 

This covers the well-structured result case, but I think there are use cases where the result may not be known or the model should decide what data it would like from the response of the tool call. 

providedArguments callbacks get ProvidedArgumentContext. Can pass that same parameter as a second argument to toModelOutput to allow the user to decide to save the output to a file via `SessionContext.getSandbox().writeTextFile()` and return the path to the agent. Happy to add to this PR or open a new one (or a research PR) for this.

PR #3236, opened 2026-09-10
[eve] Add experimental.workflow.retention

VaguelySerious:
### Summary

Agents that handle sensitive payloads need their run data gone as soon as the run
finishes, rather than kept for the world's default retention period. The Workflow
SDK shipped a `experimental_retention` start option for this
(vercel/workflow#3787, released in `@workflow/core@5.0.0-beta.50`), but eve had no
way to set it. This adds `experimental.workflow.retention` to `defineAgent` and
forwards it to `start()`.

The setting is declared once on the agent and covers a session end to end: the
session run, every turn run the session dispatches, and the activity-collector
run. It is resolved outside the run from the authored config and carried as a
primitive on the workflow input, the same way `sessionTimeoutMs` already is.

This PR also bumps the `@workflow/*` pins to the current betas (`core`/`builders`
beta.50, `world` beta.34, `world-local` beta.43, `world-vercel` beta.45,
`errors` beta.21) and drops the `@workflow/world-local` pnpm patch, whose
`withWindowsRetry` fix landed upstream in beta.43.

#### Behavior and scope

The SDK's accepted values are `0` and `"default"`, so eve accepts exactly those.
There is no arbitrary duration yet. `"default"` writes nothing and stays
byte-identical to omitting the field.

`0` deletes user data when the run finishes, and that purge races reads of the
completed run. `await run.returnValue` on such a run usually throws
`RunExpiredError`, so a finished session's output is generally unreadable. The
docs say so plainly and point at writing anything you need to keep from inside a
tool. `0` also needs a world implementing spec version 4 or later; a world that
does not recognize the value keeps the data.

Retention is per-agent, so a subagent running its own session honors its own
value. That is deliberately unlike `experimental.workflow.world`, which
`assertRootOnlyConfig` restricts to the root node: retention has no
cross-node coupling, and a subagent handling sensitive data should be able to
opt in on its own.

Runs eve starts for other purposes keep the world's default and are out of
scope: session timeouts, task runs, and workflow-tool runs. They carry no model
or transcript payloads.

Publicly exporting `AgentWorkflowRetentionDefinition` changed the hash of the
`subagent` extension capability's API surface, so its contract epoch moves 11 →
12, retained as backward compatible. `extension-contracts/compatibility/subagent/v11.ts`
is a new fixture proving a subagent authored before retention existed still
compiles.

#### Alternatives considered

- **A per-`start()` option, or a per-request/per-turn field.** Rejected as the
  primary API. Retention is a property of the data an agent handles, not of one
  caller's request, and a per-call option lets one caller silently opt a
  sensitive agent out. It stays available as a follow-up if a per-request
  override is actually wanted.
- **Session run only.** Simplest change, but wrong: turn runs hold the model
  payloads and streamed output, so a session-only setting would leave the bulk of
  the data behind.
- **Rehydrating the agent bundle inside `dispatchTurnStep` to read the config
  there.** Avoids threading a field, at the cost of a bundle load on the
  dispatch path for a value already known before the run started.
- **Threading it through `TurnWorkflowInput`.** That shape is a versioned
  durable wire contract (`TURN_WORKFLOW_INPUT_VERSION`). The turn body never
  reads retention, only `dispatchTurnStep` does, so it is a dispatch-only field
  and the version stays at 2.
- **Injecting it ambiently in the private start funnel.** Fewest call-site
  edits, but it makes every `start()` in the codebase implicitly retention-aware
  and hides which runs are covered. The explicit field-per-hop plumbing is what
  the existing `.strict()` schemas and `expectOnlyKnownKeys` allowlists are
  built for.

`workflow-entry.ts` crossed the 700-line cap with the new field, so its outer
crash-cleanup helpers moved to `workflow-entry-crash.ts`. No behavior change.

### Validation

- `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm guard:invariants`, `pnpm docs:check` — all pass.
- `pnpm test:unit` — 787 files pass.
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/workflow-entry.integration.test.ts` — 17 pass, covering the crash-helper extraction.
- New unit coverage: accept/reject cases for the authored value, a manifest
  round-trip and a literal-inference case both using `retention: 0` (falsy, so
  any truthiness-based copy on the plumbing path drops it), and a drift test
  asserting eve's accepted values still match `@workflow/world`'s
  `RETENTION_ZERO`/`RETENTION_DEFAULT`.
- No e2e run yet. Left to CI.

### Checklist


packages/eve/src/shared/agent-definition.ts
```diff
@@ -268,6 +285,28 @@ export interface AgentWorkflowDefinition {
...
+   * first-party worlds (Vercel, Local, Postgres) do; a world that does not
+   * recognize the value keeps the data.
+   *
+   * At `0` the purge races reads of a finished run and usually wins, so
+   * results and transcripts of completed sessions generally become
+   * unreadable. Persist anything you need to keep from inside a tool.
+   *
+   * @default "default"
+   */
+  readonly retention?: AgentWorkflowRetentionDefinition;
```
VaguelySerious, line 307:
```suggestion
  /**
   * How long the agent's run data is kept after the run finishes.
   * Applied to both the session run and every turn run.
   *
   * - `"default"`: same as omission. The Workflow SDK World decides.
   *   On Vercel this follows your team's plan.
   * - `0`: the world deletes run payloads and stream chunks as soon as the run
   *   completes or fails. Metadata such as run IDs, status, and timestamps may
   *   persist up to the default period.
   *
   * The World you are using with the Workflow SDK might not support every option.
   * A World that does not  value keeps the data.
   *
   * Note that with retention set to `0`, data deletion can race reads of a finished
   * run, so results and transcripts of completed sessions generally become
   * unreadable. Persist anything you need to keep from inside a tool.
   *
   * @default "default"
   */
  readonly retention?: AgentWorkflowRetentionDefinition;
```

packages/eve/src/execution/workflow-runtime.ts
```diff
@@ -155,6 +155,9 @@ export function createWorkflowRuntime(config: {
         parentTraceContext: input.parentTraceContext,
       });
       const sessionTimeoutMs = effectiveAgent.limits?.sessionTimeoutMs;
+      // Retention is always the authored value: `experimental` cannot be
+      // selected by a dynamic subagent config, so there is nothing to resolve.
```
VaguelySerious, line 159:
```suggestion
```

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -87,52 +91,6 @@ type DriverLoopOutcome =
       readonly result: WorkflowEntryResult;
     };
 
-/**
```
VaguelySerious, line 90: All of the deleted code here was just moved to `workflow-entry-crash.ts`

docs/agent-config.md
```diff
@@ -283,18 +283,58 @@ disappear in any release. See [Execution model and
 durability](./concepts/execution-model-and-durability#resuming-after-a-crash)
 for the retry behavior.
 
+## Run data retention
+
+The durable runtime keeps each run's data after the run finishes: model and tool
+payloads, streamed output, and the event log eve replays from. How long it keeps
+them is the world's decision, and on Vercel that follows your team's plan. Set
+`experimental.workflow.retention` to `0` to have it deleted as soon as the run
+finishes instead:
```
VaguelySerious:
```suggestion
The runtime keeps each run's data after the run finishes: model and tool
payloads, streamed output, and the event log eve replays from. How long it keeps
them is the World's decision, and on Vercel that follows your team's plan. Set
`experimental.workflow.retention` to `0` to have it deleted as soon as the run
finishes instead:
```

docs/agent-config.md
```diff
@@ -283,18 +283,58 @@ disappear in any release. See [Execution model and
...
+
+The value applies per agent. A [subagent](./subagents) that runs its own session
+uses its own value, unlike `experimental.workflow.world`, which is root-only.
+
+The world enforces retention, so it needs a world that implements it. The
+first-party worlds do; a world that does not recognize the value keeps the data.
+
+> ⚠️ **At `0`, a finished session's output is usually gone before you can read it.** The purge races reads of a completed run and generally wins, so results and transcripts become unreadable and a client polling for a finished session's output can see it disappear. Write anything you need to keep to your own storage from inside a tool.
+
+This option is experimental and may change or disappear in any release.
```
VaguelySerious:
```suggestion
This will be applied to the session run and every turn run the session
dispatches, as well as the run that collects session activity. Runs eve
starts for other purposes keep the world's default: session timeouts,
background tasks, and [workflow tools](./tools/workflows).

The value applies per agent. A [subagent](./subagents) that runs its own session
uses its own value, unlike `experimental.workflow.world`, which is root-only.

Custom Worlds used with eve might not support this feature, in which case
it falls back to the World's default retention period.

> ⚠️ **At `0`, a finished session's output is usually gone before you can read it.** Since data is deleted immediately before it can be read back, results and transcripts become unreadable and a client polling for a finished session's output can see it disappear.
```

packages/eve/src/shared/agent-definition.ts
```diff
@@ -254,6 +254,23 @@ export interface AgentBuildDefinition {
...
 export type AgentWorkflowWorldDefinition = string;
 
+/**
+ * Accepted values for `experimental.workflow.retention`.
+ *
+ * Declared as eve's own tuple so the authored type, the manifest schema, and
+ * the authored-definition normalizer share one source of truth. A drift test
+ * keeps it aligned with what the Workflow SDK accepts.
+ */
+export const AGENT_WORKFLOW_RETENTION_VALUES = [0, "default"] as const;
```
VaguelySerious, line 264: I messed up a little in the SDK implementation here by not exporting the values directly (just the type) so we needed to redefine it here. I'll fix this in the next SDK version so we can remove this duplicate value and remove drift, but fine to ship as-is for now IMO

AndrewBarba (approved):
Reviewed locally on branch review/pr-3236. The retention plumbing is coherent and the PR checks pass; no merge-blocking issues found.

PR #3248, opened 2026-09-10
fix(eve): preserve task child activity lineage

benpankow:
### Summary

Fixes background agent progress appearing under the root turn or a previous task.

The built-in `agent` and declared local/remote subagent tools now use one task-owned progress item, with child tool activity beneath it. Task identity is persisted in the parent-owned index and forwarded on both fresh starts and reuse/steering; the sink still comes from trusted parent context.

Custom-workflow child hierarchies and root grouping across HITL continuations are unchanged. Task activity submission now waits for its bounded, best-effort request before the durable step returns.

### Before / after

Illustrative Unicode renderer output while a local `researcher` task is running. Previously, its tool activity appeared directly under the root:

**Before**

```text
• Working
├── ✓ Searching files
├── • Reading results
└── • researcher
```

**After**

```text
• Working
└── • researcher
    ├── ✓ Searching files
    └── • Reading results
```

Local and remote agents now share this task-owned layout. Continuing or steering an existing agent attaches new tool activity to the new task, leaving the previous task's completed or cancelled status unchanged.

### Validation

Adds integration coverage using real derived identities and reduced activity snapshots for local/remote starts, task completion, and reuse after completion or cancellation. Also covers remote observer binding, disabled observation, task-index persistence, and waiting for activity submission. No live Slack validation; the existing eval harness does not expose activity collector snapshots.


packages/eve/src/execution/tasks/parent/tool-execution.ts
```diff
@@ -54,6 +55,11 @@ import { cancelBackgroundAgentTask } from "#execution/tools/subagent/task-cancel
 
 const IN_PROCESS_WORKFLOW_EXECUTOR = { data: {}, kind: "workflow-task" } as const;
 
+type ActivityBackgroundTask = BackgroundTask & {
+  readonly activityWorkIdentity?: ActivityWorkIdentityV1;
+};
+type ActivityBackgroundTaskDraft = Omit<ActivityBackgroundTask, "taskRunId">;
+
```
ruiconti, line 62: I wonder if this should be derived by the called tool metadata and thus be also part of `BackgroundTask` baggage. We would want to eventually have the flexibility to defer to the model to name tasks? e.g., 
- Investigator (as a named generic `worker` subagent)
- Investigator (as a named specific `investigator` subagent)
- CI Checks (as a generic `bash` call)

@e0 wdyt?

ruiconti: LGTM

ruiconti (approved):

PR #3256, opened 2026-09-10
fix(eve): reconnect idle TUI session streams

benpankow:
### Summary

Fixes the dev TUI dropping late approvals and questions from long-running background tasks after its idle transport stream closes. Idle session following now reconnects from the retained client cursor with bounded backoff, while still stopping for terminal session boundaries.

Before (question ask is dropped):

Screenshot 2026-09-10 at 2 20 42 PM

### Validation

Adds a unit regression where one idle transport stream ends before the next delivers `selfmod__registry_add` approval. It fails against the parent commit and passes with this fix.

### DIFF
--- a/.changeset/reconnect-idle-tui-streams.md
+++ b/.changeset/reconnect-idle-tui-streams.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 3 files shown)


prha (approved):

De-ASI-INTERFACE:
Got it.


Sent from Yahoo Mail for iPhone


On Thursday, September 10, 2026, 9:01 PM, Ben Pankow ***@***.***> wrote:



Merged #3256 into main.

PR #3257, opened 2026-09-10
test(eve): cover subagent install approval timing

benpankow:
### Summary

Adds TUI smoke controls for an approval-gated subagent installed before `eve dev` starts and during an active onboarding session. Both cases already pass on `main`; this captures the installation and approval-routing behavior separately from the idle-stream timeout fix in #3256.

### Validation

Both TUI smokes pass against `main`.

🤖


benpankow:
Closing this exploratory smoke-test PR. The description notes that the approval timing behavior already passed on `main`, while the added broader suites remained failing, so this is not providing a clear mergeable regression test.

PR #3263, opened 2026-09-10
feat(eve): run sessions in one owner workflow with deployment handoff

AndrewBarba:
### Summary

Every conversational turn currently starts a child workflow and coordinates with the long-lived session through private hooks and a driver/child protocol. That is the largest fixed cost per turn (#876) and most of the orchestration code. This PR moves turn execution into the owning workflow while preserving the public session ID, stream, cursor, and `send()`/`respond()` behavior. Idle ID-addressed sessions can move to the deployment that accepted a new delivery; alias-bearing sessions conservatively remain on their current owner until Workflow supports atomic hook takeover.

Related to #876. Implements [`research/single-workflow-session-upgrades.md`](research/single-workflow-session-upgrades.md).

The old turn engine is removed from current execution; an isolated `turnWorkflow` importer lets existing sessions upgrade directly on their next legacy dispatch, preserving committed conversation state, the triggering message, session ID, and stream. Pending work is interrupted once, and the original deployment must remain available until the old driver finishes serving as the stream anchor; rollback across the execution-model boundary still requires restarting sessions.

**New boundaries inside `execution/`.** `SessionExecution` runs a turn and coordinates its waits locally. `SessionInbox` merges every claimed address into one queue with explicit consume. `admitSessionInboxPayload` is the single decode/admit path for both the active turn and the parked wait; `TurnRouting` and `nextTurnDelivery` are policy on top. `SessionInputQueue` owns admitted input and `SessionInputLedger` its in-memory task dedupe. `SessionHandoff` owns eligibility, checkpoint → release → start → activate → recover, and the stream anchor. The harness still emits its own terminal turn events inside `turnStep`, so a warm turn is exactly one durable step.

**Behavior changes worth reviewing.**

- `continuation.rekey()` → `continuation.alias()`. Every claimed address stays live; the most recent one is `continuation.token`.
- Steering applies at the next committed step boundary while a turn is still open, without cancelling in-flight model or tool work, and preserves the turn ID and usage. Steering accepted while a blocking action settles is applied before the action result continues the model turn, without redispatching the action. Once the model has produced its answer, later input is the next turn. Background task notifications are excluded from steering so cohort batching survives.
- The inbox pump no longer pauses on queue depth. Interrupts share the hook's ordered stream with ordinary messages, so a depth gate could hold a `cancel` behind a burst of unread input.
- Every current-runtime eve-owned Workflow start requires an exact deployment ID; `"latest"` and absent IDs are rejected at ingress and at start.
- Handoff eligibility is strict: only a single conversational delivery to an idle, ID-addressed session triggers it. Anything else processes on the current owner, and the deployment signal is forgotten. Alias-bearing sessions return before releasing any hook, preventing a channel delivery from creating a replacement session during the release/claim gap. A failed ID-only candidate never duplicates work; accepted payloads are replayed FIFO ahead of later arrivals.

- Successful deployment handoffs and one-time legacy imports restart the original configured session timeout (30 days by default); custom durations and disabled timeouts carry forward. Ordinary messages and failed or skipped handoffs keep the existing deadline. Successors ignore late wakes from the previous owner's timer. Checkpoint v3 carries the duration explicitly; incompatible checkpoints reject before hook claims.

**Checkpoint compatibility.** Durable task state (including activity identities and settled task views) and stored subagent handles accept and preserve additive metadata through updates. Known fields, task lifecycle rules, handle ownership fields, and command envelopes remain validated. Source eligibility and target activation share one state inspection: parse retained tasks and handles, and require pending-work registries to be absent or empty. The target runs this before claiming hooks or signaling active, so incompatible nested state returns ownership to the source and the triggering delivery runs once. Authored state remains opaque. This uses the existing validation step and checkpoint version; it adds no workflow round trip or migration framework. Broader client and wire-response parser changes remain separate follow-ups.

**Follow-up.** When Workflow provides atomic forced hook claims through `createHook(…, { force: true })`, `SessionHandoff` can remove the temporary alias guard and transfer the complete hook set without an unowned interval. `@workflow/core` beta.51 (vercel/workflow#3941, stream-writer drain before step completion) is already on main and merged here; this branch carries no SDK patch.

### Validation

- `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm guard:invariants`, `pnpm docs:check`, and `git diff --check` — pass.
- `pnpm test:unit` — pass across all 9 workspace tasks; eve: 8,677 passed, 1 skipped. Initial generated-artifact failures cleared after rebuilding; the final full run passed.
- Both affected integration files, `workflow-entry.integration.test.ts` and `legacy-session/import.integration.test.ts` — 33 passed. Coverage includes normal handoff, default/custom/disabled timeout renewal, original-stream preservation, legacy import, and target nested-state rejection before hook creation with exactly one source-side delivery. The strengthened rejection assertion also passes independently.
- Unit coverage verifies additive task/handle metadata survives updates, known lifecycle fields remain invalid in incompatible phases, settled task entries are parsed, malformed pending work blocks handoff, and payloads from a failed activation are restored FIFO before later arrivals.
- Schema skew is simulated by injecting an incompatible candidate snapshot into the real Workflow transport; this does not run two separately compiled historical eve bundles.
- Full integration previously stalled during cleanup after 809 passing assertions; a complete successful run remains pending. Scenario and e2e tiers were not run locally. The hosted `agent-tools-sandbox` `redeploy` eval covers ID-addressed cross-deployment handoff in `e2e-vercel`.

### Checklist


packages/eve/src/execution/active-turn-inbox.ts
```diff
@@ -0,0 +1,243 @@
...
+import { findRunningAgentHandle } from "#subagents/handles/query.js";
+import { runProxySubagentEventStep } from "#subagents/event-proxy-step.js";
+
+interface TurnCancelPayload {
+  readonly tasks?: boolean;
+  readonly turnId?: string;
+}
+
+/** Buffers arrivals during a step and admits them against its committed state. */
+export class ActiveTurnInbox {
```
ruiconti: @e0 what is the responsibility relationship between 
1. session inbox
2. active turn inbox and
3. command inbox
?

ruiconti: @e0 I don't fully get the distinction or the reason for existence of both `SessionCommandInbox` and `ActiveTurnInbox`. Seems like all of command inbox's responsibilities could be also handled by the (current) `ActiveTurnInbox`? or what does "session-level arrivals are continually collected" mean that `ActiveTurnInbox` shouldn't be also responsible for handling?

AndrewBarba: Closing the loop on this now that the branch has settled. Both classes you asked about are gone. What remains, and what each one owns:

- `SessionInbox` (`session-inbox/inbox.ts`) — transport only. Merges every claimed hook into one queue; `read(mode)` hands out a lease and nothing is consumed until the caller says so. No turn policy.
- `admitSessionInboxPayload` (`session-routing.ts`) — the single decode/admit path. Both the active turn and the parked wait go through it.
- `SessionInputQueue` — the ordered, admitted input. `TurnRouting` (active turn) and `nextTurnDelivery` (parked) are policy on top of it, deciding what steers, what starts a turn, and what's a control command.

So there's one inbox and one queue, and the two things that used to be "command inbox vs. active-turn inbox" are just two policies reading the same queue.

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -98,170 +102,102 @@ type DriverLoopOutcome =
...
+    input.kind === "initial" ? (serializedContext["eve.mode"] as RunMode) : input.checkpoint.mode;
   const crashCleanupState: CrashCleanupState = {
     caller: undefined,
     callerResolved: false,
     lastSessionState: undefined,
-    serializedContext: input.serializedContext,
+    serializedContext,
     terminalEmitted: false,
   };
+  const commandInbox = createSessionInbox(sessionId);
```
ruiconti: Nit

```suggestion
  const sessionInbox = createSessionInbox(sessionId);
```

packages/eve/src/execution/workflow-entry.ts
```diff
@@ -301,72 +237,243 @@ export async function workflowEntry(input: WorkflowEntryInput): Promise<Workflow
...
+): Promise<SessionLoopOutcome> {
+  const { commandInbox, crashCleanupState, handoff } = deps;
+  const cursor = new SessionStateCursor({
+    commandInbox,
+    parentWritable: boot.sessionWritable,
+    serializedContext: boot.serializedContext,
+    sessionState: boot.sessionState,
+  });
+  const queue = new SessionInputQueue();
+  const ledger = new SessionInputLedger();
```
ruiconti: @e0 /explain what `cursor`, `queue` and `ledger` are doing with respect to the session lifetime and cross-turn mechanisms

ctgowrie (approved):

PR #3274, opened 2026-09-11
fix(eve): align internal AI SDK step lifecycle hooks with the current `onStepStart` and `onStepEnd` callback names

felixarntz:
### Summary

- The harness emitted `step.started` from `prepareStep`, even though the AI SDK exposes `onStepStart` for step-start lifecycle work.
- The harness still used the deprecated `onStepFinish` callback name, which has been replaced with `onStepEnd` a while ago.

This change moves the event emission to a returned `onStepStart` hook, renames the internal completion hook to `onStepEnd`, and wires both current callback names into `ToolLoopAgent` while preserving cache/provider preparation and step-result handling.

`onStepStart` fires right after `prepareStep`, so it is a more suitable replacement that maintains the relevant semantics for the emitted event.

### Checklist


packages/eve/src/harness/step-hooks.ts
```diff
@@ -205,10 +204,22 @@ export function buildStepHooks(input: StepHooksInput): StepHooks {
...
+        messages,
+      );
+    }
+  };
+
   return {
-    onStepFinish: async (step: StepResult<ToolSet>): Promise<void> => {
+    onStepEnd: async (step: StepResult<ToolSet>): Promise<void> => {
       resolveStep(step);
     },
```
ruiconti, line 221: @e0 curious, do we not emit `step.finish` or `step.end` here? Or do we only emit when eve's step ends?

ruiconti: @e0 okay so no user-facing impact with this change?

ruiconti (approved):

PR #3294, opened 2026-09-11
fix(eve): offer to repair failed self-modifications

benpankow:
### Summary

When a self-modification later fails or behaves incorrectly, the parent now offers to send it back to self-mod for repair. For now this is conservative: it requires confirmation and only applies to failures tied to a prior self-modification. We may want to offer self-mod repair after any tool failure once we have a good way to avoid irrelevant suggestions.

### Validation

Adds a unit test for the repair offer and confirmation requirement.

🤖

### DIFF
--- a/.changeset/offer-self-mod-repairs.md
+++ b/.changeset/offer-self-mod-repairs.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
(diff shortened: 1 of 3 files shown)


benpankow:
Closing this self-modification repair exploration. Its follow-up also closed without merging, and repair behavior should be redesigned around the current self-modification and Connect flows if the requirement remains.

PR #3295, opened 2026-09-11
feat: add non-provisioning sandbox stop lifecycle

codyss:
## Summary

Add a non-provisioning `SessionContext.stopSandbox()` lifecycle API for durable session boundaries.

The API stops the current session's owned compute without calling `getSandbox()`, creating a provider resource, resuming a stopped resource, running `onSession`, or running base-runtime initialization. Persisted session metadata remains available for the next callback.

## Implementation

- Add `SessionContext.stopSandbox()` with idempotent and ownership-aware behavior.
- Add the optional backend `stopExisting` primitive and public input/result types.
- Coalesce concurrent stop requests within one access and tolerate replayed requests.
- Add the Vercel implementation using `Sandbox.get({ resume: false })` and stop only running or pending resources.
- Preserve the existing explicit live-handle `getSandbox().stop()` behavior.
- Document the lifecycle API and add unit and integration coverage for unused, persisted, concurrent, shared, missing, already-stopped, and provider-failure cases.

## Verification

- `pnpm exec oxfmt --check .`
- `pnpm --filter eve build:js`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/sandbox/ensure.test.ts src/execution/sandbox/bindings/vercel.test.ts`
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/context/accessors.integration.test.ts`

The full unit suite currently has four environment-sensitive failures unrelated to this change: three existing redaction assertions include the checkout's `/private` path, and one existing production-server test cannot bind loopback in the restricted test environment. The focused lifecycle and Vercel suites pass.

### DIFF
--- a/.changeset/stop-sandbox.md
+++ b/.changeset/stop-sandbox.md
@@ -1,0 +1,5 @@
+---
+"eve": minor
+---
+
+Add `ctx.stopSandbox()` for stopping durable sandbox compute without opening or resuming the sandbox first.
--- a/docs/guides/hooks.md
+++ b/docs/guides/hooks.md
@@ -64,26 +64,27 @@ interface HookContext extends SessionContext {
 }
 ```
 
-That means a hook can access the current sandbox and release its backing
-compute at an application-defined boundary:
+That means a hook can release the current session's backing compute at a
+durable lifecycle boundary:
 
 ```ts title="agent/hooks/stop-after-turn.ts"
 import { defineHook } from "eve/hooks";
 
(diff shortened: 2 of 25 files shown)


codyss:
Ready for review after adding the DCO sign-off.

This API is needed by Dogsled to stop Vercel Sandbox compute at durable session boundaries without provisioning or resuming a resource.

Rchie121 (approved):

PR #3303, opened 2026-09-11
fix(eve): restore legacy session history

### LINKED ISSUE #3300: eve 0.54.2 cannot resume pre-0.54 sessions with user-message history

[Focused legacy-session integration test](https://github.com/vercel/eve/blob/eacd8aad05a4e2dc1f7d216d8b1573d7dc71972c/packages/eve/src/execution/legacy-session-history.integration.test.ts)

The test constructs a durable snapshot containing the user-message shape written before eve 0.54.0, reads the snapshot through the current durable-session boundary, and hydrates it without making a model or network call.

### Steps to reproduce

1. Clone eve and check out the public repro commit:

   ```sh
   git clone https://github.com/vercel/eve.git
   cd eve
   git fetch origin pull/3293/head:legacy-session-repro
   git switch legacy-session-repro
   pnpm install --frozen-lockfile
   ```

2. Keep the regression test but restore the affected implementation to the latest release:

   ```sh
   git restore --source=eve@0.54.2 -- \
     packages/eve/src/execution/durable-session-migrations/snapshot.ts \
     packages/eve/src/harness/messages.ts \
     packages/eve/src/harness/compaction.ts \
     packages/eve/src/tracing/agent-otel-content.ts
   ```

3. Run the focused case:

   ```sh
   pnpm --filter eve exec vitest run \
     --config vitest.integration.config.ts \
     src/execution/legacy-session-history.integration.test.ts \
     -t "restores old user and framework messages without inventing provenance"
   ```

The test fails in `hydrateDurableSession` with:

```text
TypeError: Expected every user-role model message to have a kind.
 ❯ validateHarnessModelMessages src/harness/messages.ts:98:13
 ❯ hydrateDurableSession src/execution/session.ts:271:14
```

4. Restore the repro branch implementation and rerun to demonstrate the repair:

   ```sh
   git restore --source=HEAD -- \
     packages/eve/src/execution/durable-session-migrations/snapshot.ts \
     packages/eve/src/harness/messages.ts \
     packages/eve/src/harness/compaction.ts \
     packages/eve/src/tracing/agent-otel-content.ts

   pnpm --filter eve exec vitest run \
     --config vitest.integration.config.ts \
     src/execution/legacy-session-history.integration.test.ts
   ```

Current: sessions created before eve 0.54.0 may contain retained messages shaped as `{ role: "user", content: ... }`. The 0.54.2 snapshot migration chain is empty, so those messages reach `hydrateDurableSession` unchanged. Its new strict validator rejects the missing `kind` before model execution. Durable step retries read the same incompatible snapshot and fail again, making the session unrecoverable through an ordinary follow-up.

Expected: legacy snapshots resume with their content, attachments, metadata, and existing classifications intact. Since old snapshots cannot distinguish genuine user input from framework-authored user-role messages, missing classifications should become an explicit unknown-provenance kind rather than being guessed. Messages produced by current code should remain subject to strict validation.

This occurred in a deployed GitHub-connected agent after a PR session created before the upgrade was resumed under eve 0.54.2: https://github.com/vercel/eve/pull/3132#issuecomment-5640414545

### eve version

`eve@0.54.2` (latest published release). The incompatible history was written before `eve@0.54.0`; `eve@0.53.1` and earlier accepted that history shape.

### Environment

```shell
Production occurrence: deployed GitHub-connected agent
Local source repro:
OS: macOS 26.6.2 (arm64)
Node.js: v26.8.1
pnpm: 11.25.0
eve: 0.54.2
```

### Where does the bug occur?

Production (`eve start` / deployed); also reproducible directly through the focused source integration test.

### Deployment

Production failure and error ID: https://github.com/vercel/eve/pull/3132#issuecomment-5640414545

### Build and runtime logs

```shell
FatalError: Step "step//eve@0.54.2//turnStep" failed after 3 retries: Expected every user-role model message to have a kind.

Error id: 9847cc91-b437-4833-b891-b95ae1b2ab62
```

Local repro:

```shell
FAIL src/execution/legacy-session-history.integration.test.ts > pre-0.54 durable history > restores old user and framework messages without inventing provenance
TypeError: Expected every user-role model message to have a kind.
 ❯ validateHarnessModelMessages src/harness/messages.ts:98:13
 ❯ hydrateDurableSession src/execution/session.ts:271:14
```

Restore compatibility for durable session histories written before semantic user-message kinds were introduced.

Requirements:

- Normalize legacy user-role entries lacking `kind` at the persisted-snapshot read boundary, before `hydrateDurableSession` performs strict validation.
- Preserve content, attachments, metadata, unknown fields, assistant messages, and already-valid classifications.
- Represent unrecoverable old provenance explicitly, for example as `legacy.unknown`; do not guess `user` or a framework-specific kind.
- Keep strict rejection for missing or malformed classifications produced by current code.
- Cover both embedded snapshots and the legacy `eve.session` stream read path.
- Preserve compatibility with pinned workflow drivers carrying older snapshot envelopes.

Acceptance criteria:

- A snapshot created by eve 0.53.1 or earlier resumes under the current release.
- Repaired history can be persisted and read again without changing its classification.
- Existing valid message kinds remain unchanged.
- New malformed messages still fail validation.
- Compaction and tracing handle unknown legacy provenance without replaying it as the latest known human request.
- Add a cross-version integration test that fails on eve 0.54.2 with the production error.

### Additional context

PR #3155 introduced semantic model-message kinds. Its review identified this exact compatibility failure: https://github.com/vercel/eve/pull/3155#discussion_r3981798793

A follow-up commit added a v1-to-v2 migration and `legacy.unknown`, and the review confirmed the trigger was removed: https://github.com/vercel/eve/pull/3155#discussion_r3982008357. That migration was absent from the final merged commit while strict hydration validation remained.

PR #3293 independently reproduced and repaired the issue but was closed without merge: https://github.com/vercel/eve/pull/3293

A code repair will not necessarily revive workflow runs that have already exhausted their retries; those may require explicit replay or persisted-state recovery.

chadhietala:
### Summary

Sessions created before eve 0.54 can become unrecoverable on resume because retained user-role messages lack the `kind` required by current strict validation. This classifies only persisted legacy entries as `legacy.unknown`, preserving message data and existing valid kinds while keeping validation strict for newly produced history. Compaction also avoids treating unknown legacy provenance as the latest human request. Closes #3300.

### Validation

- `git diff --cached --check`
- `node scripts/guard-invariants.mjs`
- `node_modules/.bin/oxlint --fix packages/eve/src/execution/durable-session-store.test.ts packages/eve/src/execution/legacy-session-history.integration.test.ts`
- `node_modules/.bin/oxfmt --check packages/eve/src/execution/durable-session-store.test.ts packages/eve/src/execution/legacy-session-history.integration.test.ts`
- `node_modules/.bin/tsc -p packages/eve/tsconfig.json --noEmit`
- Focused tests were not rerun for the cast-only lint follow-up.

### Checklist

### DIFF
--- a/packages/eve/src/execution/durable-session-migrations/snapshot.ts
+++ b/packages/eve/src/execution/durable-session-migrations/snapshot.ts
@@ -14,8 +14,11 @@
  * 4. Append the migration to {@link snapshotMigrations}.
  * 5. Cover it in `snapshot.test.ts`.
  */
+import type { ModelMessage } from "ai";
+
 import type { DurableSessionSnapshot } from "#execution/durable-session-store.js";
 import { DURABLE_SESSION_VERSION } from "#execution/durable-session-store.js";
+import { validateHarnessModelMessages } from "#harness/messages.js";
 
 import { runMigrationChain, type VersionMigration } from "./chain.js";
 
@@ -30,10 +33,22 @@ const snapshotMigrations: readonly VersionMigration[] = [];
  * {@link DURABLE_SESSION_VERSION}. Pure; safe to call inline.
  */
 export function migrateDurableSessionSnapshot(value: unknown): DurableSessionSnapshot {
-  return runMigrationChain<DurableSessionSnapshot>({
+  const snapshot = runMigrationChain<DurableSessionSnapshot>({
     label: "durable session snapshot",
     migrations: snapshotMigrations,
     targetVersion: DURABLE_SESSION_VERSION,
     value,
   });
+
+  // Pre-0.54 history used user-role messages for both human and framework input.
+  // Keep v1 so a pinned older driver can still forward the updated snapshot.
+  const history = snapshot.session.history.map((message: ModelMessage) =>
+    message.role === "user" && !("kind" in message)
+      ? { ...message, kind: "legacy.unknown" as const }
+      : message,
+  );
+  return {
+    ...snapshot,
+    session: { ...snapshot.session, history: validateHarnessModelMessages(history) },
+  };
 }
--- a/.changeset/restore-legacy-message-history.md
+++ b/.changeset/restore-legacy-message-history.md
@@ -1,0 +1,5 @@
+---
+"eve": patch
+---
+
+Resume sessions whose history predates user-message provenance instead of failing with a missing-kind error. Existing unclassified messages retain their content and are marked `legacy.unknown`, without treating unknown framework input as a new human request after compaction.
--- a/docs/guides/instrumentation.md
+++ b/docs/guides/instrumentation.md
@@ -154,6 +154,15 @@ Every memory span includes `gen_ai.operation.name` and `gen_ai.memory.store.id`.
 
 `gen_ai.memory.records` contains recalled record content only when `recordInputs` is enabled. eve classifies recalled records as input content because they become part of the model context. eve does not set `gen_ai.memory.query.text` because a memory provider receives structured conversation messages, not a standalone search-query string. The `agent.memory.slot` and `agent.memory.phase` attributes identify the eve slot and lifecycle boundary without adding either to the span name.
 
+Messages in `gen_ai.input.messages` include a `kind` for user-role entries.
+New human input uses `user`; framework-authored input uses namespaced kinds
+such as `context.instruction` and `execution.background_task`. When resuming
+history saved before eve 0.54, entries without a kind use `legacy.unknown`:
+the old format did not reliably distinguish human input from framework input.
+Their content and metadata remain unchanged, and already classified entries
+retain their kind. Unknown legacy entries are not replayed as the latest human
+request after compaction.
+
 ## Agent trace contract
 
 The provider layout and zero-config local tracing emit the following spans.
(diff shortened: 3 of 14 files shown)


ruiconti (approved):

packages/eve/src/execution/durable-session-migrations/snapshot.ts
```diff
@@ -30,10 +33,22 @@ const snapshotMigrations: readonly VersionMigration[] = [];
...
+  // Pre-0.54 history used user-role messages for both human and framework input.
+  // Keep v1 so a pinned older driver can still forward the updated snapshot.
+  const history = snapshot.session.history.map((message: ModelMessage) =>
+    message.role === "user" && !("kind" in message)
+      ? { ...message, kind: "legacy.unknown" as const }
+      : message,
+  );
+  return {
+    ...snapshot,
+    session: { ...snapshot.session, history: validateHarnessModelMessages(history) },
```
ruiconti, line 52: @e0 is rewriting history always the right mode instead of fixing (ie considering) at directly at validation?

PR #3327, opened 2026-09-14
[1/3] test(e2e): cover self-mod tool generation and repair

benpankow:
### Summary

This PR adds local `eve eval` cases that use the standard self-modification scaffold to create and repair tools, rebuild the agent, and assert exact tool inputs and outputs behave as expected.

The harness retires parent and child sessions, restores the full source tree, and prevents concurrent processes from mutating the same checkout.

The fixture stays out of deployed world suites because it depends on local rebuild controls. Routing-only coverage remains in `agent-subagents`.

### Validation

Adds live-model cases for incident triage, destination and weight-based shipping quotes, and repairing quantity and discount calculations. Adds fixture-local scenario coverage for scaffold copying, source restoration, session retirement, cross-process locking, failed cleanup recovery, and serialization across bundled evals.

The fixture builds and typechecks locally. Its scenario tests pass, and I reproduced the previously failing e2e extension-update command after rebasing onto `main`; it now completes without resolving the newly published `eve` package. The live-model cases still require CI.


.github/scripts/discover-e2e-fixtures.mjs
```diff
@@ -14,95 +14,123 @@
...
+//                         world's suite workflow, which runs fixtures that
+//                         select that world once with mock models
+//                         (EVE_E2E_MODEL=mock).
 import { appendFileSync, existsSync, readdirSync, readFileSync, statSync } from "node:fs";
+import { fileURLToPath } from "node:url";
 import { join } from "node:path";
 
-const roots = ["e2e/fixtures", "apps/fixtures"];
-
-const registry = JSON.parse(readFileSync("e2e/matrix.json", "utf8"));
```
prha, line 24: may want to consider running this on more models (maybe just manually at first), to figure out baseline of acceptable models + reasoning.

e2e/fixtures/agent-self-modification/evals/evals.config.ts
```diff
@@ -0,0 +1,8 @@
+import { e2eJudgeModel } from "@eve-e2e/config";
+import { defineEvalConfig } from "eve/evals";
+
+export default defineEvalConfig({
+  judge: { model: e2eJudgeModel() },
+  maxConcurrency: 1,
```
prha, line 6: are we constrained by the filesystem edits to single-process?

main pause here is adding a job that currently takes 5 min to run, that will potentially be adding many more evals.

prha (approved):

e2e/fixtures/agent-self-modification/evals/evals.config.ts
benpankow: parallelized verification turns after rebuild, which reduced local runtime by about 40%, but agree we should watch runtime closely

PR #3347, opened 2026-09-14
feat(eve): add periodic activity refreshes

benpankow:
### Summary

Activity renderers currently update only when a new activity event arrives, so time-based presentation becomes stale during long-running work. This adds an opt-in `activity.periodicRefreshIntervalMs` for Slack channels, which re-renders active snapshots on a durable timer alongside the existing event-driven renders and stops refreshing once activity settles.

This is useful for two cases:
- long-running tasks, e.g. slack status which disappears after 2 minutes
- enabling renderers to e.g. display how long they've been working on a task

```ts
slackChannel({
  activity: {
    periodicRefreshIntervalMs: 5_000,
    renderers: [experimental_slackActivityRenderer(renderer)],
  },
});
```

### Validation

Adds unit coverage for periodic refreshes, settlement with an outstanding timer, interval validation, and propagation into the collector workflow. Ran `pnpm build`, `pnpm test`, `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm guard:invariants`, and `pnpm docs:check`.


benpankow:
Closing as superseded by the later activity-duration and lifecycle work associated with #2646. If stale long-running activity still reproduces on current `main`, it should return as a focused bug against the current renderer.

PR #3356, opened 2026-09-14
feat(eve): onboarding - go from init directly to chat

AndrewBarba:
### Summary

First use asks users to choose a coding agent, link a Vercel project, and review optional setup before they can chat. This replaces that interview with scaffold → TUI → reuse a connection or `/login` → chat, keeping project creation and Vercel CLI login in deployment. Shared, compact menus and quieter transitions reduce the navigation and visual interruptions throughout setup.

### Behavior

- **Init goes directly to chat.** Interactive `eve init` opens the TUI without the continuation picker, self-modification question, channels interview, or review screen. Noninteractive and coding-agent invocations remain nonblocking; default self-modification is separate work.
- **One model login surface.** `/login` offers Vercel Account, Vercel AI Gateway API Key, ChatGPT Subscription, OpenAI API Key, and Anthropic API Key, in that order. Existing project choices win; new connections prefer explicit environment credentials, the saved machine default, then the Vercel CLI's current team. Vercel OAuth uses endpoint discovery and a searchable team picker only when needed, without creating or linking a project.
- **Credentials stay local and secure.** eve-owned keys and refresh credentials use just-secrets; CLI token rotation remains CLI-owned. Runtime credential brokers keep discovered secrets out of authored files and generated artifacts. Deployment uses explicitly provisioned credentials. Projectless account-token access remains feature-gated, and the isolated OAuth client ID temporarily belongs to fx.
- **Model changes apply immediately.** `/model` selects models and settings without a final Done step. New eve-owned `openai()` and `anthropic()` helpers sit alongside `chatgpt()`. Source edits support flat and nested layouts, verify helper imports instead of inferring ownership from endpoints, and refuse custom expressions safely. Missing defaults open the provider's picker; expired ChatGPT sessions point to `/login`.
- **Setup behaves like part of chat.** Borderless, searchable menus use a compact selection marker and dim secondary text. The shared panel owns one stable heading across loading and questions, avoids intermediate idle repaints, and dismisses routine cancellations quietly. Errors, approvals, and partial-install results remain visible. `/add` selects and installs one catalog item with its required setup; `/add <address>` goes directly to installation.
- **Deployment owns Vercel CLI setup.** Remove `/vc:install` and `/vc:login`; interactive `/deploy` installs the CLI and signs in when necessary, including for already-linked projects. Remote chat and integration setup reuse existing credentials without opening CLI login. Noninteractive deployment never starts browser login or installation.

Includes updated user documentation, a research note, an eve changeset, and connection-ready/first-response timing telemetry without credentials or prompt content.

### Validation

- `pnpm fmt`, `pnpm lint`, `pnpm typecheck`, `pnpm guard:invariants`, `pnpm docs:check`, and `git diff --check` passed during implementation. The final review fixes also passed `pnpm --filter eve typecheck`.
- `pnpm --filter eve test:unit`: **8,791 passed, one skipped** after the final fixes. Earlier full-workspace unit runs passed; a later workspace run encountered a package-resolution race during a concurrent build, and the isolated rerun passed after rebuilding.
- Targeted Vitest runs with `vitest.integration.config.ts` passed for init handoff, login → first response, all five connection selections, provider request routing, environment reload, and deployment.
- Targeted runs with `vitest.scenario.config.ts` passed for compiled model resolution, `/add` → usable skill, and model/settings edits from flat, nested, and agent-directory starting paths.
- `pnpm --filter eve build:js` and `node packages/eve/test/tui-client/tui-packed-install-model.ts` passed the packed-install smoke check. Rendering regression tests cover stable heading placement at 32/80 columns, filtering, cancellation, Unicode, resizing, streaming, and scrollback.
- Live checks validated a Vercel OAuth device response, loaded the ChatGPT model catalog, and streamed a projectless Gateway response with the current CLI account/team. This verifies that account's Gateway access, not general availability. Maintainer testing also exercised the revised TUI and login flows.
- Added a deterministic first-chat fixture eval for CI; E2E suites were not run locally. **CI is not fully green:** the current [core checks](https://github.com/vercel/eve/actions/runs/34911891945), [local E2E](https://github.com/vercel/eve/actions/runs/34911891915), and [Vercel E2E](https://github.com/vercel/eve/actions/runs/34911892033) report failures in TUI, integration, scenario, prompt-cache, and fixture-tasks jobs. Those failures have not been diagnosed as part of this description update.

### Checklist


packages/eve/src/internal/model-auth/vercel-team.ts
```diff
@@ -0,0 +1,20 @@
+import { authJson, resolveVercelSession } from "./vercel.js";
+import { readVercelCliConnection } from "./vercel-cli.js";
+
+/** Resolves display metadata without changing credentials or linking a project. */
+export async function resolveModelTeamSlug(signal?: AbortSignal): Promise<string | undefined> {
```
ruiconti, line 12: @e0 isn't `vercel*.ts` modules here duplicating existing logic already present in `setup/`?

cmpadden (approved):

packages/eve/src/cli/dev/tui/prompt-command-handler.ts
```diff
@@ -81,26 +78,6 @@ export function createPromptCommandHandler(
...
-        try {
-          ({ runRemoteAuthCommand } = await import("./remote-auth-command.js"));
-        } catch (error) {
-          return { message: `/vc:login failed: ${toErrorMessage(error)}` };
-        }
-        const message = await runRemoteAuthCommand({
-          connection: context.remoteConnection,
-          flow: options.remoteAuthFlow,
-          renderer: flow,
-        });
```
ruiconti, line 98: @e0 is this losing the automatic trusted source rule creation? can be quite disruptive for new projects

packages/eve/src/internal/model-auth/vercel-team.ts
AndrewBarba: Addressed in f6e1d1af2. I consolidated Vercel API resource name/slug parsing under `internal/vercel/api-resource.ts` and now reuse it from both project setup and model-team display. The credential transports remain separate intentionally: setup invokes `vercel api` in project/scope context, while runtime model auth must work before linking with eve OAuth or read-only CLI credentials.

PR #3370, opened 2026-09-15
feat(selfmod): add bounded local trace analysis

prha:
### Summary

Self-modification currently has raw access to local trace segments but no bounded way to find the relevant session or distinguish structural signals from large payloads. This adds conversation-scoped trace search, compact timelines, and batched span-detail inspection, plus a focused analysis skill that directs the subagent to inspect only the traces and fields needed. Trace discovery is indexed for new spans while retaining bounded compatibility with older, unindexed local traces.

### Validation

- `pnpm typecheck`
- `pnpm lint` (passed with pre-existing warnings)
- `pnpm guard:invariants`
- `pnpm build`
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/self-modification/inspect-trace.test.ts src/self-modification/local-trace-analysis.test.ts` (11 tests passed)
- `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts src/tracing/local-instrumentation-runtime.scenario.test.ts` (2 tests passed after building; the initial pre-build run could not pack stale `dist` output)

### Checklist


packages/eve/src/self-modification/extension/subagents/agent/tools/inspect_trace.ts
```diff
@@ -0,0 +1,128 @@
...
+  additionalProperties: false,
+  properties: {
+    limit: { type: "integer", minimum: 1, maximum: MAX_RECORDS },
+    offset: {
+      type: "integer",
+      minimum: 0,
+      maximum: MAX_OFFSET,
+      description: "Zero-based record offset for paging through larger traces.",
+    },
+    traceId: { type: "string", pattern: "^[0-9a-f]{32}$" },
```
chadhietala, line 30: So the entire session now if bounded by `gen_ai.conversation.id`. It is stamped to the workflow runId that started the session. We emit a new `traceId` for each turn so not sure if you want that or if you want to look at the entire session.

prha: yeah, i mostly wanted to have different tool primitives available to the selfmod subagent.

it should use search_traces to view summaries of all relevant traces (bound to the conversation), and use sort/filters to prioritize by failures / latency / etc.

then, it can get more detailed info on a turn by turn (trace by trace) basis.

chadhietala (approved):

packages/eve/src/self-modification/extension/subagents/agent/tools/inspect_trace.ts
chadhietala: Sounds good.

PR #3373, opened 2026-09-15
feat(eve): add A2A channel

ctgowrie:
### Summary

Adds an ingress-only A2A 1.0 JSON-RPC channel so external agents can discover and delegate durable work to an eve agent. `a2aChannel` publishes a cacheable Agent Card, reuses eve route auth and principal-bound invocation ownership, and supports starting, reading, continuing, and canceling tasks. The card does not advertise streaming, push notifications, or extended cards because those operations are outside this first slice; outbound eve-to-A2A calls are also out of scope.

Agent Card security is derived from `none()` or `oauthResource(oidc(...))`. OAuth-backed channels can declare `requiredScopes`, which are enforced against the verified token and reused for protected-resource and Agent Card discovery; custom auth retains an explicit `security` escape hatch.

### Validation

The official `@a2a-js/sdk` discovers the well-known Agent Card, selects the JSON-RPC interface, sends `SendMessage`, and parses eve's Task response in focused coverage. The full unit suite passed 809 files and 8,862 tests; focused coverage confirms OAuth security derivation, required-scope enforcement, malformed-part errors, and abort-listener cleanup.

### Checklist

### Diff size

**Docs** — 6 files · `+170 / -6`

Documents the Agent Card, derived auth and scope-enforcement boundary, supported task flow, and explicit first-release limitations; includes navigation and release metadata.

**Implementation** — 6 files · `+948 / -4`

Owns the A2A wire model and validation inside eve, adds the public channel and package export, and extends OAuth resource metadata with reusable required-scope enforcement. The official SDK remains a test-only dependency rather than an eve runtime dependency.

**Tests** — 3 files · `+815 / -0`

Covers discovery, official-client interoperability, auth derivation and ownership, OAuth scope enforcement, task projection and continuation, cancellation, protocol errors, malformed parts, body limits, HTTP security, and unsupported capabilities.


ctgowrie:
Addressed the auth review and both inline findings in `84b153ac1`:

- The OAuth example now enforces exact `a2a.invoke` membership after OIDC verification; regression coverage rejects missing and unrelated scopes before task creation.
- Malformed `text`, `url`, `raw`, and `data` parts now return JSON-RPC `-32602` instead of an internal error.
- Blocking poll delays remove abort listeners on both timeout and abort paths.

Focused A2A tests, type checking, docs validation, invariant guards, and lint pass.

ctgowrie:
Simplified A2A security authoring in `7fd3b63d5`:

```ts
a2aChannel({
  auth: oauthResource(oidc({ issuer, audiences: [resource] }), {
    issuer,
    resource,
    requiredScopes: ["a2a.invoke"],
  }),
});
```

`oauthResource()` now enforces `requiredScopes` against the verified `scope` claim and advertises them by default. `a2aChannel()` derives the OIDC Agent Card scheme and requirements from the same auth metadata. `none()` similarly derives public card security; explicit `security` remains only for custom auth.

packages/eve/src/public/channels/a2a.ts
```diff
@@ -0,0 +1,523 @@
...
+    {
+      readonly protocolBinding: "JSONRPC";
+      readonly protocolVersion: "1.0";
+      readonly url: string;
+    },
+  ];
+  readonly version: string;
+}
+
+export interface A2AChannelInput {
```
chadhietala, line 97: Can we make sure we have an `audience` method / field on this. When passed as a method make sure it gets called after auth and is fed enough information that  you could make a determination to return `private`, `public` or `unknown`. Set to `unknown` by defualt.

PR #3413, opened 2026-09-16
refactor(eve): unify background workflow execution

ruiconti:
### Summary

Background tasks and waiting workflow tools had separate execution and storage implementations, even when both ran the same kind of workflow body. This PR makes a background task a workflow invocation that belongs to the session and can outlive the turn that started it. It shares execution and storage with waiting workflow calls, removes the ordinary/dynamic background-tool paths, and combines overlapping tasks' final results into one report. Related to #1084.

### Before and after

A **waiting call** keeps its tool call pending until the workflow returns. A **background task** returns a task receipt first and delivers its result later.

| Concern | Before | After |
| --- | --- | --- |
| Authoring background work | `defineTool`, dynamic tools, or `defineWorkflowTool`; task-specific `TaskExec` / `postMessage` API | Authored background work uses `defineWorkflowTool({ execution: "background" })`; `return` supplies the result; background yields are consumed without publishing progress |
| Running workflow bodies | Waiting calls enter `workflowToolRunWorkflow`; background calls enter `taskRunWorkflow`. Each has its own body execution and outcome handling | Both enter `workflowToolRunWorkflow` and use one invocation reader for execution, progress drainage, and cancellation cleanup |
| Recording ownership | Waiting calls live in `eve.runtime.workflowToolRuns`; background tasks live in `eve.tasks`, with separate write paths | Both live in `eve.workflowTool` as `{ version: 3, runs }`; lifetime distinguishes turn-owned calls from session-owned tasks |
| Cancelling work | Waiting calls signal cancellation; tasks use a separate one-second forced-stop deadline | Both use shared escalation: 30 seconds for body cleanup and a 35-second caller deadline for settlement; every claimed child is signalled |
| Task-specific data | A separate task entry stores metadata, creator authentication and subagent selections, cohort membership, and a cached final view | Identity and metadata live once in the `task` payload; `task.outcome` retains only terminal status, output, and usage. Readers reconstruct the public `TaskView` |
| Reading task outcomes | The child writes an `eve.task` snapshot stream; the parent reads it and polls for cancellation commitment | The child sends outcomes through the existing inbox; the parent records the first terminal outcome and retains it for the session lifetime |
| Delivering final results | Successful tasks wait for their group; failures and cancellations can be delivered separately | Success, failure, and cancellation all wait for the same group and arrive in one automatic report |
| Subagent completion | Background admission emitted `subagent.completed` with a working receipt, while blocking completion carried a result | Both use `subagent.called` for dispatch and `subagent.completed` for a successful parent-recorded result. Working receipts remain `action.result` tool outputs; completion carries the actual output and original `callId` |

### What remains specific to background work

Sharing the runner does not make the two lifetimes identical:

- **Waiting calls:** the recorded originating turn owns the invocation. Its record is removed when the call settles or that turn is cancelled; pausing for authorization retains it.
- **Background tasks:** the session owns the invocation. Ending the initiating turn leaves it running. Completed task payloads remain for the session lifetime, including after reporting.

The parent must persist ownership before sending `ready`, which allows the background body to start. Admission happens once: repeated `ready` or rejection commands cannot mark running work settled. A repeated admission message cannot bypass invocation cleanup.

Task handles, live status, human input, child-agent ownership, and accounting remain. The removed machinery is the separate task workflow entry, executor bindings, ordinary background executors, and conversions between workflow messages and a second task-executor protocol. The child keeps execution-local admission, abort, and answer-routing state. The parent owns retained outcomes, pending input routes, and settlement accounting. The `eve.task` snapshot stream, its reader timeout, and cancellation-commit polling are removed.

Ownership reads validate invocation identity and routing without decoding unrelated saved results. Retained output is validated when consumed; checkpoint handoff validates every retained result. A malformed historical result cannot prevent cancellation of a different live invocation.

Cancellation reads the parent’s recorded outcome. An existing success or failure is returned unchanged; otherwise the control step signals cancellation, cleans up owned work, and returns the cancelled outcome for the parent to record. The existing bounded runtime-status polling remains for cleanup and forced stop. Startup and reset waits also remain.

The parent’s first recorded terminal outcome wins. Late success cannot replace cancellation, and terminal notifications deduplicate by task so competing outcomes cannot produce a second report. Reporting uses the parent’s recorded outcome. This separates task settlement from wrapper-workflow exit without adding a new wake-up mechanism.

The unused coordination acknowledgement list is removed; the background admission acknowledgement remains.

Workflow runs now always return a tool result. Previously, `resultKind` selected a second agent-result settlement path and split usage accounting between request handling and final workflow completion. Now `agent-settled` releases the exact invocation claim and records usage; the child waits for acknowledgement on its existing reply hook before returning. Duplicate settlements do nothing, including an old result arriving after the agent has been claimed for a newer call. Agent preparation still uses its node identity for receipts and reservations, without copying an agent discriminator onto workflow runs.

Parent notification variants keep their existing typed payloads and delivery IDs and share one durable delivery step. The unused `eve.task.progress` stream and writer are removed. Authored background yields are consumed without publishing progress or retaining a task-progress stream; subagent update notifications retain their existing delivery behavior.

### Reporting example

Start tasks A and B while both overlap. A fails while B is still running.

- **Before:** A's failure could be delivered to the parent while B continued, followed by a separate report for B.
- **After:** A's failure notification waits in the parent’s inbox. Once B also notifies completion, the parent records and reports both outcomes together.

The existing group of overlapping tasks is called a **cohort** and can span multiple user turns. Questions, authorization requests, and user messages remain serviceable while a cohort is open. The tradeoff is that a slow sibling delays automatic failure reporting.

### Breaking changes and rollout

This includes a minor changeset. Ordinary and dynamic background tools must migrate to `defineWorkflowTool`; `TaskExec`, `TaskMessage`, and `postMessage` are removed. Channel task views no longer expose executor bindings. Subagent event names and stream version 25 remain. The timing of `subagent.completed` changes for background calls: it now follows the parent's recorded success, not admission. Receipt consumers must read the existing `action.result` tool output. Historical receipt-marked completion events remain readable as admission. Background completion events do not wait for the cohort report; task notifications retain the existing combined-delivery policy. Failed or cancelled outcomes do not emit successful completion.

The shared run registry uses `eve.workflowTool` version 3, replacing main's version-2 task-only index and separate blocking-run store. Session checkpoints use version 6 because run ownership and result settlement have changed. The new runtime does **not** migrate the old task registries or in-flight task workflows. Checkpoint version 6 prevents handoffs between incompatible readers: the target rejects the checkpoint before taking session ownership, and the original owner resumes with the triggering message preserved. Existing deployments must remain available for their sessions and runs, or be drained. Conversation import retains history after stopping discoverable old work.

A real two-deployment upgrade/rollback and the remaining combined-path E2E cases are unproven; see the [design document](https://github.com/vercel/eve/blob/ruiconti/background-tasks-redesign/research/background-tasks-redesign.md#remaining-proof-gaps-and-migration-risks).

### Validation

- The full unit suite passed (9,038 tests, one skipped), followed by 30 focused unit tests for the outcome storage change. All 38 selected integration tests passed, covering session entry, task settlement, and cancellation; the unit suite covers registry round-trips, old-format rejection, cleanup, and handoff compatibility. Regressions cover blocking and background completion boundaries, a completed task emitting before its cohort report, duplicate and late outcomes, receipt-before-dispatch ordering in the TUI, and historical receipt handling.
- Package typecheck, production build, lint, formatting, invariant guards, and all 92 published-doc checks passed. All seven affected E2E fixtures and their shared config package typecheck.
- E2E receipt readers now inspect `action.result`. Local and remote completion evals require one actual `subagent.completed`; final-result assertions use `completed` for both execution modes. Subagent assertions reuse `TaskStatus`: admission-only observations are `working`, rejected dispatch is `failed`, and explicit cancellation is `cancelled`. Status is still inferred from captured events. E2E runs only in CI.

An infrastructure failure before the child sends an outcome still requires reconciliation, and this change does not repair the local workflow runtime’s cancel/completion race. Earlier progress-after-answer and compatible-legacy-handoff tests had intermittent failures before passing reruns; their causes remain unresolved.

### Checklist


AndrewBarba (approved):

AndrewBarba:
This lands the child-side consolidation that `executor-neutral-core.md` stalled on, and it's the right shape: one durable entry, one registry, one body runner, `defineWorkflowTool` as the only background surface. I've baselined a follow-on research doc on it (#3486) rather than in parallel — it treats this PR as a prerequisite and picks up the parent-side half (steps still polling for child state; identity decided by hook-claim race) that this PR intentionally doesn't touch.

One design point I'd like to resolve before this merges, because it's the piece the follow-on would otherwise have to undo:

**`settleWorkflowToolRunCancellation` (`tools/workflow/cancel.ts:48`) polls `getRun().status` every 250 ms for up to `WORKFLOW_CANCELLATION_SETTLE_MS` = 35 s, from the parent's step.** It replaces the task-specific 1 s force-stop, and unifying the escalation policy across both lifetimes is correct. But the *location* means `cancelOwnedTask` / `cancelDescendantTurnsStep` can now hold a step invocation open for 35 s watching the child, where before it was ~1 s. On Vercel that's a live function the whole time.

The alternative that fits `runWorkflowToolInvocation` already having the `cleanupDeadline` race: let the child own its own force-stop. The invocation loop already knows when `signal.aborted && body !== undefined` and already sleeps `WORKFLOW_CANCELLATION_CLEANUP_MS`; if the body hasn't settled by then, the child transitions to `cancelled`, publishes its terminal outcome, and exits — rather than the parent noticing from outside and calling `cancelRun`. The parent then only sends `cancel` and returns; the committed `cancelled` view arrives through the existing terminal delivery and cohort barrier, which this PR already makes the single path for terminal reporting.

That removes the parent-side poll entirely, keeps the 30 s cleanup grace you've defined, and means `task_cancel`'s tool result becomes "requested" rather than blocking on commit — which I've confirmed with Andrew is the desired semantics.

Two of the listed proof gaps ("initiating-turn cancellation racing after task admission" and "parent finalization cancelling a blocked invocation") are also races that go away once the child is the only party that decides when it's terminal, so I'd hold off writing bespoke coverage for them if you take this direction.

Happy to pair on it or take it as a stacked commit if you'd rather keep this PR's scope where it is.

PR #3417, opened 2026-09-16
feat(eve): sessions - prewarm and stream continuously

### LINKED ISSUE #1476: New-session bootstrap costs ~4s inside the workflow run before session.started (queue dispatch is only ~0.2s)

AndrewBarba:
### Summary

New chats currently pay durable workflow startup costs after the user submits their first message. This adds message-free creation and opt-in frontend prewarming so the workflow can establish its inbox earlier, while initialization still uses the first message's identity and context. Frontend bindings keep consuming the session stream across turns. Eval APIs now use explicit session ownership: every `t.session()` or `t.send()` creates a new conversation, and follow-ups use its returned session handle. This is a breaking eval API change. Closes #1476.

### API usage

Start the workflow when the chat page opens, then send when the user submits:

```ts
import { Client } from "eve/client";

const client = new Client({ host: "https://agent.example.com" });
const { session } = await client.sessions.create();

// Later, when the user submits:
const response = await session.send("Summarize account A.");
await response.result();
```

`create()` returns `{ session }` on `202 Accepted`, without starting a turn. The existing `create({ message: "..." })` returns `{ session, response }` and starts the first turn immediately. Over HTTP, send `{}` to `POST /eve/v1/session`, then `{ message: "..." }` to `POST /eve/v1/session/:sessionId`.

```text
prewarm → start workflow + establish inbox → wait
message → onMessage → initialize session → turn_0
```

The normal turn step runs session hooks, dynamic definitions, and sandbox setup using the first message's auth and context. It also sets the initial title from `onMessage.title` or the message. Later messages do not reinitialize or rename the session. There is no separate initialization step and no initial `session.started` or `session.waiting` event before the message. Prewarming moves workflow startup ahead of submission; session initialization still happens on first send.

Acceptance does not guarantee that the inbox exists yet. A send to a pending or running workflow whose inbox is unavailable returns `409 session_not_ready`; the TypeScript client retries for up to 20 seconds with exponential delays starting at 250 ms and capped at 2 seconds, resolving fresh headers for each attempt. Unknown and terminal sessions return `session_not_active` without retries or replacement. Sending never depends on receiving a readiness event.

### Frontend usage

Prewarming defaults to **false**. Opt in when the application can create its session on mount:

```tsx
import { useEveAgent } from "eve/react";

export function Chat() {
  const agent = useEveAgent({ prewarm: true });

  return (
    <>
      <button onClick={agent.reset}>New chat</button>
      <form onSubmit={(event) => {
        event.preventDefault();
        const form = new FormData(event.currentTarget);
        void agent.send(String(form.get("message") ?? ""));
      }}>
        <input name="message" required />
        <button disabled={agent.status !== "ready" && agent.status !== "error"}>Send</button>
      </form>
    </>
  );
}
```

The same hook instance handles successive chats. With `prewarm: true`, mount and reset create an owned session and open its stream. The composer remains `ready`; an early send waits for the create response, then posts using the bounded startup retries. Explicit `prewarm()` also resolves on acceptance, and concurrent calls share that request. If creation fails before acceptance, a waiting send can fall back to create-with-message.

One continuously consumed stream updates replies, authorization state, and turns started elsewhere. Completing a turn does not close the connection. Resume uses its initial durable tail index for catch-up; disconnects and the server's renewable lease reconnect from the cursor. Reset and unmount abort local transport and pending requests without cancelling durable execution. Aborting a send also interrupts preparation or a wait for prewarming; it does not cancel session creation shared by other callers. Reconnects use the latest headers and policy, including explicit retry limits, and historical readers cannot rewind the shared cursor. An explicitly supplied session is never replaced automatically. React avoids duplicate creation during Strict Mode effect replay; Vue and Svelte share the same store and opt-in default.

Message-free creation is conversation-only. Route auth, audience, and the session timeout are determined at creation; unused prewarmed sessions can expire. Applications that bind chat records or prepare required headers only on submit should leave prewarming disabled. No proxy-specific activation protocol is added.

### Eval usage

`t.session()` creates an accepted session without starting a turn; its `sessionId` and `state` are always available. `t.send(message)` creates a fresh session and its first turn in one request. Completed and live turns both expose their owning `.session`.

```ts
import { defineEval } from "eve/evals";

export default defineEval({
  async test(t) {
    const session = await t.session();
    await session.send("Alice opened this chat. Greet her briefly.");

    // A separate conversation, created with its first message in one request.
    const first = await t.send("My favorite word is marigold.");
    const next = await first.session.send("What is my favorite word?");
    next.messageIncludes("marigold");
    t.succeeded();
  },
});
```

**Migration:** replace `t.newSession()` / `t.prewarm()` with `await t.session()`. Repeated `t.send()` calls now create separate conversations; continue with `session.send()`. Identity, cursor, transcript, pending input, cancellation, and streaming operations belong to the session (`session.start()`, `session.respond()`, etc.). Read replies from `turn.message`; implicit conversation state such as `t.sessionId`, `t.reply`, and `t.transcript` is removed. Run-level assertions and judges remain on `t`; default judge output follows the most recently settled turn.

All accepted sessions participate in reporting and timeout cleanup, including sessions with no first message. Creation failures reject before exposing a handle. The prewarm fixture uses `t.session()` and `watchTurn()` without direct endpoint fetches or custom retry/stream helpers. The multi-turn fixture also verifies that another `t.send()` creates a distinct conversation while follow-ups retain their session.

### Validation

- `pnpm fmt --check`, `pnpm lint`, `pnpm guard:invariants`, and `pnpm docs:check` pass.
- `pnpm --filter eve typecheck` passes.
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts` — 8,817 passed, 1 skipped.
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/session/entry.integration.test.ts` — 27 passed.
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/client/client.integration.test.ts src/client/stream-follow.integration.test.ts src/harness/tool-loop-steering.integration.test.ts` — 17 passed.
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/turn-cancellation.integration.test.ts` — 10 passed after targeting the late cancellation at its completed turn.
- `git diff --check` passes.

Regression coverage includes aborting preparation and prewarm waits, resetting during steering, shared stream reuse, updated reconnect policies, explicit retry budgets, monotonic cursors, and preparing a delayed steering message exactly once. Integration coverage verifies that prewarming emits no initial events or lifecycle initialization, then initializes once with the first message's identity and title. E2E runs only in CI; the latest matrix is running after merging `main`.

### Pending Workflow release

The existing `cancellation/steer-pending-generation` eval exposed a Workflow Postgres queue deadlock: the client received `step.started` promptly and steering was accepted promptly, but the workflow wake waited behind the original 30-second inline generation. [Workflow #4219](https://github.com/vercel/workflow/pull/4219) removes that process-local per-run queue lock while preserving the existing bounded, per-instance message deduplication and the opt-in invocation executor.

This branch contains no dependency patch and remains on the published `@workflow/world-postgres@5.0.0-beta.42`. Upgrade the catalog and lockfile after Workflow publishes a version containing #4219; the Postgres steering eval may fail until then. Its assertions remain unchanged.

The fix was verified before removing the temporary patch: the [Postgres cancellation CI job](https://github.com/vercel/eve/actions/runs/35238629909/job/105261058851) passed all seven evals, including all nine steering assertions, and the corrected response completed within about half a second of `step.started`. Upstream queue, storage, and invocation regressions also pass against real PostgreSQL/Graphile. This is validation of the upstream fix, not a claim that the currently pinned adapter contains it.

The post-merge CI run also encountered an intermittent Anthropic `content-filter` response during the prompt-cache eval's fifth parallel tool call. That eval and its cache assertions are retained.

### Checklist


packages/eve/src/client/sessions.ts
```diff
@@ -17,11 +22,17 @@ export class ClientSessions {
     this.#context = context;
   }
 
-  /** Creates a session immediately and returns its fixed handle plus first-turn response. */
+  /** Creates a conversation session before its first turn. */
+  async create(options?: CreateSessionOptions): Promise<CreatedIdleClientSession>;
```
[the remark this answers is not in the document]
AndrewBarba: Verified fixed: the message-bearing create overload is first. Client.sessions tests pass a typed SendTurnInput<{ answer: string }> and assert CreatedClientSession<{ answer: string }>, including response. The merged branch passes TypeScript checking.

packages/eve/src/execution/session/initialize-step.ts
```diff
@@ -0,0 +1,120 @@
...
+    effectiveAgent,
+    instrumentation,
+    isFirstTurn: true,
+    sessionId: session.sessionId,
+    sessionWritable: input.sessionWritable,
+  });
+
+  try {
+    const scoped = await withContextScope(ctx, session, async (enrichedSession) => {
+      ctx.setVirtualContext(HandleEventKey, sink.handleEvent);
```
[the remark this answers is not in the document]
AndrewBarba: Addressed by removing the separate initializer. Prewarming now parks before any lifecycle events, and the first message uses the normal turn path with instrumentation.createHandleEvent. The merged branch also removes the leftover session-preamble helper. Session initialization and event ordering are covered by the session integration suite.

packages/eve/src/client/eve-agent-store.ts
```diff
@@ -179,6 +119,46 @@ export class EveAgentStore<TData> {
     };
   }
 
+  /** Creates this store's owned session without starting a turn. */
+  prewarm(): Promise<void> {
+    if (this.#session !== undefined || this.#externalSession) return Promise.resolve();
+    if (this.#prewarmPromise !== undefined) return this.#prewarmPromise;
+    if (this.#activeTurn !== undefined) {
+      return this.#activeTurn.response.then(() => {});
```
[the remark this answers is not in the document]
AndrewBarba: Verified fixed: prewarm rejects when the joined active create resolves without a response/session, preserving the creation error. The regression test “rejects prewarm when the first send it joined fails to create a session” passes on the merged branch.

docs/channels/eve.mdx
```diff
@@ -38,7 +38,21 @@ The session routes use only durable session IDs. Create a session explicitly, th
...
+
+```bash
+curl -X POST https://<deployment>/eve/v1/session \
+  -H "Content-Type: application/json" \
+  -d '{}'
+# {"ok":true,"sessionId":"wrun_A","status":"accepted"}
+```
+
+The workflow runs session-scoped initialization and parks without starting `turn_0`. Because the
+request has no message, it does not invoke `onMessage`; that hook runs when the first message is
```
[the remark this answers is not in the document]
AndrewBarba: Addressed at the runtime boundary as well as in the docs: prewarming establishes the workflow inbox, then waits before session initialization. The first message runs onMessage before session hooks, dynamic definitions, and sandbox setup, and supplies the initial auth, context, and title. Integration coverage verifies the first-message identity/title and that subsequent turns do not reinitialize or rename the session.

docs/guides/frontend/overview.mdx
```diff
@@ -77,6 +77,7 @@ export function Chat() {
 | `error`   | The last `Error` thrown, if any.                                                          |
 | `events`  | Raw eve stream events for this session.                                                   |
 | `session` | Serializable fixed session cursor (`sessionId`, `streamIndex`).                           |
+| `prewarm` | Create and initialize the durable session without starting a turn.                        |
```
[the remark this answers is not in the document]
AndrewBarba: The frontend guide now states that prewarm resolves on 202 Accepted and is not an initialization/readiness barrier. Sending waits only for an in-flight create response and uses bounded session_not_ready retries; it does not wait for stream events.

packages/eve/src/client/eve-agent-store-state.ts
```diff
@@ -1,5 +1,66 @@
...
+export interface EveAgentStoreSnapshot<TData> {
+  readonly data: TData;
+  readonly error: Error | undefined;
+  readonly events: readonly MessageStreamEvent[];
+  readonly session: ClientSessionState | undefined;
+  readonly status: EveAgentStoreStatus;
+}
+
+/**
+ * Hooks invoked while the store processes a turn. The observation callbacks
```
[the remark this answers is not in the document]
AndrewBarba: Completed in 44f2835b9. The callback reference now explicitly covers session creation, cursor advancement, and reset for onSessionChange, and creation/stream/resume/turn failures for error and onError. The store JSDoc already includes session creation. Docs checks pass.

docs/channels/eve.mdx
```diff
@@ -38,7 +38,26 @@ The session routes use only durable session IDs. Create a session explicitly, th
 
 ### Start and continue a session
 
-Start a session with an initial message, then use the returned `sessionId` for every follow-up and control operation:
+Create a conversation session before its first message by sending an empty object:
+
+```bash
+curl -X POST https://<deployment>/eve/v1/session \
+  -H "Content-Type: application/json" \
+  -d '{}'
```
ruiconti: Could we do empty request body vs empty JSON object?

```suggestion
curl -X POST https://<deployment>/eve/v1/session
```

What do you think of

- no request body: create a session with defaults, without a first message
- `{}`: still accepted (but not advertised in docs etc)
- `{"message":"..."}`: creates the session and start first turn.
- else: 400.

AndrewBarba: Done in 1951f8fdc. `POST /eve/v1/session` with no body is now the canonical message-free create request, and the TypeScript client sends it that way. `{}` remains accepted. Malformed JSON and non-object JSON still return 400, while a body with `message` creates the session and starts the first turn. The docs now show the bodyless curl form, with route and client regression coverage.

docs/channels/eve.mdx
```diff
@@ -38,7 +38,26 @@ The session routes use only durable session IDs. Create a session explicitly, th
...
+# {"ok":true,"sessionId":"wrun_A","status":"accepted"}
+```
+
+The workflow establishes its inbox and parks before session initialization or `turn_0`.
+No `session.started` or `session.waiting` event is emitted until a message arrives. The first
```
ruiconti: Alternatively, we could also be emitting a `session.created` so users could also move some initialization bits upfront e.g ~booting sandbox,~ pulling user prefs and storing in session

```ts
export default defineHook({
  events: {
    async "session.created"(_event, ctx) {
      const user = ctx.session.auth.current;
      const saved = await db.user_prefs.byId(user.principalId)
       preferences.update(() => ({ model: saved.model }));
      }
    },
  },
});
```

AndrewBarba: sandbox exposes an onSession today. I need to check if that runs here or not

AndrewBarba: I dont mind adding the event though, makes sense

docs/concepts/sessions-runs-and-streaming.md
```diff
@@ -33,16 +33,32 @@ React, Vue, and Svelte apps reach for [`useEveAgent()`](../guides/frontend/overv
 
 ## Start a session
 
+Create and park a conversation session before its first turn by omitting `message`:
+
+```bash
+curl -X POST http://127.0.0.1:2000/eve/v1/session \
+  -H 'content-type: application/json' \
+  -d '{}'
+```
```
ruiconti, line 40: Same thing here. Also these pages are somewhat redundant. Should we link the session concepts page (that is expected to have more session detail) in `channels/eve`?

docs/evals/cases.mdx
```diff
@@ -105,21 +105,56 @@ export default defineEval({
...
+Use `t.session()` to create a session before its first message:
+
+```ts title="evals/prewarm.eval.ts"
+import { defineEval } from "eve/evals";
+import { equals } from "eve/evals/expect";
+
+export default defineEval({
+  async test(t) {
+    const session = await t.session();
+    const turn = await session.send("Greet Alice briefly.");
```
ruiconti, line 136: Should this have similar semantics as the HTTP API?


```suggestion
    const { session, turn } = await t.session("Greet Alice briefly.");
```

docs/channels/eve.mdx
AndrewBarba: Checked this against the runtime boundary. `sandbox.onSession` does not run during message-free creation: the workflow establishes the inbox and parks before `createSessionStep`, session context, sandbox setup, and `session.started`. A `session.created` hook with `ctx.session.auth` and mutable preferences would therefore require a second initialization phase before `onMessage`, bringing back the split identity/initialization boundary this PR removes. I would keep prewarm limited to workflow + inbox establishment here; if we want a creation-time hook, it should be designed separately with an intentionally limited context. I clarified the canonical lifecycle boundary in c3cab547f.

docs/concepts/sessions-runs-and-streaming.md
AndrewBarba: Done in c3cab547f. The concepts page now uses the bodyless request too. The eve channel’s start section links directly to the canonical “Start a session” section for initialization, first-turn identity, readiness, and retry behavior, and I removed the duplicated lifecycle/readiness paragraphs from the channel page.

docs/evals/cases.mdx
ruiconti: Or maybe the ideal data structure resembles

```suggestion
    const session = await t.session();
    const turn = session.turns[session.turns.length-1]
```

? Seems like an unrelated and bigger change though so feel free to dismiss

AndrewBarba: The underlying TypeScript client does mirror the HTTP API: `client.sessions.create()` creates an idle session, while `client.sessions.create({ message })` returns the session plus the first-turn response. For evals we deliberately keep the intent split into two verbs: `t.session()` creates an idle session, and `t.send(message)` creates a session and returns its settled first turn with `.session` attached. Adding a message overload to `t.session()` would make it overlap with `t.send()` and introduce a second way to express the common eval path. I would keep this example as the explicit prewarm flow.

docs/guides/frontend/overview.mdx
```diff
@@ -85,6 +86,49 @@ export function Chat() {
 
 Most chat UIs only need `data.messages` and `status`. Drop down to `events` when you need the authoritative wire events directly, for example to persist an audit log or build a custom projection.
 
+Pass `prewarm: true` to start a workflow when the chat mounts and after each reset:
+
+```tsx
+import { useEveAgent } from "eve/react";
+
+export function Chat() {
+  const agent = useEveAgent({ prewarm: true });
```
ruiconti, line 95: Nice 👍🏻

ruiconti:
Found a stale credential scenario, opened up https://github.com/vercel/eve/pull/3449 describing it with a failing eval.

AndrewBarba:
Confirmed and fixed in 1c6b11169. The continuous frontend stream now resolves its per-send headers from the store’s latest turn when a transport reconnect opens, so a second send with refreshed credentials replaces the stale credentials without reconnecting between turns. I incorporated the E2E repro from #3449, added focused unit coverage, and documented the behavior. The targeted client tests, eve typecheck, fixture typecheck, lint, and docs checks pass; the E2E will run in CI.

packages/eve/src/client/session.ts
```diff
@@ -310,7 +355,7 @@ async function postSessionSend(
     try {
       return await postTurn(context, path, input, false);
     } catch (error) {
-      if (!isSessionNotActive(error) || retry >= SESSION_SEND_RETRY_COUNT) throw error;
+      if (!isSessionNotReady(error) || retry >= SESSION_SEND_RETRY_COUNT) throw error;
```
ruiconti: [P2] Allow healthy startup to outlast the current retry window. These four attempts have only 1.75 seconds of total backoff (250 ms + 500 ms + 1 s), plus request time. If a user submits immediately after prewarming and the inbox takes three seconds to appear, fast `session_not_ready` responses exhaust the retries before the message can be accepted. I reproduced that with a fake-clock client test. This follows the documented retry policy, but the policy does not cover the multi-second startup reported in #1476. Could we use a cancellable startup deadline justified by those timings and add coverage for an inbox becoming ready after three seconds? Retries should keep targeting the same accepted session.

packages/eve/src/client/session-event-stream.ts
```diff
@@ -0,0 +1,133 @@
...
+  subscribe(signal?: AbortSignal): SessionEventReader {
+    const reader = new SessionEventReader(() => this.#readers.delete(reader), signal);
+    if (this.#ended) reader.end(this.#error);
+    else this.#readers.add(reader);
+    return reader;
+  }
+
+  get ended(): boolean {
+    return this.#ended;
+  }
```
ruiconti, line 62: Nice

ruiconti (approved):

packages/eve/src/client/session.ts
AndrewBarba: Fixed in 3bdb14fb0. session.send() now retries only session_not_ready against the same accepted session for a cancellable 20-second readiness deadline, with exponential backoff capped at two seconds and fresh headers on every attempt. The client test now covers an inbox becoming ready after 3.75 seconds, verifies every retry targets the same session path, and keeps cancellation and terminal-error coverage. The documented client behavior now matches the deadline.

PR #3433, opened 2026-09-16
fix(eve): preserve idle subagents across session handoff

benpankow:
### Summary

A completed, resumable self-modification subagent leaves an idle handle in its parent session, but deployment handoff previously required the handle store to be empty. That pinned otherwise idle parents to old code and prevented newly rebuilt tools from appearing on the next turn. This adds regression coverage and allows settled `parked` conversation handles and `available` task-owned handles to cross with the parent snapshot, while active turns and leases still block handoff.

### Compatibility

The target deployment validates the complete handle store before accepting the checkpoint. Retained handles contain a stable child-session address, so the successor parent can continue addressing the child, but the child itself remains on its owning deployment. Removing or renaming the corresponding subagent makes the handle unreachable; the existing continuation path reports the permanent error and removes the handle.

`SESSION_CHECKPOINT_VERSION` remains at v4 because the serialized shape did not change. Older v4 successors safely reject checkpoints containing retained handles under their stricter eligibility rule, allowing the source owner to recover.

### Validation

- `pnpm --filter eve test:unit` (807 files, 8,725 passed, 1 skipped)
- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/session/state-compatibility.test.ts src/execution/session/handoff.test.ts` (27 passed)
- `pnpm lint`
- `pnpm guard:invariants`
- `pnpm docs:check`
- `pnpm --filter eve build`
- `MSB_HOME=/tmp/eve-selfmod-parked-handoff-msb pnpm exec eve eval self-modification/create-background-replication-check --strict --verbose` (9/9 gates passed)
- CI self-modification regression eval (9/9 gates passed with the fix)

### Checklist


cmpadden:
## Investigation results

I reproduced this locally from the PR branch using the fixture’s generated self-modification scaffold and an isolated `MSB_HOME`:

```sh
pnpm run e2e:prepare
MSB_HOME="$(mktemp -d /tmp/eve-selfmod-msb.XXXXXX)" \
EVE_E2E_MODEL=openai/gpt-5.6-sol \
pnpm exec eve eval --strict --verbose \
  self-modification/create-background-replication-check
```

The reproduction reaches the reported behavior:

```text
self-modification writes agent/tools/eval_replication_check.ts
→ eve dev detects the edit and performs a structural reload
→ a new development generation is published
→ the same parent session receives the follow-up
→ eval_replication_check is absent from that turn’s tool surface
```

This is not a TUI artifact-refresh issue and not a failed authored rebuild. The TUI deliberately keeps its existing `ClientSession` across HMR. With the single-owner session topology, the existing owner must hand off to the generation that accepted the next eligible user delivery before that turn can use newly authored tools.

The owner does see the new-generation delivery, but retains ownership before it starts a successor:

```text
[eve:execution.session-handoff] session handoff retained existing owner {
  reason: 'not-idle',
  sourceDeploymentId: '<generation A>',
  targetDeploymentId: '<generation B>',
  work: {
    handleCount: 1,
    nonterminalTaskIds: [],
    pendingKeys: [],
    pendingInputBatchCount: undefined,
    workflowToolRunCount: undefined,
    proxyInputRequestCount: undefined
  }
}
```

The earlier `missing-deployment` event is an internal child/task notification without ingress deployment metadata; it is unrelated to the user follow-up. The relevant user follow-up is stamped for generation B and is rejected only because the parent is not idle.

The remaining handle is intentional under the current subagent lifecycle, rather than an unperformed terminal cleanup:

- `settleAgentTurn()` deletes a handle only for an explicit `terminal` child outcome; it transitions a `parked` outcome to a resumable parked handle.
- `resolvePendingCoordination()` already applies this transition and has coverage for terminal-result deletion.
- The self-modification child is a normal dynamic `defineAgent()` subagent. After its source-editing turn completes, its child session remains parked and resumable, so the parent retains one `eve.agent.handles` entry.
- The single-owner handoff contract deliberately refuses every retained task, subagent, and workflow-tool handle because checkpoints do not transfer their live ownership or callback state.

The complete causal chain is therefore:

```text
self-modification child finishes its editing turn
→ child session remains parked/resumable
→ parent retains its parked subagent handle
→ parent cannot satisfy handoff-idle invariant
→ generation-B user follow-up stays with generation-A owner
→ generation-A tool projection does not contain the newly authored tool
```

I added temporary, local-only diagnostics while investigating both pre-activation retention branches and candidate activation failures. Candidate activation did not fail in this reproduction; the diagnostic establishes that the handoff is blocked by the parked self-modification handle before candidate startup.

Focused validation of the instrumentation passed:

```text
  src/execution/session/handoff.test.ts \
  src/execution/session/handoff-steps.test.ts

17 tests passed
```

The real-model repro still fails as expected after reaching this path: `eval_replication_check` is not offered or called on the next turn of the existing session.

AndrewBarba (approved):

PR #3468, opened 2026-09-17
[1/3] feat(eve): define external resource snapshot

benpankow:
### Summary

Defines a small, versioned internal snapshot for external resources discovered while compiling an eve agent. The snapshot describes provider-independent resource identity, interfaces, credential requirements, routes, and provider-native configuration; it intentionally does not define or emit the final Connect manifest.

```json
{
  "kind": "eve-external-resources",
  "schemaVersion": 1,
  "generator": { "name": "eve", "version": "..." },
  "resources": []
}
```

The snapshot is designed as an ephemeral handoff to build integrations rather than a published deployment artifact. Replaces #2495.

### Validation

Adds focused unit coverage for snapshot construction and versioned output.

🤖


packages/eve/src/internal/external-resources-snapshot.ts
```diff
@@ -0,0 +1,52 @@
...
+
+interface ExternalResourceBase {
+  readonly credentials: ExternalCredentialRequirement;
+  readonly logicalPath: string;
+  readonly name: string;
+}
+
+export interface ExternalConnectionResource extends ExternalResourceBase {
+  readonly connection: {
+    readonly type: "mcp" | "openapi";
```
dvoytenko, line 21: This would be better as a discriminated union. E.g. mcp can have transport info, openapi might have its own metadata. This metadata is pretty useful.

packages/eve/src/internal/external-resources-snapshot.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+interface ExternalResourceBase {
+  readonly credentials: ExternalCredentialRequirement;
+  readonly logicalPath: string;
+  readonly name: string;
+}
+
+export interface ExternalConnectionResource extends ExternalResourceBase {
+  readonly connection: {
```
dvoytenko, line 20: this is more "protocol" or "interface", or even "resource". "connection" is overloaded.

dvoytenko (approved):

packages/eve/src/internal/external-resources-snapshot.ts
dvoytenko: One thing OpenAPI is likely to need from OAuth is "scopes". With MCP scopes are usually always defaulted. But OpenAPIs still very much need them. In this case it's a requirement and somewhat similar to Slack channel where the intent is "For me to work correctly, I need permissions to read the channel". I'm just not sure in case of OpenAPI connection if this should be on the connection level or Connect level. I'd tend to think of latter for now.

benpankow: Updated to discriminated union.

packages/eve/src/internal/external-resources-snapshot.ts
benpankow: Renamed to `protocol`.

dvoytenko (approved):

PR #3478, opened 2026-09-17
feat(eve): configure background task reporting on message sends

ruiconti:
### Summary

Useful background results should not have to wait for unrelated work to finish. Add `taskDeliveryPolicy: "auto" | "cohort"` to `send(...)`, so the caller can choose whether a session reports independently useful results as they arrive or waits for a combined report. New channel sessions default to `auto`; new scheduled sessions default to `cohort`. Related to #1084.

### Before → after

Background task outcomes currently wait for every task in their cohort, the group of overlapping tasks, to settle before waking the parent. That suits scheduled digests, but holds back results that could already answer part of a user's request.

Suppose A is useful on its own, while B and C are needed for a combined answer:

| Event | Before | With `auto` |
| --- | --- | --- |
| A finishes | Hold A until the cohort settles. | Wake the parent so it can report A. |
| B finishes while C runs | Hold B too. | Wake the parent; it can withhold B and stay silent. |
| C finishes | Report A+B+C together. | Report B+C without repeating A. |

`cohort` keeps the existing grouped behavior. With `auto`, the model decides whether a result is independently useful. Each reporting turn receives the whole cohort's pending status and available outputs, including results withheld earlier. A silent turn still costs a model call; completions already queued together may share one wake.

### Configure the send

```ts
await from(address).send(message, { auth, taskDeliveryPolicy: "cohort" });
await to(slack, target).send(message, { auth, taskDeliveryPolicy: "auto" });
await session.send(message, { auth, taskDeliveryPolicy: "cohort" });
```

The option belongs to the send that starts or continues the work. Channel definitions, wrappers such as `slackChannel(...)`, and schedule definitions expose no policy setting. Client message sends, Chat SDK `bridge.send(...)`, and Slack event sends accept the same option. Compile-time assertions cover every channel wrapper and schedule definitions. Chat SDK declarations now reference the installed `chat` package: copied declarations had lost configuration checks and created incompatible private class fields. eve and docs use Chat SDK 4.41.0, with a dependency guard against version drift. The runtime remains bundled; `chat` is an optional peer for this integration.

An explicit value updates the session policy, including pending tasks. Omission preserves an existing session's policy. For a new session, ordinary sends default to `auto` and scheduled sends default to `cohort`. A schedule handler can select `auto` on its `send(...)`; Markdown schedules use the `cohort` default. Cross-channel sends carry their setting through the receiver's `from(...).send(...)` unless that call supplies an explicit value.

The runtime persists one policy value in `eve.runtime.taskDeliveryPolicy` and reads outputs from the existing workflow tool run registry. Scheduled task-mode sessions stay alive while they own pending tasks, so a partial report cannot finish the session and cancel its siblings. Failures and cancellations update lifecycle state immediately; their automatic reports follow the selected policy. Input requests and explicit task messages remain eligible for delivery under either policy. Internal child sessions default to `cohort`.

### Validation

Rebased onto `main` at `dea2ced59`.

- On this revision: the package build, complete docs registry check (including TypeScript), source typecheck, all 1,077 channel tests, dependency-version check, targeted lint, and invariant checks pass. Type assertions verify that the bridge and external handlers share `Message` and `Thread` types.
- The earlier full unit run passed 9,080 tests with one skipped; it has not been repeated on this revision.
- Channel epoch 28 retains 27; schedule epoch 15 retains 14. Existing contract history is unchanged.
- The [A/B/C eval on the previous head `c5444a5b6`](https://github.com/vercel/eve/actions/runs/35460817121/job/105944257812) passed all 21 assertions. Its launch assertion explicitly checks three `working` children; completion and reporting are checked after their approvals. It tests delivery behavior with a deterministic mock, not a real model's judgment of usefulness. CI must validate the new head; no E2E was run locally.

### Checklist


AndrewBarba (approved):
Approved based on updated direction, removing from channel api and only keeping on send/receive

PR #3507, opened 2026-09-18
feat(eve): step state - persist delta returns for workflow replay

AndrewBarba:
### Summary

Accumulating conversations repeat their full history and context in every turn-step input and output. Return state deltas and apply them in the owning workflow during normal replay, preserving cancellation and background-task checkpoints. Full inputs remain persisted; the 32-step serializer benchmark reduces combined state bytes by 47.8%. The state argument is separated from streams, signals, and delivery so adopting [replay-derived inputs](https://github.com/vercel/workflow/pull/4239) becomes `turnStep.replayInputs = [0]` after upgrading Workflow.

Deltas capture state before authored mutations and represent nested edits, appends, removals, and truncation. Aliased/cyclic subtrees travel together; rich values retain full replacement through Workflow's serializer. Initialization, handoffs, and other coordination steps keep their existing checkpoints. No new state store or runtime history retrieval is introduced. Total storage and diff work can still grow quadratically while full inputs remain.

### Validation

- `pnpm build`
- `pnpm test:unit`; final session tests: `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/execution/session` (288 passed)
- `pnpm test:integration` (885 passed); final serializer tests: `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/execution/session/turn-step-delta.integration.test.ts` (3 passed)
- `pnpm typecheck`; final package check: `pnpm --filter eve exec tsc -p tsconfig.json --noEmit`
- `pnpm fmt`, `pnpm lint`, `pnpm guard:invariants`, `pnpm docs:check`
- Real session replay with Node's retained VM enabled/disabled, stored-output inspection, in-place authored state mutation, cancellation/background checkpoints, handoff/import coverage, interrupted-input retry simulation, and rich-value serialization. eve currently excludes the QuickJS runtime.
- Expanded the existing reconnect e2e eval; e2e runs only in CI.

Fixed 2 KiB history additions, actual Workflow serialization with compression disabled (state payload bytes, excluding streams/events):

| Steps | Persisted inputs | Full output baseline | Delta outputs | Combined reduction |
| 8 | 61,891 | 78,497 | 20,377 | 41.4% |
| 16 | 257,540 | 290,721 | 40,786 | 45.6% |
| 32 | 1,050,116 | 1,116,449 | 81,618 | 47.8% |

### Checklist


packages/eve/src/execution/session/state-delta.ts
```diff
@@ -0,0 +1,154 @@
...
+      return keep;
+    return arrays
+      ? { kind: "array", beforeLength: before.length, length: after.length, remove, set }
+      : { kind: "object", remove, set };
+  }
+  return { kind: "replace", value: after };
+}
+
+/** Copy-on-write keeps the pre-step checkpoint available for cancellation and background work. */
+export function applyValueDelta(before: unknown, delta: ValueDelta): unknown {
```
ruiconti, line 104: What do you think of fuzzy testing this module APIs eg with `fast-check`? Ran this and found a bug in `diff ` when reversing props

https://github.com/vercel/eve/pull/3509

packages/eve/src/execution/session/state-delta.ts
```diff
@@ -0,0 +1,154 @@
...
+      while (common < previous.length && previous[common] === path[common]) common++;
+      atomic.add(JSON.stringify(path.slice(0, common)));
+      return;
+    }
+    seen.set(current, path);
+    if (
+      (!isRecord(current) && !Array.isArray(current)) ||
+      Object.getOwnPropertySymbols(current).length > 0
+    ) {
+      atomic.add("[]");
```
ruiconti, line 46: One unchanged rich value disables delta savings for the entire projected state. `atomic.add("[]")` marks the root, so a `Date` anywhere in serialized context causes the context, session history, and metadata to be replaced together. This is broader than replacing only the rich subtree.

I reran the existing 32-step, 2 KiB append benchmark with compression disabled, adding only `createdAt: new Date(0)` to the initial serialized context and retaining it in subsequent results:

| Case | Input bytes | Full-output bytes | Delta-output bytes | Combined reduction |
| Original plain state | 1,050,116 | 1,116,449 | 81,618 | 47.76% |
| With one unchanged Date | 1,051,907 | 1,118,209 | 1,119,905 | -0.08% |

All 32 deltas in the Date case were root replacements. The fallback preserves correctness, but this should be an explicit benchmark and documented optimization boundary. If narrowing the fallback, preserve aliases across the rich value and the rest of the graph.

packages/eve/src/execution/session/turn-step-delta.integration.test.ts
```diff
@@ -0,0 +1,174 @@
...
+async function encode(value: unknown): Promise<Uint8Array> {
+  return (await dehydrateStepReturnValue(
+    value,
+    runId,
+    undefined,
+    [],
+    globalThis,
+    false,
+    false,
+    false,
```
ruiconti, line 27: The uncompressed result demonstrates reduced duplication, but it does not yet establish the production storage/computation tradeoff. I reran this same 32-step fixture with compression enabled for both step arguments and return values:

| Compression | Input bytes | Full-output bytes | Delta-output bytes | Combined reduction |
| Disabled | 1,050,116 | 1,116,449 | 81,618 | 47.76% |
| Enabled | 14,127 | 13,893 | 9,620 | 15.25% |

The fixture repeats `"x"` heavily, so neither percentage should be treated as an estimate for real conversations. Capture and diff also traverse accumulated state, adding computation that the byte benchmark does not measure.

Please add compressed measurements using representative conversation payloads, plus execution-time and peak-memory measurements for the full-result and delta paths. The decision criterion is storage saved versus added computation and maintenance cost, while preserving serialization behavior.

ruiconti (commented):
Two performance observations from rerunning the serializer benchmark at this revision. These concern the scope of the storage savings and the evidence for net benefit; neither establishes another state-corruption defect.

PR #3527, opened 2026-09-18
refactor(slack): route every outbound Slack call through the channel transport

pcomans:
> Groundwork for #3605, which adds a configurable Slack API base URL. Merge this first.

### Why

The channel sends every outbound Slack call through one transport, which resolves the bot token for the installation workspace, builds the URL and signs the request. Two calls did not: the freeform modal's `views.open` and the answered-card `chat.update` each built their own `fetch("https://slack.com/api/…")`. Nothing configurable can apply while those exist.

### What

Both now call the same function the rest of the channel uses. No new abstraction and no public API change. Their bodies go out form-encoded like every other call the channel makes; Slack accepts both encodings for these methods.

Behaviour is identical at both sites and each case has a test. The subtlety worth checking: the transport parses the response body before it reads the status, so a non-2xx carrying an HTML error page surfaces as a parse failure rather than a Slack error. The modal's handler catches both, because Slack retries any interaction ack that is not a 2xx. A network failure and an unresolvable token still fail the request, as before.

One thing to know rather than review: `log.error("Slack views.open returned non-2xx", { status })` is now `log.error("Slack views.open failed", { error })`, which renames a logged string and drops a structured field.


pcomans:
Adversarial review of commit 08052c3c:

### P2 — URL normalization is string manipulation pretending to be URL handling

`packages/eve/src/public/channels/slack/api-transport.ts:56–60`

For `https://sim.example/api?fixture=demo`, appending `/` modifies the query, not the pathname. Resolving `chat.postMessage` then produces `https://sim.example/chat.postMessage`, losing both `/api` and the query. Fragment-bearing bases have the same path-normalization problem.

Parse the URL at the shared boundary and normalize its pathname. Either reject query/fragment components explicitly or define how they are preserved. Silently routing somewhere else is the worst outcome. Cover this through both the vendored caller and `postSlackApiJson`.

### Nonblocking coverage gap — The upload test doesn't upload anything

`packages/eve/src/public/channels/slack/api.test.ts:1075–1099`

`slack.uploadFiles([])` returns without making a request. It adds the appearance of coverage while proving nothing about upload routing.

Use one small file and verify `getUploadURLExternal`, the returned upload URL, and `completeUploadExternal` all use the injected fetch. Assert that global fetch remains untouched. The implementation looks correctly wired; the test doesn't establish that.

### Nonblocking architecture — Moving duplicated transport code doesn't eliminate the duplication

`packages/eve/src/public/channels/slack/api-transport.ts:149–163`

`postSlackApiJson` repeats URL construction, authorization, encoding, and fetch selection. The vendored primitive already supports JSON requests.

Use that primitive behind a small eve-owned wrapper while preserving intentional response handling. The duplication predates this PR, so this is cleanup feedback rather than a new regression. The module split itself is reasonable; there is no need for a transport class, registry, or another abstraction layer.

### Verification

The review subagent reported:

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/public/channels/slack` — PASS, exit 0; 20 files, 470 tests.
- `git diff --check HEAD^ HEAD` — PASS, exit 0.
- An in-memory URL-resolution check — confirmed the path-loss behavior, exit 0.

The cited changes were also checked against the local checkout. No files changed. Typecheck, integration tests, docs checks, and e2e were not rerun for this review.

pcomans:
Replying to the adversarial review of 08052c3c, finding by finding.

**P2 — URL normalization is string manipulation pretending to be URL handling.** Already fixed in 73f2ccbb5, before this reply: `resolveSlackApiUrl` parses the base and normalizes `parsed.pathname`, so `https://sim.example/api` resolves `chat.postMessage` to `https://sim.example/api/chat.postMessage`. A base carrying a query string or fragment is rejected outright rather than silently rerouted, and a non-absolute base is rejected with a message naming what to pass instead. Both the vendored caller and the JSON path go through that one boundary; `api.test.ts` covers the pathname normalization, the query/fragment rejection through both callers, and the relative-base rejection.

**Nonblocking coverage gap — The upload test doesn't upload anything.** Already fixed in 73f2ccbb5: the test now uploads one small CSV and asserts all three legs — `files.getUploadURLExternal`, the upload URL Slack hands back, and `files.completeUploadExternal` — go through the injected fetch, with `expect(fetchMock).not.toHaveBeenCalled()` proving the global one stays untouched.

**Nonblocking architecture — Moving duplicated transport code doesn't eliminate the duplication.** Fixed in 83c81575e: `postSlackApiJson` is gone, replaced by `callSlackApiJson`, a wrapper over the vendored `callSlackApi` with `contentType: "json"` and options from the existing `createSlackApiOptions` — no more hand-rolled URL construction, authorization header, encoding, or fetch selection. Response handling is preserved per surface: `views.open` still never fails the interaction ack (it catches `SlackApiError` and logs the status), while the answered-card `chat.update` now lets `SlackApiError` surface instead of a bare `Error`, so the method and Slack's parsed error body come with it. No transport class, registry, or extra abstraction layer was added.

One knock-on: `interactions.ts` was at 699 lines against the 700-line structural cap, and the import plus the try/catch pushed it over. Rather than take an allowlist entry the rule reserves for files already above the cap, I moved `findPromptBlocks` and `readPromptTextFromBlocks` to `hitl.ts` next to the `buildAnsweredBlocks` that consumes them — `findPromptBlocks` existed as a byte-equivalent copy in both `interactions.ts` and `interaction-cards.ts`, so that deletes a duplicate too. `interactions.ts` is now 680 lines.

Local: `pnpm typecheck`, `pnpm lint`, `pnpm fmt`, `pnpm test` (unit 8915 passed / integration 881 passed), `pnpm guard:invariants`, `pnpm docs:check`, `pnpm --filter eve build` all pass.

pcomans:
**CI status at `03be10d1c`: green.** 137 checks pass, 0 fail, 1 skipped.

`test-dev-windows` is the only non-pass and it is `skipping`, not a failure — the CI workflow's own path filter declines to run it.

Three reds along the way, all re-run and now green, none in code this PR touches:

- `test-integration (windows-latest)` — `src/execution/session/entry.integration.test.ts > parks before initialization and initializes with the first message identity and title`: `Error: Test timed out in 30000ms.` on a runner that had already spent 375s in tests. Passed on re-run. The CI workflow is green on `main` across its last five runs (35619398769, 35614999973, 35558786825, 35544920559, 35455044341), so this is runner load, not a standing failure.
- `e2e-local (agent-prompt-cache, anthropic-opus)` — a provider refusal, not an assertion: `model call failed — parking session for retry by the user { finishReason: 'content-filter' }`, which then cascaded into `notEvent(step.failed)` and `calledSubagent(reviewer)`. Passed on re-run. This is the real-model nondeterminism the e2e guidance in `AGENTS.md` warns about.
- `test-scenario` — `test/scenarios/nitro-bundle-report.scenario.test.ts > groups symlinked Vercel functions by their real payload and includes publish metrics`: `Error: Test timed out in 15000ms.` on a run that spent 1224s in tests. Passed on re-run. It had also passed on the immediately preceding head `83c81575e`, whose only difference is one line of changeset markdown.

The bundle-analysis comment above flags `Runtime: Total function bytes +1.63 MB ⚠️`. That warning is not this PR: it was already `+1.56 MB` at the first push, it tracks drift in the `main (61d98c3)` baseline rather than the diff, `Packed tarball` actually moves `-204.6 kB ✅`, and Socket reports "Pull request contains no net changes to dependencies". The branch adds no dependency and no new runtime import — the diff is six Slack source files.

Vercel Agent Review: "Suggestions made: 0". Vercel Security Review: "No findings."

pcomans:
@benpankow I pushed the separate constructor, then went looking for precedent before asking for a re-read, and found three that point the other way:

- `discordChannel` — `readonly api?: Omit<DiscordApiOptions, "credentials">` (`apiBaseUrl`, `fetch`)
- `telegramChannel` — the same, plus `fileBaseUrl`
- `githubChannel` — `readonly api?: GitHubApiOptions` (`apiBaseUrl`, `fetch`)

There is no `mockDiscordChannel` or `mockGithubChannel`, and GitHub Enterprise Server is the most obvious non-mock reason a base URL is configurable. The vendored Slack adapter documents the same knob for a production flavour of Slack:

> `/** Override the Slack API base URL (e.g. "https://slack-gov.com/api/" for GovSlack). */`

So "mock" may be the wrong label for the capability. Under the current shape a GovSlack deployment calls `mockSlackChannel()` in production.

Two ways to go, and I am happy with either:

1. **`slackChannel({ api: { apiBaseUrl, fetch } })`**, the same key and field names as the three siblings. Slack stops being the odd one out, and the diff shrinks a lot: no second constructor, no resolved-config type threaded through eight files.
2. **Keep `mockSlackChannel`.** Then it is worth deciding whether "a production channel's transport cannot be redirected" should hold for discord, telegram and github too, which is a bigger conversation than this PR. It also puts a new top-level export under the research-doc and e2e bar in `AGENTS.md`, which I am happy to write if that is the direction.

Your call. Say which and I will push it.

Unrelated to the shape, a review of this branch turned up a regression I am fixing either way: folding `views.open` into the shared transport changed its failure mode. The vendored primitive parses the response body before checking the status, so a non-2xx with a non-JSON body (a proxy's HTML error page, say) raises `SyntaxError` rather than `SlackApiError`, and the handler's `instanceof SlackApiError` guard rethrows it. That turns a logged warning into a 500 on the interaction ack, which Slack then retries — exactly what the comment above the guard says it prevents. The previous code checked `response.ok` without parsing, so it was immune.

packages/eve/src/public/channels/slack/interactions.ts
```diff
@@ -570,16 +568,17 @@ async function openFreeformModal(input: {
...
-    headers: {
-      authorization: `Bearer ${token}`,
-      "content-type": "application/json; charset=utf-8",
-    },
-    body: JSON.stringify({ trigger_id: triggerId, view }),
-  });
-  if (!response.ok) {
-    log.error("Slack views.open returned non-2xx", { status: response.status });
+  // The token is resolved above rather than handed to the transport, so a
+  // throwing resolver cannot land in this catch, which only wants response errors.
```
pcomans: Where is it resolved?

pcomans: Four lines up, at the `resolveSlackBotToken` call just above the `try`. The comment is about which side of the `try` that call sits on: resolving before it means a `credentials.botToken` resolver that throws surfaces to the caller, while the catch only ever sees errors from the `views.open` response. Moving the resolve inside would put a resolver failure in the same catch, and since it only rethrows what is not a response error, a throwing resolver would be silently swallowed and the click acked with no modal. I will reword it to name the line rather than say "above".

packages/eve/src/public/channels/slack/api-errors.ts
```diff
@@ -0,0 +1,20 @@
...
+import { SlackApiError } from "#compiled/@chat-adapter/slack/api.js";
+
+/**
+ * True when `error` is Slack's answer to a call rather than a failure to
+ * make one. The transport parses the response body before it reads the
+ * status, so a non-2xx carrying an HTML error page raises a `SyntaxError`
+ * rather than a `SlackApiError`; both mean a response came back.
+ */
+export function isSlackResponseError(error: unknown): boolean {
+  return error instanceof SlackApiError || error instanceof SyntaxError;
```
pcomans: P2 — Classify response failures at the transport boundary instead of guessing from exception types.

This predicate misses failures while consuming a response body. The pinned transport awaits `response.json()` before checking HTTP status. If the connection drops after headers arrive, that read can throw `TypeError: terminated`. `openFreeformModal` then rethrows it at `interactions.ts:580`, so the handler never reaches the interaction acknowledgement. Previously it checked `response.ok` without reading the body and acknowledged the same response.

Confirmed with the pinned adapter and this classifier: both HTTP 200 and HTTP 502 with an interrupted body were acknowledged before this change and now reject. The complete HTML-error-page test covers `SyntaxError`, but misses this failure mode.

The root fix belongs in the transport: preserve an explicit distinction between failing before a response arrives and failing while processing a received response. Don't add `TypeError` to this predicate; that also catches genuine pre-response fetch failures that this refactor promises to preserve. Add interrupted-body coverage for both successful and non-2xx responses, alongside the existing fetch-rejection case.

pcomans (commented):
One confirmed regression to fix before merge: received responses with interrupted bodies can now prevent interaction acknowledgement. Details inline. Verified with an isolated probe using the pinned @chat-adapter/slack 4.34.0 transport and this commit's actual error classifier; HTTP 200 and 502 both reproduce it. Full channel tests were not run locally because dependencies and generated artifacts are absent. The reported CI checks pass.

.changeset/slack-single-api-transport.md
```diff
@@ -0,0 +1,7 @@
+---
+"eve": patch
+---
+
+The Slack channel's `views.open` and answered-card `chat.update` now go through the channel's own Slack API transport. Slack sees the same calls, and a `views.open` that fails after Slack responds is still logged and acknowledged.
+
+A failed `views.open` is now logged as `Slack views.open failed` with the error attached, where it was `Slack views.open returned non-2xx` with a `status` field. Alerts or log queries keyed on either need updating.
```
pcomans: Should we mark this as BREAKING? Is that the repo convention?

packages/eve/src/public/channels/slack/api-errors.ts
pcomans: Fixed in `501dd0bef`. The transport now hands the vendored primitive a `fetch` that observes the response, and anything thrown from that point on is recorded, so `isSlackResponseError` reads a fact rather than inferring one from the error class. `TypeError` was not added to the predicate, so a request that never reached Slack still surfaces to the caller and still fails the interaction. Tests cover an interrupted body on a 200 and on a 502, alongside the existing fetch-rejection case.

.changeset/slack-single-api-transport.md
pcomans: `patch` matches the convention. `AGENTS.md` says eve is pre-1.0, so `patch` covers most changes including new features, and `minor` is reserved for a change that breaks a public API. No exported type or signature moves in this PR, so it stays a patch. The eight pending changesets are all `patch` too.

There is no `BREAKING` marker in this repo: nothing in `.changeset/` or the changelog uses one, and the bump level is what produces the "Minor Changes" and "Patch Changes" headings. The log rename is the second paragraph of the note, so it reaches the release notes either way.

benpankow (approved):
Thanks, this is a useful unification

PR #3593, opened 2026-09-21
test(slack): drive the Slack tests off a strict, typed Slack double

pcomans:
### Summary

eve's Slack tests each hand-rolled their own `fetch` mock, and unrecognised calls fell through to a friendly `{ ok: true, ts: … }`. A test could pass having declared nothing about what it expected Slack to do.

This replaces that with a **strict, typed Slack test double**, modelled on rspec-mocks:

- **Strict by default.** A message the test did not stub is a loud failure naming the method, not a canned success.
- **Verified against a contract.** A typed method → request/response map plays the role of rspec's verifying doubles: only real methods typecheck, requests and responses are shape-checked. Request types are a *wire projection* — form encoding turns every scalar into a string, so that is what the types say, and a constraint like `.with({ limit: 50 })` is a compile error rather than a matcher that silently never fires.
- **Interaction assertions, not a simulated workspace.** Tests assert which messages were received with which arguments; they do not query a fake world.

`mock-slack-api.ts` (387) and `mock-slack-workspace.ts` (567) are deleted. An earlier revision of this branch replaced them with a 954-line stateful in-memory Slack; that is deleted too. The problem with simulating Slack was not that it was wrong — an audit found it wrong in one branch production never calls — but that establishing whether it was honest took a dedicated investigation, and that cost recurs for every future reader. A simulation of an API we do not own cannot be verified; a contract we do own can be.

This is a **test-only** change. No production source and no published types are modified.

### What it caught

Not hypotheticals — each of these was green before:

- **Four tests in the mention and receive pipelines were asserting about the public branch while exercising the "treat as private" one.** A test helper threw a plain `Error` for an unexpected channel id; production deliberately swallows a throwing `conversations.info` and fails closed, so the turn silently took the private path. Surfaced when the blanket stub was split.
- **An assertion about posted text could not fail.** `thread.post` sends `markdown_text`, not `text`. The old fake's accessor normalised both into one stored field, so the assertion had nothing to bite on — the convenience accessor had erased a distinction production makes.
- **A matcher that could never fire.** The contract typed form-encoded values as numbers and booleans; on the wire they are strings. Fixed by typing the wire shape.

It also found two production bugs, now their own PRs: #3592 (`thread.post` returns a file id where the contract promises a message `ts`) and #3591 (a Slack 429 throws mid-turn; eve has no outbound rate-limit handling).

### Scope of the strictness

A blanket `allowChannelTurn` helper originally pre-stubbed 13 of the 14 contract methods across ~127 tests, which quietly reduced strictness to "strict about methods that would not typecheck anyway". It is now split into `allowOutboundPost` and `allowInboundResolve`. **Only 11 of 135 tests need anything beyond those** — that ratio is what the wide baseline was suppressing.

### Stated limits

- A **parity check** asserts the contract's keys match the methods the suite stubs. It only sees methods a test reaches, and it counts *stubbed* rather than *called*, so a baseline stub nothing exercises would pass. Current entries were hand-checked as genuinely driven.
- The contract encodes what we believe Slack and production do. It cannot enforce that production conforms, because the transport seam is still `fetch`.
- `with()` cannot express two argument-discriminated stubs for one method; `andRespond` covers that case. Documented at the call site.
- Four production methods (`chat.startStream`, `chat.appendStream`, `chat.stopStream`, `chat.delete`) are deliberately unmodelled, since no test drives them. The first test to touch one will not compile until its entry is added.

### Coverage

No coverage lost. Assertions deleted were ones that only restated the double's own behaviour and could not have failed for a reason involving eve — nine tests of the old workspace's state machine, plus three claims the fake had itself declared true. Assertions that got stronger: cursor pagination checks the exact echoed cursor rather than `expect.any(String)`; post→update linkage is asserted rather than emergent; successive posts return distinct timestamps.

### Note on the branch name

`test/stateful-slack-fake` is a misnomer — the double is not stateful. Renaming would retarget two dependent PRs for no benefit.

### Validation

`pnpm test` (8953 unit + 881 integration), `pnpm typecheck`, `pnpm build`, `pnpm lint`, `pnpm fmt` all pass.

### Stacked on

Based on #3527, which adds the injectable transport this depends on. #3591 and #3592 stack on top of this branch; both were found or made testable by this work.

### Checklist

- [x] I have run the checks
- [x] I have added tests
- [ ] I have added a changeset — not required: `AGENTS.md` exempts internal-tooling and test changes, and nothing published changes
- [x] DCO sign-off on all commits


pcomans:
Superseded by #3597 — same commits, branch renamed to `test/strict-slack-double` because the double is not stateful. Closing this one; no review history is lost.

PR #3597, opened 2026-09-21
test(slack): drive the Slack tests off a strict, typed Slack double

pcomans:
### Summary

eve's Slack tests each hand-rolled their own `fetch` mock, and unrecognised calls fell through to a friendly `{ ok: true, ts: … }`. A test could pass having declared nothing about what it expected Slack to do.

This replaces that with a **strict, typed Slack test double**, modelled on rspec-mocks:

- **Strict by default.** A message the test did not stub is a loud failure naming the method, not a canned success.
- **Verified against a contract.** A typed method → request/response map plays the role of rspec's verifying doubles: only real methods typecheck, requests and responses are shape-checked. Request types are a *wire projection* — form encoding turns every scalar into a string, so that is what the types say, and a constraint like `.with({ limit: 50 })` is a compile error rather than a matcher that silently never fires.
- **Interaction assertions, not a simulated workspace.** Tests assert which messages were received with which arguments; they do not query a fake world.

`mock-slack-api.ts` (387) and `mock-slack-workspace.ts` (567) are deleted. An earlier revision of this branch replaced them with a 954-line stateful in-memory Slack; that is deleted too. The problem with simulating Slack was not that it was wrong — an audit found it wrong in one branch production never calls — but that establishing whether it was honest took a dedicated investigation, and that cost recurs for every future reader. A simulation of an API we do not own cannot be verified; a contract we do own can be.

This is a **test-only** change. No production source and no published types are modified.

### What it caught

Not hypotheticals — each of these was green before:

- **Four tests in the mention and receive pipelines were asserting about the public branch while exercising the "treat as private" one.** A test helper threw a plain `Error` for an unexpected channel id; production deliberately swallows a throwing `conversations.info` and fails closed, so the turn silently took the private path. Surfaced when the blanket stub was split.
- **An assertion about posted text could not fail.** `thread.post` sends `markdown_text`, not `text`. The old fake's accessor normalised both into one stored field, so the assertion had nothing to bite on — the convenience accessor had erased a distinction production makes.
- **A matcher that could never fire.** The contract typed form-encoded values as numbers and booleans; on the wire they are strings. Fixed by typing the wire shape.

It also found two production bugs, now their own PRs: #3592 (`thread.post` returns a file id where the contract promises a message `ts`) and #3591 (a Slack 429 throws mid-turn; eve has no outbound rate-limit handling).

### Scope of the strictness

A blanket `allowChannelTurn` helper originally pre-stubbed 13 of the 14 contract methods across ~127 tests, which quietly reduced strictness to "strict about methods that would not typecheck anyway". It is now split into `allowOutboundPost` and `allowInboundResolve`. **Only 11 of 135 tests need anything beyond those** — that ratio is what the wide baseline was suppressing.

### Stated limits

- A **parity check** asserts the contract's keys match the methods the suite stubs. It only sees methods a test reaches, and it counts *stubbed* rather than *called*, so a baseline stub nothing exercises would pass. Current entries were hand-checked as genuinely driven.
- The contract encodes what we believe Slack and production do. It cannot enforce that production conforms, because the transport seam is still `fetch`.
- `with()` cannot express two argument-discriminated stubs for one method; `andRespond` covers that case. Documented at the call site.
- Four production methods (`chat.startStream`, `chat.appendStream`, `chat.stopStream`, `chat.delete`) are deliberately unmodelled, since no test drives them. The first test to touch one will not compile until its entry is added.

### Coverage

No coverage lost. Assertions deleted were ones that only restated the double's own behaviour and could not have failed for a reason involving eve — nine tests of the old workspace's state machine, plus three claims the fake had itself declared true. Assertions that got stronger: cursor pagination checks the exact echoed cursor rather than `expect.any(String)`; post→update linkage is asserted rather than emergent; successive posts return distinct timestamps.

### Validation

`pnpm test` (8953 unit + 881 integration), `pnpm typecheck`, `pnpm build`, `pnpm lint`, `pnpm fmt` all pass.

### Stacked on

Based on #3527, which adds the injectable transport this depends on. #3591 and #3592 stack on top of this branch; both were found or made testable by this work.

### Checklist

- [x] I have run the checks
- [x] I have added tests
- [ ] I have added a changeset — not required: `AGENTS.md` exempts internal-tooling and test changes, and nothing published changes
- [x] DCO sign-off on all commits


packages/eve/src/internal/testing/mocks/slack-api-contract.test.ts
```diff
@@ -0,0 +1,42 @@
...
+ * call path no test reaches is invisible to it.
+ *
+ * Scanning the tests rather than channel source is what makes the check
+ * possible at all: 7 of these methods never appear as literals in
+ * channel source because the vendored Slack adapter issues them, while
+ * every stub is a literal by construction.
+ */
+describe("Slack API contract parity", () => {
+  it("has an entry for every method the Slack tests stub, and no others", async () => {
+    const files = (await readdir(SLACK_SOURCE_DIR)).filter((name) => name.endsWith(".test.ts"));
```
pcomans, line 29: This is such a dumb test. If we really want this it should be in CI

packages/eve/src/internal/testing/mocks/slack-api-contract.ts
```diff
@@ -0,0 +1,219 @@
...
+ * `request` is declared in Slack's own logical types — `limit` is a
+ * number because Slack documents a number — but the double never sees
+ * those types. It sees what came off the wire, which is what
+ * {@link SlackApiRequest} projects them to.
+ */
+
+/** A Slack message as the Web API returns it inside a response. */
+export type SlackRawMessage = Record<string, unknown>;
+
+export interface SlackApiContract {
```
pcomans, line 32: Does Slack export their types? Can we use those instead?

packages/eve/src/public/channels/slack/activity-progress.test.ts
```diff
@@ -9,6 +10,26 @@ import {
...
+ * call from a render that legitimately did not happen.
+ */
+function slackDouble(): MockSlack {
+  const slack = mockSlack();
+  doubles.push(slack);
+  return slack;
+}
+
+afterEach(() => {
+  for (const slack of doubles.splice(0)) slack.assertNoViolations();
```
pcomans, line 30: Shouldn't violations just throw?

packages/eve/src/internal/testing/mocks/slack-api-contract.test.ts
pcomans: Agreed, deleting it.

It is also weaker than it looks. It cannot see the direction that matters — production calling a method the contract lacks — because 7 of the methods never appear as literals in channel source; the vendored adapter issues them. So all it actually prevents is a contract entry nothing stubs, which is a few lines of unused type. It scrapes source text with a regex to do that.

The real guard against an unmodelled method is the double itself: an unstubbed call fails loudly at runtime. That one works and needs no meta-test.

packages/eve/src/internal/testing/mocks/slack-api-contract.ts
pcomans: They do, and this is the right question — it goes at the drift problem I claimed to solve and my hand-written contract only half solves.

What I checked:

- `@slack/web-api` has typed arguments, and since v6.2 typed responses too.
- eve does not have the package. The vendored `.generated/compiled/@chat-adapter/slack/_slack-web-api.d.ts` looks like Slack types but is an 8-line auto-generated stub — `WebClient` with `[key: string]: any`. So adopting them means a new devDependency, test-only, nothing at runtime.

The part I would not do blind is responses. Ours are deliberately narrow — only the fields eve reads — which is what makes `andReturn({ ok: true, ts })` fail to compile without the `ts`. Slack’s generated response types cover each method’s full surface, and I have not verified how much of that is optional; if it is mostly optional, swapping them in would loosen stubs while tightening arguments.

So my suggestion is a hybrid: take Slack’s argument types as the source of truth and keep the wire projection on top, since form encoding makes every scalar a string either way; keep hand-written response shapes narrow to what eve reads.

Want that in this PR, or as a follow-up? It is a real chunk of work and it changes the shape of the contract file, so I would rather not fold it in silently.

packages/eve/src/public/channels/slack/activity-progress.test.ts
pcomans: It does throw — `reject()` in `mock-slack.ts` records and then throws, return type `never`.

The recording exists because eve catches it. `refresh()` and `startTyping()` swallow transport errors, and the privacy path deliberately fails closed when `conversations.info` throws. So in exactly the places a violation matters most, the throw is absorbed and the test goes green. The recorded list is the backstop for that.

But the boilerplate you are looking at is not defensible, and that is the real answer to your question: every file declaring its own `slackDouble()` factory plus an `afterEach` is ceremony a reader has to remember, and forgetting it silently removes the backstop — `activity-progress.test.ts` was missing it until a review caught it.

eve already has the right mechanism. `vitest.unit.config.ts` has `setupFiles` with `unit-guard.ts`, whose stated job is making a whole class of violation "surface as a loud error". Moving the sweep there makes it automatic and deletes the per-file factory and hook. Doing that.

Issue #3284, opened 2026-09-11
Dynamic skill refresh leaves removed package files in the sandbox


komly:
### Link to a minimal reproduction

[Focused regression tests against eve source](https://github.com/komly/eve/blob/fix/dynamic-skill-package-refresh/packages/eve/src/context/dynamic-skill-refresh.integration.test.ts).

### Steps to reproduce

1. Use Node.js 24 and clone the reproduction branch:

   ```sh
   git clone --branch fix/dynamic-skill-package-refresh https://github.com/komly/eve.git
   cd eve
   pnpm install --frozen-lockfile --filter 'eve...' --filter eve-monorepo
   pnpm --filter eve build:js
   ```

2. Restore the affected implementation to the upstream base, keeping the regression tests:

   ```sh
   git restore --source=df62d7361afd7e4ee01b57ed8ce7e74ae9fe40ca -- packages/eve/src/context/dynamic-skill-lifecycle.ts
   pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/context/dynamic-skill-refresh.integration.test.ts
   ```

3. Restore the branch implementation and rerun the same test:

   ```sh
   git restore --source=HEAD -- packages/eve/src/context/dynamic-skill-lifecycle.ts
   pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/context/dynamic-skill-refresh.integration.test.ts
   ```

The test resolves a package on `turn.started`, then returns a new revision from the same resolver in the same session. The new revision changes text and binary files and omits `references/obsolete.txt`. It uses the real just-bash backend and `ContextContainer`, without sandbox or resolver mocks.

### Current vs. expected behavior

The new files and `SKILL.md` update, but `references/obsolete.txt` remains readable in the sandbox. A file-to-directory or directory-to-file change also fails because the old path survives. Expected: after a successful refresh, the package directory contains exactly the current resolver's files.

`writeSkillPackageToSandbox` currently only writes the returned file list. The lifecycle removes an entire package when its skill name disappears, but does not remove omitted siblings when the name stays the same.

### eve version

`eve@0.54.0`; source baseline `df62d7361afd7e4ee01b57ed8ce7e74ae9fe40ca`.

### Environment

macOS arm64; Node.js v24.21.0; pnpm 11.25.0; just-bash 3.1.0.

### Where does the bug occur?

Production (`eve start`); also reproducible through the focused source integration test above.

### Build and runtime logs

[...]

Issue #1972, opened 2026-08-12
Markdown-only dynamic skills should not require opening a sandbox session


gabrielmmh:
A dynamic skill resolver is a function in the app that produces a skill's markdown at the start of a session. It runs in-process and returns a string. To serve that string, eve opens a sandbox: the markdown is written into the sandbox filesystem, and `load_skill` later reads it back from there.

#969 already fixed this exact problem for static skills. Since 0.26.2 a static skill is served from memory, and the changelog spells out what was left behind: "Listing or loading static skills no longer requires opening a sandbox. Dynamic skills and access to supporting skill package files, such as references, assets, and scripts, remain sandbox-backed." That leftover makes sense for skills that ship extra files, because files need a filesystem. A skill that is only markdown does not need one.

This is the dynamic path as of 0.33.1 (both excerpts prettified from the published, minified dist):

```js
// dist/src/context/dynamic-skill-lifecycle.js
const results = await Promise.allSettled(eventResolvers.map(async (resolver) => {
  const handler = resolver.events[event.type];
  if (handler === undefined) return null;
  const produced = await handler(event, resolveContext);
  if (produced == null) return { resolver, named: [] }; // a null result still flows on
  // ...
}));
// every fulfilled result lands in `resolved`, skills or none
if (resolved.length === 0) return;
// ...
const sandbox = await ctx.require(SandboxKey).get(); // opens a session on the configured backend
if (sandbox !== null) {
  for (const { skills } of resolved)
    for (const skill of skills) await writeSkillPackageToSandbox({ sandbox, skill });
}
ctx.set(DynamicSkillManifestKey, manifest); // entries keep only name and description
```

Two things stand out. The sandbox opens even when the resolver returns `null`: an empty answer is treated like any other and the code continues to the sandbox step. And the markdown itself is dropped: the manifest eve keeps in memory has only the name and description, so the only copy of the body lives inside the sandbox.

That second point is why `load_skill` cannot serve a dynamic skill without a sandbox, while a static one comes straight from memory:

```js
// dist/src/runtime/framework-tools/skill.js
if (dynamicSkillNames.includes(skill)) {
  const sandbox = ctx.get(SandboxKey);
  if (sandbox === undefined)
    throw Error(`The dynamic skill "${skill}" requires sandbox access on the runtime context.`);
  return await loadSkillFromSandbox(sandbox, skill, availableSkillNames);
}
const compiled = compiledSkills.find((entry) => entry.name === skill);
[...]

PR #3604, opened 2026-09-21
fix(eve): skills - defer dynamic package materialization

### LINKED ISSUE #3602: Optimize native skill materialization and refresh

Native dynamic skills add avoidable sandbox and filesystem work before the first model call and on later turns, even when the agent never loads a skill.

In eve 0.62.0, and still in the current upstream [`dynamic-skill-lifecycle.ts`](https://github.com/vercel/eve/blob/main/packages/eve/src/context/dynamic-skill-lifecycle.ts):

- A resolver returning `null` or an empty map still reaches `SandboxKey.get()`.
- Formatting or rebuilding the available-skill announcement also acquires the sandbox to resolve its skill root.
- Every returned package is written again on refresh, even when its contents are unchanged.
- Dynamic markdown loading relies on the staged sandbox copy, tying progressive disclosure to eager filesystem materialization.

These are framework lifecycle costs independent of an application's discovery, ranking, or storage implementation. This issue tracks only optimizations to native skills.

### Proposed solution

1. **Keep availability and markdown loading independent of sandbox startup.** Advertising skill names and descriptions should not provision a sandbox. Empty/no-op refreshes should avoid sandbox work when there is no previously materialized content to reconcile. Serve dynamic markdown without a sandbox, following the existing static-skill behavior.
2. **Defer supporting-file materialization until filesystem access needs it.** Making a package available should not require writing all of its files before inference. Preserve the standard `load_skill`, `ctx.getSkill(...)`, and sandbox access behavior, including usable script/reference paths when the files are actually needed.
3. **Skip unchanged package writes.** Repeated resolution of an identical package against the same materialized sandbox should not rewrite it. Compare complete package contents, including supporting files; equal names or descriptions do not establish that a package is unchanged. Rematerialize when the sandbox is replaced or its prior materialization is no longer valid.

Prefer improving the existing lifecycle over adding another model-facing loader or requiring application-owned caches. These optimizations can land independently; they do not require a remote-source API.

### Acceptance criteria

- A resolver returning `null` or `{}` with no prior package opens no sandbox.
- Advertising or loading a markdown-only dynamic skill opens no sandbox, including after a durable workflow step resumes.
- An available but unused package with supporting files causes no eager filesystem materialization; its files become accessible when required.
- An unchanged refresh performs no package writes. A changed revision updates the affected package, including removed files and file/directory shape changes, without disturbing unrelated packages.
- Removing a skill immediately updates availability and prevents stale dynamic content from being loaded. Cleanup of already-materialized files must still happen where necessary.
- Cold resume, sandbox replacement, dynamic-over-static overrides, and shared sandbox readers retain correct behavior. Materialization failures must not publish a misleading successful state.

Validate with focused lifecycle/load tests and a fixture covering the user-visible load/file-access flow. Measure sandbox acquisitions and package writes as well as turn-start latency.

### Related work and scope

- #1972 already covers empty results and sandbox-free markdown-only dynamic skills; that is the first part of this work. The additional scope here is lazy supporting-file materialization and avoiding unchanged refresh writes.
- Coordinate refresh correctness with #3284; an optimization must not preserve stale files or worsen partial-update behavior.
- #2674 proposes remote skill sources. External catalog/body-fetch laziness, semantic discovery, tags, databases, evaluator choice, and application UI changes are outside this issue.

AndrewBarba:
### Summary

Dynamic skill resolution opened a sandbox even for empty results, discarded markdown after writing it, and rewrote every package on refresh. Static and dynamic `load_skill` now share one in-memory lookup: dynamic instruction bodies are prepared during resolution and retained with package bytes in durable context. Materialize files through the existing sandbox accessor when needed, skip unchanged writes, and reconcile changed or withdrawn packages while restoring authored overrides and keeping failures retryable. Closes #3602; related to #1972, #3284, and the overlapping refresh fix in #3285.

The issue is grounded in the original [announcement acquisition](https://github.com/vercel/eve/blob/e110eb2af4e4611bf12309960085b1eff79648a4/packages/eve/src/context/dynamic-skill-lifecycle.ts#L73), [unconditional acquisition and write loop](https://github.com/vercel/eve/blob/e110eb2af4e4611bf12309960085b1eff79648a4/packages/eve/src/context/dynamic-skill-lifecycle.ts#L209-L235), and [sandbox-backed dynamic loader](https://github.com/vercel/eve/blob/e110eb2af4e4611bf12309960085b1eff79648a4/packages/eve/src/execution/tools/load-skill.ts#L31-L38).

Dynamic package state is capped at 1 MiB of serialized JSON across resolvers, with at most 128 files per package and 256 KiB per file. This includes base64 expansion and the preprocessed instruction body. Static instructions continue to use the compiled agent, with supporting files supplied by the sandbox template. The checkpoint contract advances to version 7: incompatible handoffs are rejected so older sessions retain their original owner.

Shared-child preparation persists materialization ownership into the parent checkpoint without importing task-scoped authentication. Sandbox-local receipts and one in-sandbox checksum command verify declared files, repairing mixed concurrent writes without downloading package bytes. Generated sibling files survive unchanged access. Directory replacement remains non-atomic for concurrent readers; this does not add filesystem transactions or a public authoring API.

### Validation

- `pnpm test` — full unit and integration suites pass, including regressions for resume, replacement sandboxes, binary content, file/directory changes, removals, shared-child checkpoint cleanup, mixed concurrent writes, generated files, static restoration, and failed writes. Focused unit tests also cover incompatible checkpoints and exact durable-payload limits before sandbox access or publication.
- `pnpm typecheck` — all 50 workspace tasks pass, including the updated skills fixture build.
- `pnpm --filter eve typecheck` and `pnpm --filter eve exec vitest run --config vitest.scenario.config.ts test/scenarios/app-runtime-dependencies.scenario.test.ts -t 'rewrites framework tool executors into hosted Vercel output'` pass. The hosted bundle assertion now checks a current loader diagnostic instead of its removed sandbox error.
- `pnpm fmt`, final targeted `pnpm exec oxfmt --check`, `pnpm lint`, `pnpm guard:invariants`, and `pnpm docs:check` pass. Lint reports existing warnings outside the changed files.
- Controlled before/after measurement: five runs with a simulated 25 ms sandbox startup reduced median cold resolver dispatch from 27.1–27.5 ms to 0.01–0.05 ms. Null, markdown-only, and supporting-file results now make zero sandbox accessor calls before use; an unchanged three-file package refresh goes from three writes to zero. This measures lifecycle overhead, not a live VM or model.
- Added deterministic `agent-skills` eval flows for `load_skill` followed by `ctx.getSkill(...).file(...)` or `read_file` across durable steps. The previous commit passed skills E2E on local, Postgres, and Vercel; CI will rerun after the review fixes. The prior run has one unrelated prompt-cache/Anthropic failure with model finish reason `content-filter`. E2E execution is reserved for CI.

### Checklist


AndrewBarba:
Closing in favor of a smaller approach that keeps sandbox verification off the per-tool-call path and avoids persisting package bytes in durable context. A replacement PR will follow for #3602.

PR #3605, opened 2026-09-21
feat(slack): let slackChannel target another Slack API host

pcomans:
> Stacked on #3527; merge that first.

### Why

Testing an agent's Slack surface end to end means standing up a stand-in for Slack, and the channel has no seam for that: the API base URL is a default buried in the transport. The same seam is what a Slack flavour on another host needs.

### What

```ts
slackChannel({
  api: { apiBaseUrl: "http://localhost:3000/api/slack" },
});
```

`api` and all three of its fields are optional. Omitted, the channel talks to `https://slack.com/api/` exactly as today. `fetch` supplies the fetch implementation for those calls, for a stand-in that needs a header attached. `fileBaseUrl` says where it serves `url_private` attachments, falling back to `apiBaseUrl`.

Two things worth a reviewer's eye. The option mirrors what `discordChannel`, `telegramChannel` and `githubChannel` already expose, including the `fileBaseUrl` fallback. And Slack's own hosts always keep the global `fetch`, on downloads and on the upload legs, so a wrapper carrying stand-in credentials cannot send them to Slack.

### Consumer

[vercel/internal-agents#2867](https://github.com/vercel/internal-agents/pull/2867) heads a stack adding a Slack simulator to the `v` agent. It vendors a build of this today and switches to a published `eve` once this lands.


packages/eve/src/public/channels/slack/api-transport.ts
```diff
@@ -117,21 +180,47 @@ export function createSlackTransport(input: { readonly botToken?: SlackBotToken
...
     return body as Record<string, unknown>;
   }
   return {};
 }
+
+/**
+ * Confines a caller-supplied `fetch` to one origin, so a wrapper attaching stand-in
+ * credentials cannot carry them to a host the stand-in merely named — notably the
+ * `upload_url` that `files.getUploadURLExternal` answers with.
+ */
```
pcomans: This is gibberish, what does it mean?

packages/eve/src/public/channels/slack/slackChannel.ts
```diff
@@ -641,6 +642,8 @@ export type SlackApprovalChannelResolver = (
 ) => SlackApprovalChannel | Promise<SlackApprovalChannel>;
 
 export interface SlackChannelConfig {
+  /** API transport overrides (base URLs, custom fetch). Credentials are supplied separately via `credentials`. */
```
pcomans: Why would you need to say that if credentials is right below it?

packages/eve/src/public/channels/slack/api-transport.ts
```diff
@@ -14,6 +14,43 @@ import {
...
+const SLACK_API_BASE_URL = "https://slack.com/api/";
+
+/**
+ * Slack API transport overrides. `apiBaseUrl` defaults to `https://slack.com/api/`;
+ * `fileBaseUrl` falls back to `apiBaseUrl` for attachment downloads. `fetch` overrides
+ * the global fetch for those two bases only — Slack's own hosts keep the global fetch.
+ */
+export interface SlackApiOptions {
+  readonly apiBaseUrl?: string;
+  readonly fetch?: SlackFetch;
```
pcomans: Why do we have to Over ride the slack fetch here?

packages/eve/src/public/channels/slack/api-transport.ts
pcomans: Agreed, and the function is coming out. It confined a caller-supplied `fetch` to the configured origin so that a wrapper attaching credentials for the stand-in could not send them to `files.slack.com`, whose upload URL Slack returns mid-handshake. That protects other consumers and does nothing for the one driving this PR, so it leaves with the rest of the trim. I will raise it with the eve team separately rather than drop it silently.

packages/eve/src/public/channels/slack/slackChannel.ts
pcomans: No reason. `credentials` is the next line, so the sentence answers a question nobody is asking. Cut to a single clause naming what the field is.

packages/eve/src/public/channels/slack/api-transport.ts
pcomans: So a caller can attach a per-call header. The stand-in that motivates this PR runs behind Vercel Deployment Protection, which answers an unauthenticated request with a redirect to SSO, so every call to it carries a bypass header. There is no other seam for that: the base URL says where the call goes, not what it carries.

`discordChannel` and `telegramChannel` both expose the same field alongside their base URL, so this is their shape rather than a new one. It also serves instrumenting or stubbing the calls in a test.

docs/channels/slack.mdx
```diff
@@ -419,7 +420,25 @@ if (!response.ok) throw new Error(String(response.error));
...
+slackChannel({
+  api: {
+    apiBaseUrl: "http://localhost:3000/api/slack",
+    fileBaseUrl: "http://localhost:3000/files",
+  },
+  credentials: { botToken: process.env.SLACK_BOT_TOKEN },
+});
+```
+
+`apiBaseUrl` is the base Slack Web API method names resolve against, and defaults to `https://slack.com/api/`. `fileBaseUrl` is the base attachment downloads are authorized against and falls back to `apiBaseUrl`: a download URL that starts with it is fetched with the bot token, Slack's own file hosts stay downloadable, and every other URL passes through to the model provider unchanged. Both bases must be absolute `http:` or `https:` URLs carrying no query string or fragment; a missing trailing slash is added, so `https://sim.example/api` sends `chat.postMessage` to `https://sim.example/api/chat.postMessage`.
```
pcomans: IS this correct?  `https://slack.com/api/`. `fileBaseUrl` is the base attachment downloads are authorized against and falls back to `apiBaseUrl`:

docs/channels/slack.mdx
```diff
@@ -419,7 +420,25 @@ if (!response.ok) throw new Error(String(response.error));
...
+
+`api.fetch` replaces the global `fetch` for Slack Web API calls and for downloads from the file base — use it to attach the per-call header a host behind an authenticating proxy requires.
```
pcomans: where is this?

packages/eve/src/public/channels/slack/api-options.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * The host the channel's Slack traffic goes to: the Web API base, the
+ * attachment-download base, and the fetch both travel on.
+ *
+ * Separate from `api.ts` because that file sits on the 700-line cap
+ * `internal/structural-tests/file-length.test.ts` enforces.
+ */
+
+import type { SlackApiOptions as SlackPrimitiveApiOptions } from "#compiled/@chat-adapter/slack/api.js";
```
pcomans: Does that match what we're doing for telegram and discord? And also I could have sworn the previous PR introduced something very similar

docs/channels/slack.mdx
pcomans: Checked each clause against the code. The default is right: the vendored transport falls back to `https://slack.com/api/`, and `slackBaseHref` only appends a trailing slash. The fallback is right: `fileBaseUrl ?? apiBaseUrl`, matching `telegramChannel`. The prefix behaviour is right, and so is Slack's own hosts staying downloadable.

One word is wrong. "Authorized against" reads as though the base replaces Slack's file hosts. It adds to them. I will say that plainly: a download URL under this base is fetched with the bot token, in addition to `files.slack.com` and Enterprise Grid.

docs/channels/slack.mdx
pcomans: In the example directly above, and that is the problem: the example sets `apiBaseUrl` and `fileBaseUrl` but not `fetch`, so the paragraph describes a field the reader cannot see. I will add `fetch` to the example.

The behaviour itself is in `slackApiEndpoint` and `slackFileDownloadFetch` in `api-options.ts`, which hand it to the vendored helpers and to the attachment download.

packages/eve/src/public/channels/slack/api-options.ts
pcomans: Yes on both counts, and the second one is a fair hit.

On matching: `TelegramApiOptions` is `{ apiBaseUrl?, credentials?, fetch?, fileBaseUrl? }` and `DiscordApiOptions` is `{ apiBaseUrl?, credentials?, fetch? }`. Ours is Telegram's without `credentials`, because Slack already carries those on `SlackChannelConfig.credentials`. Same names, same optionality, same `fileBaseUrl` fallback.

On the previous PR: #3527 added `api-errors.ts` carrying the same header, that it is separate from `api.ts` only because that file sits on the 700-line cap. This adds a second module with the same apology, so the channel now splits by what fits rather than by concern. I will fold the two into one module so there is one seam beside `api.ts` instead of two.

docs/channels/slack.mdx
```diff
@@ -418,8 +420,36 @@ if (!response.ok) throw new Error(String(response.error));
...
+
+### Point the channel at a different Slack API host
+
+```ts
+slackChannel({
+  api: {
+    apiBaseUrl: "http://localhost:3000/api/slack",
+    fetch: (input, init) => {
+      const proxyKey = process.env.SLACK_PROXY_KEY;
+      if (proxyKey === undefined) throw new Error("SLACK_PROXY_KEY is required.");
```
pcomans: I'm losing my mind. Having error handling in A. Block of example code is fucking wild

pcomans: Also what is SLACK_PROXY_KEY?

packages/eve/src/public/channels/slack/api.ts
```diff
@@ -501,7 +511,7 @@ export function buildSlackBinding(input: {
       }
 
       const response = await postSlackMessage(
-        buildPostMessageOptions(message, input.channelId, currentThreadTs, input.botToken, context),
+        buildPostMessageOptions(message, input.channelId, currentThreadTs, apiOptions),
```
pcomans: Why are we no longer passing the bot token here?

pcomans: It still is — it moved inside `apiOptions`, two lines up in `buildSlackBinding`:

```ts
const token = () => resolveSlackBotToken(input.botToken, context);
const apiOptions = slackApiOptions(input.api, token); // → { apiUrl, fetch, token }
```

`buildPostMessageOptions` spreads `...apiOptions` where it spread `createSlackApiOptions(botToken, context)` before, so the token is the same lazy closure and a rotated credential still resolves per call. `token` is required on the vendored `SlackApiOptions`, and `SlackMessageOptions` extends it, so a dropped token fails typecheck rather than reaching Slack unsigned.

The call now carries three transport concerns — host, fetch, token — and threading them positionally through four sites is worse than one value. The cost is that `apiOptions` does not say the token is in it, which is why the line reads as a removal.

docs/channels/slack.mdx
pcomans: Agreed, and the guard was covering for a missing guarantee rather than teaching anything.

The example attached its header on every request the fetch saw. eve hands that fetch two URLs it does not choose — the `upload_url` from `files.getUploadURLExternal`, and the `url_private` a download arrives on — so copying the example sent the proxy secret to `files.slack.com`, and to any other host a stand-in cared to name. The three lines you are looking at were the least of it.

Fixed at the seam instead: `resolveSlackTransportOptions` now confines a supplied fetch to the origins of `apiBaseUrl` and `fileBaseUrl`, and everything else travels on the global fetch. With no base configured the supplied fetch still carries every call, so instrumenting or stubbing traffic in a test is unchanged.

The example is now the header-setter with no conditional and no throw:

```ts
fetch: (input, init) => {
  const headers = new Headers(init?.headers);
  headers.set("x-vercel-protection-bypass", process.env.SLACK_SIMULATOR_BYPASS_SECRET ?? "");
  return fetch(input, { ...init, headers });
},
```

In cd3c2bc0e.

pcomans: Nothing. It was invented for the example — the only three occurrences of `SLACK_PROXY_KEY` and `x-proxy-key` in the repo were those lines, and nothing reads either.

The real mechanism is Vercel Deployment Protection, which the simulator driving this PR sits behind, so the example now sets `x-vercel-protection-bypass` from `SLACK_SIMULATOR_BYPASS_SECRET`. The header is the one eve's own docs already reference; the env var is named for the simulator's project, since the secret belongs to that deployment and not to the agent's.

The host in the example moved off `localhost:3000` too — the origin is what the confinement is keyed on now, so a plausible deployed host reads better than a port.

docs/channels/slack.mdx
```diff
@@ -418,8 +420,34 @@ if (!response.ok) throw new Error(String(response.error));
 `callSlackApi` resolves function-form tokens (secret managers, Connect
 rotation) with the supplied `context` at call time and form-encodes the body — the only safe default,
 since Slack's JSON support is partial (`conversations.replies` rejects
-JSON). `resolveSlackBotToken` materializes a `SlackBotToken` to a string
-when you need the bearer token itself.
+JSON). It has no channel to inherit
+[`api`](#point-the-channel-at-a-different-slack-api-host) from, so pass the same one the
+channel was given, through `resolveSlackTransportOptions` — that is where the bases are
+checked and resolved. An `apiBaseUrl` of `undefined` keeps Slack's own host.
```
pcomans: This seems very inelegant. Is there really no way around it

pcomans: And what does it even mean to have a channel to inherit API from

pcomans: It is, and it was worse than inelegant — it was a hole. A caller who forgot `resolveSlackTransportOptions` lost the base check, the trailing slash and the fetch confinement, with nothing to say so: either a bare `TypeError` out of the vendored transport, or a credential sent wherever the far end pointed. Documenting a step that silently drops a guarantee when skipped is not a seam worth keeping.

`callSlackApi` now resolves its own options, and resolving an already-resolved value returns it unchanged. So `slackChannel()` still resolves once at construction, where a bad base belongs, and a call made without a channel gets the same checks and the same confinement. `resolveSlackTransportOptions` is off the public surface — there is nothing left for a caller to remember.

The example is just the option now:

```ts
const response = await callSlackApi({
  api: { apiBaseUrl: process.env.SLACK_API_BASE_URL },
  botToken,
  ...
});
```

In 567062e7a.

pcomans: Nothing — it was describing eve's plumbing rather than anything you have. The channel holds a resolved `api`, and the sentence was saying a bare `callSlackApi` has no such channel behind it, so you must hand it one yourself. That is an internal detail leaking into the reference.

Gone with the fix above: `callSlackApi` takes the same `api` the channel takes and resolves it itself, so the paragraph no longer needs to explain where options come from.

.changeset/slack-api-base-url-option.md
```diff
@@ -0,0 +1,5 @@
+---
+"eve": patch
+---
+
+`slackChannel` accepts an `api` option, `{ apiBaseUrl?, fileBaseUrl?, fetch? }`, that points its Slack Web API calls and attachment downloads at another host, matching the option `discordChannel`, `telegramChannel` and `githubChannel` expose. The bases say where a call may go, and `api.fetch` says how that traffic travels: it is confined to the configured bases' origins, so a wrapper can attach credentials for the host without checking each destination itself. `slackChannel()` checks each base at construction and throws on one that is not an absolute http or https URL, naming the option. Omitting `api` keeps `https://slack.com/api/` and the global `fetch`.
```
pcomans: Rephrase this plainly without flowery prose

packages/eve/src/public/channels/slack/api-errors.ts
```diff
@@ -1,64 +0,0 @@
-/**
```
pcomans: Why was this deleted?

.changeset/slack-api-base-url-option.md
pcomans: Rewritten to two plain sentences saying what you set:

> `slackChannel` and `callSlackApi` accept an `api` option, `{ apiBaseUrl?, fileBaseUrl?, fetch? }`, that sends Slack Web API calls and attachment downloads to another host. `fileBaseUrl` defaults to `apiBaseUrl`, `fetch` is called only for URLs on those hosts, each base must be an absolute http or https URL with no query string or fragment, and omitting `api` keeps `https://slack.com/api/` and the global `fetch`.

The "bases say where a call may go / `fetch` says how the traffic travels" framing belongs in the docs page, not a release note. In 8dc4433e6.

packages/eve/src/public/channels/slack/api-errors.ts
pcomans: It moved rather than went away — into `transport.ts`, in this PR's first commit (dfe368c9d). `callSlackApiTrackingResponse`, the `slackResponseErrors` WeakSet and `isSlackResponseError` are all there unchanged.

That is the fold I said I would do in the thread above: `api-errors.ts` opened with "Separate from `api.ts` because that file sits on the 700-line cap", and adding a second module beside it would have split the channel by what fits rather than by concern. One seam now sits beside `api.ts` — the host and fetch a call travels on, and whether a thrown error means Slack answered — and its header gives that reason instead of citing the cap.

The cap did bite again straight afterwards: `api.ts` reached 702 lines and `test-unit` went red. Fixed by splitting on a real boundary rather than restoring a cap-shaped module — the helpers that shape a `SlackPostInput` into the transport's `SlackMessageOptions` moved into `api-encoding.ts`, which already owns request-body encoding.

pcomans: Correction on the last paragraph: I moved the shaping helpers into `api-encoding.ts` to get under the cap, which is the same cap-driven structure you objected to here. Reverted in 23a9b3c2b — the helpers are back in `api.ts` beside the binding that posts, and `api.ts` is on the file-length allowlist. The cap does not get to decide the module boundaries.

benpankow (approved):
Thanks, this looks good 👍

PR #3637, opened 2026-09-22
feat(eve): add inline TUI argument completion

OwenKephart:
### Summary

Typing a `/model` or `/add` argument currently requires either knowing its exact identifier or leaving the composer for a picker. Add cached, inline catalog suggestions so users can filter, navigate, and complete a model or registry item without interrupting the command. Bare commands keep their existing picker flows, while a loading row makes asynchronous catalog retrieval visible.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/cli/dev/tui/{argument-typeahead,terminal-renderer}.test.ts`

### Checklist


OwenKephart:
Addressed the outstanding bot findings in ba9225f7d: catalog fields are sanitized at the typeahead boundary before rendering or completion, and the eight-row viewport now follows the selected result (with coverage for a ninth entry). The loading-state report was already resolved by the catalog-state refactor.

AndrewBarba (approved):

PR #3541, opened 2026-09-18
feat(eve): mount bundled development extensions

prha:
### Summary
Local development should expose eve's self-modification capabilities without requiring every project to install an extension mount first. This change makes `eve dev` mount bundled development extensions by default, initially self-modification, while letting an authored
 `self-modification` mount take precedence and providing `--no-default-extensions` to disable the defaults. The selection persists across local server processes and rebuilds, while production bundles do not initialize development-only mounts.

### Validation

- `pnpm --filter eve exec vitest run --config vitest.unit.config.ts src/compiler/source-graph.test.ts src/discover/bundled-extension.test.ts src/self-modification/extension/subagents/agent/instructions.test.ts src/cli/run.test.ts` (78 tests passed)
- `pnpm --filter eve exec vitest run --config vitest.integration.config.ts src/compiler/development-extensions.integration.test.ts` (2 tests passed)
- `git diff --check`

### Checklist

- [ ] This change was requested or approved by a maintainer
- [x] I ran the relevant checks from `CONTRIBUTING.md`
- [x] I added tests and documentation where relevant
- [x] I added a changeset if this touches the published `eve` package
- [x] DCO sign-off passes for every commit (`git commit --signoff`)

PR #3639, opened 2026-09-22
fix(eve): install just-bash in scenario apps

ctgowrie:
### Summary

Related to #3541. Default development extensions now give generated scenario apps a self-modification subagent backed by just-bash, but those apps did not install the optional engine up front. `eve dev` consequently attempted a runtime install that pnpm rejected under strict build-script policy; the scenario harness now pins and installs its workspace just-bash version alongside its existing default dependencies.

### Validation

The scenario-app materialization suite confirms the generated manifest and installed dependency, and the previously failing dev-server shutdown scenario passes with default extensions enabled. Workspace typecheck and invariant checks also passed.

### Checklist

- [x] Changeset not applicable; this only changes internal scenario fixtures

### Diff size

Not applicable.

**Implementation** — 1 file · `+13 / -6`

Keeps generated scenario applications deterministic by resolving and declaring the installed just-bash version before their initial dependency install.

**Tests** — 1 file · `+4 / -0`

Verifies both the generated dependency declaration and the installed package.


ctgowrie:
Going to revert change instead

